PR #6, opened 2018-04-11
[TST] Resolve np.allclose error in integration testing

emdupre:
Make sure we're getting consistent results !

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: python
+sudo: false
+
+python:
+  - 3.6
+
+before_install:
+  - python -m pip install --upgrade pip
+  - pip install "flake8<3.0" flake8-putty
+
+script:
+- flake8 tedana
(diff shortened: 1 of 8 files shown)


emdupre:
To facilitate in this, made it such that CircleCI saves out build artifacts (for comparison / inspection), and added PEP8 linting with TravisCI on PRs, to encourage good development style !

emdupre:
Went back to https://github.com/me-ica/me-ica/ and used the `tedana.py` script to generate a new testing data set, since I lost confidence that the CI errors weren't just due to a forgotten seed value. The resulting files were, perhaps unsurprisingly, still different from those I had been generating.

It ended up that this was from a few differences in the re-coded script and in py27 vs py3, particularly in the handling of global variables. I'd love to just get rid of globals altogether, but for now (after almost crying in frustration over it), comp_table.txt at least aligns with the original script's outputs 🎉 

The one complication (of course) is that this is currently true locally, but not on the circle build. I'm going to go ahead and merge this, and then open another PR to see where there's an issue in circle-- maybe the cache ?

KirstieJane:
I’m no help to you on the technical front @emdupre but big hugs on the frustrations of tracking this down! Keep going! :muscle

PR #22, opened 2018-04-27
[WIP, ENH] Add ability to read surface .gii files

rmarkello:
Working to add the ability to seamlessly accept gifti files in addition to nifti files.

This is still a long ways from ready, but I want to use circle testing to ensure I don't screw up anything in the process of making changes. As I'm modifying the functions to accept surface files, I'm making some stylistic changes along the way and beefing up doc-strings, as applicable.

### DIFF
--- a/tedana/interfaces/t2smap.py
+++ b/tedana/interfaces/t2smap.py
@@ -3,6 +3,9 @@ import nibabel as nib
 from tedana.utils import (niwrite, cat2echos,
                           makeadmask, unmask, fmask)
 
+import logging
+lgr = logging.getLogger(__name__)
+
 
 def t2sadmap(catd, mask, tes, masksum, start_echo):
     """
@@ -120,16 +123,16 @@ def main(options):
     catd = cat2echos(catim.get_data(), ne)
     nx, ny, nz, Ne, nt = catd.shape
 
-    print("++ Computing Mask")
-    mask, masksum = makeadmask(catd, min=False, getsum=True)
+    lgr.info('++ Computing Mask')
+    mask, masksum = makeadmask(catd, minimum=False, getsum=True)
 
-    print("++ Computing Adaptive T2* map")
+    lgr.info('++ Computing Adaptive T2* map')
     t2s, s0, t2ss, s0vs, t2saf, s0vaf = t2sadmap(catd, mask, tes, masksum, 2)
     niwrite(masksum, aff, 'masksum%s.nii' % suf)
     niwrite(t2ss, aff, 't2ss%s.nii' % suf)
     niwrite(s0vs, aff, 's0vs%s.nii' % suf)
 
-    print("++ Computing optimal combination")
+    lgr.info('++ Computing optimal combination')
     tsoc = np.array(optcom(catd,
                            t2s,
                            tes,
--- a/tedana/interfaces/t2smap.py
+++ b/tedana/interfaces/t2smap.py
@@ -7,102 +7,113 @@ import logging
 lgr = logging.getLogger(__name__)
 
 
-def t2sadmap(catd, mask, tes, masksum, start_echo):
+def t2sadmap(data, tes, mask, masksum, start_echo):
     """
-    t2sadmap(catd,mask,tes,masksum)
+    Parameters
+    ----------
+    data : (S x E x T) array_like
+        Multi-echo data array, where `S` is samples, `E` is echos, and `T` is
+        time
+    tes : (E, ) list
+        Echo times
+    mask : (S, ) array_like
+        Boolean array indicating samples that are consistently (i.e., across
+        time AND echoes) non-zero
+    masksum : (S, ) array_like
+        Valued array indicating number of echos that have sufficient signal in
+        given sample
+    start_echo : int
+        First echo to consider
 
-    Input:
-
-    catd  has shape (nx,ny,nz,Ne,nt)
-    mask  has shape (nx,ny,nz)
-    tes   is a 1d numpy array
-    masksum
+    Returns
+    -------
+    t2sa : (S x E x T) np.ndarray
+        Limited T2* map
+    s0va : (S x E x T) np.ndarray
+        Limited S0 map
+    t2ss : (S x E x T) np.ndarray
+        ???
+    s0vs : (S x E x T) np.ndarray
+        ???
+    t2saf : (S x E x T) np.ndarray
+        Full T2* map
+    s0vaf : (S x E x T) np.ndarray
+        Full S0 map
     """
-    nx, ny, nz, Ne, nt = catd.shape
-    echodata = fmask(catd, mask)
-    Nm = echodata.shape[0]
 
-    t2ss = np.zeros([nx, ny, nz, Ne - 1])
-    s0vs = t2ss.copy()
+    n_samp, n_echos, n_vols = data.shape
+    t2ss, s0vs = np.zeros([n_samp, n_echos - 1]), np.zeros([n_samp, n_echos - 1])
 
-    for ne in range(start_echo, Ne + 1):
-
-        # Do Log Linear fit
-        B = np.reshape(np.abs(echodata[:, :ne]) + 1, (Nm, ne * nt)).transpose()
+    for echo in range(start_echo, n_echos + 1):
+        # perform log linear fit of echo times against MR signal
+        B = np.reshape(np.abs(data[:, :echo, :]) + 1,
+                       (n_samp, echo * n_vols)).T
         B = np.log(B)
-        neg_tes = [-1 * te for te in tes[:ne]]
-        x = np.array([np.ones(ne), neg_tes])
-        X = np.tile(x, (1, nt))
-        X = np.sort(X)[:, ::-1].transpose()
+        neg_tes = [-1 * te for te in tes[:echo]]
+        x = np.array([np.ones(echo), neg_tes])
+        X = np.tile(x, (1, n_vols))
+        X = np.sort(X)[:, ::-1].T
 
         beta, res, rank, sing = np.linalg.lstsq(X, B)
-        t2s = 1 / beta[1, :].transpose()
-        s0 = np.exp(beta[0, :]).transpose()
+        t2s = 1 / beta[1, :].T
+        s0 = np.exp(beta[0, :]).T
 
-        t2s[np.isinf(t2s)] = 500.
-        s0[np.isnan(s0)] = 0.
+        t2s[np.isinf(t2s)] = 500.  # why 500?
+        s0[np.isnan(s0)] = 0.      # why 0?
 
-        t2ss[:, :, :, ne - 2] = np.squeeze(unmask(t2s, mask))
-        s0vs[:, :, :, ne - 2] = np.squeeze(unmask(s0, mask))
+        t2ss[..., echo - 2] = np.squeeze(t2s)
+        s0vs[..., echo - 2] = np.squeeze(s0)
 
-    # Limited T2* and S0 maps
-    fl = np.zeros([nx, ny, nz, len(tes) - 2 + 1])
-    for ne in range(Ne - 1):
-        fl_ = np.squeeze(fl[:, :, :, ne])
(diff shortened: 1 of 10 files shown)


rmarkello:
I realized my original comment vastly understated the amount of refactoring that I'm doing to integrate gifti support. 

This is going to be a bit of a beast of a PR, so perhaps some reviews along the way might be helpful to ensuring that I'm on the right track? I *think* that my refactoring of `utils.py` is mostly done (though there will be a few minor tweaks to accommodate L/R hemisphere denotations for gifti files), if someone would like to start there!

tedana/utils/utils.py
```diff
@@ -1,198 +1,452 @@
...
+    mask : (S,) array_like
+        Boolean array of `S` samples that was used to mask `data`
+
+    Returns
+    -------
+    out : (S x E x T) np.ndarray
+        Unmasked `data` array
+    """
+
+    out = np.zeros((mask.shape + data.shape[1:]))
```
emdupre: Is there a reason this is not `out = np.zeros(mask.shape + data.shape[1:])` ?

tedana/utils/utils.py
```diff
@@ -1,198 +1,452 @@
...
+    return fdata.squeeze()
+
+
+def unmask(data, mask):
+    """
+    Unmasks `data` using non-zero entries of `mask`
+
+    Parameters
+    ----------
+    data : (M x E x T) array_like
```
emdupre: This notation is slightly confusing. Are `M` and `S` of equal size ?

tedana/utils/utils.py
```diff
@@ -1,198 +1,452 @@
...
+    mask : (X x Y x Z) array_like or img_like object
+        Boolean array or mask file
+
+    Returns
+    -------
+    fdata : (S x E x T) np.ndarray
+        Masked `data`, where `S` is samples, `E` is echoes, and `T` is time
+    """
+
+    if mask is not None and not type(data) == type(mask):
```
emdupre: Should we confirm that they're of the same shape, not just the same type ?

tedana/utils/utils.py
```diff
@@ -1,198 +1,452 @@
...
+    if isinstance(data, str):
+        root, ext, addext = splitext_addext(data)
+        if ext == '.gii':
+            # mask need not apply for gii files
+            fdata = np.column_stack([f.data for f in nib.load(data).darrays])
+        else:
+            # use nilearn for other files
+            data = check_niimg(data)
+            if mask is not None:
+                # TODO: check that this uses same order to flatten
```
emdupre: Same order as what ?

tedana/utils/utils.py
```diff
@@ -1,198 +1,452 @@
...
-        Boolean array of shape (nx, ny, nz)
+    data : (X x Y x Z x E x T) array_like
+        Multi-echo data array
 
     Returns
     -------
-    ndarray
-        Array of shape (nx, ny, nz[, Ne[, nt]])
+    fdata : (X x Y x M x T) np.ndarray
+        Z-concatenated multi-echo data array, where M is Z * number of echos
```
emdupre: If we drop zcat compatibility (as discussed in #26), will we drop this function as well ? For now, it's probably worth keeping, but thinking ahead....

tedana/utils/utils.py
```diff
@@ -1,198 +1,452 @@
...
+    fdata = load_image(img.get_data().reshape(nx, ny, nz, n_echos, -1))
+
+    # create reference image
+    ref_img = img.__class__(np.zeros((nx, ny, nz)), affine=img.affine,
+                            header=img.header, extra=img.extra)
+
+    return fdata, ref_img
+
+
+def makeadmask(data, minimum=True, getsum=False):
```
emdupre: Should we rename this? I'm still not sure what `ad` in `makeadmask` stands for.

tedana/utils/utils.py
```diff
@@ -1,198 +1,452 @@
...
     """
 
-    nx, ny = data.shape[0:2]
-    nz = data.shape[2] * Ne
-    if len(data.shape) > 4:
-        nt = data.shape[4]
-    else:
-        nt = 1
-    return np.reshape(data, (nx, ny, nz, nt), order='F')
+    if isinstance(data, list):
```
emdupre: Should we raise an error for the `else` condition, here? I'm not sure how it could be triggered, but just in case....

tedana/utils/utils.py
rmarkello: The `uncat2echos` function is actually no longer needed, I think -- even with retained zcat compatibility! A combination of `load_data()` and `filewrite()` should be enough to ensure we can handle everything. I'll delete this code to remove any confusion!

tedana/utils/utils.py
rmarkello: Definitely can be renamed. I believe it stood for **make** **ad**apative **mask**, but since that's clear to exactly no one, a more pythonic naming convention (`make_adaptive_mask()`) might be more applicable? I'm happy to do whatever works best!

tedana/utils/utils.py
rmarkello: Sorry! This was when I was waffling back and forth about what that first dimension of the data arrays should be called. I will fix this and enforce `S` (for data **S**amples).

tedana/utils/utils.py
rmarkello: Nope! Good catch -- will fix in the next commit.

tedana/utils/utils.py
rmarkello: This is worth some discussion about how we envision people using the tedana interface. As it stands, the primary mechanism for doing that is through `argparse` in `cli/run.py`; if that's the case, `data` will *always* be a list. However, if the goal is to make this a bit more modular, then we should probably add in some other checks. 

Right now even if `data` is a `str` pointing to a nifti file this function should still work (without an explicit `else` clause). But it might be good to enforce some requirements! I'm open to comments/suggestions and will be happy to build them in.

tedana/utils/utils.py
rmarkello: Apologies for the confusion; I'll beef up the doc-string here. `M` refers to the number of `True` values in the mask that had been applied to the data.

tedana/utils/utils.py
emdupre: Yes, I agree there's a broader discussion to have about how to use the tedana interface-- I'll punt on that for now. It sounds like this can handle the two most likely cases, so we can revisit when it comes time to talk about how we expect people to run the code.

handwerkerd: It's unclear how necessary this is, but I'm fairly sure some of the current component selection criteria use spatial information. For example, in tedana.py, see the current code starting at line 652: "Step 2: Compute 3-D spatial FFT of Beta maps to detect high-spatial
    frequency artifacts"

The simple solution would be to just skip 3-D space dependent criteria with gii files. I suspect that won't cause much harm, but it would require a some close examination of results processed as volumes vs surfaces. 

My hunch is that the spatial transforms and interpolations across voxels to go from a volume to a surface might cause some undesired issues with echo-dependent estimations, but there might also be some issues with non-trivial transforms & interpolations in volume space as well. I also don't have a good intuitive sense of how tedana classification mistakes would look in surface space, so there might need to be some thought on how to best visualize results to help users identify issues.

rmarkello: Good point, and thanks for raising it! Yes, there are going to have to be some differences between the pipelines for volumetric + surface data. Thankfully, having CircleCI running will ensure that none of the changes I'm making in this PR impact the volumetric results in any way, for now. 

A lot of this PR is actually just refactoring to ensure that data can be seamlessly passed between functions without regard for the **type** of data being passed. This will hopefully be useful later in making the code more modular.

As for visualization of surface-based results: I completely agree. I think there will have to be some significant testing to ensure that `tedana` is actually suitable for surface data, given the limited spatial information. Hopefully this PR will be the first step in that process!

rmarkello:
It passes! Which means this should be ready for review. 🎉

I haven't actually tested whether surface data can be run through yet, but it's in a *much* better place to incorporate those data. I think it might be beneficial to merge this in and then work out the surface bugs (and add some tests to ensure they work) afterwards.

Note: in the process of making these changes this I experienced an error wherein taking the mean of a large, float32 numpy array gave varying results across Python instances, despite the underlying data being **exactly** the same to an incredibly conservative tolerance. Because of this, this PR includes a (very slight) breaking change; the CI was updated to reflect these new (hopefully more accurate) outputs.

tedana/interfaces/tedana.py
```diff
@@ -399,28 +498,25 @@ def fitmodels_direct(catd, mmix, mask, t2s, t2sG, tes, combmode, head,
...
-    nx, ny, nz, _, n_components = betas.shape
+    # compute Betas and means over TEs for TE-dependence analysis
+    betas = get_coeffs(catd, np.repeat(mask[:, np.newaxis], len(tes), axis=1), mmix)
+    n_samp, n_echos, n_components = betas.shape
     n_voxels = mask.sum()
     n_data_voxels = (t2s != 0).sum()
-    mu = catd.mean(axis=-1)
-    tes = np.reshape(tes, (n_echoes, 1))
-    fmin, fmid, fmax = getfbounds(n_echoes)
+    mu = catd.mean(axis=-1, dtype=float)  # BUG: THIS IS THE BAD PLACE
```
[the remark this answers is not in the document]
rmarkello: Yes, this can definitely be removed since there is a trace of the rationale behind changing it. It might be worth following numpy development and reverting this change eventually, but I think that will be further down the line.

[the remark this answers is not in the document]
rmarkello: Yes, this can definitely be removed since there is a trace of the rationale behind changing it. It might be worth following numpy development and reverting this change eventually, but I think that will be further down the line.

tedana/interfaces/tedana.py
```diff
@@ -63,145 +68,122 @@ def do_svm(X_train, y_train, X_test, svmtype=0):
...
     """
+
     if infile is None:
         data = data.copy()
         data[data < thr] = 0
-        niwrite(unmask(data, mask), aff, '__clin.nii.gz', header)
-        infile = '__clin.nii.gz'
+        infile = filewrite(unmask(data, mask), '__clin.nii.gz', ref_img, gzip=True)
+
+    # FIXME: ideally no calls to os.system!!! (or AFNI, for that matter)
```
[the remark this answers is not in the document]
rmarkello: Done! See #32.

tedana/interfaces/tedana.py
```diff
@@ -63,145 +68,122 @@ def do_svm(X_train, y_train, X_test, svmtype=0):
     elif svmtype == 2:
         clf = svm.SVC(kernel='linear', probability=True)
     else:
-        raise ValueError('Input svmtype not in range (3)')
+        raise ValueError('Input svmtype not in [1, 2, 3]')
```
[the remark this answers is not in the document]
rmarkello: Yes! Will fix asap.

tedana/interfaces/t2smap.py
```diff
@@ -67,130 +68,133 @@ def fit(data, mask, tes, masksum, start_echo):
...
+
+    for echo in range(start_echo, n_echos + 1):
+        # perform log linear fit of echo times against MR signal
+        # make DV matrix: samples x (time series * echos)
+        B = np.log((np.abs(data[:, :echo, :]) + 1).reshape(len(data), -1).T)
+        # make IV matrix: intercept/TEs x (time series * echos)
+        x = np.column_stack([np.ones(echo), [-te for te in tes[:echo]]])
+        X = np.repeat(x, n_vols, axis=0)
+
+        beta, res, rank, sing = np.linalg.lstsq(X, B)
```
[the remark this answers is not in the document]
rmarkello: No! And indeed, I think a more parsimonious solution than either of the above would probably be:

```
beta = np.linalg.lstsq(X, B)[0]
```

Not sure my rationale here, so I'll change it to reflect this.

emdupre:
LGTM ! Merging !

Issue #10, opened 2018-04-20
Generate FIT R2* and S0 timeseries


tsalo:
The R2* and S0 time series produced by fitting a monoexponential decay model at each voxel and each timepjoint are used in the recent [Power paper](http://www.pnas.org/content/early/2018/02/07/1720985115) on removing motion-related noise from ME data, so it might be worth it to include that in tedana?

emdupre:
Closed with #23 !

PR #23, opened 2018-04-27
[ENH] Add FIX method for estimating T2/S0

tsalo:
To address #10. The new function, `tedana.interfaces.t2smap.fit`, is meant to estimate T2/S0 per voxel and per volume (as opposed to estimating T2/S0 across volumes as in `tedana.interfaces.t2smap.t2sadmap`). The function `tedana.interfaces.t2smap.optcom` should also now accept 4D T2 estimate maps in addition to 3D ones.


emdupre:
😄 Thanks for opening !

Is this still WIP, or ready for review ?

emdupre:
This looks great ! At some point I'd like to see if we can refactor these so that they're made of several smaller functions (so they're not duplicating quite as much code), but I'll open that as a separate issue. Thanks ! 🎉

PR #24, opened 2018-04-28
[WIP, ENH] Allow separate data files as input.

tsalo:
This should make it easier to use tedana in Python.

Also:
- Split arguments in tedana.main
- Switch from calling os.system to using shutil and specific os
functions
- Change variable dirname to out_dir
- Use full paths when saving files. Should eliminate need to chdir into
out_dir halfway through the function
- Add docstring for main

### DIFF
--- a/tedana/interfaces/tedana.py
+++ b/tedana/interfaces/tedana.py
@@ -1558,7 +1558,7 @@ def main(data, tes, mixm=None, ctab=None, manacc=None, strict=False,
                                                            mask, t2s, t2sG,
                                                            tes, combmode, head,
                                                            fout=fout)
-        if options.ctab is None:
+        if ctab is None:
             acc, rej, midk, empty = selcomps(seldict, mmix, head, manacc,
                                              strict_mode=strict)
         else:
--- a/tedana/interfaces/tedana.py
+++ b/tedana/interfaces/tedana.py
@@ -1017,7 +1017,8 @@ def selcomps(seldict, mmix, head, manacc, debug=False, olevel=2, oversion=99,
     return list(sorted(ncl)), list(sorted(rej)), list(sorted(midk)), list(sorted(ign))
 
 
-def tedpca(combmode, mask, stabilize, head, ste=0, mlepca=True):
+def tedpca(combmode, mask, stabilize, head, tes, kdaw, rdaw, ste=0,
+           mlepca=True):
     nx, ny, nz, ne, nt = catd.shape
     ste = np.array([int(ee) for ee in str(ste).split(',')])
     if len(ste) == 1 and ste[0] == -1:
@@ -1399,7 +1400,7 @@ def main(data, tes, mixm=None, ctab=None, manacc=None, strict=False,
         Either a single z-concatenated file (str or single-entry list) or a
         list of echo-specific files, in ascending order.
     tes : :obj:`list`
-        List of echo times associated with data in seconds.
+        List of echo times associated with data in milliseconds.
     mixm : :obj:`str`, optional
         File containing mixing matrix. If not provided, ME-PCA and ME-ICA are
         done.
@@ -1444,7 +1445,6 @@ def main(data, tes, mixm=None, ctab=None, manacc=None, strict=False,
     fixed_seed : :obj:`int`, optional
         Seeded value for ICA, for reproducibility.
     """
-    # global tes,
     global ne, catd, head, aff
     tes = [float(te) for te in tes]
     ne = len(tes)
@@ -1476,7 +1476,6 @@ def main(data, tes, mixm=None, ctab=None, manacc=None, strict=False,
     else:
         fout = None
 
-    global kdaw, rdaw
     kdaw = float(kdaw)
     rdaw = float(rdaw)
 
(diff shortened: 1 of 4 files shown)


tsalo:
It looks like it's breaking because `options` is being fed into `selcomps`, but it's unclear what variables from options are being used [here](https://github.com/tsalo/tedana/blob/9d609dc508654852cf0904b5e4bd6630866a9188/tedana/interfaces/tedana.py#L569-L571). 

Also, not sure why `sys.argv` is being used [here](https://github.com/tsalo/tedana/blob/9d609dc508654852cf0904b5e4bd6630866a9188/tedana/interfaces/tedana.py#L1552-L1555), or even how it's being passed in.

emdupre:
This is great ! 🎉

I think we can get rid of both of those sections (`knobargs` and `sys.argv`); they're left over from the previously implemented ability to [pass unnamed arguments](https://github.com/ME-ICA/me-ica/blob/master/meica.py#L191), which I don't think is going to help create reproducible executions.

tsalo:
Yeah, after digging a bit harder it became clear that the only fields in `knobargs` that `selcomps` uses are already named arguments, so it was redundant. Regarding `sys.argv`, I have no clue what `group0_flag` does or how the switch is meant to work, but `sys.argv` will be empty if `tedana` is called in Python. Not sure if `group0_flag` should be set to `False` in that case, but I'm pretty sure it will be.

On an unrelated note, is there any reason the output files can't be gzipped (`.nii.gz`)? I was thinking of adding that into this PR.

emdupre:
The output files could definitely be gzipped, and probably should be. In #22 I know that @rmarkello was working to re-do a lot of the IO more generally, so I might wait until we get a little further on that PR to rework the out file format. 

It's a good point, though, that calling in python could lead to odd behavior for `sys.argv`. I'd be happy to just get rid of the `group0` option, since that was undocumented behavior anyway.

tedana/interfaces/tedana.py
```diff
@@ -1365,165 +1343,214 @@ def writeresults(OCcatd, comptable, mmix, nt, acc, rej, midk, empty, head):
...
                                                            mask, t2s, t2sG,
                                                            tes, combmode, head,
-                                                           fout=options.fout,
+                                                           fout=fout,
                                                            reindex=True)
-
-        np.savetxt('meica_mix.1D', mmix)
+        np.savetxt(op.join(out_dir, 'meica_mix.1D'), mmix)
 
         if 'GROUP0' in sys.argv:
```
emdupre: What do you think of getting rid of `group0` here and in `selcomps` ? I'm ok to remove the option since it was undocumented behavior, anyway. We can open an issue about documenting and adding it back in eventually, but I'd like to get rid of this `sys.argv` call sooner than later.

tedana/interfaces/tedana.py
```diff
@@ -1365,165 +1343,214 @@ def writeresults(OCcatd, comptable, mmix, nt, acc, rej, midk, empty, head):
...
+         fixed_seed=42):
     """
-
-    Args (and defaults):
-    data, tes, mixm=None, ctab=None, manacc=None, strict=False,
-             no_gscontrol=False, kdaw=10., rdaw=1., conv=2.5e-5, ste=-1,
-             combmode='t2s', dne=False, initcost='tanh', finalcost='tanh',
-             stabilize=False, fout=False, filecsdata=False, label=None,
-             fixed_seed=42
+    Parameters
```
emdupre: This is beautiful 😍

tedana/interfaces/tedana.py
tsalo: I think that's a great idea. I'll remove it posthaste.

tsalo:
The three remaining flake errors (`midkfailed`, `dice_rej`, and `svm_acc_fail` are unused) aren't really errors, because they _are_ used [here](https://github.com/tsalo/tedana/blob/11282c5f409fa44130917085aba5aebe3e3689a8/tedana/interfaces/tedana.py#L984-L990), so I think it's ready to merge.

emdupre:
It'd be a shame to break linting over this. And, looking at it, the format we're saving out really isn't ideal for later programmatic access: 
screen shot 2018-05-04 at 16 03 45
What would you think of converting this to a dictionary and writing out as a json-- We could even give more relevant fields than just the variable names. It might kill two birds with one stone. WDYT ?

emdupre:
(If you'd rather, I could merge this in and update it after, too ! Just LMK 😄 )

tsalo:
I definitely like the idea, I don't know what several of those variables actually are, so I don't think I could come up with better fields.

Similar approaches are used above, [here](https://github.com/tsalo/tedana/blob/11282c5f409fa44130917085aba5aebe3e3689a8/tedana/interfaces/tedana.py#L546-L551) and [here](https://github.com/tsalo/tedana/blob/11282c5f409fa44130917085aba5aebe3e3689a8/tedana/interfaces/tedana.py#L806-L811), so it might be good to do the same thing there.

emdupre:
Sounds good. I've opened issues for the points we've raised, so I'll merge this in now !

PR #31, opened 2018-05-10
Update catd mean to float64

emdupre:
To aid in merging #22, where we've collapsed the xyz dimensions into an S dimension. Tracing back on a numpy core bug with large arrays: https://github.com/numpy/numpy/issues/8869#issuecomment-369620242


emdupre:
This was a breaking change-- the added dimensionality from float64 results in one component that was previously ignored being rejected. Implemented commit https://github.com/ME-ICA/tedana/pull/31/commits/074b03cbb1cd582b7803944cd0fb334e594afd08 in #22

PR #40, opened 2018-05-11
[RF] Remove os.system calls to AFNI

### LINKED ISSUE #32: Remove os.system calls in tedana.py
There are a few calls to `os.system` in `tedana.py` that might be worth removing. Most of them could be replaced by nilearn functionality: thresholding images could use [`nilearn.image.threshold_img`](http://nilearn.github.io/modules/generated/nilearn.image.threshold_img.html#nilearn.image.threshold_img) and clustering could use [`nilearn.regions.connected_regions`](http://nilearn.github.io/modules/generated/nilearn.regions.connected_regions.html#nilearn.regions.connected_regions). 

This would take a bit of work to finesse because the calls to AFNI are not incredibly straightforward. Different sub-bricks are used for thresholding and clustering at various points throughout, but the AFNI documentation for the calls that are being made (thresholding uses [`3dcalc`](https://afni.nimh.nih.gov/pub/dist/doc/program_help/3dcalc.html) and clustering uses [`3dmerge`](https://afni.nimh.nih.gov/pub/dist/doc/program_help/3dmerge.html)) is relatively verbose and should be helpful in doing this.

rmarkello:
I believe this gets rid of all the `os.system` calls and removes AFNI as a dependency! It also writes out 5 fewer files (no `.cc_temp.nii.gz`, `.fcl_in.nii.gz`, `.fcl_out.nii.gz`, `__clin.nii.gz`, `__clout.nii.gz`), so I've removed those from testing.

Closes #32.


emdupre:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 
afaik, we are a pure python package, now !!! 
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

emdupre:
(will confirm and code review this tmrw)

KirstieJane:
😱😱😱😱

Incredible!! Way to go @emdupre and ALL THE _tedana_ contributors 🙌🏻

🌺🌸🍾🥂🍹🍰🎂🍭🎊🎉🦄

tedana/model/fit.py
```diff
@@ -496,51 +485,64 @@ def gscontrol_raw(catd, optcom, n_echos, ref_img, dtrank=4):
...
-        addopts = '-doall'
-    else:
-        addopts = '-1dindex {0} -1tindex {1}'.format(str(dindex), str(tindex))
-
-    cmd_str = '3dmerge -overwrite {0} -dxyz=1 -1clust 1 {1:d} ' \
-              '-1thresh {2:.02f} -prefix __clout.nii.gz {3}'
-    os.system(cmd_str.format(addopts, int(csize), float(thr), infile))
+    img = niimg.copy_img(check_niimg(img, atleast_4d=True))
+
+    # make sure voxel sizes are 1mm^3
```
emdupre: What happens if this is not true ?

tedana/tests/test_tedana.py
```diff
@@ -47,11 +47,11 @@ def test_outputs():
      'T1gs.nii',
      'tsoc_orig.nii',
      'tsoc_nogs.nii',
-     '.cc_temp.nii.gz',
-     '.fcl_in.nii.gz',
-     '.fcl_out.nii.gz',
-     '__clin.nii.gz',
-     '__clout.nii.gz',
+     # '.cc_temp.nii.gz',
```
emdupre: Can we just delete these, rather than commenting them out ?

rmarkello: Definitely! I suppose I should have done that in the first place, but I wasn't confident this would work in a single shot. I'll remove them now.

tedana/model/fit.py
rmarkello: It's not so much "make sure" as it is "temporarily set voxel sizes to 1mm^3". That way the `min_cluster_size` argument represents the number of voxels we want in a cluster rather than the actual size of the cluster in mm^3. This was originally accomplished using `3dmerge -dxyz=1`; `nilearn.regions.connected_regions` does't have a comparable parameter.

We could instead do some calculation of the actual voxel volume, but I wasn't sure how rounding errors would be treated. Moreover, temporarily resetting the affine should be okay because it (the affine) is discarded when the clustered data is loaded back in to memory: 
```
clustered = utils.load_image(niimg.concat_imgs(clout).get_data()) != 0
```

emdupre: Ah, got it. Sorry, I misinterpreted the code comment ! I think that method totally makes sense. Would it be possible to update the comment ?

rmarkello: Definitely! Apologies for being so minimalist with them. I'll try to be a bit more descriptive in the future :smile:

Issue #30, opened 2018-05-07
Reorganize package


tsalo:
Here is the proposed structure, from conversations on [gitter](https://gitter.im/ME-ICA/tedana):

- `tedana.workflows`: Common pipelines combining functions/classes for performing TE-dependent analysis
    - `tedana.workflows.t2smap`: Fit monoexponential decay curve to each voxel, estimating S0 and T2*, and then optimally combine echoes based on T2* estimate.
- `tedana.io`: File IO functions/classes
- `tedana.model`: Model fitting
    - `tedana.model.t2sadmap`
    - `tedana.model.fit`
    - `tedana.model.optcom`
- `tedana.decomposition`: Functions/classes for decomposing multi-echo data
    - `tedana.decomposition.tedica`
    - `tedana.decomposition.tedpca`
- `tedana.selection`: Component selection functions/classes
    - `tedana.selection.characterize`: Component characterization (in terms of TE-dependence and -independence)
    - `tedana.selection.select_components`
- `tedana.qc`: Quality control functions/classes
- `tedana.report`: Report generation
    - `tedana.report.static_report`
    - `tedana.report.interactive_report`
- `tedana.utils`: Miscellaneous utility functions

emdupre:
Started in #35 ! I made two horrible typos and lost a few hours figuring that out, but I'm hoping to get this done tonight 🕵️‍♀️

tsalo:
Is it considered awkward or bad form to include submodule-specific utils? E.g., not just `tedana.utils`, but also `tedana.model.utils`, `tedana.select.utils`, etc.?

tsalo:
I think that, in its current form (i.e., 1-2 files per submodule), @rmarkello is probably right, but I also think it depends on how much `tedana` will grow. The submodules `decomp`, `model`, and `select` are going to include potentially many more methods in the future, so I think that the current organization is just preparing for that. Then again, if it turns out that not many new methods will be added, then this organization could be a bit much.

rmarkello:
I agree that this is setting up the room for growth (which is definitely a good thing!), but I don't know if it needs to be quite so fragmented. 

Is it possible that `decomp`, `model`, and `select` could all go under one submodule: `modeling`? And it might be reasonable to move `io` to the `utils` submodule. I was thinking something like:

* `tedana.modeling`
    * `tedana.modeling.model`
    * `tedana.modeling.decomposition`
    * `tedana.modeling.select` -- this could potentially be a sub-module as well!
[...]

PR #43, opened 2018-05-11
[DOC] Update style guide and auto documentation

tsalo:
This started as an attempt to update the style guide per @KirstieJane's suggestion in #30, but it ended up sprawling a bit as I tried to get the API documentation working as well. It looks pretty good, because I was able to re-use a lot of code I used in another site, but there are a couple of problems that I don't know how to fix. Namely:
- How do we want to handle the functions that are initialized a level above where they're defined? Not sure that's how to describe what we're doing, but hopefully it makes sense.
- Can we remove the module links in the tables?
- Can we get the `source` links to point to GitHub instead of the code snippets?
 
---

screen shot 2018-05-12 at 8 46 23 pm

---

screen shot 2018-05-12 at 8 46 37 pm

---

screen shot 2018-05-12 at 8 47 15 pm


KirstieJane:
Yay!! Thank you @tsalo! I’m not knowledgable enough about the details to review the PR, but I do want to say thank you for adding so much to the documentation! So so useful 😍😍😍

emdupre:
This is (another) great PR, @tsalo !!! 

- re documenting only top-level imports: it looks as though we can use the [`..automodule:: pkg` option](https://stackoverflow.com/questions/15115514/how-do-i-document-classes-without-the-module-name). I haven't tried it, personally, but it would be worth a shot !
- You don't like the module links ?? 😢 Or are they not linking out appropriately ?
- To get the source code links I think we'll need to add a sphinx extension&mdash; either [this one](http://www.sphinx-doc.org/en/master/ext/viewcode.html) or [this one](https://github.com/scikit-learn/scikit-learn/blob/master/doc/sphinxext/github_link.py)

tsalo:
@emdupre Unfortunately, I don't think some of those work (or I'm not using them correctly). I'm currently using `..automodule:: pkg` [here](https://github.com/tsalo/tedana/blob/93c38e5ad8e10b327e8d51349331c713b7e48d2d/docs/api.rst), and `sphinx.ext.viewcode` [here](https://github.com/tsalo/tedana/blob/93c38e5ad8e10b327e8d51349331c713b7e48d2d/docs/conf.py#L43), with no luck.

The sklearn extension looks promising, but I have no clue how to use it.

emdupre:
I opened [a PR on your fork](https://github.com/tsalo/tedana/pull/2) with one idea ! I'll tweak more on there to see if we can get this in line with what you're looking for, though it's a great PR as is 😄

tsalo:
Someone else will need to keep working on the Notes in the docstrings, given that I'm not that knowledgeable about the math behind tedana, but other than that I think it's ready to merge.

docs/sphinxext/github_link.py
```diff
@@ -0,0 +1,84 @@
+from operator import attrgetter
```
emdupre: This is totally my oversight, but would you mind adding something here to reference that this was taken from sklearn ? I believe their license is BSD, but I'm sure attribution is appreciated !

tsalo: Good catch! Should be good as of 6b1dfb7.

emdupre:
LGTM ! Merging ! 🎉

Issue #14, opened 2018-04-25
Replace mdp.FastICANode with sklearn.decomposition.FastICA


emdupre:
_From @emdupre on November 15, 2017 16:38_

After refactoring, if integration tests show that we're still achieving similar results, we should remove the `mdp` dependency as this library is no longer actively maintained. Instead, this should be replaced with FastICA as implemented in `sklearn`. 

_Copied from original issue: emdupre/tedana#7_

Issue #18, opened 2018-04-25
Create non-interative docker container


emdupre:
In testing, it seems that some OS variables are leaking into the docker environment-- even though we're running tedana in a container, we get different results across operating systems.

The current thought is that this is because we're running the container interactively and just mounting the code in. If `mdp` (which runs the ICA) is calling on some OS environmental variable in its random seed setting, this configuration could be exposing variables from the host system and yielding different results. 

The short term solution is to just treat the artifacts generated in CI as correct (since #6 resolves differences between ME-ICA and tedana in any one environment). Longer term, though, it might be better to have docker run non-interactively by default and see if this alleviates cross-platform differences.

emdupre:
Implemented the described short term solution in #19 ! Just in case anyone is wondering why tests pass, now. But they pass ! 🎉 Getting this resolved next would be great.

emdupre:
This seems to actually be a more general issue than the interactivity of the docker container. It's likely related to #14 (or some other, related step where host OS information might be used in setting internal variables-- random seeds just seem like a possible choice). Renaming the issue to reflect this!

emdupre:
Looking through Alexandre Gramfort's recent [ICML presentation on reproducible ML](https://www.slideshare.net/agramfort/icml-2018-reproducible-machine-learning-a-gramfort), I came across something which likely explains this issue (though it doesn't provide a solution): specifically, ICA results are [known to differ](https://image.slidesharecdn.com/icmlrepromlgramfort2018-180714064427/95/icml-2018-reproducible-machine-learning-a-gramfort-5-638.jpg?cb=1531550937) across numerical libraries, i.e., standard BLAS, OpenBLAS, and MKL. 

I'm not thrilled that we have to just accept that our result will fundamentally differ based on OS, but this will likely become less of a concern as we continue to increase our unit tests and reduce reliance on integration testing. 

Knowing this, I'm fine to close this issue. Please LMK if you disagree @tsalo @rmarkello @handwerkerd !

rmarkello:
Thanks for sharing the presentation! Does this also explain why the results differed even when run in an e.g., Docker container? I am on board with the idea that the different numerical libraries would generate variable results, but am not sure why containers would fail to remedy such an issue. Perhaps this is just my naivety about how containers operate!

emdupre:
This is something I'm also new to, but it looks as though [numerical libraries are optimized to the host](https://stackoverflow.com/questions/27919866/docker-images-with-architecture-optimisation).

PR #44, opened 2018-05-13
[ENH] update ICA to sklearn from mdp

emdupre:
Pending discussion in #14.

This is a **breaking change**, since the implementation of FastICA is slightly different across [MDP](http://mdp-toolkit.sourceforge.net/) and [sklearn](http://scikit-learn.org/stable/).


tedana/decomposition/eigendecomp.py
```diff
@@ -216,14 +214,36 @@ def tedica(n_components, dd, conv, fixed_seed, cost, final_cost):
...
+
+        Parameters
+        ----------
+        cost:  {'tanh', 'pow3', 'gaus', 'skew'} str
+        """
+        def gaus(x):
+            return x * np.exp(-1 * x**2 / 2)
+
+        def skew(x):
+            return x**2
```
tsalo: According to the [FastICA documentation](http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.FastICA.html), the cost functions need to return the derivative too.

tedana/decomposition/eigendecomp.py
```diff
@@ -216,14 +214,36 @@ def tedica(n_components, dd, conv, fixed_seed, cost, final_cost):
...
+
+        if cost is 'tanh':
+            return 'logcosh'
+        if cost is 'pow3':
+            return 'cube'
+        if cost is 'gaus':
+            return gaus
+        if cost is 'skew':
+            return skew
```
tsalo: Why not change `tanh` to `logcosh` and `pow3` to `cube` here? It's a breaking change anyway, right?
Also, can you add an exception if `cost` is none of those?

tedana/decomposition/eigendecomp.py
emdupre: Do you think we should support passing a non-standard cost function ? This was not supported in `mdp`, although there were more cost functions available there....

tedana/workflows/tedana.py
```diff
@@ -159,8 +157,12 @@ def tedana(data, tes, mixm=None, ctab=None, manacc=None, strict=False,
         n_components, dd = decomposition.tedpca(catd, OCcatd, combmode, mask, t2s, t2sG,
                                                 stabilize, ref_img,
                                                 tes=tes, kdaw=kdaw, rdaw=rdaw, ste=ste)
+<<<<<<< HEAD
+        mmix_orig = decomposition.tedica(n_components, dd, conv, fixed_seed, cost=cost)
+=======
         mmix_orig = decomposition.tedica(n_components, dd, conv, fixed_seed, cost=initcost,
                                          final_cost=finalcost, verbose=debug)
+>>>>>>> upstream/master
```
tsalo: I think this might be a mistake from when you resolved merge conflicts.

tsalo:
I don't know if backwards compatibility is a requirement for new versions before `1.0.0`, but could this PR be added to a `0.0.2` or `1.0.0` milestone? I don't think it belongs in `0.0.1`.

tedana/workflows/tedana.py
emdupre: Definitely. I'm about to push a fix for this and several other syntax errors ! Thanks for the heads up 👍

emdupre:
Yes, agreed ! This is a `0.1.0` feature, in my mind&mdash; the main one, I think. 
I don't think I have it on the `0.0.1` milestone, but I'll go ahead and make the `0.1.0` milestone now !

rmarkello:
No big surprise here: from the most recent Circle build it looks like the first file saved out post-ICA is different. It might be worth doing some manual inspection to see *how* different these are, and if it's to a tolerance we're comfortable with...

emdupre:
Yes, agreed. @prantikk specifically asked that we 
> try to make sure final SNR and contrast stay comparable. 

which I think is a good starting point for manual inspection !

emdupre:
I've also updated the testing suite to not create a python environment on top of an existing python environment 🤦‍♀️ This _might_ help with #18

emdupre:
Looking at a voxel-wise level I'm seeing very similar values in the `dn_ts_OC.nii`, which makes me pretty confident with this PR. I'd love another review, though, since I also added in a pretty big change to the testing environment !

emdupre:
sklearn_v_mdp

Thanks to @rmarkello, we have a direct comparison of sklearn vs mdp ICA in the test dataset. As you can see, the created components don't fall exactly along the diagonal between the two methods, but they are close (particularly until component 80 or so). Working off this, I'm planning to merge in these changes this week unless @handwerkerd, @tsalo or anyone else voices a concern !

KirstieJane:
Sorry - I’m doing this on my phone so not actually reviewing the code - is this going to get rid of the option to run analyses with mdp ICA? Does it replace it completely w sklearn or introduce an option?

emdupre:
The idea was to switch entirely -- we'd be trading one (largely unmaintained) dependency for another (actively developed) one. Do you think we should support both, @KirstieJane ?

KirstieJane:
Ok! @emdupre and I had a chat on gitter (💖gitter💖) and I think I’m up to date.

I have reservations about breaking the link to MEICA as processed using AFNI, I’m nervous that it will make it harder for labs that have been using MEICA to switch to tedana🎉 because they won’t know exactly how big the changes are. Once this is merged the results will never be the same again 🙀 

BUT, I do understand two points that a) different component selection criteria in different versions of MEICA likely make a greater difference than the choice of ICA decomposition and that b) at some point we need to move from an unsupported library (mdp) to a much better maintained standard library (sklearn).

I think the plan in gitter is for us to keep this conversation going, so anyone else who is swinging by: please add your thoughts!

handwerkerd:
I have no theoretical problem making a breaking change by switching the PCA & ICA algorithms. On the plus side, the slowness of the current ICA and the frequent tweaking of parameters to get convergence are a byproduct of the current algorithm. ICA has some randomness in it anyway so I don't think the differences are a bit deal. To decrease this concern, have you tried running each on a dataset with multiple random seeds? I predict that you'll see a similar range of variation across seeds as you will across algorithms. There might be slightly more variance across algorithms, but the the normal variation across seeds might make the differences in the above matrix look less worrisome. (This is just an educated guess, so please test!)

That said, the one part of the ICA step that does concern me is that we're not doing a traditional component # and selection process. The original and new algorithms are calculating kappa & rho values on the PCA components and using that information to rank the PCA components and decide which PCA components are used for running ICA. This is an interesting concept, but I've had more than a few datasets where I've worried it's doing some unintended harm. Given how lightly commented this section is I do have concerns that a different PCA might interact with this selection process differently. Just to help others get oriented,  this PCA selection starts at: 
https://github.com/emdupre/tedana/blob/44ef0fafb164d35a9d56ec183f4f3d49df70bdce/tedana/decomposition/eigendecomp.py#L161
`ctb` includes Kappa and Rho values which are thresholded in various ways with a few other criteria to go into: https://github.com/emdupre/tedana/blob/44ef0fafb164d35a9d56ec183f4f3d49df70bdce/tedana/decomposition/eigendecomp.py#L219 
Which is used to select the PCA components to keep. This is essentially another selcomps function. Since many of these thresholds are arbitrary, if there are any systematic changes in the PCA results, I'm slightly worried that this might alter the PCA selection process.

Instead of focusing on what the PCA or ICA components precisely look like, I'd be happy to see that the total variance kept in the PCA is very similar for each algorithm and, ideally, that a recreated volume using all the kept PCA (or ICA) components is nearly identical. I'm not sure this is necessary, but you can take his one step further and compare the denoised time series using the different ICAs. If those are nearly identical, even if the accepted or rejected sources of variance are parsed slightly different across components, I don't think anyone would hesitate adapting over this issue.

handwerkerd:
I'll also add that it would be helpful to add more inline comments about the selection process in the tedpca function. This isn't essential before the PR is done, but it might it might help identify any potential issues.

emdupre:
PCA is actually already done using `sklearn` and would therefore not be updated in this. I do agree that the current PCA component selection process could be re-visited, and I've opened #101 to continue that discussion.

Regarding the ICA, though, I'd be happy to re-run the results across a few different seeds and look at the resulting variance ! Will report back.

tsalo:
Now that we've committed to merging this, I wanted to check in about it. Other than dealing with the conflicts, is there anything that needs to be done before this one can be merged?

emdupre:
Thanks for checking in on this, @tsalo !! I think I just need to resolve the conflicts (which I'm happy to do later today !), but I wanted to give a little more time for feedback on the roadmap.

Also, I realized I never did the promised comparison :disappointed: Do you think it would still be useful to do ?

tsalo:
I agree that varying the seed will probably result in variability on par with the differences between the `mdp` and `sklearn` implementations. I personally don't think we need those comparisons to merge. We'll need to run the new version and inspect the results when we update the integration tests anyway, right?

tedana/workflows/tedana.py
```diff
@@ -337,11 +327,10 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
                                                 t2s, t2sG, stabilize, ref_img,
                                                 tes=tes, kdaw=kdaw, rdaw=rdaw,
                                                 ste=ste, wvpca=wvpca)
-        # Perform ICA on dimensionally reduced data (*without* thermal noise)
-        mmix_orig, fixed_seed = decomposition.tedica(n_components, dd, conv, fixed_seed,
-                                                     cost=initcost, final_cost=finalcost,
-                                                     verbose=debug)
+        mmix_orig = decomposition.tedica(n_components, dd, conv, fixed_seed,
```
tsalo: The updated version of tedica is still returning `fixed_seed`, so you'll need to grab that here as well.

```suggestion
        mmix_orig, fixed_seed = decomposition.tedica(n_components, dd, conv, fixed_seed,
```

emdupre: That was the whole merge error !! Thanks for catching it :tada:

tsalo:
I know we've probably discussed this before, but I can't see where- Does it matter that FastICA whitens the data?

emdupre:
MDP was already whitening the data -- we had not supplied a `whitened` param, so [by default we were whitening](https://github.com/mdp-toolkit/mdp-toolkit/blob/511447ae72fc93bbf5da9028fe696c45a1a89a95/mdp/nodes/ica_nodes.py#L95) the data !

tsalo:
Is it okay that MDP was doing it too though?

tsalo:
If the whitening is performed by FastICA, and we aren't using a decision tree to select PCA components during that whitening stage, do we still need the PCA step?

emdupre:
@tsalo and I chatted off-line and got a better handle on this point (since it _is_ quite confusing !).

We do need the PCA step. The reason this is confusing is because if we allow for whitening within the ICA we are introducing a _second_ PCA, which makes the first seem redundant. But it's not ! The first PCA allows us to dimensionally reduce the data, by taking dimensions that meet some criteria -- we can (and I think should !) keep discussing what those criteria are in #101.

But the second PCA, performed inside the `fastICA` call, does not dimensionally reduce the data. Instead, it just orthogonalizes components which can then be statistically whitened to help the ICA converge.

There's [a CrossValidated answer making exactly this point](https://stats.stackexchange.com/questions/97704/does-ica-require-to-run-pca-first), and might be a useful reference !

emdupre:
Merging this, since #151 is merged in ! Thanks everyone for all of your feedback, here !

Issue #48, opened 2018-05-14
Enable text wrapping in RTD tables


tsalo:
Currently, long lines in tables on the RTD site require horizontal scrolling, but we want to just have them wrap around to new lines. The most obvious way to do this is to override the theme by uncommenting [these lines](https://github.com/ME-ICA/tedana/blob/d7807548871348993d7daf849e7499bbdb64cd88/docs/conf.py#L121-L125), but this also removes the RTD theme. The solution appears to be described [here](https://github.com/rtfd/readthedocs.org/issues/2116). It just needs to be implemented in the tedana documentation code.

emdupre:
Closed with #55

PR #55, opened 2018-05-15
[DOC] css wrap

emdupre:
Fix for #48


docs/conf.py
```diff
@@ -67,17 +59,18 @@
 
 # General information about the project.
 project = 'tedana'
-copyright = '2017-2018, Elizabeth DuPre'
-author = 'Elizabeth DuPre'
+copyright = '2017-2018, tedana developers'
+author = 'tedana developers'
```
KirstieJane: Do we have a list of tedana developers somewhere? Would you like me to make an issue to discuss where this should go, and what the criteria for inclusion should be?

emdupre: That would be great ! Yes, it was listed as just me, and I think that's not quite right, thankfully 😄

Issue #59, opened 2018-05-17
Implement checksums for CircleCI integration tests


rmarkello:
The current mechanism for integration testing with CircleCI is to run the full `tedana` pipeline and then compare the content of the output files to previously generated files using [`np.allclose`](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.allclose.html). This procedure ensures that the results of the `tedana` pipeline haven't changed in any appreciable way, which is critical as refactoring and updating occurs.

I was recently inspired by @WhitakerLab's [BrainNetworksInPython](https://github.com/WhitakerLab/BrainNetworksInPython) repository, however, which uses Python's [`hashlib`](https://docs.python.org/3/library/hashlib.html) library to implement [checksum comparisons](https://en.wikipedia.org/wiki/Checksum) of their [generated outputs](https://github.com/WhitakerLab/BrainNetworksInPython/blob/master/tests/write_fixtures.py). I think this procedure could be applied with great success to the `tedana` repository!

Rather than having CircleCI re-download the previously generated outputs from Dropbox each time, they could be hashed once and kept in a JSON file in the `tests/data` folder. The outputs generated during testing could then be hashed and compared against this reference. This would also make testing locally a lot easier (no need to download the 1.9GB output directory!).

KirstieJane:
Eeeeeexcellent - I’m glad that’s helpful! I’m tagging @islast because she did all that work so she could potentially pass on any advice etc :sparkles:

Islast:
I'm thrilled that you like our method @rmarkello. 
My warning is that because hashing doesn't allow you to use tolerances as `np.close` does you have to be very delicate about random states (unless `tedana` is deterministic). We have a different hash for each environment we test in.
This is manageable but it often feels like our continuous integration is more affective at alerting us to dependency updates interfering with the random state or the formatting of written outputs than it is at checking for behavioural changes in BrainNetworksInPython.

rmarkello:
This point is well met, @Islast! Thank you so much for raising it (and thanks for pinging her in, @KirstieJane!).

While `tedana` is not deterministic, we do try to ensure consistent seeding for test purposes. However, you're completely right: given the sheer number of floating point operations that occur on the data in `tedana`, we might need `numpy`'s ability to set tolerance thresholds!

[...]

Issue #50, opened 2018-05-14
Modularize functions for easier testing


rmarkello:
I ran into an issue when writing a few of the tests in `test_utils.py` that I think is only going to get worse as we encourage writing new tests; namely, we assume that all the functions in the repo are strung together as in `tedana.workflows.tedana`. That is, we assume (1) the data are loaded (`tedana.utils.load_data`), (2) a mask is made (`tedana.utils.make_adaptive_mask`), (3) the optimal combination is created (`tedana.model.make_optcom`), (4) the T2* and S0 map are generated (`tedana.model.fit_decay`), and so on and so forth.

For the purposes of testing, this ends up becoming incredibly cumbersome because if I want to test, say, `tedana.model.fit_decay`, I need to perform the first three steps to get the appropriate input data to do so. As we move down the workflow, the individual unit tests are going to get rather bulky&mdash;we'll basically have to run the entirety of `tedana.workflows.tedana` to test, for example, `tedana.select.selcomps`.

It might be worth considering modularizing the functions a bit more to ease this burden. That is, maybe we can adopt a `nilearn._utils.check_niimg` style convention, where, regardless of what data is passed to a given function, we check that is in the appropriate shape (samples x echos x time) and then do all the necessary transformations on it within the function of interest (i.e., make a mask, generate the T2* map), rather than requiring that users supply all the various data beforehand. This still won't necessarily work for e.g., `tedana.select.selcomps`, but in that case we might consider pre-generating some of the input data (rather than going from the "raw" echos).

tsalo:
@rmarkello Regarding conversion/masking of data to the samples x echoes x time shape, do you think that those are functions that should be incorporated into `nilearn` (e.g., `check_niimg_me`, `apply_mask_me`, and `unmask_me`)?

rmarkello:
I don't think any of this is particularly suitable for `nilearn`, given the relatively limited use case... But perhaps that's just my perspective on what `nilearn`'s functionality is supposed to be.

The issue we're running into here is more that we're not putting *any* restrictions on the type of data users are supplying (i.e., it could be a z-concatenated nifti, it could be three niftis, it could be three giftis), so we need to have a function (currently, `tedana.utils.load_data`) to handle that. I think it might be beneficial to perhaps split that into `check_data` and `load_data` functions; the former could be used similarly to `nilearn._utils.check_niimg`, and the latter would be called if data needed to actually be loaded / coerced into the right format.

[...]

Issue #25, opened 2018-05-02
Refactor `t2smap`


emdupre:
In #23, a `fit` function was added to allow for estimation of a T2*/S0 time series. This shares a lot of code with the original `t2sadmap` function, since it's a bit monolithic in its current implementation. It would be great to refactor these two functions so we have less duplicated code !

TomMaullin:
Hi @emdupre, I would be happy to give this a go! Do you have any testing functions or example inputs I could use to verify I'm not changing the function of the code?

TomMaullin:
Hi @emdupre , I am a Mozsprint participant and would be happy to give this a go! Do you have any tests or example data I could use to check I haven't affected the functionality of the code?

emdupre:
That would be fantastic, @TomMaulin ! ☀️

Unfortunately we don't yet have unit tests, here, but that would be another great addition !! There is continuous integration on the repository, which is slower (as it runs the whole code base) but will confirm if your changes affect the outputs. Please let me know if I can answer any other questions !

tsalo:
I don't think that the CI will catch errors in the `fit` method though, right? I don't think it's actually tested by anything at the moment.

emdupre:
Oh, sorry, I missed your question about example data ! The file is too big to store on github, so we're grabbing it off Dropbox [here](https://www.dropbox.com/s/ljeqskdmnc6si9d/zcat_ffd.nii.gz?dl=0).

tsalo:
Related to this, now that there's a submodule called `tedana.model.fit`, I think having the FIT method named `tedana.model.t2smap.fit` is a little confusing. I only used FIT because that's how it's referred to in the Power multi-echo motion paper, but it's not the most informative name. Are there any good alternatives?

tsalo:
What about naming the two methods similarly? I'm not a huge fan of the names `t2sadmap` and `t2smap`, both because they're not very Pythonic and because I'm still not sure what `t2sadmap` means (`t2smap` is a bit clearer), but whatever it is, the FIT method can just be that plus `_ts` (e.g., `ts2admap_ts`).

What do you think about `fit_decay` or `fit_me_decay` and `fit_decay_ts`/`fit_me_decay_ts`? Or, if brevity's not important, `fit_monoexponential_decay` and `fit_monoexponential_decay_ts`?

emdupre:
[...]

PR #61, opened 2018-05-19
[TST] Add tests for tedana.model.monoexponential

tsalo:

- Add test for tedana.model.monoexponential.fit_decay that compares results to results used for integration test of full workflow.
- Add smoke test for tedana.model.monoexponential.fit_decay_ts that just runs the function and checks that the resulting numpy arrays are in the right shape.

### DIFF
--- a/tedana/model/monoexponential.py
+++ b/tedana/model/monoexponential.py
@@ -27,17 +27,17 @@ def fit_decay(data, tes, mask, masksum, start_echo):
 
     Returns
     -------
-    t2sa : (S x E) :obj:`numpy.ndarray`
+    t2sa : (S) :obj:`numpy.ndarray`
         Limited T2* map
-    s0va : (S x E) :obj:`numpy.ndarray`
+    s0va : (S) :obj:`numpy.ndarray`
         Limited S0 map
-    t2ss : (S x E) :obj:`numpy.ndarray`
+    t2ss : (S) :obj:`numpy.ndarray`
         ???
-    s0vs : (S x E) :obj:`numpy.ndarray`
+    s0vs : (S) :obj:`numpy.ndarray`
         ???
-    t2saf : (S x E) :obj:`numpy.ndarray`
+    t2saf : (S) :obj:`numpy.ndarray`
         Full T2* map
-    s0vaf : (S x E) :obj:`numpy.ndarray`
+    s0vaf : (S) :obj:`numpy.ndarray`
         Full S0 map
 
     Notes
@@ -122,13 +122,13 @@ def fit_decay_ts(data, mask, tes, masksum, start_echo):
 
     Returns
     -------
-    t2sa_ts : (S x E x T) :obj:`numpy.ndarray`
+    t2sa_ts : (S x T) :obj:`numpy.ndarray`
         Limited T2* map
-    s0va_ts : (S x E x T) :obj:`numpy.ndarray`
+    s0va_ts : (S x T) :obj:`numpy.ndarray`
         Limited S0 map
-    t2saf_ts : (S x E x T) :obj:`numpy.ndarray`
+    t2saf_ts : (S x T) :obj:`numpy.ndarray`
         Full T2* map
-    s0vaf_ts : (S x E x T) :obj:`numpy.ndarray`
+    s0vaf_ts : (S x T) :obj:`numpy.ndarray`
         Full S0 map
     """
     n_samples, _, n_trs = data.shape
--- a/tedana/model/monoexponential.py
+++ b/tedana/model/monoexponential.py
@@ -73,7 +73,7 @@ def fit_decay(data, tes, mask, masksum, start_echo=1):
         x = np.column_stack([np.ones(echo), [-te for te in tes[:echo]]])
         X = np.repeat(x, n_vols, axis=0)
 
-        beta = np.linalg.lstsq(X, B, rcond=None)[0]
+        beta = np.linalg.lstsq(X, B)[0]
         t2s = 1. / beta[1, :].T
         s0 = np.exp(beta[0, :]).T
 
--- a/tedana/model/monoexponential.py
+++ b/tedana/model/monoexponential.py
@@ -31,9 +31,9 @@ def fit_decay(data, tes, mask, masksum, start_echo=1):
         Limited T2* map
     s0va : (S) :obj:`numpy.ndarray`
         Limited S0 map
-    t2ss : (S, ?) :obj:`numpy.ndarray`
+    t2ss : (S, E-1) :obj:`numpy.ndarray`
         ???
-    s0vs : (S, ?) :obj:`numpy.ndarray`
+    s0vs : (S, E-1) :obj:`numpy.ndarray`
         ???
     t2saf : (S) :obj:`numpy.ndarray`
         Full T2* map
@@ -73,7 +73,7 @@ def fit_decay(data, tes, mask, masksum, start_echo=1):
         x = np.column_stack([np.ones(echo), [-te for te in tes[:echo]]])
         X = np.repeat(x, n_vols, axis=0)
 
-        beta = np.linalg.lstsq(X, B)[0]
+        beta = np.linalg.lstsq(X, B, rcond=None)[0]
         t2s = 1. / beta[1, :].T
         s0 = np.exp(beta[0, :]).T
 
--- a/tedana/model/monoexponential.py
+++ b/tedana/model/monoexponential.py
@@ -5,7 +5,7 @@ import numpy as np
 from tedana import utils
 
 
-def fit_decay(data, tes, mask, masksum, start_echo):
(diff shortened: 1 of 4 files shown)


tsalo:
The Travis coverage tests cannot find `/home/neuro/data/zcat_ffd.nii.gz` in `test_model_exponential.nii.gz`, but I'm not sure why.

emdupre:
The travis environment doesn't have that file, only the circle ! We're not yet shipping with a downsampled `zcat` file&mdash; we could fake one by concatenating all of the downsampled echos, or you could use / downsample the averaged `zcat` volume included in circle.

emdupre:
Right now we're having [travis run the linting and unit tests](https://github.com/ME-ICA/tedana/blob/master/.travis.yml#L34), and [circle run the integration test](https://github.com/ME-ICA/tedana/blob/master/circle.yml#L40) (i.e., [`test_tedana`](https://github.com/ME-ICA/tedana/blob/master/tedana/tests/test_tedana.py)). 

The idea was that unit tests are usually quick, and our integration test clocks around half an hour. My worry is that we'll overload the circle testing environment and end up with two hours of testing there and only linting on travis, which will increase the overall testing time than if we share resources a little more effectively.

Basically, this is totally a judgement call ! We're early enough along in testing that I don't think we're going to get near two hours either way, so the current situation might be over optimizing. We can also use the cache on circle a little more effectively than we are right now (see #59). WDYT, @tsalo ?

tsalo:
That makes sense and I'm happy to follow the current convention. Still, we are going to need data for unit tests as well, right?

emdupre:
Absolutely ! We uploaded some downsampled [individual echo data](https://github.com/ME-ICA/tedana/tree/master/tedana/tests/data) so it could be shipped with the tests. I'd probably work off the [`zcat` file](https://github.com/ME-ICA/tedana/blob/master/circle.yml#L18) and add something new to the data folder for these tests&mdash; with something like

```
import numpy as np
from nilearn import image

file = 'zcat_ffd.nii.gz'
target_aff = np.eye(4) * 4
image.mean_img(file, target_aff).to_filename("avg_zcat_ffd.nii.gz")
```

Or, if you want to provide some sort of timeseries, we could use `image.index_img` instead ! WDYT ?

tsalo:
Everything is passing now, so I think it's ready for review.

tedana/tests/test_model_monoexponential.py
```diff
@@ -0,0 +1,43 @@
...
+from tedana.model import monoexponential as me
+from tedana.tests.utils import get_test_data_path
+
+
+def test_fit_decay():
+    tes = np.array([14.5, 38.5, 62.5])
+    in_files = [op.join(get_test_data_path(), 'echo{0}.nii.gz'.format(i+1))
+                for i in range(3)]
+    data, _ = utils.load_data(in_files, n_echos=len(tes))
+    mask, mask_sum = utils.make_adaptive_mask(data, minimum=False, getsum=True)
```
rmarkello: Is there a way to refactor this to avoid duplicating the same 5 lines (`tes = ...` through `mask, mask_sum = ...`)? Options include making them globals or creating a class structure with those lines in the `__init__` and having `test_fit_decay` and `test_fit_decay_ts` as methods.

Indeed, this goes back to #50... Having to repeat these steps for testing everything seems wasteful / redundant.

tedana/tests/test_model_monoexponential.py
```diff
@@ -0,0 +1,43 @@
...
+    t2sv, s0v, t2ss, s0vs, t2svG, s0vG = me.fit_decay(data, tes, mask,
+                                                      mask_sum, start_echo=1)
+    assert len(t2sv.shape) == 1
+    assert len(s0v.shape) == 1
+    assert len(t2ss.shape) == 2
+    assert len(s0vs.shape) == 2
+    assert len(t2svG.shape) == 1
+    assert len(s0vG.shape) == 1
```
rmarkello: This is partially preference, but I think `assert data.ndim == X` is a bit more parsimonious than `assert len(data.shape) == X`. I'm open to disagreement here, though! @emdupre do you have thoughts?

emdupre: I think you're probably right, @rmarkello, that `.ndim` is more elegant (and, importantly, I think equally easy to read !). Would you mind updating, @tsalo ?

tedana/tests/test_model_monoexponential.py
emdupre: Oof. This might be outside the scope of this PR, since #50 is such an extensive issue. 

Personally, I'd rather repetition than re-introducing globals at this point&mdash; I know that their usage is less problematic here, but we just went through so much trouble to get rid of them ! It's mostly my being dogmatic, but still.

@tsalo, if you can see a clean way to cut down on the repetition, that would be great. If you want to postpone tackling this until we modularize the functions (much) better, I'm fine with that, too.

tedana/model/monoexponential.py
```diff
@@ -22,22 +22,22 @@ def fit_decay(data, tes, mask, masksum, start_echo):
...
     Returns
     -------
-    t2sa : (S x E) :obj:`numpy.ndarray`
+    t2sa : (S) :obj:`numpy.ndarray`
         Limited T2* map
-    s0va : (S x E) :obj:`numpy.ndarray`
+    s0va : (S) :obj:`numpy.ndarray`
         Limited S0 map
-    t2ss : (S x E) :obj:`numpy.ndarray`
+    t2ss : (S, ?) :obj:`numpy.ndarray`
```
emdupre: Is it unclear what the returned dimensions are, here ?

emdupre:
Rather than dropping the `rcond` argument, could we instead force the travis environment to use an up-to-date numpy ? Maybe [here](https://github.com/ME-ICA/tedana/blob/master/.travis.yml#L18) with `pip install --upgrade requirements.txt`

tedana/model/monoexponential.py
rmarkello: Yes, I'd noticed that too when going through the code a while back. You'll notice that actually overwrites itself; since it's `range(start_echo, n_echos + 1)` when it gets to n_echos it will index the same location. Completely bonkers and I just didn't have the headspace to figure out *why* it was happening...

tedana/tests/test_model_monoexponential.py
tsalo: Good catch @rmarkello. I'll just add a test-specific pytest fixture to load data.

rmarkello: That's a great solution!!!

tedana/model/monoexponential.py
emdupre: You're right, and I think that's incorrect behavior. Going back to the [original code](https://github.com/ME-ICA/me-ica/blob/master/meica.libs/t2smap.py#L235), zero should be the first value. Nice catch !

emdupre:
We've been testing on `numpy >= 1.14`. It looks like that's the minimum, since otherwise we hit this error !

tedana/model/monoexponential.py
rmarkello: I think the `start_echo` kwarg was added in the first place to address this discrepancy (but you did that, @emdupre, so you would know better than any of us!). That being said, it might be worth digging into *why* they were ever different. For now we should retain this nonsense functionality so that `tedana.workflows.tedana()` is the same. Indeed, `tedana.workflows.t2smap()` [handles the discrepancy correctly](https://github.com/ME-ICA/tedana/blob/master/tedana/workflows/t2smap.py#L48).

emdupre: Yes, I'm remembering (slowly) that I added the kwarg there to try and align them. @rmarkello, are you saying we should keep the bug ? I guess this is something we could address in #25 ... we should make a note on that issue, if so.

rmarkello: No, no, we shouldn't keep it, you're right. But I do think it should potentially be addressed in #25 / a separate PR, just to retain some semblance of order as we try and make PRs more feature-based and less gigantic overhauls of the code.

To @KirstieJane: thank you! Your continued support for all things `tedana` is :100: :blush: :+1:

emdupre: I think that's a reasonable course of action, @rmarkello ! @tsalo, would you mind adding that issue to #25 so we maintain a record ?

Thanks everyone 💖

tsalo:
Sorry for the delay, but I've updated #25 with information about the discrepancy between the two functions. Please feel free to add to it if I got anything confused.

I've also added the minimum numpy version to the requirements file and have re-added the `rcond` argument to the `lstsq` call, so I think that this PR is ready for review again.

emdupre:
This LGTM ! Thanks so much for updating it, @tsalo.

@rmarkello, any other comments ?

rmarkello:
Looks great! I'm excited to see some coverage increases :smile: :chart_with_upwards_trend:

PR #66, opened 2018-05-24
[DOC] Add citations for Zenodo release to code and docs.

tsalo:

- Add module-level duecredit citation for Zenodo release to tedana/__init__.py
- Add citations for Kundu et al. (2013) and Zenodo release to RTD index page.
- Add instructions for generating citations with duecredit to RTD index page.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -32,7 +32,8 @@ When using tedana, please include the following citations:
   *Zenodo*. doi:10.5281/zenodo.1250562.
 
 Alternatively, you can automatically compile relevant citations by running your
-tedana code with `duecredit`_, e.g.:
+tedana code with `duecredit`_. For example, if you plan to run a script using
+tedana (in this case, ``tedana_script.py``):
 
 .. code-block:: bash
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -18,11 +18,30 @@ Citations
 
 When using tedana, please include the following citations:
 
-  Kundu, P., Inati, S.J., Evans, J.W., Luh, W.M. & Bandettini, P.A. (2011).
+  Kundu, P., Inati, S. J., Evans, J. W., Luh, W. M. & Bandettini, P. A. (2011).
   `Differentiating BOLD and non-BOLD signals in fMRI time series using multi-echo EPI.`_
   *NeuroImage*, *60*, 1759-1770.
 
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -18,11 +18,30 @@ Citations
 
 When using tedana, please include the following citations:
 
-  Kundu, P., Inati, S.J., Evans, J.W., Luh, W.M. & Bandettini, P.A. (2011).
+  Kundu, P., Inati, S. J., Evans, J. W., Luh, W. M. & Bandettini, P. A. (2011).
```
emdupre: Stylistic differences 🤣 This is easier to read, though, for sure !

docs/index.rst
```diff
@@ -18,11 +18,30 @@ Citations
...
+  DuPre, E. M., Salo, T., Markello, R. D., Kundu, P., Whitaker K. J. (2018).
+  `ME-ICA/tedana: Initial tedana release (Version 0.0.1).`_
+  *Zenodo*. doi:10.5281/zenodo.1250562.
+
+Alternatively, you can automatically compile relevant citations by running your
+tedana code with `duecredit`_, e.g.:
+
+.. code-block:: bash
+
+  python -m duecredit tedana_script.py
```
emdupre: `tedana_script.py` ? Would it not just be `tedana` ?

tsalo: I was just using `tedana_script.py` as a stand-in for any script that uses `tedana`. I tried it as `python -m duecredit tedana`, but got an error:
```bash
2018-05-24 22:34:31,198 [ERROR  ] Cannot run file 'tedana' because:
[Errno 2] No such file or directory: 'tedana' (__main__.py:88)
```

I'm not sure how one would use `duecredit` with the CLI tools. Maybe it can be added as an argument and `duecredit` can be called within the Python script somehow?

emdupre: Hmm. In the short term, could you clarify what you mean by `tedana_script.py` ? We could then open an issue about using `duecredit` with CLI&mdash; I think it would be nice to be able to generate citations for `tedana` itself, if possible ! WDYT ?

tsalo: How about something like:

```html
Alternatively, you can automatically compile relevant citations by running your
tedana code with `duecredit`_. For example, if you plan to run a script using
tedana (in this case, ``tedana_script.py``):

.. code-block:: bash

  python -m duecredit tedana_script.py
```

emdupre: That looks great to me 😄

tsalo: 👍

PR #67, opened 2018-05-25
[DOC] Remove installation RTD

### LINKED ISSUE #51: Update installation documentation
The [installation docs](https://tedana.readthedocs.io/en/latest/installation.html) currently list AFNI as a requirement, which is no longer true as as of #40. It would be great to update that!

Moreover, I think that suggesting docker for use/development might be a bit overkill, since `tedana` is now a pure-Python package. Perhaps it would be better to suggest simply installing via `pip install https://github.com/ME-ICA/tedana/archive/master.tar.gz` as suggested in the README? I'm also game to proselytize `conda` (as is currently being done), but regardless I think making it as simple as possible to use `tedana` would probably be best!

emdupre:
Closes #51.

- Removes installation instructions from RTD website


emdupre:
@tsalo I don't _think_ this should cause a merge conflict with #66, but I can merge this after to be safe !

PR #76, opened 2018-06-04
[FIX] Add fname to TED dirname

emdupre:

- Sets a unique name for each TED directory, so the wrong pcastate is not accidentally loaded

### DIFF
--- a/tedana/tests/test_tedana.py
+++ b/tedana/tests/test_tedana.py
@@ -63,4 +63,5 @@ def test_outputs():
         'sphis_hik.nii'
     ]
     for fn in nifti_test_list:
-        compare_nifti(fn, Path('/home/neuro/data/TED/'), Path('/home/neuro/code/TED/'))
+        compare_nifti(fn, Path('/home/neuro/data/TED/'),
+                      Path('/home/neuro/code/TED.zcat_ffd/'))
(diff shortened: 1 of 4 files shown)


tedana/tests/test_tedana.py
```diff
@@ -63,4 +63,5 @@ def test_outputs():
         'sphis_hik.nii'
     ]
     for fn in nifti_test_list:
-        compare_nifti(fn, Path('/home/neuro/data/TED/'), Path('/home/neuro/code/TED/'))
+        compare_nifti(fn, Path('/home/neuro/data/TED/'),
+                      Path('/home/neuro/code/TED.zcat_ffd/'))
```
rmarkello: You should also change the artifact upload process in `circle.yml` to point to `/home/neuro/code/TED.zcat_ffd`

tedana/workflows/tedana.py
```diff
@@ -102,10 +102,15 @@ def tedana(data, tes, mixm=None, ctab=None, manacc=None, strict=False,
 
     kdaw, rdaw = float(kdaw), float(rdaw)
 
+    try:
+        ref_label = os.path.basename(ref_img).split('.')[0]
+    except TypeError:
+        ref_label = os.path.basename(str(data)).split('.')[0]
```
rmarkello: I believe `data` will always be a list, so should you be subsetting it before coercing to `str`?

emdupre: Testing locally, I'm only hitting the `TypeError` statement if we're provided with a non-list data type like `zcat` and the `ref_img` is nifti-like rather than a true string (as described in the `load_data` docstring). This is a good reminder to update that docstring !

tedana/tests/test_tedana.py
emdupre: Done ! Thanks for catching.

Issue #88, opened 2018-07-11
F threshold is rounded down


handwerkerd:
The following lines of code make a binary spatial map based on the F values for the model fit. The threshold is floored using int()
https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/model/fit.py#L201-L202

fmin (the minimum signif F value) is set using an echo-dependent lookup table here:
https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/utils/utils.py#L74
https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/model/fit.py#L99

These thresholds have decimals that are rounded down for no reason that is obvious to me. The threshold parameter is supposed to be a float and takes floats in other calls. I'm assuming this is a relatively minor bug?

PR #90, opened 2018-07-16
[ENH, FIX] Add wavelet denoising and fix t2smap workflow

### LINKED ISSUE #80: Add option for wavelet denoising
Wavelet denoising code did exist in tedana [at some point](https://github.com/ME-ICA/me-ica/blob/da6ac4c23a59145530a16bd6f25c676460fe9436/meica.libs/tedana.py#L514-L518), but was never turned on. We probably need to test the code before re-incorporating it, but it does seem like a useful option to have.

tsalo:
Closes #80.

- Re-implements wavelet denoising from an older version of [MEICA](https://github.com/ME-ICA/me-ica/blob/da6ac4c23a59145530a16bd6f25c676460fe9436/meica.libs/tedana.py#L514-L518), with minor changes to work with pywt's current API.
    - Adds `wvpca` argument to `tedana.decomposition.eigendecomp.tedpca`, `tedana.workflows.tedana`, and the `tedana` workflow CLI.
- Fixes the `t2smap` workflow, which hadn't been updated to work with the recent changes to the package.
    - Also adds a new argument, `fitmode`, which can fit the mono exponential model per timepoint if desired.


tedana/decomposition/_utils.py
```diff
@@ -40,3 +41,51 @@ def eimask(dd, ees=None):
...
+    Parameters
+    ----------
+    mmix : :obj:`numpy.ndarray`
+        Data to wavelet transform.
+
+    Returns
+    -------
+    mmix_wt : :obj:`numpy.ndarray`
+        Wavelet-transformed data.
+    cAlen : :obj:`int`
```
emdupre: Is there a reason to call it `cAlen` here and `cAl` in `idwtmat()` ?

tedana/decomposition/_utils.py
```diff
@@ -40,3 +41,51 @@ def eimask(dd, ees=None):
...
+        Index of some kind?
+    """
+    llt = len(np.hstack(pywt.dwt(mmix[0], 'db2')))
+    mmix_wt = np.zeros([mmix.shape[0], llt])
+    for ii in range(mmix_wt.shape[0]):
+        wtx = pywt.dwt(mmix[ii], 'db2')
```
emdupre: From the [pywavelets documentation](http://pywavelets.readthedocs.io/en/latest/regression/dwt-idwt.html#discrete-wavelet-transform), it seems that `wtx` would be the approximation coefficients for the specified row of `mmix`. Then `cAlen` would be the length of the first entry of `wtx` in the last for-loop.

If you agree with that assessment, could we improve these variable names ?

tedana/decomposition/_utils.py
```diff
@@ -40,3 +41,51 @@ def eimask(dd, ees=None):
...
+
+def idwtmat(mmix_wt, cAl):
+    """
+    Invert wavelet transform data.
+
+    Parameters
+    ----------
+    mmix_wt : :obj:`numpy.ndarray`
+        Wavelet-transformed data.
+    cAl : :obj:`int`
```
emdupre: See previous comments ^

tedana/workflows/t2smap.py
```diff
@@ -21,48 +22,80 @@ def t2smap(data, tes, combmode='t2s', label=None):
...
+
+    # set a hard cap for the T2* map/timeseries
+    # anything that is 10x higher than the 99.5 %ile will be reset to 99.5 %ile
+    cap_t2s = stats.scoreatpercentile(t2s_limited.flatten(), 99.5,
+                                      interpolation_method='lower')
+    LGR.debug('Setting cap on T2* map at {:.5f}'.format(cap_t2s * 10))
+    t2s_limited[t2s_limited > cap_t2s * 10] = cap_t2s
 
     LGR.info('Computing optimal combination')
-    tsoc = np.array(model.make_optcom(catd, t2s, tes, mask, combmode),
```
emdupre: Can you explain these changes for me ?

emdupre:
Also, it looks like this may partially address #25 ?

tedana/workflows/t2smap.py
tsalo: Most of these changes are just taken from the current version of the `tedana` workflow, except for changing the output variable names from the `fit_decay` functions, which I did because I think the new ones are more informative. I also dropped the part where it writes out the `t2ss` and `s0vs` maps because `fit_decay_ts` doesn't return those and I honestly have no clue what they are.

Are there any changes in particular that don't make sense?

tedana/decomposition/_utils.py
tsalo: I just copied the functions from the old version of tedana and removed the `correct_size` argument in [`idwtmat`](https://github.com/ME-ICA/me-ica/blob/da6ac4c23a59145530a16bd6f25c676460fe9436/meica.libs/tedana.py#L480), which `pywt` must have dropped at some point.

tedana/decomposition/_utils.py
tsalo: I think it's the approximation and detail coefficients, not that I'm sure what those are. `cAlen` is the number of coefficients (i.e., `len` of `cA`). Oh, what about `n_coefs_approx` instead of `cAlen`, `coefs_tup` instead of `wtx`, `n_coefs_total` instead of `llt`, `mmix_coefs` instead of `mmix_wt`?

What does `mmix` stand for again?

tsalo:
The tests are now passing, and there are several working (though basic) tests for the t2smap workflow. Yay increased coverage!

I guess it does address #25 a bit, though my changes are mostly to the workflow rather than the fitting functions.

I'm not sure whether the wvpca works or not, but other than that I think it looks good. WDYT?

tsalo:
Oh and I did notice two situations where `model.optcom` fails:
1. If the limited T2* map is used (how it was written, but not how tedana is) instead of the full one (`combmode='t2s'` and `fitmode='all'`, both of which are the defaults)
2. If the full T2* timeseries (how it's hardcoded) is used with `combmode='ste'` and `fitmode='ts'`

tedana/workflows/t2smap.py
emdupre: Yes, sorry, that wasn't a helpful question on my part ! Your explanation solved it anyway, though: I was confused why we were renaming `tsoc` to `OCcatd`.

emdupre:
This is great ! :tada: 

Do you think we should have some testing for the `wvpca` functions (i.e., `dwtmat` and `idwtmat`) ? I know that they're pulled almost directly from `pywt` &ndash; so I'm not sure how much sense a unit test makes &ndash; but now is the time I'd prefer to add them, if we're going to !

Regarding failing options: I had no idea the `@pytest.mark.xfail` decorator existed ! But more importantly: can you explain the first failure a little more for me ? What do you mean "not how tedana is" ?

tsalo:
I can definitely add some simple tests. Something to make sure that the data are returned in the right shape at least. I don't know enough about wavelets to come up with anything better atm.

Honestly, I had to hunt down a way to allow a test to fail. I didn't want to hide this weird condition that was failing but didn't want to have to deal with a continually breaking test, which would throw off the PRs once merged.

By "not how tedana is" I meant that `t2smap` was using the limited t2s map for `optcom` [before](https://github.com/ME-ICA/tedana/blob/ec80e16ec05e6974152778ac6a515ddd6143ccdc/tedana/workflows/t2smap.py#L53), but the [`tedana` workflow](https://github.com/ME-ICA/tedana/blob/ec80e16ec05e6974152778ac6a515ddd6143ccdc/tedana/workflows/tedana.py#L157) was using the full t2smap. So I guess how `t2smap` was set up originally, it would fail (at least on the test data in the repository). These failures could all be entirely dependent on the data for all I know though.

emdupre:
Do you think the failure has anything to do with [this](https://github.com/ME-ICA/tedana/issues/25#issuecomment-391727444) ? I can't find if we ever resolved that.

tsalo:
We continued talking about the differences somewhere (maybe on Gitter?), but I can't find the discussion. Anyway, I don't think it could be the source of the failure, because my current version uses `start_echo=1` in `t2smap` to match the `tedana` workflow.

I just realized that `ste` doesn't use the `t2s` estimates (of course), so dealing with the `alpha` weights for the averaging as if they're 3D (SxExT) (as one would have when using `fit_decay_ts` instead of `fit_decay`) instead of 2D (SxE) (the default) would cause an error.

I am changing how `make_optcom` decides to handle the `alpha` weights, but I also think we should change how `make_optcom` is organized overall, because `t2s` and `combmode` are _kind_ of redundant. At the very least, `t2s` should be an optional argument, since it isn't used when `combmode = 'ste'`. I'm going to commit my proposed changes so you can see what I was thinking, and then we can discuss them in the code comments.

tedana/model/combine.py
```diff
@@ -8,26 +8,27 @@
 LGR = logging.getLogger(__name__)
 
 
-def make_optcom(data, t2s, tes, mask, combmode, verbose=True):
+def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
```
tsalo: @emdupre I think we could probably consolidate these into one argument, but in the interest of clarity two is fine. If we move `t2s` so it's a keyword argument, then it can default to `None` (which works when `combmode` is `ste` but not when it's `t2s`). Then below we check the inputs to make sure they make sense. WDYT?

emdupre: Just to clarify: what would happen when `combmode` is `ste` if we make t2s the one (keyword) argument ? Or am I misunderstanding the suggestion ?

emdupre:
I had the same thought, but I can't find it anywhere ! I was having this same issue with the discussion we had on #88...

I see your point about having `t2s` as a keyword argument -- I think your idea makes good sense ! I am, though, a little confused about the suggestion to reduce it down to one argument; see q above.

For now, I'm happy to get circle passing again, move forward with the two argument idea, and reorganize `optcom` in another PR. I just wasn't sure if we could nail down the pytest failure !

tedana/model/combine.py
emdupre: That makes sense to me, unless we foresee adding other `combmode` options..  @handwerkerd might have thoughts on this !

tsalo:
Whoops, I missed one of the `make_optcom` calls. Hopefully circle will pass now.

Issue #84, opened 2018-07-06
Add component selection documentation


emdupre:
It would be great to have more documentation on what component selection entails. This is only covered in a very general way at this point and could be enormously expanded.

As @jbpoline had suggested, it would be great at some point to have automatically generated jupyter notebooks (using [sphinx-gallery](https://sphinx-gallery.readthedocs.io/en/latest/syntax.html#notebook-styled-examples-syntax)), but to start with I'd love to have a break down of the kappa, rho calculations and how various points in the decision tree impact those metrics. 

One other piece: I know @handwerkerd has some examples from his `meica-reports` -- this section could be a great place to include those, until we generate some of our own !

emdupre:
@jwe080, it'd be great to have your expertise here, too !

jbteves:
Since the reorganization happening in #247 I think this issue should be paused.

jbteves:
With #262 merged I am unpausing this issue.

jbteves:
In line with many of the other issues I've commented on, I think that we should come up with a discrete set of tasks to close this issue. From the above, I could envision something like the following:
- [ ] Add kappa, rho calculations and decisions to readthedocs
- [ ] (Optional) Generate jupyter notebooks

jbteves:
Given the interest in rewriting the decision tree, it seems that this issue is likely going to be solved as new code is written and documented. I think this issue is on probation.

PR #91, opened 2018-07-16
[DOC] Better commenting for component selection algorithm in selcomps

handwerkerd:
Progress towards #84 .

- The first half of the selcomps function is commented in much greater detail. There is more commenting and documenting work to be done in selcomps. Still, since this doesn't affect a single line of code, I figured it is better to submit a PR now, so that others can have easier access to this information.

### DIFF
--- a/tedana/selection/select_comps.py
+++ b/tedana/selection/select_comps.py
@@ -74,9 +74,31 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
     the decision tree for the selection process and other sections that
     are part of the decision tree. Certain comments are prefaced with METRIC
     and variable names to make clear which are metrics and others are prefaced
-    with SELETION to make clear which are for applying metrics. These
+    with SELECTION to make clear which are for applying metrics. METRICs tend
+    to be summary values that containt a signal number per component. These
     distinctions will hopefully make it easier to better modularize this
     function.
+
+    Note there are some variables that are calculated in one section of the code
+    that are later transformed into another metric that is actually part of a
+    selection criterion. This running list is an attempt to summarize
+    intermediate metrics vs the metrics that are actually used in decision
+    steps. For applied metrics that are made up of intermediate metrics defined
+    in earlier sections of the code, the constituent metrics are noted
+
+    Intermediate Metrics:  dice_tbl countnoise
+        counts_FR2_Z tt_table mmix_kurt mmix_std
+        spr fproj_arr_val fdist
+        spz, Rtz, Dz
+
+    Applied Metrices:
+        seldict['Rhos']
+        seldict['Kappas']
+        countsigFS0
+        countsigFR2
+        fz (a combination of multiple z-scored metrics: tt_table,
+            seldict['varex'], seldict['Kappa'], seldict['Rho'], countnoise,
+            mmix_kurt, fdist)
     """
 
     """
@@ -102,7 +124,10 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
     """
     List of components
     nc and ncl start out as an ordered list of the component numbers
-    nc is constant throughout the function. ncl changes based on comp selection
+    nc is constant throughout the function.
+    ncl changes through his function as components are assigned to other
+    categories (i.e. components that are classified as rejected are removed
+    from ncl)
     """
     midk = []
     ign = []
@@ -137,22 +162,23 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
     METRICS: dice_tbl
     dice_FR2, dice_FS0 are calculated for each component and the concatinated
     values are in dice_tbl
-    Br_clmaps_R2 and Br_clmaps_R2 are binarized clustered Z_maps.
+    Br_clmaps_R2 and Br_clmaps_S0 are binarized clustered Z_maps.
     The volume being clustered is the rank order indices of the absolute value
     of the beta values for the fit between the optimally combined time series
     and the mixing matrix (i.e. the lowest beta value is 1 and the highest is
     the # of voxels).
     The cluster size is a function of the # of voxels in the mask.
     The cluster threshold are the voxels with beta ranks greater than
-    countsigFS0 or countsigFR2.
+    countsigFS0 or countsigFR2 (i.e. roughly the same number of voxels will be
+    in the countsig clusters as the ICA beta map clusters)
     These dice values are the Dice-Sorenson index for the Br_clmap and the
     F_??_clmap.
     If handwerkerd understands this correctly, If the voxels with the above
     threshold F stats are clustered in the same voxels with the highest beta
-    values, then the dice coefficient will be 1. If the high F or beta values
+    values, then the dice coefficient will be 1. If the thresholded F or betas
     aren't spatially clustered (i.e. the component map is less spatially smooth)
     or the clusters are in different locations (i.e. voxels with high betas)
-    are also noisers so they have lower F values, then the dice coefficients
+    are also noiser so they have lower F values, then the dice coefficients
     will be lower
     """
     dice_tbl = np.zeros([nc.shape[0], 2])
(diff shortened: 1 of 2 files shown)


tedana/selection/select_comps.py
```diff
@@ -61,10 +61,59 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
+        are NOT removed during the denoising process
     """
 
+    """
+    There are sections of this code that calculate metrics that are used in
+    the decision tree for the selection process and other sections that
+    are part of the decision tree. Certain comments are prefaced with METRIC
+    and variable names to make clear which are metrics and others are prefaced
+    with SELECTION to make clear which are for applying metrics. METRICs tend
+    to be summary values that containt a signal number per component. These
```
emdupre: typo: "containt" --> "contain"

tedana/selection/select_comps.py
```diff
@@ -61,10 +61,59 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
+        These components are typically removed from the data during denoising
     ign : list
         Indices of ignored components in `mmix`
+        Ignored components are considered to have too low variance to matter.
+        They are not processed through the accept vs reject decision tree and
+        are NOT removed during the denoising process
     """
 
+    """
+    There are sections of this code that calculate metrics that are used in
```
emdupre: What do you think of moving these metrics into the sphinx documentation, rather than as code comments ?

tedana/selection/select_comps.py
```diff
@@ -79,28 +128,65 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
+    The cluster threshold is based on the # of echos acquired
     """
     countsigFS0 = seldict['F_S0_clmaps'].sum(0)
     countsigFR2 = seldict['F_R2_clmaps'].sum(0)
     countnoise = np.zeros(len(nc))
 
     """
     Make table of dice values
+    METRICS: dice_tbl
+    dice_FR2, dice_FS0 are calculated for each component and the concatinated
```
emdupre: typo: "concatinated" --> "concatenated"

tedana/selection/select_comps.py
```diff
@@ -189,7 +331,38 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
+
+    fz is matrix of multiple other metrics described above and calculated
+    in this section. Most are all of these have one number per component and
+    they are z-scored across components
+    Attempted explanations in order:
+    Tz: The z-scored t statistics of the spatial noisiness metric in tt_table
+    Vz: The z-scored the natural log of the non-normalized variance explained
+        of each component
+    Ktz: The z-scored natural log of the Kappa values
+    (the '/ 2' doesnot seem necessary beacuse it will be removed by z-scoring)
```
emdupre: typo: "doesnot" --> "does not"

tedana/selection/select_comps.py
```diff
@@ -210,50 +383,98 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
-    KRcut = np.median(KRcutguesses)
 
-    # only use exclusive when inclusive is extremely inclusive - double KRcut
+    # Default to the most inclusive kappa threshold (_cons) unless:
+    # 1. That threshold is more than twice the median of Kappa & Rho thresholds
+    # 2. and the moderate elbow is more inclusive than a p=0.01
+    # handwerkerd: This actually seems like a way to avoid using the theoretically
+    #   most liberal threshold only when there was a bad estimate and _mod is
+    #   is more inclusive. My one concern is that it's an odd way to test that
+    #   the _mod elbow is any better. WHy not at least see if _mod < _cons?
```
emdupre: typo: "WHy" --> "Why"

tedana/selection/select_comps.py
```diff
@@ -210,50 +383,98 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
     cond1 = getelbow_cons(seldict['Kappas']) > KRcut * 2
     cond2 = getelbow_mod(seldict['Kappas'], val=True) < F01
     if cond1 and cond2:
         Kcut = getelbow_mod(seldict['Kappas'], val=True)
     else:
         Kcut = getelbow_cons(seldict['Kappas'], val=True)
-    # only use inclusive when exclusive is extremely exclusive - half KRcut
```
emdupre: Is there a reason to remove this comment ? It might be helpful to have the original explanation, even if we decide that it doesn't make sense (for us).

tedana/selection/select_comps.py
```diff
@@ -264,12 +485,19 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
+    # Throw lower-risk bad components out based on 3 criteria all being true:
+    #  1. tt_table (a spatial noisiness metric) <0
+    #  2. A components variance explains is greater than the median variance
+    #     explained
+    #  3. The component index is greater than the KRcut index. Since the
+    #      components are sorted by kappa, this is another kappa thresholding)
     rejB = ncl[utils.andb([tt_table[ncl, 0] < 0,
                            seldict['varex'][ncl] > np.median(seldict['varex']), ncl > KRcut]) == 3]
     rej = np.union1d(rej, rejB)
+    # adjust ncl again to only contain the remainng non-rejected components
```
emdupre: typo: "remainng" --> "remaining"

tedana/selection/select_comps.py
```diff
@@ -264,12 +485,19 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
     ncl = np.setdiff1d(ncl, rej)
 
+    # This is where handwerkerd has paused in hypercommenting the function.
```
emdupre: :rofl:

emdupre:
Also, I (finally) fixed the circle error. Would you be able to merge in the current master branch ? LMK if I can help in that !

tedana/selection/select_comps.py
handwerkerd: I think that's the eventual goal, but it's a bit tricky when the code is only half-commented. Until there's full documentation of how each metric is or isn't used, making something seem definitive int he sphinx documentation could cause more harm than help. For example, there are some metrics which I thought would be used one way, but they were transformed into another metric & used in more component selection steps than I realized. In that case, saying Metric X is used in selection step Y would be wrong. My hope is, once the full process is better commented in the code, it will be relatively easy to summarize those comments in end-user documentation.

rmarkello:
I just wanted to jump in to say how unbelievable this is, @handwerkerd ! :tada: :100: Having tried to work through pieces of `selcomps()` before I know how cumbersome it can be, so to see this level of detail is really amazing !! This will be a huge boon as we figure out how best to modularize / modify the function. :woman_technologist:

tedana/selection/select_comps.py
handwerkerd: I tried to expand the purpose of that comment since it's not clear how inclusive & exclusive are defined. I felt like the following blocks of comments added more detail and replaced this one. If you think the line I deleted contains unique and relevant info, I can add it back in.
https://github.com/handwerkerd/tedana/blob/4c65baf8873ebbc4965820101b006c2cc061c4b7/tedana/selection/select_comps.py#L399-L409
https://github.com/handwerkerd/tedana/blob/4c65baf8873ebbc4965820101b006c2cc061c4b7/tedana/selection/select_comps.py#L415-L419
https://github.com/handwerkerd/tedana/blob/4c65baf8873ebbc4965820101b006c2cc061c4b7/tedana/selection/select_comps.py#L426-L432

emdupre: What about keeping it with something like "Originally written by PK:" at the start of the comment ? That way, if in finishing commenting this code we decide that it adds something helpful, we can work it back in -- and otherwise delete it. Does that seem reasonable, or do you think it's confusing to have even now ?

handwerkerd: OK. I also removed some comments on Step numbers since it was completely unclear to me why a bunch of stuff was or wasn't considered part of each step. I'll add those back in as well.

emdupre: That seems reasonable to me, at least for the time being (where we're denoting them as "originally written by" or whatever you think best). Thanks again, @handwerkerd !

handwerkerd:
I've added info to the doctoring and done another round of proofreading. If no one sees any additional issues, it should be possible to merge.

PR #94, opened 2018-07-24
[DOC, REF] Add files generated to function docstrings

tsalo:

- Add files-generated tables to docstrings wherever I could track them (especially in `tedana.utils.io`).
- Use `:obj:` for parameter and output data types (even builtins), as this standardizes the formatting of the Sphinx documentation somewhat.
    - This can totally be reverted if it's not desired. It's necessary for inter-package linking, but not for Python builtins (I just think it looks better on the RTD site).
- Switch API and contributing pages on RTD side panel, so it goes **Introduction**, **Usage**, **API**, and then **Contributing to tedana**.
    - This just makes more sense, I think.
- Fix subheaders in contributing page.
- Add information to Usage page for each of the workflows.
- Move workflows to top of API page (since I figure that they're more relevant to most users than the individual functions).
- Update tedana Zenodo DOI to match current version in `__init__.py`.
- Move ste and t2s in `tedana.model.combine.make_optcom` to separate private functions (`tedana.model.combine._combine_ste` and `tedana.model.combine._combine_t2s`). This lets us cite the appropriate paper (Posse vs Poser) depending on the method used with duecredit.
- Add a lot of input checks to `tedana.model.combine.make_optcom`.


tedana/__init__.py
```diff
@@ -36,6 +36,6 @@
          version=__version__, path='tedana', cite_module=True)
 
 # Citation for package version.
-due.cite(Doi('10.5281/zenodo.1250562'),
-         description='The tedana package version 0.0.1',
-         version='0.0.1', path='tedana', cite_module=True)
+due.cite(Doi('10.5281/zenodo.1289081'),
```
emdupre: Is there a way to just cite the general package, without specific versions ? Or do we need to update this every version release ? If the latter, we should star putting together a "version release" checklist, to make sure items like this aren't missed !

tedana/model/combine.py
```diff
@@ -3,11 +3,94 @@
...
+
+        # If all values across echos are 0, set to 1 to avoid
+        # divide-by-zero errors
+        ax0_idx, ax2_idx = np.where(np.all(alpha == 0, axis=1))
+        alpha[ax0_idx, :, ax2_idx] = 1.
+    combined = np.average(mdata, axis=1, weights=alpha)
+    return combined
+
+
+@due.dcite(Doi('10.1002/mrm.20900'),
```
emdupre: Should we expand this out with the author list and title ?

tedana/model/combine.py
```diff
@@ -45,45 +128,47 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
...
+                         'voxels/samples: {0} != {1}'.format(mask.shape[0],
+                                                             data.shape[0]))
+
+    if combmode not in ['t2s', 'ste']:
+        raise ValueError("Argument 'combmode' must be either 't2s' or 'ste'")
+    elif combmode == 't2s' and t2s is None:
         raise ValueError("Argument 't2s' must be supplied if 'combmode' is "
                          "set to 't2s'.")
     elif combmode == 'ste' and t2s is not None:
         LGR.warning("Argument 't2s' is not required if 'combmode' is 'ste'.")
```
emdupre: Can we modify the warning a bit, here ? I imagine that if someone supplies `t2s` and `combmode=ste` they might have misunderstood something.

Maybe we can add "Please make check if calling arguments match input data!" or something to that effect ?

tedana/model/combine.py
```diff
@@ -45,45 +128,47 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
...
+               'estimates')
         ft2s = t2s[mask, :, np.newaxis]
 
     if verbose:
         LGR.info(msg)
 
     if combmode == 'ste':
-        alpha = mdata.mean(axis=-1) * tes
-        alpha = np.tile(alpha[:, :, np.newaxis], (1, 1, n_vols))
+        combined = _combine_ste(mdata, tes)
```
emdupre: I love that we're splitting these up, here ! Just to confirm, do we now have unit test for each option, too ?

tedana/__init__.py
tsalo: We can use the Zenodo DOI for the general package and drop the version argument, but I think it makes more sense to keep this up to date with the version releases. I'm not sure how to do that, since the Zenodo DOI is generated only after we have made the release, but I'm sure someone has dealt with this issue before.

If we do use the version-specific DOI here and in the README, it might make it easier for us to track version-specific usage, assuming we can figure out how to track Zenodo DOI citations in the literature.

A "version release" checklist is a great idea, although I don't know what actually goes into the releases.

tedana/model/combine.py
tsalo: I can add the first author and year to the description, but as long as duecredit can find the DOI we don't need all of the BibTeX info. My concern with the T2* citation was that the DOI looked weird, but after testing it that one rendered too, so I can switch out the clunky BibTeX citation for just the DOI.

emdupre: Got it, thanks for clarifying ! Yes, being consistent either with only the DOI or with the whole BibTeX citation sounds great to me :smile:

tedana/model/combine.py
tsalo: I've added tests for these two and for `make_optcom`, but they just check the output array shapes. We'll want to add ones that use valid input data and check the actual results at some point, I'm sure.

tedana/model/combine.py
tsalo: I just realized I committed changes without responding to this. What about the following? It's informative but not very strong.

```python
LGR.warning("Argument 't2s' is not required if 'combmode' is 'ste'. "
            "'t2s' array will not be used.")
```

tedana/__init__.py
emdupre: I think fMRIPrep keeps [theirs in their wiki](https://github.com/poldracklab/fmriprep/wiki/Release-checklist), although I had shut down the wiki since I wanted to make sure information made it on to [our RTD site](https://tedana.readthedocs.io), first. Do you think it's something we could have as a section in the "Contributing" section of RTD ? Happy to handle this in another PR !

emdupre:
I think this is ready to merge in, and I'll open a new issue for creating a "release checklist" in the documentation. Does that work for you, @tsalo  ?

tsalo:
Sounds good to me. Thanks!

PR #95, opened 2018-07-24
Remove hardcoded numbers of echoes.

### LINKED ISSUE #77: Hard coded number of echoes in Kappa & Rho Estimates
I was going through some of the code and I think I've noticed a bug in the kappa & rho calculations both here and in the meica code. If you look in Appendix A of Olafsson et al NeuroImage 2015 ( https://doi.org/10.1016/j.neuroimage.2015.02.052 ), you can see that the degree of freedom in the denominator of the F tests are (number of echoes - 1)/1. When there are 3 echoes, this becomes the hard coded '* 2' in the following two lines of code. I think it should be (n_echos-1). Does this look right to others?
https://github.com/ME-ICA/tedana/blob/30762df3824752f55b9add787f32e0bddee7758e/tedana/model/fit.py#L136
https://github.com/ME-ICA/tedana/blob/30762df3824752f55b9add787f32e0bddee7758e/tedana/model/fit.py#L143

In practice, this would make all kappa & rho values larger by a constant value based on the # of echoes, which shouldn't affect the elbow thresholds in the selection step, but I don't know  if this might affect some of the non-elbow-based selection criteria.

tsalo:
Closes #77.

- Remove hardcoded number of echoes from Kappa/Rho calculations.
- Remove hardcoded number of echoes from adaptive mask creation.

### DIFF
--- a/tedana/model/fit.py
+++ b/tedana/model/fit.py
@@ -135,14 +135,14 @@ def fitmodels_direct(catd, mmix, mask, t2s, t2sG, tes, combmode, ref_img,
         coeffs_S0 = (B * X1).sum(axis=0) / (X1**2).sum(axis=0)
         SSE_S0 = (B - X1 * np.tile(coeffs_S0, (n_echos, 1)))**2
         SSE_S0 = SSE_S0.sum(axis=0)
-        F_S0 = (alpha - SSE_S0) * 2 / (SSE_S0)
+        F_S0 = (alpha - SSE_S0) * (n_echos - 1) / (SSE_S0)
         F_S0_maps[:, i] = F_S0
 
         # R2 Model
         coeffs_R2 = (B * X2).sum(axis=0) / (X2**2).sum(axis=0)
         SSE_R2 = (B - X2 * np.tile(coeffs_R2, (n_echos, 1)))**2
         SSE_R2 = SSE_R2.sum(axis=0)
-        F_R2 = (alpha - SSE_R2) * 2 / (SSE_R2)
+        F_R2 = (alpha - SSE_R2) * (n_echos - 1) / (SSE_R2)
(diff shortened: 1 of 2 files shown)


handwerkerd:
I don't think the removal of '/ 3' in utils.py is a number of echoes issue.
https://github.com/ME-ICA/tedana/blob/bfaec0632b2e395b31cc830f62d16e4c057addc9/tedana/utils/utils.py#L200
Looking at the code above and below, this function is a simple, but arbitrary way to create a voxelwise mask. It makes the assumption that at least 1/3 of the volume isn't in the brain and it finds the voxel with the 33rd percentile magnitude in the first echo. It then uses the same voxel for all 3 echoes since it's assumed to be an out-of-brain voxel. It takes a threshold from each echo and uses that to define the echo-specific masking threshold. The '/ 3' seems to be used to set the threshold to 1/3 of the selected voxel in each echo so that you only exclude voxels that are really noise. Both the 33rd percentile and the divide by 3 steps seem like arbitrary thresholds, but the division by 3 doesn't look like it has anything to do with the number of echoes so it shouldn't be changed.

Also, this is passing CI, but does any of the sample data this is being tested on include more than 3 echoes? If not, then this change wouldn't alter the output.

tsalo:
That makes sense. I'll change that function back.

emdupre:
WDYT of flagging that value as arbitrary in the code comments, @tsalo ? Maybe with a `# NOTE:` ?

tsalo:
@emdupre That's a great idea! Adding notes now.

emdupre:
LGTM ! I'm ok to merge this in and then re-visit when we add additional tests from other 4+ echo datasets. WDYT @handwerkerd, @tsalo ?

PR #100, opened 2018-07-29
[REF] Merge CLI into workflows

tsalo:
This reorganization would make it easier to update workflows, as the argument parser and the workflow function are consolidated in one file for each workflow (instead of spread between `tedana.cli` and `tedana.workflows`.

- Move argument parsing for workflows from `tedana.cli` to `tedana.workflows`. Each workflow submodule now contains a function named `[name]_workflow` (e.g., `tedana_workflow` and `t2smap_workflow`) as well as private functions such as `_main` and `_get_parser` for the argument parsing and CLI.
- Update `setup.py` to call the `_main` functions in the workflows for the CLI.
- Update tests to use new workflow naming convention.
- Update docs to match new workflow and CLI organizations.


emdupre:
We'll need the CI passing before merging, but I don't have a strong feeling on this reorganization either way. I can see the appeal of it, though, and am happy to move forward if it makes things more intuitive ! I imagine it will also make @rmarkello happy, since he thinks we have too many modules :smile:

tedana/tests/test_tedana.py
```diff
@@ -28,7 +27,7 @@ def compare_nifti(fn, test_dir, res_dir):
     """
     res_fp = (res_dir/fn).as_posix()
     test_fp = (test_dir/fn).as_posix()
-    assert np.allclose(nb.load(res_fp).get_data(), nb.load(test_fp).get_data())
+    assert np.allclose(nib.load(res_fp).get_data(), nib.load(test_fp).get_data())
```
emdupre: Are we consistently using `nib` instead of `nb`, now ? I just want to make sure we don't end up with conflicting imports !

tsalo: I can change it back, but my understanding was that `nib` was the alias used in the nibabel documentation (e.g., [here](http://nipy.org/nibabel/gettingstarted.html#getting-started)). On the other hand, while they may use that alias in at least _some_ of their documentation, I've never seen anything official about it, so we can use `nb` if you prefer.

emdupre: It looks like that's our only import of it for now, and I know I'm the minority opinion so happy to change :) Just wanted to check !

emdupre:
This LGTM ! Ready to merge, @tsalo ?

tsalo:
I think so, unless anyone has any concerns.

emdupre:
I think @rmarkello is most likely to have feedback here (since he was involved in the original organization decisions) !

emdupre:
:100: Merging, then ! Thanks everyone !

PR #104, opened 2018-08-01
[DOC] Update docstring for fixed_seed to ref None option

### LINKED ISSUE #87: seed for ICA is never random
I noticed that the seed for the ICA step is always defined
Default set in 2 ways:
https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/cli/run_tedana.py#L154
https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/workflows/tedana.py#L32

The random number generator is seeded based on the default or a user defined seed:
https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/decomposition/eigendecomp.py#L262

There's actually no way to run the code with a randomly assigned seed. I'm not sure this is a good or bad design choice, but it's not clearly documented that the seed is always fixed. (I had assumed it would be random unless I set the variable). I think it does make sense to set the default to a fixed seed, but I could imagine situations where someone might want a random seed. Whether or not an unseeded option is added, the fact that the code defaults to a fixed seed should be better noted.

I can take care of this, but I wanted to make sure I wasn't missing something first & get feedback on whether it's worth added a non-fixed seed option.

emdupre:
Closes #87 .

- Updates the docstring for the `fixed_seed` argument in eigendecomp.py and tedana.py


emdupre:
I cheated and fixed the typos with an `git commit --amend` :laughing: 

The awkwardness of `fixed_seed=None` is a good point, though. Do you think it's too much ? We could do a couple of things, in my mind:
* Have `None` be the default and specify that users should set a seed
* Go ahead with the dual :obj:int or :obj:None identity, as you suggested
* Try for a special value like -1, but I am worried that could actually be a valid seed

WDYT ? Do you have a preference here ?

tsalo:
Smart!

Regarding the fixed_seed values- I know that -1 breaks `numpy.random.seed`, so I don't expect it to be a valid value. I do think that `None` _should_ be the default, although that will involve changing the order of the arguments in `tedica`, but that will still involve changing the docstrings. I don't know if setting the default value in the argument parser to `None` will conflict with the type being `int`, but my intuition is that it won't be a problem (unlike feeding in `--seed None` directly).

emdupre:
I just checked with `scipy` (what this is really a wrapper for), and -1 also breaks, so we could coerce the value. I think the choice then comes down to whether or not `None` should be the default.

I'm not enthusiastic about it, since I'd like to have results be reproducible out of the box unless a user knows and wants to specifically request non-replicable results. Happy to hear other thoughts, though ! Could you explain why you think `None` should be the default ?

emdupre:
@kirstiejane might have an opinion here, too, since her group also implemented a `fixed_seed` argument !

KirstieJane:
To be clear, the question is which should be the default, right? `None` or an integer value.

I can see both sides. There's a danger that groups won't mix up the random seed and therefore every run will be weirdly similar.....but I think the bigger danger is that people **don't** fix the random seed and every time they re-run the code they get (slightly) different answers.

So, my not-very-strong vote is for the default to be an integer rather than None....but with clearly documentation to say - go ahead and change this value as you choose!

handwerkerd:
The changes look fine to me. I'm fine merging, as is, but I'll make some comments in response to other's concerns regarding the default option and the awkwardness of --fixed_seed=None

Earlier, I agreed that the default should be a fixed seed so that the default is replicable results, but understand why default randomness could be useful. If that's the general consensus, for replicability, we'd just need to save the random seed somewhere in the output. The easiest way to do this would be to use a random number generator to assign a value to `fixed_seed` and then make sure to save that number in the output.
For example, replace 
https://github.com/ME-ICA/tedana/blob/0c5f93dc06e5fbbbe49811fa5eb717f7ea87ec22/tedana/decomposition/eigendecomp.py#L284
with

```
if fixed_seed == None:
    fixed_seed=np.random_integers(low=1)
mdp.numx_rand.seed(fixed_seed)
```


and then make sure that fixed_seed is actually part of the output of the code. I'd recommend this in general because, even if people use None, they might want to replicate a specific iteration at some point in the future.

KirstieJane:
Endorse @handwerkerd's suggestion, even if we do keep a fixed seed as the default! 💯

emdupre:
Ok, I've edited this to take `-1` as the value for non-reproducible calls, but following @handwerkerd's suggestion I'm also writing out the seed that was actually used !

Happy to hear any other suggestions, here. And I'd always appreciate another review, @tsalo :)

emdupre:
Merging !

PR #108, opened 2018-08-06
[ENH] Add mask argument to t2smap, tedana

### LINKED ISSUE #96: Allow for an inputted mask
If I'm reading the code correctly, the volumes are always masked within the code:
https://github.com/ME-ICA/tedana/blob/bfaec0632b2e395b31cc830f62d16e4c057addc9/tedana/workflows/tedana.py#L203
There is no way for a user to create their own mask and use that instead. This issue came up when I was answering a question from J Dalenberg & realized I needed to make an entire dataset to make sure certain voxels were removed from the analysis (and I still wasn't be able to control if additional voxels would be removed).

I suggest adding an option for users to apply their own masks. I can take care of this if others agree it would be useful. My only hesitation on doing this immediately is I'm not sure how it might interact with @tsalo's pull requests, like #70

emdupre:
Closes #96.

- Allow for the input of a binary, spatially aligned mask in `tedana` and `t2smap` workflows


tedana/utils/utils.py
```diff
@@ -169,6 +169,9 @@ def make_adaptive_mask(data, minimum=True, getsum=False):
     data : (S x E x T) array_like
         Multi-echo data array, where `S` is samples, `E` is echos, and `T` is
         time
+    mask : :obj:`str`, optional
+        Binary mask for voxels to consider in TE Dependent ANAlysis. Default is
+        to generate mask from data with good signal across echoes
```
tsalo: This can be a string (filename), a numpy array, or an image, right?

emdupre: Right now I'm only accepting a file name, since it's user-specified. Can you explain how it could be an image or numpy array ?

emdupre:
OK, merging ! :tada:

PR #114, opened 2018-08-20
[FIX] Drop gifti support

### LINKED ISSUE #34: Spatial info for surface data
There are a few points throughout `tedana.py` where spatial information is required. Namely:

1. `tedana.interfaces.tedana.fitmodels_direct()`
2. `tedana.interfaces.tedana.spatclust()`
2. `tedana.interfaces.tedana.selcomps()`

The function `fitmodels_direct()` calls `spatclust()` to do spatial clustering on the independent components generated from ICA. While spatial clustering is possible with surface data, it would require information about the location of the vertices in 3D space; this info is generally kept in a separate datafile. We could maybe ask users to provide that file if they want spatial clustering to occur?

On the other hand, the function `selcomps()` makes a call to `np.fft.rfftn()`, which operates over the 3D volumetric data to detect high-spatial frequency artifacts. Figuring out how to get this working with surface-based data (or indeed, if it is even reasonable to do so) is up for debate.

Relatedly, there are a few other steps in `selcomps()` where surface data may run into issues (e.g., with generating the vein masks) that will require some thought.

emdupre:
Closes #34.

- Drops gifti support. Images must instead be tedana-denoised before projecting to the surface.

### DIFF
--- a/tedana/model/fit.py
+++ b/tedana/model/fit.py
@@ -197,9 +197,6 @@ def fitmodels_direct(catd, mmix, mask, t2s, t2sG, tes, combmode, ref_img,
             out[:, 2] = np.squeeze(utils.unmask(F_S0_maps[:, i], t2s != 0))
             out[:, 3] = np.squeeze(utils.unmask(Z_maps[:, i], mask))
 
-            if utils.get_dtype(ref_img) == 'GIFTI':
-                continue  # TODO: pass through GIFTI file data as below
-
             ccimg = utils.new_nii_like(ref_img, out)
 
             # Do simple clustering on F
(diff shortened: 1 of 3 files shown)


emdupre:
@tsalo @rmarkello please let me know if you foresee any issues with merging this !

tsalo:
Nothing jumps out at me in the code, and since tests are passing I think it should be good.

emdupre:
Dropping gifti support is a little bit of an extreme response to the discussion in #34, but I think we agreed on gitter that this was the best course of action in the long run. Just trying to make sure I'm not misrepresenting / misremembering that discussion !!

rmarkello:
As @tsalo said, nothing in the code strikes me as particularly concerning! And I think this is probably the right way forward, so :+1: from me.

emdupre:
Great, thanks for the feedback ! Merging !

PR #115, opened 2018-08-20
[FIX] Explicitly strip extensions

### LINKED ISSUE #82: Change number of volumes when saving images
The function, filewrite, uses a ref_img to get all the parameters for the output image. Looking through the code, it seems like the same ref_img is used for almost all outputs. I think this is causing some problems because, the ref_image has the same number of volumes as the time series, but some of the outputs have a different number of volumes, yet are still coded having more.
https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/utils/utils.py#L241

The quick solution is to add lines into https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/utils/utils.py#L302 and https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/utils/utils.py#L333 that will adjust the number of volumes in the image to match the number of volumes in the new data set.

I'm not sure if there's an easier way to do this within the nilearn I/O framework or if there are other data-specific parameters we should also be examining and changing as well.

Also, related to this, any reason why the volume I/O functions are in utils.py rather than io.py?

emdupre:
Closes #82.

- Reformats `load_data` to explicitly strip extensions for both zcat images and listed echo files

### DIFF
--- a/tedana/utils/utils.py
+++ b/tedana/utils/utils.py
@@ -133,7 +133,9 @@ def load_data(data, n_echos=None):
                              '{}'.format(data))
         else:  # individual echo files were provided (surface or volumetric)
             fdata = np.stack([load_image(f) for f in data], axis=1)
-            return np.atleast_3d(fdata), data[0]
+            ref_img = check_niimg(data[0])
+            ref_img.header.extensions = []
+            return np.atleast_3d(fdata), ref_img
 
     img = check_niimg(data)
     (nx, ny), nz = img.shape[:2], img.shape[2] // n_echos
--- a/tedana/tests/test_utils.py
+++ b/tedana/tests/test_utils.py
@@ -123,8 +123,8 @@ def test_load_data():
     # list of filepath to images
     d, ref = utils.load_data(fnames, n_echos=len(tes))
     assert d.shape == exp_shape
-    assert isinstance(ref, str)
-    assert ref == fnames[0]
+    assert isinstance(ref, nib.Nifti1Image)
+    assert np.allclose(ref.get_data(), nib.load(fnames[0]).get_data())
 
     # list of img_like
     d, ref = utils.load_data(fimg, n_echos=len(tes))


emdupre:
@rmarkello it'd be great to get your review, here !

tedana/utils/utils.py
```diff
@@ -133,17 +133,19 @@ def load_data(data, n_echos=None):
                              '{}'.format(data))
         else:  # individual echo files were provided (surface or volumetric)
             fdata = np.stack([load_image(f) for f in data], axis=1)
-            return np.atleast_3d(fdata), data[0]
+            fdata = np.atleast_3d(fdata)
+            ref_img = check_niimg(data[0])
+    else:
```
rmarkello: I don't think this will work as-is for z-concatenated images. The `if isinstance(data, list)` will _always_ be triggered because of argparse, so the `else` clause here never will. I think we can get rid of this whole `if`-`else` block and move the code below this `else` section up under `if len(data) == 1`.

tedana/utils/utils.py
```diff
@@ -133,17 +133,19 @@ def load_data(data, n_echos=None):
...
+        (nx, ny), nz = img.shape[:2], img.shape[2] // n_echos
+        fdata = load_image(img.get_data().reshape(nx, ny, nz, n_echos, -1, order='F'))
 
-    img = check_niimg(data)
-    (nx, ny), nz = img.shape[:2], img.shape[2] // n_echos
-    fdata = load_image(img.get_data().reshape(nx, ny, nz, n_echos, -1, order='F'))
+        # create reference image
+        ref_img = img.__class__(np.zeros((nx, ny, nz)), affine=img.affine,
+                                header=img.header, extra=img.extra)
+        ref_img.header.set_sform(ref_img.header.get_sform(), code=1)
```
rmarkello: This should probably be moved back next to `ref_img.header.extensions = []` so that it triggers regardless of the input data!

Issue #58, opened 2018-05-17
Track selection criteria provenance across workflows


handwerkerd:
This has been directly or indirectly discussed in a few other issues (  #16 #35 #50 ), but I think it's worth it's own issue. A problem I've repeatedly come across when trying to understand the output of the original tedana is: When a component is rejected or in midK for non-obvious reasons, it is sometimes very hard to figure out what selection criterion was used for it's placement without digging into the code. For example see "Reasons for Fork" at https://bitbucket.org/BenGutierrez/me-ica which was made because the default high variance selection criteria in meica meant that a component representing the visual cortex response to a block design flashing checkerboard was regularly getting rejected not because of it's TE dependence but because it's variance was above a hard-coded threshold.

As part of the #16 efforts to break up the selection components into multiple functions, it would be useful to create a data structure (or add fields to an existing data structure) that tracks which selection criteria are applied and where each component falls within each criteria. This might get a bit tricky because, for some selection methods, including the one in the original tedana, several selection criteria are combined across a decision tree, so it might be necessary to store the decision tree in some way.

I think there is a separate issue regarding how to give this information to the end user in an easily digestible way, but I wanted to put this idea out there as something to consider including at a low level rather than trying later retrofit this into the code.

I will comment here, and I may be able to personally work on this code, but don't count on me for substantive code changes before late June.

tsalo:
@handwerkerd @emdupre I'm really not familiar with the component selection code (yet), but can we make `comp_table.txt` a csv or tab-delimited text file with this information? I think that a spreadsheet makes more sense to track metrics and selection criteria than the current text file format.

The header contains information, such as number of rejected and accepted components, that can mostly be easily extrapolated from the statistic table. The columns at the bottom of the file are "comp", "Kappa", "Rho", "Var", and "Var (norm)". If we add another column (say, "label" or "classification status" or whatever) with values of "accepted", "rejected", and "ignored", along with a column for the rationale, we could log all of the necessary information in a more concise and parsable format (especially if more metrics are ever added). Tracking the rationale will take some time to figure out and is probably better handled by someone other than me, but I'd be happy to get the rest of it working and to lay the groundwork for next steps. WDYT?

handwerkerd:
I see this as two separate parts. The more significant part is how do we set up a data structure to keep track of this information across multiple functions and workflows? Once that is figured out, deciding how to output the information will require discussion, but will be relatively easy.

[...]

PR #119, opened 2018-08-23
[ENH] Switch to Selcomps 2.5

### LINKED ISSUE #89: fproj_arr is never used
In the selcomps function, a variable fproj_arr is created, which is a ranked list of value of a max projection of the FFT of each component. I don't think this variable is ever used again. The following line contains the creation of fproj_arr_val, which is used later in the code in an SVM classifier of rejected components to identify some additional components as midk. I don't completely get this part of the code, but it seems like fproj_arr_val might have been intended to contain the ranked list of values in fproj_arr and not the unranked values from fproj_z
https://github.com/ME-ICA/tedana/blob/2fa0136605f52a54fc11f361cd883af0e6a6fb8f/tedana/selection/select_comps.py#L178-L179

emdupre:

- Closes #89 
- Initial commit of the 2.5 selection criteria workflow

To Do: 

- [x] Add initial documentation
- [x] Update testing


tsalo:
I think that further documentation can wait until @handwerkerd has the time for it, but do you want to handle the testing in this PR or later?

emdupre:
Yes, I think I'd prefer to handle testing in this PR -- it will make it quite large but I'm also uncomfortable merging in a non-passing PR and breaking CI for the rest of the codebase. WDYT ?

tsalo:
That makes sense. 👍

tedana/selection/select_comps.py
```diff
@@ -105,794 +78,283 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
-    that are in clusters (signal) and the number outside of clusters (noise)
-
-    tt_table is a bit confusing. For each component, the first index is
-    some type of normalized, log10, signal/noise t statistic and the second is
-    the p value for the signal/noise t statistic (for the R2 model).
-    In general, these should be bigger t or have lower p values when most of
-    the Z values above threshold are inside clusters.
-    Because of the log10, values below 1 are negative, which is later used as
-    a threshold. It doesn't seem like the p values are ever used.
+    Step 3a
```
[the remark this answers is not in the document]
tsalo: I don't know what all the "steps" are referring to, but that is from [here](https://bitbucket.org/prantikk/me-ica/src/8cc47cfed0203b3d6d187935ad3c2823b3e36a88/meica.libs/select_model.py?at=master&fileviewer=file-view-default#select_model.py-252) and Step 1 is from [here](https://bitbucket.org/prantikk/me-ica/src/8cc47cfed0203b3d6d187935ad3c2823b3e36a88/meica.libs/select_model.py?at=master&fileviewer=file-view-default#select_model.py-277). I'd be happy to see all references to "steps" removed, which could make it less confusing.

tedana/model/fit.py
```diff
@@ -88,6 +88,8 @@ def fitmodels_direct(catd, mmix, mask, t2s, t2s_full, tes, combmode, ref_img,
                              '({0}) does not match second dimension of '
                              't2s ({1})'.format(catd.shape[2], t2s.shape[1]))
 
+    mask = t2s != 0  # Override mask because problems
```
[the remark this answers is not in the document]
tsalo: Not really... this is a hacky way I was able to get around the fact that different arrays were being masked differently (i.e., either with `mask` or with `t2s != 0`) and then were being compared element-wise, which was raising errors. This is something we should dig into before merging this PR, but I'm not sure why the arrays were being masked differently to begin with, so I don't know if it's better to make _everything_ with `mask`, with `t2s != 0`, or to change which comparisons are made.

tedana/selection/select_comps.py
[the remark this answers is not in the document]
emdupre: Ah, I see, you had just moved the code comment up -- I wasn't sure if it was a bad merge or had been in the original code ! I'll remove it, then. Thanks for clarifying !

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
+                          (Z_clmaps[:, i_comp] == 0))
+        comptable.loc[i_comp, 'countnoise'] = np.array(
+            comp_noise_sel, dtype=np.int).sum()
+        noise_FR2_Z = np.log10(np.unique(F_R2_maps[comp_noise_sel, i_comp]))
+        signal_FR2_Z = np.log10(np.unique(
+            F_R2_maps[Z_clmaps[:, i_comp] == 1, i_comp]))
+        counts_FR2_Z[i_comp, :] = [len(signal_FR2_Z), len(noise_FR2_Z)]
+        (comptable.loc[i_comp, 'tt0'],
+         comptable.loc[i_comp, 'tt1']) = stats.ttest_ind(
+             signal_FR2_Z, noise_FR2_Z, equal_var=False)
```
tsalo: Given that tt0 and tt1 are t- and p-values, respectively, from a t-test comparing z-transformed R2 F-values from what looks like signal and noise voxels... I think we can maybe called them something like "signal-noise_t" and "signal-noise_p". WDYT?

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
+        comptable.loc[comptable['kappa'] > getelbow(comptable['kappa'], return_val=True),
+                      'variance explained'])
+    ncls = acc.copy()
+    # NOTE: We're not sure why this is done, nor why it's specifically done
+    # three times. Need to look into this deeper, esp. to make sure the 3
+    # isn't a hard-coded reference to the number of echoes.
+    for nn in range(3):
+        ncls = comptable.loc[ncls].loc[
+            comptable.loc[
+                ncls, 'variance explained'].diff() < varex_ub_p].index.values
```
tsalo: This PR is a chance to take a look at this part as well.

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
-    mmix_kurt_z = (mmix_kurt-mmix_kurt[group0].mean()) / mmix_kurt[group0].std()
-    # smaller is worse - drift
-    mmix_std_z = -1 * ((mmix_std-mmix_std[group0].mean()) / mmix_std[group0].std())
-    mmix_kurt_z_max = np.max([mmix_kurt_z, mmix_std_z], 0)
+    kappa_rate = ((np.max(comptable.loc[good_guess, 'kappa']) -
+                   np.min(comptable.loc[good_guess, 'kappa'])) /
+                  (np.max(comptable.loc[good_guess, 'variance explained']) -
+                   np.min(comptable.loc[good_guess, 'variance explained'])))
+    kappa_ratios = kappa_rate * comptable['variance explained'] / comptable['kappa']
+    varex_lb = stats.scoreatpercentile(
```
emdupre: `_lb` is lower bound, right ? Would you be opposed if I changed this to `_lower` and `_upper` ?

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
-                            [0] * len(group0) + [1] * len(rej),
-                            fproj_arr_val[:, to_clf].T,
-                            svmtype=2)
-    midk = np.setdiff1d(
-        to_clf[utils.andb([
-            midk_clf == 1,
-            (comptable['variance explained'][to_clf] >
-             np.median(comptable['variance explained'][group0]))]) == 2],
-        np.union1d(toacc_hi, toacc_lo))
+    max_good_d_score = EXTEND_FACTOR * len(good_guess) * n_decision_metrics
```
emdupre: Is there a better term than `EXTEND_FACTOR` to explain what's going on, here ?

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
+    midk = acc[(comptable.loc[acc, 'd_table_score'] > max_good_d_score) &
+               (comptable.loc[acc, 'variance explained'] > EXTEND_FACTOR * varex_ub)]
     comptable.loc[midk, 'classification'] = 'rejected'
     comptable.loc[midk, 'rationale'] += 'midk;'
```
emdupre: Is there a reason `midk` gets a semi-colon and `rejected` does not ?

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
+        comptable['d_table_score_scrub'] = np.nan
+        comptable.loc[acc, 'd_table_score_scrub'] = d_table_rank.sum(1)
+        num_acc_guess = int(np.mean([
+            np.sum((comptable.loc[acc, 'kappa'] > kappa_elbow) &
+                   (comptable.loc[acc, 'rho'] < rho_elbow)),
+            np.sum(comptable.loc[acc, 'kappa'] > kappa_elbow)]))
+        conservative_guess = num_acc_guess * n_decision_metrics / RESTRICT_FACTOR
+
+        # candartA
+        candartA = np.intersect1d(
```
emdupre: `candartA` ? Any idea what that stands for ?

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
+            acc[comptable.loc[acc, 'd_table_score_scrub'] > conservative_guess],
+            acc[kappa_ratios[acc] > EXTEND_FACTOR * 2])
+        candartA = np.intersect1d(
+            candartA,
+            candartA[comptable.loc[candartA, 'variance explained'] > varex_ub * EXTEND_FACTOR])
+        comptable.loc[candartA, 'classification'] = 'rejected'
+        comptable.loc[candartA, 'rationale'] += 'candartA;'  # TODO: Better rationale
```
emdupre: From this line, I'm assuming the answer to my previous question is "no" :joy:

tedana/selection/select_comps.py
tsalo: @handwerkerd requested that we eliminate `midk` as a classification [here](https://github.com/ME-ICA/tedana/issues/58#issuecomment-433975211), so here we just note it as the rationale for rejection. Classifications are set, while rationales are appended to (hence `+=` and the semicolon to delineate reasons). I tried to append the rationale every time components are rejected or ignored, so that any times when a component is rejected or ignored for multiple reasons they're all logged.

tedana/selection/select_comps.py
tsalo: Rejection candidate based on artifact type A, is my guess.

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
-    version 3.2 of MEICA at:
-    https://github.com/ME-ICA/me-ica/blob/b2781dd087ab9de99a2ec3925f04f02ce84f0adc/meica.libs/select_model_fft20e.py
-    Many of the early publications using and evaulating the MEICA method used a
-    different selection algorithm by prantikk. The final 2.5 version of that
-    algorithm in the selcomps function in select_model.py at:
+    It is from selcomps function in select_model.py in version 2.5 of MEICA at:
     https://github.com/ME-ICA/me-ica/blob/b2781dd087ab9de99a2ec3925f04f02ce84f0adc/meica.libs/select_model.py
-
+    Some later publications using and evaluating the MEICA method used a
+    different selection algorithm by prantikk. The 3.2 version of that
```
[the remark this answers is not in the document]
emdupre: Done ! https://gist.github.com/emdupre/ca92d52d345d08ee85e104093b81482e

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
+    if len(good_guess) == 0:
+        LGR.warning('No BOLD-like components detected')
+        ign = sorted(np.setdiff1d(all_comps, rej))
+        comptable.loc[ign, 'classification'] = 'ignored'
+        comptable.loc[ign, 'rationale'] += 'no good components found;'
+
+        # Move decision columns to end
+        comptable = comptable[[c for c in comptable if c not in cols_at_end] +
+                              [c for c in cols_at_end if c in comptable]]
+        return comptable
```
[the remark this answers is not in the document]
emdupre: It looks like this is in fact the end of the file, the rendering on GitHub is just a little off !

tedana/selection/select_comps.py
```diff
@@ -71,828 +46,298 @@ def selcomps(seldict, comptable, mmix, mask, ref_img, manacc, n_echos, t2s, s0,
...
+        signal_FR2_Z = np.log10(np.unique(
+            F_R2_maps[Z_clmaps[:, i_comp] == 1, i_comp]))
+        counts_FR2_Z[i_comp, :] = [len(signal_FR2_Z), len(noise_FR2_Z)]
+        (comptable.loc[i_comp, 'tt0'],
+         comptable.loc[i_comp, 'tt1']) = stats.ttest_ind(
+             signal_FR2_Z, noise_FR2_Z, equal_var=False)
+
+    comptable.loc[np.isnan(comptable['tt0']), 'tt0'] = 0
+    comptable.loc[np.isnan(comptable['tt1']), 'tt1'] = 0
+    del noise_FR2_Z, signal_FR2_Z
```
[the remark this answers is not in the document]
tsalo: TBH we probably don't _need_ to optimize, given how much simpler and faster version of the function is. I'm totally cool with removing the `del`s.

tsalo:
Would it be possible to split the 3-echo Cornell data and tar it like the 5-echo NIH dataset? If possible, it would even be nice to have the files within the tarred folders follow BIDS convention, although we may not have the necessary information (e.g., subject ID, session ID, task).

emdupre:
I suppose it's a question of whether we want to explicitly test the zcat functionality, since a lot of the code still supports it. If we _don't_ test it, I'd argue that we should also drop support !

emdupre:
This is probably a discussion we should have in #152 -- I don't know that zcat files _have_ to be BIDS incompatible, but we should definitely think through that angle, maybe in consultation with the BIDS folks.

emdupre:
Currently hitting a snag where the `tedana` command is not recognized. I've added an SSH key to circle to try and check in on the environment, but it's not yet recognized. Hoping to pick back up as soon as it's live, but if anyone looks over the file and is able to quickly spot the error, please let me know !

.circleci/config.yml
```diff
@@ -0,0 +1,195 @@
...
+          name: Run three-echo dataset
+          no_output_timeout: 40m
+          command: |
+            source activate venv
+            cd /tmp/data/three_echo_Cornell/
+            tedana -d three_echo_Cornell_zcat.nii.gz -e 14.5 38.5 62.5
+      - run:
+          name: Checking outputs
+          command: |
+            find /tmp/data/three-echo/TED.three_echo_Cornell_zcat/ \
```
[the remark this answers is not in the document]
emdupre: I just went through and standardized the dirnames (I think) -- we'll see how it goes !

.circleci/config.yml
```diff
@@ -0,0 +1,195 @@
...
+          name: Download test three-echo data
+          command: |
+            mkdir -p /tmp/data
+            curl -L --create-dirs -o \
+              /tmp/data/three_echo_Cornell/three_echo_Cornell_zcat.nii.gz https://osf.io/e3hsn/download
+      - run:
+          name: Download test five-echo data
+          command: |
+            curl -L -o five_echo_NIH.tar.xz https://osf.io/ea5v3/download
+            tar xf five_echo_NIH.tar.xz -C /tmp/data/
```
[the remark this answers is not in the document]
emdupre: Thanks for catching !  :hugs:

emdupre:
It's passing now-- WDYT, @ME-ICA/tedana-devs ? I'd argue to merge this now, cut a release, and move forward from here to improve everything :)

emdupre:
OK, I'm going to merge this ! We can (and I'm sure will !) revisit going forward.

Thanks everyone ! :tada:

Issue #106, opened 2018-08-04
Replace/supplement label argument with outdir/prefix argument


tsalo:
I would really like the ability to specify an output directory for the workflows and each of the functions that write out files. Output file generation is spread out over several functions, which I think rely on `cd`-ing into the output directory to make sure they're written out where the workflow wants them written out, so this will probably be a substantial refactor. Still, I think it's important.

Does anyone agree or disagree?

rmarkello:
+1 (but agreed this will likely be a substantial refactor)

The file IO definitely leaves something to be desired and I think it would be great to abstract away from what we currently have to make it a bit more intuitive.

handwerkerd:
I don't completely understand what you're suggesting here. Currently the output is in TED.[ref_label] if `label` is none and TED.[ref_label].label if `label` is an input. I could see the default being TED.[ref_label] and, if a user defines `label` then use TED.label (without ref_label)

If I'm reading your suggestion correctly, you want to use `outdir` instead of TED.label so that a user could output files to a directory that doesn't begin with "TED"? I'd lead towards keeping TED because it's a relatively benign label that adds some naming consistency, but I won't be stubborn about this.
Having a separate prefix option for the outputted file names would be useful since, having multiple directories containing files with identical names can cause headaches... even if it's nice to be able to hard-code file names.

In general, I agree that I/O naming can use some more thought.

emdupre:
@tsalo I'm also a bit confused on the suggested naming convention -- is there any chance you could mock-up what you're imagining the output directory would look like ?

Otherwise, yes, I think being able to create outputs without `cd`'ing into the directory is important, particularly as we continue abstracting away !

tsalo:
I personally would prefer not to put any limits on the output directory name, but I don't have a real issue with keeping "TED" in the output directory name. What I'd prefer is to not use the input filename at all, because that means users have to name their input data according where they want their results to go.

Here's one way it could work, if we dropped `ref_label`:
- `tedana -d something.nii.gz` would output to `TED/` if we kept the TED thing or just to `.` if not.
- `tedana -d something.nii.gz -label LABEL` would output to `TED.LABEL/` or `LABEL/` if we wanted to keep a label argument.
- `tedana -d something.nii.gz -outdir /home/tsalo/` would output to `/home/tsalo/` if we dropped the TED thing and `/home/tsalo/TED` if not.
[...]

Issue #101, opened 2018-07-30
Should PCA component selection leverage multi-echo information?


emdupre:
Before running [independent components analysis (ICA)](http://scikit-learn.org/stable/modules/decomposition.html#independent-component-analysis-ica), we first run [principal components analysis (PCA)](http://alexhwilliams.info/itsneuronalblog/2016/03/27/pca/) and select a limited number of the resulting principal components. This is a common approach across the community, and makes a great deal of sense -- the [data need to be whitened, and the dimensionality reduction helps us avoiding over-fitting to noise](https://stats.stackexchange.com/a/99507).

**Edited to add:** See this conversation for an explanation of why we need *both* PCA and ICA steps to dimensionality reduce and whiten: https://github.com/ME-ICA/tedana/pull/44#issuecomment-437694589

However, the way that principal components are selected here is not by a noise threshold, but rather via a decision tree similar to the process by which independent components are later selected. This can yield very high dimensional datasets that have a low probability of converging with ICA. Is it worth revisiting whether principal components should be selected by [a hard threshold](https://ieeexplore.ieee.org/document/6846297/)?

dowdlelt:
I have been frustrated by failure to converge in the past, and the tunable parameters kdaw and rdaw also add greatly to complexity. However, I would be concerned about selecting using a hard threshold, based primarily on Figure 10 from the [2017, Multiecho fMRI paper](https://www.sciencedirect.com/science/article/pii/S1053811917302410?via%3Dihub), which suggest that a kappa/rho based PCA approach captures more of the useful variance. 
![image](https://user-images.githubusercontent.com/15126366/43402560-09689f6c-93e1-11e8-9bfb-f829ffb5408e.png)
Maybe this is only a particular concern for those that use the High Kappa Timeseries or the tedana coefficient maps for analyses - if BOLD signals are not extracted during PCA, then they would never make it to the ICA. For the optimally combined and denoised timeseries I think this would be a less of a concern, and the benefit of more stable convergence might be more important than less variance explained. 

Not sure how often folks just use the denoised vs high kappa data, or component maps. I stick with the denoised data, myself, as the high kappa timeseries can miss brain components or include things that are BOLD but not 'brain activity' (breathing related).

handwerkerd:
[...]

Issue #17, opened 2018-04-25
Reduce intermediate outputs


emdupre:
_From @emdupre on November 15, 2017 16:23_

The `niwrite` function is called throughout `tedana` to output many intermediate files; however, these are poorly documented and of unclear value to the user. It would make more sense to only have these intermediate files output if the user provides a `--verbose` flag.

_Copied from original issue: emdupre/tedana#2_

emdupre:
I don't like having an `invalid` label ! I just learned that was an option by accidentally applying it here 😆 This is still a very valid concern !

tsalo:
I'd like to continue the discussion started in #133 regarding intermediate outputs and the possible addition of a new argument to control their generation here.

Here is a hopefully comprehensive list of files that are generated. I've taken the liberty of bolding files that I _think_ should be optional.

| Filename | Content |
| ------- | --------- |
| t2sv.nii | Limited estimated T2* 3D map. |
| s0v.nii | Limited S0 3D map. |
| ts_OC.nii | Optimally combined time series. |
| dn_ts_OC.nii | Denoised optimally combined time series. |
| lowk_ts_OC.nii | Combined time series from rejected components. |
| midk_ts_OC.nii | Combined time series from "mid-k" rejected components. |
| hik_ts_OC.nii | High-kappa time series. |
| **t2ss.nii** | Voxel-wise T2* estimates using ascending numbers of echoes, starting with 2. |
| **s0vs.nii** | Voxel-wise S0 estimates using ascending numbers of echoes, starting with 2. |
| **t2svG.nii** | Full T2* map/time series. |
| **s0vG.nii** | Full S0 map/time series. |
| **comp_table_pca.txt** | TEDPCA component table. |
| **mepca_mix.1D** | Mixing matrix (component time series) from PCA decomposition. |
| **__meica_mix.1D** | Mixing matrix (component time series) from ICA decomposition. |
| **meica_mix.1D** | Mixing matrix (component time series) from ICA decomposition. The only differences between this mixing matrix and the one above are that components may be sorted differently and signs of time series may be flipped. |
| **betas_OC.nii** | Full ICA coefficient feature set. |
| **betas_hik_OC.nii** | High-kappa ICA coefficient feature set |
| **feats_OC2.nii** | Z-normalized spatial component maps |
| **comp_table_ica.txt** | TEDICA component table. |
| **veins_l0.nii** | ??? |
| **veins_l1.nii** | ??? |

## If global signal correction is employed:

[...]

PR #122, opened 2018-09-01
[WIP, ENH] Track PCA and ICA component selection decisions

tsalo:
Ref #58.

- Use pandas DataFrames to store MEPCA and MEICA component metrics, rather than numpy arrays or dictionaries.
- Track component selection in "classification" column as well as reason behind a given classification (e.g., rejected, ignored) in "rationale" column.


tsalo:
Tests are now passing, though there are a couple of things that could be improved- especially column names and the rationale documentation within `selcomps`.

tedana/decomposition/eigendecomp.py
```diff
@@ -143,21 +143,22 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
             from sklearn.decomposition import PCA
             ppca = PCA(n_components='mle', svd_solver='full')
             ppca.fit(dz)
-            v = ppca.components_
```
emdupre: `u`, `v`, and `s` are pretty standard for an SVD. I'm ok to keep them as is, but if you think this is easier to understand  I'm happy to update !

tedana/decomposition/eigendecomp.py
```diff
@@ -143,21 +143,22 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
...
+            voxel_comp_weights, varex, comp_ts = np.linalg.svd(dz, full_matrices=0)
 
         # actual variance explained (normalized)
-        sp = s / s.sum()
-        eigelb = getelbow_mod(sp, return_val=True)
+        varex_norm = varex / varex.sum()
+        eigenvalue_elbow = getelbow_mod(varex_norm, return_val=True)
 
-        spdif = np.abs(np.diff(sp))
+        spdif = np.abs(np.diff(varex_norm))
```
emdupre: Can we change these names (`spdif`, `spdifh`, `spdthr`) to match the others ?

tedana/decomposition/eigendecomp.py
```diff
@@ -167,20 +168,24 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
...
-                                                     tes, combmode, ref_img,
-                                                     mmixN=vTmixN, full_sel=False)
-        ctb = ctb[ctb[:, 0].argsort(), :]
-        ctb = np.vstack([ctb.T[:3], sp]).T
+        _, ct_df, betasv, v_T = model.fitmodels_direct(catd, comp_ts.T, eimum,
+                                                       t2s, t2sG,
+                                                       tes, combmode, ref_img,
+                                                       mmixN=vTmixN,
+                                                       full_sel=False)
+        ct_df['variance explained 2 (normalized)'] = varex_norm
```
emdupre: I don't love `variance explained 2 (normalized)`. Could we just do `normalized variance explained` ? Is there a reason to keep the `2` ?

tedana/decomposition/eigendecomp.py
```diff
@@ -167,20 +168,24 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
...
 
         # Save state
         fname = op.abspath('pcastate.pkl')
         LGR.info('Saving PCA results to: {}'.format(fname))
-        pcastate = {'u': u, 's': s, 'v': v, 'ctb': ctb,
-                    'eigelb': eigelb, 'spmin': spmin, 'spcum': spcum}
+        pcastate = {'u': voxel_comp_weights, 'varex': varex, 'comp_ts': comp_ts,
```
emdupre: Why are we keeping the `u` here ?

tedana/decomposition/eigendecomp.py
```diff
@@ -191,62 +196,107 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
         LGR.info('Loading PCA from: pcastate.pkl')
         with open('pcastate.pkl', 'rb') as handle:
             pcastate = pickle.load(handle)
-        u, s, v = pcastate['u'], pcastate['s'], pcastate['v']
-        ctb, eigelb = pcastate['ctb'], pcastate['eigelb']
-        spmin, spcum = pcastate['spmin'], pcastate['spcum']
+        voxel_comp_weights, varex = pcastate['u'], pcastate['varex']
```
emdupre: See above question !

tedana/decomposition/eigendecomp.py
```diff
@@ -191,62 +196,107 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
...
+
+    np.savetxt('mepca_mix.1D', comp_ts)
+
+    # write component maps to 4D image
+    comp_maps = np.zeros((OCcatd.shape[0], comp_ts.shape[0]))
+    for i_comp in range(comp_ts.shape[0]):
+        temp_comp_ts = comp_ts[i_comp, :][:, None]
+        comp_map = utils.unmask(model.computefeats2(OCcatd, temp_comp_ts, mask), mask)
+        comp_maps[:, i_comp] = np.squeeze(comp_map)
+    utils.filewrite(comp_maps, 'mepca_OC_components.nii', ref_img)
```
emdupre: This is the first time writing out `mepca_OC_components.nii` ? I don't see it elsewhere..

tedana/model/fit.py
```diff
@@ -142,113 +146,117 @@ def fitmodels_direct(catd, mmix, mask, t2s, t2s_full, tes, combmode, ref_img,
...
+        F_S0_maps = F_S0_maps[:, sort_idx]
+        F_R2_maps = F_R2_maps[:, sort_idx]
+        Z_maps = Z_maps[:, sort_idx]
+        WTS = WTS[:, sort_idx]
+        PSC = PSC[:, sort_idx]
+        tsoc_B = tsoc_B[:, sort_idx]
+        tsoc_Babs = tsoc_Babs[:, sort_idx]
     else:
-        comptab = comptab_pre
+        comptab = comptab
```
emdupre: Is this line necessary ?

tedana/selection/select_comps.py
```diff
@@ -63,18 +67,7 @@ def selcomps(seldict, mmix, mask, ref_img, manacc, n_echos, t2s, s0, olevel=2,
...
-        Indices of rejected (non-BOLD) components in `mmix`
-    midk : :obj:`list`
-        Indices of mid-K (questionable) components in `mmix`
-        These components are typically removed from the data during denoising
-    ign : :obj:`list`
-        Indices of ignored components in `mmix`
-        Ignored components are considered to have too low variance to matter.
-        They are not processed through the accept vs reject decision tree and
-        are NOT removed during the denoising process
+    comptable
```
emdupre: Is there a docstring, here?

tedana/decomposition/eigendecomp.py
tsalo: There is a normalized variance explained in `fitmodels_direct`, although I don't think it's ever used:

https://github.com/tsalo/tedana/blob/869b0842cb95b21f67cd79bd08dd995db6e48c39/tedana/model/fit.py#L207-L211

And then there's the second one we see above in `tedpca`. `variance explained (normalized)` is on a percentage scale (0-100), while `variance explained 2 (normalized)` is on a 0-1 scale. Even accounting for that, though, they aren't the same.  I can definitely rename one or both of them, and it might be worth it to drop the one in `fitmodels_direct`.

tedana/decomposition/eigendecomp.py
tsalo: It wasn't originally written out, but I thought it might be worth it to do so. Ultimately, when we get around to #17, it would be good to make this output optional (e.g., with `--verbose`).

tedana/model/fit.py
tsalo: Probably not. Good catch.

tedana/decomposition/eigendecomp.py
tsalo: I'll replace `u` with `voxel_comp_weights` (or something better if you can think of it).

tedana/decomposition/eigendecomp.py
tsalo: I'd be happy to, though I don't know what they are. Any thoughts on what they could be called?

tedana/decomposition/eigendecomp.py
tsalo: I'd prefer something more readable than single letters, but it doesn't necessarily have to be anything as long as `voxel_comp_weights`.

emdupre: What about `left_singular_vectors`, `right_singular_vectors`, and `singular_values` ? :laughing: 

This is a tough one, really. I usually am strongly against single variable names, but here it's such a strong convention that to break from it is almost confusing.

Maybe we could have @rmarkello or @handwerkerd weigh in.

tedana/decomposition/eigendecomp.py
```diff
@@ -143,21 +143,22 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
...
-            u = np.dot(np.dot(dz, v.T), np.diag(1. / s))
+            comp_ts = ppca.components_
+            varex = ppca.explained_variance_
+            voxel_comp_weights = np.dot(np.dot(dz, comp_ts.T), np.diag(1. / varex))
         else:
-            u, s, v = np.linalg.svd(dz, full_matrices=0)
+            voxel_comp_weights, varex, comp_ts = np.linalg.svd(dz, full_matrices=0)
 
         # actual variance explained (normalized)
-        sp = s / s.sum()
```
emdupre: As an alternative, maybe what we should be doing is -- starting here-- break the above into a separate function, returning `u`, `s`, and `v`, so they can be imported as more meaningful variable names where they're actually used.

tedana/decomposition/eigendecomp.py
rmarkello: Oof, this is tough. I've struggled with this in my own code that uses SVD, and I eventually settled on abiding by the mathematical conventions (using `u`, `s`, and `v`) with copious comments above the assignment detailing what they represent.

I think @emdupre's [comment below](https://github.com/ME-ICA/tedana/pull/122#discussion_r215481102) about splicing this out into its own function might be the best move!

tedana/decomposition/eigendecomp.py
rmarkello: :+1: for moving this into its own function!

tedana/decomposition/eigendecomp.py
emdupre: Suggested names:
 - `diff_varex_norm` for `spdif` -- it's the difference between normalized, successive singular values
 - `lower_diff_varex_norm` for `spdifh` -- it's the second half of `spdif`
 - `varex_norm_thr` for `spdthr` -- it's a threshold determined by finding the average of the maximum of `spdifh` and the minimum of `spdif`

tedana/decomposition/eigendecomp.py
emdupre: Agreed ! I feel like this could be addressed (at least in part) alongside #106 ...

tedana/decomposition/eigendecomp.py
tsalo: New function (with `u`, `v`, and `s`) created.

tedana/decomposition/eigendecomp.py
tsalo: Let me know what you both think of the new function.

rmarkello: :+1: looks good to me.

tedana/decomposition/eigendecomp.py
```diff
@@ -18,6 +18,23 @@
 Z_MAX = 8
 
 
+def run_pca(data, mlepca=True):
+    """
+    Run PCA or SVD.
```
emdupre: This PCA is an SVD, so I find this docstring / function name a little confusing. Maybe we could say "run_svd" instead ?
The difference afaict seems to be that `mle` option automatically estimates the number of components using the method from Thomas P. Minka: Automatic Choice of Dimensionality for PCA. NIPS 2000: 598-604.

tsalo: Done. Do you want me to include the Minka citation with duecredit? It will involve splitting off parts of `run_svd` into a `_run_mlepca` (with the citation) and a `_run_svd` (without). Sort of like what we did with `make_optcom`:
https://github.com/ME-ICA/tedana/blob/cb01d725d05f0102e2da89ad50c32905c68a79fd/tedana/model/combine.py#L153-L156

tsalo: In a perfect world, the citation would be generated by `sklearn`, but I don't see why we can't do it ourselves. Done!

tedana/decomposition/eigendecomp.py
```diff
@@ -11,13 +11,68 @@
...
+    from sklearn.decomposition import PCA
+    ppca = PCA(n_components='mle', svd_solver='full')
+    ppca.fit(data)
+    v = ppca.components_
+    s = ppca.explained_variance_
+    u = np.dot(np.dot(data, v.T), np.diag(1. / s))
+    return u, s, v
+
+
+def run_svd(data, mlepca=True):
```
emdupre: A more general question here -- we don't call this with `mlepca=False` anywhere in the codebase. What do you think is the benefit of keeping this flag, if any ? I'm genuinely unsure...

tsalo: I noticed the same thing too. I just assumed that we would support `mlepca=False` with a new argument at the command line at some point, but I don't know if it's worth it.

emdupre: I don't think so, personally. At this point honestly I'm worried we'll need to add more dimensionality constraints rather than fewer (which the `np.linalg.svd` would point towards). How would you feel about just getting rid of that option ? Sorry for all of the iterating, here !

tsalo: Happy to.

emdupre:
After chatting with @handwerkerd it sounds as though we might want to drop the PCA decision tree entirely. Given the obvious relevance to this PR, I thought I would bring it up here. I can also open a separate issue for broader discussion !

tsalo:
That fits in with #101, right? Depending on the results (which I don't know anything about), it sounds like it makes sense to use a single threshold, but I think that the PCA component selection should be treated the same way that ICA component selection is (or will be, really). Namely, different approaches can be implemented in different functions, even if they aren't all incorporated as arguments in the tedana workflow.

emdupre:
Ok, re-visiting this, and thinking in the context of #101, it sounds like we should re-enable the `non-mlepca` option s.t., if dimensionality is not automatically estimated via sklearn, the "classical" decision tree is employed.

We can make that a user-accessible option, or re-purpose the existing "stabilize" or "convergence" options for that purpose. I'd argue we should have `mlepca` as the default, unless the user requests otherwise. What do you think @tsalo (and @handwerkerd) ?

tsalo:
I like this, although I don't think we should rely on a boolean argument to specify the TEDPCA decomposition and selection methods, because we'll be expecting more to be added over time. How about something like `method` with the options `'mlepca'` and `'kundu_v2_5'`? That way, it would match what you're doing in #119.

emdupre:
I actually hope that, over the long term, we're  going to be collapsing `kundu_v2.5` and `kundu_v3` as we better understand and validate v3 of the selection criteria. This is, of course, something that we should be discussing in the context of a larger tedana roadmap, which I am very slowly in the process of drafting for discussion. Thanks to literally everyone for bearing with me.

Regardless, I'm really not sure this in particular (i.e., the PCA selection) is a step where we expect more decision trees to be added, or to just refine the existing decision tree. I'm ok with the inconvenience of waiting until a later stable version release (e.g., 0.2) to change this behavior if we need to, but I think for our first stable release a boolean flag to trigger the "short circuit" of the decision tree is probably easiest to introduce. Does that seem reasonable, @tsalo ? Anxious to hear what you think !

tsalo:
It's easy enough for me to implement the `method` switch, I think. I've done 90% of the work locally (without committing). It's also pretty easy to use the boolean switch too. It's also possible to make a boolean flag in the workflow correspond to a specific option in `tedpca`, so that `tedpca` could implement multiple methods but only a couple would be accessible within the canonical workflow.

I guess it goes back to the extent to which `tedana` will be a toolbox vs. a workflow. I don't know how many different kinds of PCA or ICA component selection are possible/reasonable, nor do I know how many we would want to implement in `tedana` specifically. I think that the roadmap and discussions in #131 will help clarify this. The structures we've been discussing in #131 would make it easy to have as many PCA or ICA decomposition or selection methods as we want, but that's only if we want them all. In the meantime, I can implement the switch however's easiest.

KirstieJane:
Hi @tsalo! This is super cool - I'm really excited about making this process more transparent!

Would you mind changing the precision in the file write out to something like 5 or 6 decimal places? We probably don't need 18 dp, right?

tsalo:
@KirstieJane Absolutely. It should be good now.

emdupre:
Thanks for being so quick to get to this, @tsalo !! We're a few time zones ahead of you, so I know this is quite early your time :smile:

@KirstieJane reached out on gitter with a question about the `%g` formatting, but other than that I'm happy to merge. Looking forward to having this in !

Issue #39, opened 2018-05-10
orthogonalization of ME-ICA components


afni-rickr:
As it stands, tedana.py fits all components to the time series and removes the resulting fit of the bad, naughty terms.  But the 'bad' components are not orthogonal to the 'good' ones, meaning this regression still removes some of the good components (though less than it would had the good components not been in the model).

Being more conservative, one might think that if the good components are really good BOLD, it would be preferable to not remove any of them.  To avoid removing any of the good BOLD components from the data one can project the good components out of the bad ones, and then project the resulting "very bad" (clean bad? pure evil?) out of the data.

This is what afni_proc.py now does with 'tedort' (tedana.py orthogonalized terms) -combine_method choices of 'OC_tedort' (AFNI's OC (~same as tedana.py but with no masking), and passing the "pure evil" terms along to the typical linear regression block) and 'tedana_OC_tedort' (tedana.py's OC, plus passing the tedort terms to the final regression block).

What do you think @emdupre, @CesarCaballeroGaudes, @dowdlelt and @handwerkerd (to drag you all into this thread... :).  I am planning to run a comparison of 'tedana' and 'tedana_OC_tedort' tomorrow, though it will be hard to put any measure on it.  But it seems preferable to not remove any of the signal that we are really interested in.

dowdlelt:
I like this, and look forward to contributing on the testing of this idea. I would like to add some questions and concerns to this however. 

Sometimes I notice that tedana will identify components that correlate highly with motion regressors but not remove them - thus I am inclined to include standard motion regressors (and their derivatives) in the GLM. In other cases, tedana will find a portion of a motion regressor. That is, it finds a components, identifies it as noise, but in comparing it to the motion regressors it matches to only a portion of the motion estimates (say a spike, or a part after a sharp delta in motion). At this junction, I am nervous about including motion regressors in the model - part of their timeseries is in a tedana component. Thus, if I include them in the model, they will not truly fit the noise that I want them to remove - they may in fact add noise back in!

This extends to ideas like high pass filtering - tedana tends to remove low frequency components (though, sometimes they are ignored :( ...) so should include polorts in my model, or no? Should we project out the tedana outputs from all the baseline (noise) regressors in the model? I'm not an expert with GLMs so I may be making a large, embarrassing mistake here...but it just seems foolish to have these multiple methods of 'denoising' in a model (highpass filtering, motion regression, tedana) all interacting in unknown ways. 

[...]

PR #125, opened 2018-09-09
[FIX] Include ignored components in ME-DN T1c time series

tsalo:
The ME-DN time series should include the ignored ICA components, but does not. The current approach is to regress out all ICA components from the data to get `resid`, and then to combine that with the BOLD-T1c (GS-controlled accepted components) time series as the ME-DN T1c time series. With this PR, the residuals time series will not have the ignored components regressed out, and thus the ME-DN time series will include those components.

- Add `rej` (index of rejected components) argument to `gscontrol_mmix`.
- Compose ICA residuals time series  (`resid`) of the data minus all of the ICA components _except_ the ignored components.
- There are some minor improvements to documentation as well.

### DIFF
--- a/tedana/utils/io.py
+++ b/tedana/utils/io.py
@@ -40,11 +40,11 @@ def gscontrol_mmix(optcom_ts, mmix, mask, acc, ign, ref_img):
     ======================    =================================================
     Filename                  Content
     ======================    =================================================
-    sphis_hik.nii             T1-like effect.
-    hik_ts_OC_T1c.nii         T1 corrected time series.
-    dn_ts_OC_T1c.nii          Denoised version of T1 corrected time series
-    betas_hik_OC_T1c.nii      T1-GS corrected components
-    meica_mix_T1c.1D          T1-GS corrected mixing matrix
+    sphis_hik.nii             T1-like effect
+    hik_ts_OC_T1c.nii         T1-corrected BOLD (high-Kappa) time series
+    dn_ts_OC_T1c.nii          Denoised version of T1-corrected time series
+    betas_hik_OC_T1c.nii      T1 global signal-corrected components
+    meica_mix_T1c.1D          T1 global signal-corrected mixing matrix
     ======================    =================================================
     """
     optcom_masked = optcom_ts[mask, :]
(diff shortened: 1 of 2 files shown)


emdupre:
This is a great catch, @tsalo ! I'd really love @handwerkerd's review and to hear his opinion on the T1c series, generally, if he has time.

I'm just not sure I see the benefit of regressing the global signal instead of, say, the rejected comments. So, I'm thinking it might make sense to repurpose this into a kind of aggressive denoising, as discussed in #39. WDYT ?

tsalo:
Isn't the removal of T1 global signal largely there based on the [Power paper](http://www.pnas.org/content/early/2018/02/07/1720985115.short)? I know that this particular approach, while recommended, hasn't been evaluated in any publications, which might be one reason to avoid making it mandatory. When we add in GODEC in #120, we'll be adding an argument (currently `ws_denoise`) to allow users to specify what kind of post-tedana denoising, if any, they'd like to implement in order to remove "widespread noise." We can make `None` the default at that point, I think.

I know that it's probably a good idea to orthogonalize the bad components with respect to the good ones, and we can do that with `lstsq` in `writeresults` pretty easily, right? I don't know how that interacts with the T1c GSR, although I wasn't involved in #39 and couldn't figure out where GSR was discussed. Could you elaborate on that a bit more?

handwerkerd:
I've been a bit behind at looking into code, but please please please do not regress out the whole brain global signal across the brain as a default setting. I'm familiar with the Power papers, but there are serious issues with setting GSR as a default. I can talk more about this at some point, but not this week.

tsalo:
Just to be clear, are you referring to the standard GSR that is done to the concatenated and optimally combined data [here](https://github.com/ME-ICA/tedana/blob/ad7b30d953572f5f874284d5d5e4ac836f7d50c1/tedana/workflows/tedana.py#L390-L391) or to the minimal image regression (aka mmix GSR, aka T1-correction, aka who-knows-what because there's no canonical name afaik) done [here](https://github.com/ME-ICA/tedana/blob/ad7b30d953572f5f874284d5d5e4ac836f7d50c1/tedana/workflows/tedana.py#L435) (or both)?

Both are done by default, but the T1-correction outputs additional files (and so doesn't have to affect results, depending on which files people use), while the standard GSR impacts all downstream outputs.

I'm hoping to re-integrate GODEC in #120, which will (in my mind) involve modularizing the post-TEDICA processing and will offer an alternative to the T1-correction. We'll need to add an argument for the post-TEDICA processing to perform, but I _do_ have the default as `None`, which I think is fair.

handwerkerd:
I think both. I don't remember GSR as a standard step in earlier pipelines, but I'm not sure I specifically looked for it.  If I'm following the first one correctly regressing out the global signal before ICA seems particularly problematic.

tsalo:
As best I can tell, both GSR steps were first added [here](https://bitbucket.org/prantikk/me-ica/diff/meica.libs/tedana.py?diff2=906bd1f6db70&at=v3.2#Lmeica.libs/tedana.pyT746), on 2017/05/23, in the initial shift from v2.5 to v3.2, so I don't think either step is used in any "official" releases of the pipeline.

Given that we're planning to roll back to v2.5 in #119, perhaps we can use that PR to change the defaults to what they were at that time. We can still retain the T1-correction, but can turn it off by default (as is the plan in #120).

emdupre:
It looks like this is ready -- to be clear, this is just updating the ME-DN-T1c to (theoretically) match the creation of the ME-DN, no ?

I know we had a lot of changes since this was first created, so I wanted to make sure it's not intended to do anything else !

tsalo:
That's right, that's all the changes are meant to do.

I just realized that I forgot to update the docstring for `gscontrol_mmix`, so there will be one more commit in just a second to fix that.

PR #126, opened 2018-09-09
[DOC] Add Python version info

### LINKED ISSUE #124: Versions of python that tedana works with
Hiya,

I can't seem to find in the documentation but it seems that the latest version works with python 3.6 , and from testing It didn't work with python2.7 -- I was getting importing errors -- I assume the syntax to import modules is slightly different in python 3.

tsalo:
Closes #124.

- Specify minimum Python version in README.
- Add Python version badges tied to PyPi to README and docs index page.
- Make duecredit’s import warning sound more like a warning and less
like an error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 
 python:
     - 3.6
-    - 3.7
+    - 3.7-dev
 
 matrix:
   include:
@@ -19,9 +19,11 @@ before_install:
     - pip install --upgrade virtualenv
     - if [ "${STYLE}" == "1" ]; then
       pip install "flake8<3.0" flake8-putty;
+      else
+      pip install pytest;
       fi
     - if [ "${COVERAGE}" == "1" ]; then
-      pip install coverage coveralls codecov pytest pytest-cov;
+      pip install coverage coveralls codecov pytest-cov;
       fi
 
 script:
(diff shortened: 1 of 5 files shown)


tsalo:
@KirstieJane Thank you for reviewing! I'll wait until @emdupre has a chance to review the code itself (not that much is changed), but I think that the documentation is the most important aspect so I'm very glad that the result is readable.

emdupre:
Thanks so much @tsalo ! I actually don't know if it works 3.7, too, now, but I'd imagine so !

Depending on how you feel about the need to check that, I have one suggestion: Can we have the badge and the README text match (i.e., either change to say "3.6, 3.7" in the badge or "= 3.6" in the text) ?

tsalo:
I think it makes the most sense to include 3.7 in the Travis CI. I can do that in this PR.

I don't know if PyPi automatically detects supported Python versions from CI or something, or if you have to manually specify, but you might have to change tedana's PyPi settings to support 3.7. That should update the badge.

emdupre:
Thanks ! Happy to see it passes with 3.7 :smile: I think you need to change [this line](https://github.com/ME-ICA/tedana/blob/cb01d725d05f0102e2da89ad50c32905c68a79fd/tedana/info.py#L68) to update it on PyPi (for the next release)

emdupre:
Thanks for adding it in ! This LGTM -- merging :tada:

PR #128, opened 2018-09-13
[FIX] Accept non-NIFTI files without complaining

rmarkello:
- If non-NIFTI files (e.g., BRIK/HEAD) are provided, don't complain and just handle it as needed. This involves re-writing `utils.filewrite` and `new_nii_like` to accommodate the fact that `ref_img` may not be a nifti image.


emdupre:
Thank you !!! :tada:

PR #133, opened 2018-09-23
[DOC] Improve documentation for pipeline

tsalo:

- Rename Code_of_Conduct.md to CODE_OF_CONDUCT.md so that GitHub can detect it.
- ~~Change README over to rst so that we can use it in both the repository and in the RTD site.~~
- Add HBM 2018 poster to README.
- Add `Processing pipeline details` page to docs with general pipeline figure. This figure can definitely be improved, but it's a start.
- Reorganize RTD contents to better mirror other projects, like fMRIPrep.

The README will need to be fixed up a bit in order for the poster to show up on GitHub. Also, GitHub won't run the function used to build the package reference.


README.rst
```diff
@@ -0,0 +1,154 @@
+tedana: TE Dependent ANAlysis
+=============================
+
+The ``tedana`` package is part of the ME-ICA pipeline, performing TE-dependent
+analysis of multi-echo functional magnetic resonance imaging (fMRI) data.
+``TE``-``de``\pendent ``ana``\lysis (_tedana_) is a Python module for denoising
```
emdupre: I think the underscores around `_tedana_` will need to be converted to asteriks (i.e., `*tedana*`) to render properly in rst

README.rst
```diff
@@ -0,0 +1,154 @@
...
+
+.. image:: https://img.shields.io/pypi/pyversions/tedana.svg
+   :target: https://pypi.python.org/pypi/tedana/
+   :alt: PyPI - Python Version
+
+.. image:: https://zenodo.org/badge/110845855.svg
+   :target: https://zenodo.org/badge/latestdoi/110845855
+   :alt: DOI
+
+.. image:: https://circleci.com/gh/ME-ICA/tedana.svg?style=svg
```
emdupre: Can we adopt the shield-style badge, here ? Before it was a shield-style in the README and an svg-style badge in the docs, but I think I prefer the shield ! So it would be:
```
.. image:: https://circleci.com/gh/ME-ICA/tedana.svg?style=shield
```

emdupre:
Yes, it's unfortunate that we can't have the citation-pulling javascript execute on both the RTD site and the main README.

Given that both the citation as well as the poster won't display, is there a strong reason to have the RTD site and README pull from the same source file ? I'm happy to have the README as a subset of the RTD site, since I think they serve slightly different purposes.. Let me know what you think.

README.rst
```diff
@@ -0,0 +1,154 @@
...
+You can then install ``tedana`` with:
+
+.. code-block:: bash
+
+  pip install tedana
+
+Getting involved
+----------------
+
+We :yellow_heart: new contributors!
```
tsalo: RST doesn't support emojis, unfortunately ☹️. We can swap this out with `<3` or "love" or something, though.

tsalo:
The poster will show up once this PR is merged if we switch to using the URL [like fMRIPrep does](https://github.com/poldracklab/fmriprep/blob/master/README.rst). The file just needs to exist within the repository, which will require uploading it, getting the stable link to the image, and then updating the README.

They get around the citation rendering by having the citation on a separate page ([Citing fMRIPREP](https://fmriprep.readthedocs.io/en/stable/citing.html)). It might be worth it for us to do the same thing.

I think using the same source is a good idea because it reduces the number of things we have to update when we make changes to either the site or the repository. If you would like the README to be more comprehensive than the RTD index page, in order to summarize more of the site (but more concisely), I can definitely understand that. I just think that reducing duplication will help keep one page from ending up out-of-date.

README.rst
tsalo: 👍

README.rst
tsalo: 👍

tsalo:
I'm adding a series of plots to the pipeline description page, which will hopefully make tedana's approach easier to understand for new users. It's hard to evaluate the raw rst files in `docs/`, so I exported the html file to a pdf, which is attached. I'd love to know what everyone thinks.

[Processing pipeline details — tedana 0.0.5-rc2+26.g37b0384.dirty documentation.pdf](https://github.com/ME-ICA/tedana/files/2444278/Processing.pipeline.details.tedana.0.0.5-rc2%2B26.g37b0384.dirty.documentation.pdf)

dowdlelt:
@tsalo That is a fantastic doc, and adds a lot of clarity to the entire process. I feel like it is an excellent introduction for folks who want to know what tedana is doing.

tsalo:
@dowdlelt I'm glad it looks good. Going through step by step, writing out new files, and generating these figures definitely helped me understand some of the finer points of certain steps (especially the data whitening, ICA decomposition, and model fitting), so hopefully it will help newcomers as well.

The figures require the 5-echo dataset and some adapted outputs from the tedana workflow run on those data, but I've uploaded the notebooks that made them anyway. I guess we can figure out how to make it reproducible later. In the meantime I can adjust any of the figures as needed in this PR.

I guess all we need now is a way to generate an overall workflow diagram. The one I have in this PR was made in PowerPoint, so it's not the best solution.

emdupre:
@KirstieJane if you have time it'd be fantastic to have your feedback here on the workflow walk-through !

tsalo:
@KirstieJane Here is the current version with the adaptive mask information and figure as well.

[Processing pipeline details.pdf](https://github.com/ME-ICA/tedana/files/2446675/Processing.pipeline.details.pdf)

dowdlelt:
I like that notebook. I may take take a stab at an addition or two - one thing that I think future users may benefit from is a better appreciation of some of the output. The timeseries plots are very illustrative for 'what' is happening, but a spatial view from the betas_OC.nii output might give folks an understanding of the 'where' those time series correspond to. Not critical, and there are only a few outputs where this may be useful. Perhaps that would be better for a separate document, rather than this one.

README.rst
KirstieJane: I think it works if you just copy and paste in the emoji (eg: from https://emojipedia.org/yellow-heart) (based on [this stackoverflow answer](https://stackoverflow.com/a/42256239)) could you give that a bash @tsalo and see if it works? Otherwise I'm totally fine with `<3` as you suggested!

KirstieJane:
So sorry I missed this last week. I'm having a reasonably quick read through now (thank you for making the PDF - that makes reviewing **so** much easier!)

First comment - I'm wildly into `TEDPCA` and `TEDICA` 😆 - fab branding.



Ok, no, really the first comment is that this is fantastic. Really really valuable and so helpful to read through all the various steps.

I like the idea of linking to the jupyter notebook from the write up to show that these figures didn't come from nowhere. Could you "binderise" the repo? (https://mybinder.org) It looks like you'd need to host the data somewhere - is that possible? Even a dropbox link is fine for a demo notebook in my opinion.

The README as `.rst` gives me a little pause because it makes it a little harder for people to add to or edit. But I'm unhelpfully very much on the fence about whether it's important for the README to be part of the read the docs site or duplicated information (see @emdupre's comment above: https://github.com/ME-ICA/tedana/pull/133#issuecomment-424067728). I can totally see both sides: less likely to go out of date if they're the same file, potentially fulfilling different use cases otherwise. 🤷‍♀️ sorry I'm no help there! (Except for saying, there's no wrong answer - go with your favourite in my opinion.)

The last thing I'll say is that I don't think this PR has to be perfect in order for it to be merged. Having people add to this part of the documentation is a great first contribution (eg from current researchers starting to use `tedana`:tada:

And REALLY well done on getting so much clarity up in the docs! 💖

README.rst
tsalo: It works! Thanks!

tsalo:
@KirstieJane Thank you so much for taking a look! I'm glad it's looking good.

Binderizing the notebooks is a great idea. If we can make the data accessible somewhere, we also move the notebooks into `examples/` instead of keeping them hidden in `docs/_static/`. @emdupre I know that you were looking to upload the single-run 5-echo dataset to OpenNeuro at some point. Is that still the plan? If so, we can use nistats to download the files in the notebooks. Otherwise, is there a good place to upload those files? I know you wanted to move away from Dropbox for the integration tests, so I assume that holds for the 5-echo dataset as well.

Regarding the README, the consensus seems to be that consolidating the RTD and GitHub READMEs is more effort (especially for new contributors) than it's worth. I can split it back up into a markdown file for the README and an rst file in `docs/` for the site. I think it's worth it to keep the content synchronized, even if that's just something we check for in PRs rather than something automated. Does that sound good to everyone?

tsalo:
I think that this PR is almost ready to merge, but there are a couple of things I'd like to figure out before we do it:
1. Can we add an argument (e.g., `verbose`) that will determine whether a lot of the extra files are generated? There are a handful of new files I'd like to output to make the figure notebooks reproducible, but I don't want to increase the number of unused workflow outputs.
2. While it _was_ my idea to document the workflow derivatives in the respective workflows' docstrings, in order to keep the list of outputs closer to the actual code to reduce the chances of one becoming out-of-date over time, it seems like people would prefer just to look at a Derivatives page on the RTD site (which is currently _way_ out of date). How would everyone feel about removing the output file tables from the docstrings and just using the RTD site to list and describe the outputs?

emdupre:
Sorry for not directly responding on whether it's worth keeping the content synchronized !!! I totally agree that it is worth checking in the PRs directly. Do you think we should add it to [the HTML of the PR template](https://github.com/ME-ICA/tedana/blob/master/.github/pull_request_template.md) ?

Regarding a `--verbose` flag, I'm not yet ready to commit to that mostly since we need to do so many other steps to get there (like deciding on defaults to know which files are usually going to be generated). But I absolutely agree that it needs to be added later ! Can we open another issue to make sure that's on the to-do list ?

Regarding removing the file strings, I think it would actually be valuable to keep them *and* add the derivatives description to the RTD site !

Other than that, I'm ready to merge this in. LMK what you think !

KirstieJane:
I've made an issue #149 so that the binder request doesn't hold up this merge 😸. So everything looks great to me ✨

README.md
```diff
@@ -1,6 +1,10 @@
-# tedana
+tedana: TE Dependent ANAlysis
+=============================
 
-`TE`-`de`pendent `ana`lysis (_tedana_) is a Python module for denoising multi-echo functional magnetic resonance imaging (fMRI) data.
+The ``tedana`` package is part of the ME-ICA pipeline, performing TE-dependent
+analysis of multi-echo functional magnetic resonance imaging (fMRI) data.
+``TE``-``de``\pendent ``ana``\lysis (``tedana``) is a Python module for denoising
```
emdupre: Can we remove the escape characters ('\') here ? They're being rendered as text.

README.md
```diff
@@ -11,55 +15,48 @@
...
+The ME-ICA pipeline originally performed both pre-processing and TE-dependent
+analysis of multi-echo fMRI data; however, ``tedana`` now assumes that you're
+working with data which has been previously preprocessed.
+If you're in need of a preprocessing pipeline, we recommend
+[fmriprep](https://github.com/poldracklab/fmriprep/), which has been tested
+for compatibility with multi-echo fMRI data and ``tedana``.
 
-## About
+![http://tedana.readthedocs.io/](https://user-images.githubusercontent.com/7406227/40031156-57b7cbb8-57bc-11e8-8c51-5b29f2e86a48.png)
+![](https://github.com/tsalo/tedana/blob/doc-pipeline/docs/_static/tedana-poster.png)
```
emdupre: Could we remove this from the README and put it on the landing page of the RTD site ? Maybe under citations ?

emdupre:
I made a few review comments, but they're small enough that I went ahead and just did them myself. If you dislike any of them, though, just let me know and I'm happy to revert !

tsalo:
I think that managing two lists of outputs in parallel (on the RTD site and in the workflow docstring) is going to be difficult in the long run. I even managed to get them out of sync with one another over the course of this one PR!  Still, I'm happy to wait on dealing with new or optional outputs until a future PR.

I think it's ready to merge if you're happy with it.

emdupre:
Oh no ! That's a great point, but I'm anxious to see this in and am happy to deal with the trouble for now -- especially since all of the output names are likely to change soon and we'll need to revisit this in depth.

Maybe we can make a new issue to make a decision on which of the output descriptions we keep ?

tsalo:
I don't think it requires a separate issue. Few enough people will look at the API section of the site compared to an Outputs/Derivatives page that I think that's the one we should ultimately drop.

emdupre:
OK, but the API should be kept up-to-date when interacting with the code directly, right ? And sorry, I botched the merge conflict resolution -- fixing now.

tsalo:
AFAIK, most function docstrings don't list all of the files that are generated, so I always saw the inclusion of those tables as something cool, but extra. We can include a link to the Derivatives page in the docstring, so that the API links to the relevant Derivatives page without duplicating the full list of files. That's what we're doing with the Usage page (i.e., Usage links to API). A solution could be to link both Usage and API to Derivatives.

emdupre:
Ok, that seems fine as a future direction. Let's go ahead and remove those for now, then ? After that, if Travis passes I'll merge :tada:

tsalo:
Awesome! I think it's ready to merge now.

PR #139, opened 2018-10-08
[NO-MRG] Pin environment

emdupre:
Testing environmental changes to patch persistent CircleCI failures. 

Not intended to be merged.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy>=1.14
-scikit-learn
+scikit-learn==0.19.0
 scipy
 mdp
 nilearn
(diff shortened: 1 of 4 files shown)


emdupre:
Since I highly doubt the current pin will work, documenting my understanding of the situation in case others can chime in:

- The docker image hash for our passing builds and failing builds is identical
- Diff'ing the two environment creation text logs from circle suggest that all pulled Python dependencies are identical
- All outputs from `tedana` UP TO `tedpcc` are identical
- In the pickle file output from `tedpca` the original decomposition is different, specifically in the *last component*

emdupre:
Per a decision on the last developer call, we have decided to allow breaking changes, and I'm therefore closing this PR. Check the (incoming) Roadmap for more details.

PR #136, opened 2018-10-04
[ENH] Rename modules

frodeaa:
Closes # https://github.com/ME-ICA/tedana/issues/135.

Changes proposed in this pull request:
 - renaming `model.monoexponential` to `decay`
 - renaming `model.combine to combine`
 - reorganizing `io` and `utils`
   - move `new_nii_like()`, `filewrite()`, `load_data()` from `utils` to `io`

remains to fix the documentation `docs/api.rst`

PR #152, opened 2018-11-01
[ENH] Make outputs BIDS derivatives-compatible

### LINKED ISSUE #146: Update output names to match BIDS derivatives RC2 conventions

### LINKED ISSUE #180: Output component tables as BIDS Derivatives-compatible jsons
After a discussion with Chris M., we think we should use jsons to store the component tables. We'll need functions to load DataFrames from jsons and dump DataFrames to jsons as well, I think. We're pushing this out of #152 because it's a bit too much for the one PR.

### LINKED ISSUE #252: T2* maps should be outputted in seconds, not milliseconds
### Summary
Per [BEP001](https://github.com/bids-standard/bep001/blob/e661031158c462c7c7ba9e11d42b99a67fca591e/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md), T2* maps should be in units of seconds, not milliseconds.

### Additional Detail
BEP001 hasn't been merged yet, but it seems likely that the units for these kinds of maps will be set in stone. 

### Next Steps
1. Divide T2* estimates by 1000  when they're written out in the tedana and t2smap workflows.

tsalo:
Closes #146, closes #180, and closes #252.

~~We should wait until #136 is merged to deal with this PR, as this will involve editing every function that generates files.~~

- Add function `gen_fname` to generate BIDS derivatives-compatible filenames from a base file (input file corresponding to first echo), a description for the output file, the data type, and the file extension.
- Replace argument `label` with `out_dir` in workflows.
- Add `bf` (base filename) argument to all file-generating functions.
- Get automatic gzipping working in `filewrite`.
- Remove unused `ctabsel` function.


tedana/io.py
```diff
@@ -16,6 +17,50 @@
...
+
+    Returns
+    -------
+    out_file : :obj:`str`
+        BIDS derivatives-compatible filename
+    """
+    if not all([isinstance(v, str) for k, v in kwargs.items()]):
+        raise ValueError("All keyword arguments must be strings")
+
+    # Remove echo field from filename
```
emdupre: Will this break if given a file without an 'echo' key ? For example, in `fmriprep` we currently remove the echo key after optimal combination, but prior to denoising.

tsalo: It seems to run fine when there's no echo key. My test:
```python
>>> import re
>>> echo_regex = re.compile('_echo-[0-9+]_')
>>> basefile1 = 'sub-01_ses-01_echo-1_run-01_bold.nii.gz'
>>> test1 = re.sub(echo_regex, '_', basefile1)
>>> test1
'sub-01_ses-01_run-01_bold.nii.gz'
>>> basefile2 = 'sub-01_ses-01_run-01_bold.nii.gz'
>>> test2 = re.sub(echo_regex, '_', basefile2)
>>> test2
'sub-01_ses-01_run-01_bold.nii.gz'
```

tsalo:
I think this is ready for review, although we're going to need to fix the integration tests and I'm sure the output filenames will change.

docs/outputs.rst
```diff
@@ -4,97 +4,99 @@ Outputs of tedana
...
+                                           map has a NaN.
+*_desc-optcom_bold.nii.gz                  Optimally combined time series.
+*_desc-optcomDenoised_bold.nii.gz          Denoised optimally combined time series.
+                                           **Recommended dataset for analysis.**
+*_desc-optcomRejected_bold.nii.gz          Combined time series from rejected components.
+*_desc-optcomRejectedMidK_bold.nii.gz      Combined time series from "mid-k" rejected components.
+*_desc-optcomAccepted_bold.nii.gz          High-kappa time series. This dataset does not
+                                           include thermal noise or low variance components.
+                                           Not the recommended dataset for analysis.
+*_desc-TEDPCA_comptable.tsv                TEDPCA component table. A tab-delimited file with
```
tsalo: Replace with json.

docs/outputs.rst
```diff
@@ -4,97 +4,151 @@ Outputs of tedana
...
-betas_hik_OC.nii          High-kappa ICA coefficient feature set
-feats_OC2.nii             Z-normalized spatial component maps
-comp_table_ica.txt        TEDICA component table. A tab-delimited file with
-                          summary metrics and inclusion/exclusion information
-                          for each component from the ICA decomposition.
-======================    =====================================================
+========================================   =====================================================
+Filename                                   Content
+========================================   =====================================================
+*_desc-limited_t2s.nii.gz                  Limited estimated T2* 3D map.
```
tsalo: Suffices `t2s` and `s0` are pending integration into BIDS Derivatives.

docs/outputs.rst
emdupre: Resolving this since we opened #180

docs/outputs.rst
emdupre: The discussion is in https://github.com/bids-standard/bids-specification/issues/127

.circleci/tedana_outputs_verbose.txt
```diff
@@ -1,45 +1,60 @@
...
+p06.SBJ01_S09_Task11_echo-4_desc-accepted_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-4_desc-denoised_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-4_desc-rejected_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-5_desc-accepted_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-5_desc-denoised_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-5_desc-rejected_bold.nii.gz
+
+p06.SBJ01_S09_Task11_desc-TEDICAPredictedR2Values_something.nii.gz
+p06.SBJ01_S09_Task11_desc-TEDICAPredictedS0Values_something.nii.gz
+p06.SBJ01_S09_Task11_echo-1_desc-TEDICA_components.nii.gz
```
tsalo: Here I'm proposing that we split `betas_catd` up into echo-specific maps. I believe that these maps should be 4D with as many volumes as there are components from the associated decomposition. That's why I used the `components` datatype, but I don't think it corresponds directly to the true components definition. WDYT?

.circleci/tedana_outputs_verbose.txt
```diff
@@ -1,45 +1,60 @@
...
+p06.SBJ01_S09_Task11_echo-3_desc-rejected_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-4_desc-accepted_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-4_desc-denoised_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-4_desc-rejected_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-5_desc-accepted_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-5_desc-denoised_bold.nii.gz
+p06.SBJ01_S09_Task11_echo-5_desc-rejected_bold.nii.gz
+
+p06.SBJ01_S09_Task11_desc-TEDICAPredictedR2Values_something.nii.gz
+p06.SBJ01_S09_Task11_desc-TEDICAPredictedS0Values_something.nii.gz
```
tsalo: I just have no idea what to call these. They were "pred_R2" and "pred_S0" in verbose (currently merged).

.circleci/tedana_outputs_verbose.txt
```diff
@@ -1,45 +1,60 @@
...
+p06.SBJ01_S09_Task11_echo-2_desc-TEDPCA_components.nii.gz
+p06.SBJ01_S09_Task11_echo-3_desc-TEDPCA_components.nii.gz
+p06.SBJ01_S09_Task11_echo-4_desc-TEDPCA_components.nii.gz
+p06.SBJ01_S09_Task11_echo-5_desc-TEDPCA_components.nii.gz
+
+
 meica_R2_pred.nii
 meica_S0_pred.nii
 meica_betas_catd.nii
 meica_metric_weights.nii
```
tsalo: I've got no idea what to call the metric_weights files either. They are used for weighted averaging across the brain, so in a sense they're like a nonbinary mask of some kind.

docs/outputs.rst
tsalo: It sounds like we're good to go with T2Starmap and S0map. Should I update this PR with those suffixes?

.circleci/tedana_outputs_verbose.txt
tsalo: I went with `_echo-1_desc-predictedR2ModelTEDICA_weights.nii.gz` and `_echo-1_desc-predictedS0ModelTEDICA_weights.nii.gz`

.circleci/tedana_outputs_verbose.txt
tsalo: I went with `_desc-metricAveragingTEDICA_weights.nii.gz`

tsalo:
This is currently breaking with `png` because the figures are saved to the current working directory instead of the output directory, but once an `out_dir` argument to added to the visualization functions this should work fine.

docs/outputs.rst
```diff
@@ -4,97 +4,151 @@ Outputs of tedana
...
+=================================================   =====================================================
+Filename                                            Content
+=================================================   =====================================================
+*_desc-adaptiveGoodSignal_mask.nii.gz               Adaptive mask. Each voxel has value corresponding to
+                                                    number of echoes with good signal.
+*_desc-ascendingEstimates_t2s.nii.gz                Voxel-wise T2* estimates using ascending numbers
+                                                    of echoes, starting with 2.
+*_desc-ascendingEstimates_s0.nii.gz                 Voxel-wise S0 estimates using ascending numbers
+                                                    of echoes, starting with 2.
+*_desc-full_t2s.nii.gz                              Full T2* map/time series. The difference between
```
emdupre: I think we can just get rid of this `desc`. It should be clear that it's "full" in comparison to the `desc-limited` files above !

docs/outputs.rst
```diff
@@ -4,97 +4,151 @@ Outputs of tedana
...
+                                                    the limited and full maps is that, for voxels
+                                                    affected by dropout where only one echo contains
+                                                    good data, the full map uses the single echo's
+                                                    value while the limited map has a NaN. Only used
+                                                    for optimal combination.
+*_desc-full_s0.nii.gz                               Full S0 map/time series. Only used for optimal
```
emdupre: See previous comment on dropping this `desc` key-value pair !

docs/outputs.rst
```diff
@@ -4,97 +4,151 @@ Outputs of tedana
...
-tsoc_nogs.nii             Optimally combined time series with global signal
-                          removed.
-======================    =====================================================
+=========================================   =====================================================
+Filename                                    Content
+=========================================   =====================================================
+*_desc-T1gs_bold.nii.gz                     Spatial global signal
+*_globalSignal_regressors.tsv               Time series of global signal from optimally combined
+                                            data.
+*_desc-optcomWithGlobalSignal_bold.nii.gz   Optimally combined time series with global signal
```
emdupre: Isn't this already covered on L22 ?

docs/outputs.rst
```diff
@@ -4,97 +4,151 @@ Outputs of tedana
...
-sphis_hik.nii             T1-like effect
-hik_ts_OC_T1c.nii         T1 corrected high-kappa time series by regression
-dn_ts_OC_T1c.nii          T1 corrected denoised time series
-betas_hik_OC_T1c.nii      T1-GS corrected high-kappa components
-meica_mix_T1c.1D          T1-GS corrected mixing matrix
-======================    =====================================================
+==================================================   =====================================================
+Filename                                             Content
+==================================================   =====================================================
+*_desc-optcomAccepted_min.nii.gz                     T1-like effect
```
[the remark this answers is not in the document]
emdupre: Should we comment on the derivs PR to suggest it ?

tedana/decomposition/eigendecomp.py
```diff
@@ -185,6 +185,8 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG,
         Boolean mask array
     ref_img : :obj:`str` or img_like
         Reference image to dictate how outputs are saved to disk
+    bf : :obj:`str`
```
emdupre: Can we rename this to something a little more intuitive ? You mean everything before like `echo-` in the original filename, right ?

emdupre:
I think there are a few general discussion points that we might want to rope in others for, as they extend more broadly into BIDS. In particular, there are two immediate things I can think of that will make our file names much more intuitive.

1. Moving `optcomb` out of `desc`. It would be great if instead we could denote this under the `echo` key, with something like `echo-combined` or `echo-all`. This actually might be something that BEP001 folks already have ideas about (@KirstieJane @handwerkerd ?).

2. Moving accepted/rejected/ignored classifications out of `desc`. I know AROMA shares this problem, and @effigies has already thought about it in that context. It won't make it into the current derivs PR, but we could trial our own a solution here !

tedana/decomposition/eigendecomp.py
tsalo: That part of the filename and the output directory as well. Do you have a preference for the new name? `bf` stands for base filename, so that's a possibility.

docs/outputs.rst
tsalo: I only don't like this because `full` is only used for optimal combination. It contains T2* estimates for voxels with only one good echo, which in my mind makes it less of a "full" map and more of a... liberal map? I'm not sure what to call it, but I think we should rename it. What about changing `limited` to nothing (dropping `desc`) and changing `full` to `liberal`?

docs/outputs.rst
```diff
@@ -4,97 +4,151 @@ Outputs of tedana
...
-                          decomposition.
-hik_ts_e[echo].nii        High-Kappa time series for echo number ``echo``
-midk_ts_e[echo].nii       Mid-Kappa time series for echo number ``echo``
-lowk_ts_e[echo].nii       Low-Kappa time series for echo number ``echo``
-dn_ts_e[echo].nii         Denoised time series for echo number ``echo``
-======================    =====================================================
+=================================================   =====================================================
+Filename                                            Content
+=================================================   =====================================================
+*_desc-adaptiveGoodSignal_mask.nii.gz               Adaptive mask. Each voxel has value corresponding to
```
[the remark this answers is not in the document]
tsalo: 👍

docs/outputs.rst
tsalo: The `gsr` method outputs two files that contain the same contents, I guess, although I think the duplicate is actually `*_desc-optcomNoGlobalSignal_bold.nii.gz`.

tsalo:
@emdupre I've added the decomposition files now. Please let me know what you think of them from the integration test artifacts.

jbteves:
@tsalo and @emdupre is this a WIP? If so, could we update the tag? If not, is there anything I could help review on?

jbteves:
@tsalo You may want to update this branch after the refactor.

tsalo:
@jbteves I'm planning to hold off on updating until we deal with #251, since that will also probably break things.

tsalo:
I'm going to close this. I'll probably circle back and take another crack at it at a later date.

PR #155, opened 2018-11-12
[ENH] Add tedort argument to tedana workflow

### LINKED ISSUE #39: orthogonalization of ME-ICA components

tsalo:
Closes #39.

- Orthogonalize time series TEDICA mixing matrix for rejected components w.r.t. retained components when requested, after component selection but prior to denoising.
- ~~Change classification label for ignored components to "acceptedLowVariance", in line with @handwerkerd's preference and to make it easier to select all retained components without hard-coding in categories.~~ Removed to prevent PR bloat.


tsalo:
I'm not seeing anything. Could GitHub be acting up?

emdupre:
Literally always :laughing: Yes, I refreshed and it looks fine ! I think I saw the previous commit, somehow.

PR #158, opened 2018-11-14
[FIX] Loop through volumes in FIT method

tsalo:
Does not close any issues.

- The FIT method currently fits to the full dataset each time. This should fix that by looping through the volumes in the dataset.


emdupre:
Can you explain why this update is necessary, @tsalo  ? It looks like we could sneak this in to the next release !

tsalo:
No, but the outputs can. That might make it easier to read.

emdupre:
Merging ! Thanks ! :tada:

PR #159, opened 2018-11-23
[DOC] Remove fMRIPrep recommendation from usage page

tsalo:
Given a [recent NeuroStars question](https://neurostars.org/t/using-fmriprep-with-tedana-for-echo-combination/2883) which referenced our documentation, I thought it might be good to clarify the usage documentation regarding fMRIPrep (i.e., that it does not currently support `tedana`, but that there **are** future plans).


tsalo:
Moving this one minor change into #160 and closing.

Issue #83, opened 2018-07-06
Add acquisition documentation


emdupre:
~.@dowdlelt had compiled [a google spreadsheet](https://docs.google.com/spreadsheets/d/1WERojJyxFoqcg_tndUm5Kj0H1UfUc9Ban0jFGGfPaBk/edit?usp=sharing) of published multi-echo scanning parameters. It would be great to include this in a more general, motivating section of the [readthedocs](tedana.readthedocs.io) on what mutli-echo EPI acquisition entails, such as echo times, effective echo spacing, in-plane acceleration, etc.~

~I _think_ that @dowdlelt is getting started on this ! If there's anything else that could / should be added, it'd be great to keep track of it here.~ 

This is done, now ! :tada: :sparkles: 

Following our May tedana developers calls, @jbteves came up with the following actionables:
- [x] link datasets and their sequence parameters (fixed by #284)
- [ ] add parameter adjustment locations for various scanner manufacturers
- [ ] link to CMRR for guidance on getting sequences at institutions

I think @handwerkerd @dowdlelt and @tsalo (currently) have the most expertise on this, but it would of course be great to get feedback from anyone interested in acquiring multi-echo data !

dowdlelt:
Strong motivation to make some progress here - I've nearly finished adding details to what is on the spreadsheet - but I'm quite certain I am still missing some papers. I'm not strong with the idea of adding to the documentation, but I'm still motivated to figure it out. I'll keep chipping away at it - for delays, please blame the world cup. Thanks for the shout out

tsalo:
To consolidate discussion from Gitter, here are some recommendations for doing multi-echo with the [CMRR MB sequence](https://www.cmrr.umn.edu/multiband/). They are Siemens-specific (and maybe even 3T-specific as well), but may be useful generally:
- Number and timing of echoes can be set in "Contrasts" in Sequence, Part 1 tab.
- If you want short TEs and are using GRAPPA acceleration, use an excitation pulse of 4000-4200 and enable LeakBlock (per @handwerkerd).
    - LeakBlock is set in the Sequence, Special tab.
    - If banding occurs with the above settings or you don't care about short TEs, consider extending the excitation pulse to 8180 (per @dowdlelt) 
- If using GRAPPA acceleration, use GRE with default FA=12 (good performance per @handwerkerd) or use FLEET with FA=15 and 5 dummy scans (MGH multi-echo recommendations and used by @dowdlelt), but do not use single-shot EPI.
[...]

PR #160, opened 2018-11-24
[DOC] Add FAQ and link to ME papers spreadsheet

tsalo:
References #83.

- Add link to @dowdlelt's Google spreadsheet of multi-echo papers and acquisition parameters to Multi-echo resources page.
- Add FAQ section to **Support and communication** with some common questions.
- Move information about integrating `tedana` into preprocessing pipelines from **Usage** into **FAQ**.
- Change title-case page and section titles to sentence case.


docs/outputs.rst
```diff
@@ -23,7 +23,8 @@ ts_OC.nii                 Optimally combined time series.
 dn_ts_OC.nii              Denoised optimally combined time series.
 lowk_ts_OC.nii            Combined time series from rejected components.
 midk_ts_OC.nii            Combined time series from "mid-k" rejected components.
-hik_ts_OC.nii             High-kappa time series.
+hik_ts_OC.nii             Time series of BOLD-like accepted components constructed
```
emdupre: Can we add that this is missing thermal noise, here ?

Otherwise I feel like it seems as though we're recommending this output, when I personally (and I think @handwerkerd said something to this effect, too) would not recommend it for most use cases.

docs/usage.rst
```diff
@@ -1,17 +1,19 @@
...
+  #. Functional datasets equal to the number of acquired echoes
 
 But you can supply many other options, viewable with ``tedana -h`` or
 ``t2smap -h``.
 
 For most use cases, we recommend that users call tedana from within existing fMRI preprocessing
-pipelines such as `fMRIPrep`_ or `afni_proc.py`_..
-Users can also construct their own preprocessing pipelines from which to call tedana;
+pipelines such as `fMRIPrep`_ or `afni_proc.py`_.
+fMRIPrep currently supports optimal combination through ``tedana``, but not the full multi-echo
```
emdupre: Should we link to this section in our docs, on what the optimal combination is ?

docs/usage.rst
```diff
@@ -106,5 +98,59 @@ If this is not considered, resulting intensity gradients (e.g., in the case of s
...
+Why isn't v3.2 of the component selection algorithm supported in ``tedana``?
+````````````````````````````````````````````````````````````````````````````
+There is a lot of solid logic behind the updated version of the TEDICA component
+selection algorithm, first added to the original ME-ICA codebase `here`_ by Dr. Prantik Kundu.
+However, we (the ``tedana`` developers) have encountered certain difficulties
+with this method (e.g., misclassified components) and the method itself has yet
+to be validated in any papers, posters, etc., which is why we have chosen to archive
+the v3.2 code, with the goal of revisiting it when ``tedana`` is more stable.
+
+Anyone interested in using v3.2 may compile and install an earlier release of ``tedana``.
```
emdupre: Should we specify any release numbers, here ?

docs/usage.rst
```diff
@@ -106,5 +98,59 @@ If this is not considered, resulting intensity gradients (e.g., in the case of s
...
+(i.e., there are too few components) or too lenient (there are too many).
+
+In our experience, this may happen when preprocessing has not been applied to
+the data, or when improper steps have been applied to the data (e.g., distortion
+correction, rescaling, nuisance regression).
+If you are confident that your data have been preprocessed correctly prior to
+applying tedana, and you encounter this problem, please submit a question to `NeuroStars`_.
+
+
+I think that some BOLD ICA components have been misclassified as noise.
```
emdupre: I'm not sure that I'm yet ready to support this -- I think once we modularize the selection criteria and create visual reports I'll feel more comfortable doing so, but not as of yet. WDYT ?

docs/outputs.rst
tsalo: Absolutely.

emdupre:
@tsalo I fixed the merge conflict -- could you pull before adding ? Sorry for the hassle !

docs/usage.rst
tsalo: I think it's fair to drop the second paragraph, since we don't have the visual reports ready yet:

> However, if you feel that ``tedana`` is consistently misclassifying certain BOLD-related patterns, please feel free to submit a question to `NeuroStars`_.
We may request access to relevant visual reports, metadata, and even the input data in order to figure out what's happening.
We are always looking for ways to improve ``tedana``'s component selection methods, and having access to data with consistent issues may help us do so.

But I think the first paragraph is fair, since `tedana` includes the ability to manually accept components:

> ``tedana`` allows users to manually specify accepted components when calling the pipeline.
You can use the ``--manacc`` argument to specify the indices of components to accept.

docs/usage.rst
tsalo: Any release from 0.0.4 or earlier should work, right? I'll add that info.

tsalo:
Everything looks good now! CI is taking a bit longer now that we're waiting until the ICA converges.

emdupre:
It's showing that the ipython notebooks and figures were updated as well -- was that from the merge, or were those actual additions ?

emdupre:
It's showing that the ipython notebooks and figures were updated as well -- was that from the merge, or were those actual additions ?

tsalo:
I moved the voxel to a V1 area and the fit quality decreases, which I think is actually really great because it makes the plots a lot more interesting and shows how noisy real data can be.

emdupre:
OK -- are we discussing that in the RTD text ?

tsalo:
Yeah, these changes will only show up in the Processing pipeline details page.

emdupre:
OK, will merge and we can re-visit if necessary :smile: Thanks !!

Issue #162, opened 2018-11-29
Audit workflow arguments


tsalo:
In addition to the arguments we're considering modifying or adding in #161, I think we should audit several of the other arguments in the workflow.

## FastICA parameters
Now that we're using `sklearn` for the ICA, we get a warning in cases of non-convergence instead of an error. Overall, I think this is a good thing, but it also obscures potential problems. The two arguments to `sklearn.decomposition.FastICA` that are relevant are `max_iter` (maximum number of iterations) and `tol` (convergence tolerance). The default for `tol` is 0.0001, while our default is 0.000025, and is set by the `conv` argument. My questions here are, is `conv` something anyone ever plays with (i.e., do we actually need an argument) and is there a reason to use the much stricter default threshold we have over the default from `sklearn`? If not, then I propose that we drop the `conv` argument from `tedana` and let `sklearn` use its default in `tedica`.

On a related note, the default for maximum number of iterations is 200, while `mdp`'s was 5000, IIRC. The problem is that `FastICA` is failing to converge for both of our test datasets (Cornell 3-echo rest and NIH 5-echo task). I think we should probably increase `max_iter` when we call `FastICA` to at least 1000, which will probably increase the amount of time `tedica` takes, but will probably also still be faster than `mdp`.

## Other workflow parameters
I think that `--filecsdata` and `--strict` are currently unused. Unless we plan to re-implement them, I think we can drop them.

Does anyone use `--sourceTEs`, `--denoiseTEs`, or `--stabilize`?
- I don't recall anyone ever talking about `sourceTEs` or `denoiseTEs`, but neither seems very useful, unless I'm misunderstanding them.
- `sourceTEs`: While I maybe understand switching between using `catd` vs `optcom` for fitting models (though I've never tried the former), if we wanted to fit to just some of the echoes, why not _just_ feed those echoes into the workflow?
- `denoiseTEs`: This just determines whether additional files are written out. Unless they're regularly used, I propose that we incorporate this switch into a general `verbose` argument as we've been discussing in #17.
[...]

Issue #161, opened 2018-11-28
Change defaults for global signal regression and minimum image regression


tsalo:
We have discussed changing the defaults for global signal regression and T1c-GSR [here](https://github.com/ME-ICA/tedana/pull/125#issuecomment-434400354), but I think it would be good to address this in its own new issue.

I propose that we do the following:
1. Change default for GSR from running it by default to not running it. We can do this by changing the argument `--no_gscontrol` to `--gscontrol`.
2. Add argument to select a global noise removal method. This is implemented in #120 as `--ws_denoise`, but that PR is not going to be merged any time soon, and I'd like to change these defaults sooner than later. Perhaps `--global_denoise` or `--postproc` would also be good names? Ultimately, we plan to support multiple options for this, including, at minimum, T1c-GSR and GODEC, so it would be better to plan for that now. The default for this argument would be `None`, and to turn T1c-GSR on we could use `t1c`, `mir`, or something else entirely.

Does anyone have any issues with turning off GSR and T1c-GSR (i.e., minimum image regression) by default?

handwerkerd:
No issues with turning this off.

emdupre:
To be clear, I think we are _not_ running gs control by default -- see https://github.com/ME-ICA/tedana/blob/63d643aa79e8b7c4a9865ec57c3263e93bfeb3cd/tedana/workflows/tedana.py#L122

Am I misunderstanding this ?

tsalo:
Perhaps one argument against `--global_denoise` is that we don't want to have a submodule called `global` (for Python reasons) or `denoise` (tedana reasons). An argument against `--postproc` is that that doesn't include the standard GSR that is run on the input data rather than the denoised data, although maybe we don't want to roll GSR in with GODEC and T1c-GSR. Does anyone have any ideas for a better name?

emdupre:
Ok, for some reason I remember us changing the default because we had turned it off, but I'm having trouble tracking down that history right now. I'm very pro removing it.

tsalo:
Closed by #163.

Issue #144, opened 2018-10-23
No convergence after 5000 steps


daraucla:
Hi, 

We are getting an issue with convergence during the call to icanodes.py - see below.  We're running the following in Python 3.6.1:

`
tedana -d BOLD_SmokeCues_MultiTE_10.nii BOLD_SmokeCues_MultiTE_10a.nii BOLD_SmokeCues_MultiTE_10b.nii -e 13.0 35.9 58.0 
`

The job ran for 18 hours with 36 GBs RAM. Each 4D image file contains 288 volumes. 

Any ideas what may be the root of the issue?

> /u/local/apps/python/3.6.1/lib/python3.6/site-packages/h5py/__init__.py:34: FutureWarning: Conversion of the second argument of is
subdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
/u/local/apps/python/3.6.1/lib/python3.6/site-packages/scipy/stats/stats.py:1633: FutureWarning: Using a non-tuple sequence for mu
ltidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an a
rray index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return np.add.reduce(sorted[indexer] * weights, axis=axis) / sumval
/u/home/d/darag/.local/lib/python3.6/site-packages/sklearn/cross_validation.py:41: DeprecationWarning: This module was deprecated 
in version 0.18 in favor of the model_selection module into which all the refactored classes and functions are moved. Also note th
at the interface of the new CV iterators are different from that of this module. This module will be removed in 0.20.
  "This module will be removed in 0.20.", DeprecationWarning)
/u/home/d/darag/.local/lib/python3.6/site-packages/sklearn/ensemble/weight_boosting.py:29: DeprecationWarning: numpy.core.umath_te
sts is an internal NumPy module and should not be imported. It will be removed in a future NumPy release.
  from numpy.core.umath_tests import inner1d
/u/home/d/darag/.local/lib/python3.6/site-packages/sklearn/grid_search.py:42: DeprecationWarning: This module was deprecated in ve
rsion 0.18 in favor of the model_selection module into which all the refactored classes and functions are moved. This module will 
be removed in 0.20.
  DeprecationWarning)
[...]

Issue #149, opened 2018-10-31
Add binder examples


KirstieJane:
In #133 @tsalo added some really lovely jupyter notebooks to explain some aspects of the `tedana`:tada: processing.

It would be really cool if these were accessible through [mybinder](https://mybinder.org).

I don't have time to work on this so any help would be greatly appreciated! ✨ 😻 👾 

I think this requires:

- [x] an update to the [`info.py`](https://github.com/ME-ICA/tedana/blob/master/tedana/info.py) file to include (for example) `matplotlib`,  `seaborn` and `nilearn`.
  * Side note, thank you to Minn in the binder team for helping understand where binder goes looking for its requirements 😸 
  * @emdupre would prefer this to be in the "DOCS REQUIRES" section so these packages aren't added when a regular user installs `tedana`.
  * An alternate idea is to make a `binder/requirements.txt` file that contains everything needed for the notebook. Note that this option won't run `setup.py` so *everything* needs to be added into that file, not just additional packages on top of those defined for `tedana`. 
- [x] moving the jupyter notebooks to somewhere more accessible, eg an `examples` folder
- [ ] uploading some data to OpenNeuro so it's easy to access
- [ ] creating the binder link
- [ ] updating the documentation to make it clear that this information exists!!

souravsingh:
@tsalo @KirstieJane I would like to take a stab at the issue. Where do I start?

emdupre:
Hi @souravsingh ! If you're not yet familiar with Binder, they have some really great resources to get started !  [These tutorials](https://mybinder.readthedocs.io/en/latest/) might be the best first place to look.

As @tsalo said, the notebooks and data are already hosted online, so the issue would be pulling them together in the right environment. Which Binder is perfect for ! Let us know if you have any specific questions about next steps :)

jbteves:
I see that the original PRs were closed for this. @KirstieJane do you mind if I edit your comment so that your task list is at the bottom and we can track progress?

jbteves:
Note: we already have data hosted on OSF for our CI. Relevant links:
- [five-echo task](https://osf.io/e3hsn/download)
- [three-echo rest](https://osf.io/e3hsn/download)

KirstieJane:
Totally fine to edit!! Go for it 🚀🌟

jbteves:
Thanks! Done, now we can track progress directly on the issue title/tracker

jbteves:
[...]

PR #163, opened 2018-11-30
[REF] Remove unused arguments and simplify CLI

tsalo:
Ref. #161, #162.

- Change maximum number of iterations for `tedica` from 200 to 5000, to match how many we ran with `mdp` (see #144).
- Remove `conv` argument from `tedica` and workflow, to use FastICA's default tolerance.
- Remove `cost` argument, to use FastICA's default (`logcosh`).
- Change argument `denoiseTEs` to `verbose`, which will be expanded to toggle output of many intermediate and less-used files, as discussed in #17.
- Change default behavior from performing GSR to _not_ performing GSR.
- Remove arguments `filecsdata` and `strict`.
- Add `gscontrol` argument to toggle minimum image regression (`t1c`) and global signal regression (`gsr`). One or more algorithms can be used, but the default is to use none.


tsalo:
CircleCI is failing because GSR is no longer being performed, which means that the files `T1gs.nii`, `glsig.1D`, `tsoc_nogs.nii`,  and `tsoc_orig.nii` are no longer generated.

tedana/decomposition/eigendecomp.py
```diff
@@ -202,11 +202,9 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
         vTmix = comp_ts.T
         vTmixN = ((vTmix.T - vTmix.T.mean(0)) / vTmix.T.std(0)).T
         LGR.info('Making initial component selection guess from PCA results')
-        _, ct_df, betasv, v_T = model.fitmodels_direct(catd, comp_ts.T, eimum,
-                                                       t2s, t2sG,
-                                                       tes, combmode, ref_img,
-                                                       mmixN=vTmixN,
-                                                       full_sel=False)
+        _, ct_df, betasv, v_T = model.fitmodels_direct(
```
emdupre: I find this convention (where all of the arguments go on a newline) really hard to read -- none of them will fit on the first line, though ?

tedana/workflows/tedana.py
```diff
@@ -109,32 +104,28 @@ def _get_parser():
...
-                        dest='dne',
+    parser.add_argument('--verbose',
+                        dest='verbose',
                         action='store_true',
-                        help='Denoise each TE dataset separately.',
+                        help='Generate intermediate and additional files.',
                         default=False)
-    parser.add_argument('--strict',
-                        dest='strict',
+    parser.add_argument('--gscontrol',
```
emdupre: I'm not sure why we have both  `gscontrol` and `global_denoise` arguments. Could you explain what's the benefit, here ?

tedana/model/fit.py
```diff
@@ -188,6 +199,19 @@ def fitmodels_direct(catd, mmix, mask, t2s, t2s_full, tes, combmode, ref_img,
         kappas[i_comp] = np.average(F_R2, weights=norm_weights)
         rhos[i_comp] = np.average(F_S0, weights=norm_weights)
 
+    if verbose:
```
emdupre: It looks like you're writing out _new_ files with `verbose` enabled ? That is, things that had never been written out before, in any previous version ?

tedana/decomposition/eigendecomp.py
```diff
@@ -202,11 +202,9 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
...
+            catd, comp_ts.T, eimum, t2s, t2sG, tes, combmode,
+            ref_img, mmixN=vTmixN, full_sel=False, label='pca')
```
emdupre: I'm not sure that we want to add an additional label argument here, since we already have a label argument one level up !

tedana/workflows/tedana.py
```diff
@@ -109,32 +104,28 @@ def _get_parser():
                               'logcosh (default), cube, exp'),
```
emdupre: While we're cleaning up the CLI, I'd say we should remove the option to pass different cost functions. This is an example we had originally listed in the Roadmap of something we thought was too complex for the non-power user (since power users can always update the cost function themselves)

tedana/workflows/tedana.py
```diff
@@ -327,30 +320,31 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
         LGR.info('Making second component selection guess from ICA results')
         # Estimate betas and compute selection metrics for mixing matrix
         # generated from dimensionally reduced data using full data (i.e., data
         # with thermal noise)
-        seldict, comptable, betas, mmix = model.fitmodels_direct(catd, mmix_orig,
-                                                                 mask, t2s, t2sG,
-                                                                 tes, combmode,
-                                                                 ref_img,
-                                                                 reindex=True)
+        seldict, comptable, betas, mmix = model.fitmodels_direct(
```
emdupre: Same comment about the indentation style

tedana/workflows/tedana.py
tsalo: I thought the distinction between power and non-power users was the defaults, not the actual arguments. Am I misunderstanding that?

Still, I agree that the cost function is probably not something we need an argument for. I doubt it affects much.

tedana/model/fit.py
tsalo: In order to binderize the walkthrough notebooks (from #149), we need these additional files. I figured that I would roll that into this PR since we're adding `verbose`, but I can push it to another PR.

tedana/workflows/tedana.py
tsalo: GSR (`gscontrol_raw`) and T1c-GSR (`gscontrol_mmix`) are both run by default on the data, rather than treated as competing alternatives to one another. Also, since GSR is run within the pipeline, while T1c-GSR (and someday GODEC, dGSR with rapidtide, etc.) is run after, as a sort of post-processing step.

That said, we can clean this up. Maybe we could allow multiple global denoising steps? E.g.,
`--global_denoise gsr t1c` if you want to run both.

tedana/workflows/tedana.py
emdupre: I think it's both, to an extent. Removing arguments implies a very strong default, whereas having them implies a weaker one. I think this is something we can have a very strong default on -- I've never needed to change the ICA cost function, for example. Not to say you couldn't need to, ever ! But it's reasonable as a strong default.

tedana/model/fit.py
emdupre: Do you mind if we push it off ? I'd like to walk through that whole binderizing process, and I think I'll get confused having it in several different PRs.

tedana/workflows/tedana.py
emdupre: Yes, I think I'd prefer that ! Since they're both aimed at the same outcome, it would make more sense to group them together, in my mind. WDYT ?

tedana/decomposition/eigendecomp.py
tsalo: The two ways we've done it in the past are either:
```
_, ct_df, betasv, v_T = model.fitmodels_direct(
    catd, comp_ts.T, eimum, t2s, t2sG, tes, combmode,
    ref_img, mmixN=vTmixN, full_sel=False, label='pca')
```
or
```
_, ct_df, betasv, v_T = model.fitmodels_direct(catd, comp_ts.T, eimum,
                                               t2s, t2sG, tes,
                                               combmode, ref_img,
                                               mmixN=vTmixN,
                                               full_sel=False,
                                               label='pca')
```

I prefer the former because the latter looks kind of bunched up and generally takes up more lines. I hadn't thought about readability though. Do you prefer the second one?

tedana/decomposition/eigendecomp.py
tsalo: `fitmodels_direct` is a general function that's run for both PCA and ICA, so any files that are generated need to be identified as either PCA- or ICA-related. I'll move it off to another PR though.

emdupre: No no, sorry, I understand the point ! I do think we should move it to another PR for the binder-related reasons, but I just mean the argument itself could be called something other than "label". Otherwise we have two, unrelated "label" arguments floating around for writing out files !

tedana/workflows/tedana.py
tsalo: That's fair. Dropping `cost` now.

tedana/decomposition/eigendecomp.py
emdupre: I do, personally, but I totally see your point ! Maybe there's a compromise. Could we indent the first one a little more, to show that they're arguments ?
```
_, ct_df, betasv, v_T = model.fitmodels_direct(
                catd, comp_ts.T, eimum, t2s, t2sG, tes, combmode,
                ref_img, mmixN=vTmixN, full_sel=False, label='pca')
```

I don't think that will trigger pylint...

tedana/decomposition/eigendecomp.py
tsalo: Just reverted the walkthrough-related changes, so we can revisit the setup for `fitmodels_direct` in that PR.

tedana/model/fit.py
tsalo: Done.

tedana/workflows/tedana.py
tsalo: I can do that, but I'm still a little wary of the argument name. Can you think of anything better that `global_denoise`, `denoise`, or `ws_denoise`?

emdupre: I liked `gs_control` ! Since it's "controlling" global signal, through various means :)

tsalo: That works, I think. The other methods we'd ultimately consider include rPCA, GODEC, dynamic GSR with rapidtide, aCompCor, and tCompCor. The only ones where `gs_control` (maybe `gscontrol`?) would seem kind of weird are aCompCor and tCompCor, but it's not too much of a stretch.

emdupre: Also, if we're integrating into larger pipelines, many of those already do aCompCor and tCompCor !

tedana/decomposition/eigendecomp.py
tsalo: It looks like that's safe. I've changed the fitmodels_direct calls.

tedana/workflows/tedana.py
tsalo: That's true, and aCompCor actually can't be integrated into tedana without information we don't want to require, although I imagine that a number of denoising steps will need to go _inside_ of tedana the same way that GSR and T1c-GSR are done.

emdupre:
It looks like we're still adding new individual echo outputs:
```
> dn_ts_e1.nii
> dn_ts_e2.nii
> dn_ts_e3.nii
> dn_ts_e4.nii
> dn_ts_e5.nii
```
I thought we moved that to the binder-related PR ? Or is this an older CI output ?

tedana/workflows/tedana.py
```diff
@@ -248,6 +228,10 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
     tes = [float(te) for te in tes]
     n_echos = len(tes)
 
+    # Coerce gscontrol to list
+    if not isinstance(gscontrol, list):
```
emdupre: If any values were supplied here it will be a list since you used `nargs='+'`. Is this just to check that it's not `None` ?

tsalo:
That's from moving `--denoiseTEs` into `--verbose`. It's unrelated to the binder stuff.

emdupre:
Ah I see -- yeah I guess by using `--verbose` we are losing some granularity ! This is ok with me, and I suppose we can always revisit later :)

tedana/workflows/tedana.py
tsalo: That's right, since `None` should be the default. Plus I'm using `if x in gscontrol` which would break if `gscontrol` was left as `None`.

emdupre: Could we make this explicit by having None return an empty list, rather than a list with a value of `None` ? Like:
```
if gscontrol is None:
    gscontrol = ['']
```

tsalo: That will make things easier for the CLI, but not for the Python function, since users could still input strings like `tedana_workflow(..., gscontrol='gsr')`.

emdupre:
OK, I just want to review the generated artifacts and then I think this would be good to merge ! Do you mind if I squash and merge ?

tsalo:
There are a lot of mini-commits, so that's fine with me. To be honest, since we're working to make PRs contained and sort of single-effect changes, I'd love to switch to squash and merge generally. I always make so many commits to check the CI that I feel bad about merging normally.

emdupre:
OK, I checked the files and I think this looks good !

Only remaining issue: we need to update the expected outputs [here](https://github.com/ME-ICA/tedana/blob/master/.circleci/tedana_outputs.txt). Since these will now differ by dataset, we should probably create a separate file for each dataset with the correct filenames -- this will actually be better in the long-run, since it means we'll test more options !

tsalo:
Okay, it should be good now.

PR #164, opened 2018-11-30
[ENH] Split automatic dimensionality detection from decision tree in TEDPCA

tsalo:
Ref. #101.

We can also revisit `kdaw` and `rdaw` if the decision tree is no longer going to be the default, since that's the only case in which they're used.

- Add argument (`tedpca`) to CLI that specifies method for PCA component selection. Current options are `'mle'` (uses MLE to automatically detect dimensionality of data), `'kundu'` (decomposes data into standard number of components, then uses decision tree to select components), and `'kundu-stabilize'` (same as `'kundu'`, but uses the additional steps from the `stabilize` argument). The default is `'mle'`.
- Remove `stabilize` argument.


tedana/decomposition/eigendecomp.py
```diff
@@ -32,7 +32,7 @@
     """),
     description='Introduces method for choosing PCA dimensionality '
                 'automatically')
-def run_svd(data):
+def run_svd(data, mle=True):
```
tsalo: We'll need to split this somehow so that the MLE paper is only cited when MLE is used, but I'm not sure what the best way to do that is.

tedana/decomposition/eigendecomp.py
```diff
@@ -175,21 +281,30 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
...
+            LGR.warning('Method from PCA state file ({0}) does not match '
+                        'requested method ({1}).'.format(pcastate['method'],
+                                                         method))
+            state_found = False
+        else:
+            state_found = True
+    else:
+        state_found = False
+
+    if not state_found:
```
emdupre: You could clean this up as:

```
perform_mle = (method == 'mle')
voxel_comp_weights, varex, comp_ts = run_svd(dz, mle=perform_mle)
```

tsalo: That's a good catch! We need to split run_svd into two functions though, in order to cite the MLE paper iff MLE is used.

tedana/decomposition/eigendecomp.py
```diff
@@ -51,16 +51,123 @@ def run_svd(data):
...
     ppca.fit(data)
     v = ppca.components_
     s = ppca.explained_variance_
     u = np.dot(np.dot(data, v.T), np.diag(1. / s))
     return u, s, v
 
 
-def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG, stabilize,
-           ref_img, tes, kdaw, rdaw, ste=0, wvpca=False):
+def kundu_tedpca(ct_df, n_echos, kdaw, rdaw, stabilize=False):
```
tsalo: I think we should move this into `selection`, given that it doesn't actually run PCA and is just used for component selection. Thoughts?

tedana/decomposition/eigendecomp.py
emdupre: Opened an issue on duecredit ! https://github.com/duecredit/duecredit/issues/141

tedana/decomposition/eigendecomp.py
emdupre: Strongly agree ! :+1:

tsalo: I spent some time figuring out how best to do this, and I think I'd like to push that to a separate PR actually. There are some elements of tedpca that I think we should discuss in more detail before we can reorganize it to better mirror tedica. Just as an example, I'm not sure why we use an `eimask`-generated mask when computing Kappa and Rho values for tedpca, but not for tedica.

tedana/workflows/tedana.py
```diff
@@ -162,13 +149,30 @@ def _get_parser():
                         help=argparse.SUPPRESS,
                         action='store_true',
                         default=False)
+    tedpca_args = parser.add_argument_group('Options for controlling TEDPCA')
+    tedpca_args.add_argument('--kdaw',
```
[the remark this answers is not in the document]
tsalo: Okay, I'll drop these arguments, but will use retain them as arguments to `tedpca` (with the default values) rather than hardcoding them into `kundu_tedpca`. That way, power users can play with them pretty easily when they want to.

tsalo:
The five echo dataset is timing out because ICA takes >40 minutes. Perhaps we should lower the maximum number of iterations back down to something like 1000? The ICA will definitely fail to converge, but at least tests will pass.

tsalo:
So I switched to using the decision tree for TEDPCA here to get around the time limit. Ironically ICA still failed to converge, but at least it finished in time.

docs/outputs.rst
```diff
@@ -94,6 +94,68 @@ betas_hik_OC_T1c.nii      T1-GS corrected high-kappa components
...
+I002   rejected         Rho greater than Kappa or        Kundu v2.5, Kundu v3.2
+                        more significant voxels
+                        in S0 model than R2 model
+I003   rejected         S0 Dice is higher than R2 Dice   Kundu v2.5, Kundu v3.2
+                        and high variance explained
+I004   rejected         Noise F-value is higher than     Kundu v2.5, Kundu v3.2
+                        signal F-value and
+                        high variance explained
+I005   retained         No good components found         Kundu v2.5
+I006   rejected         Mid-Kappa component              Kundu v2.5, Kundu v3.2
```
emdupre: I don't think we should be putting in the ICA selection in the docs, since we're not supporting v3.2 right now. Should we put these in a gist for now, so we can come back to them ?

.circleci/config.yml
```diff
@@ -196,7 +196,7 @@ jobs:
           command: |
             source activate venv
             cd /tmp/data/five-echo/
-            tedana -d p06.SBJ01_S09_Task11_e[1,2,3,4,5].sm.nii.gz -e 15.4 29.7 44.0 58.3 72.6 --verbose
+            tedana -d p06.SBJ01_S09_Task11_e[1,2,3,4,5].sm.nii.gz -e 15.4 29.7 44.0 58.3 72.6 --verbose --sourceTEs 0
```
emdupre: The optimal combination is probably what we should be using, here !

tedana/decomposition/eigendecomp.py
tsalo: That would be great, thanks!

docs/outputs.rst
```diff
@@ -94,6 +94,68 @@ betas_hik_OC_T1c.nii      T1-GS corrected high-kappa components
...
+
+=====  ===============  ===============================  =======================
+Code   Classification   Description                      Algorithm(s)
+=====  ===============  ===============================  =======================
+P001   rejected         Low Rho, Kappa, and variance     Kundu decision tree
+                        explained
+P002   rejected         Low variance explained           Kundu decision tree
+P003   rejected         Kappa equals fmax                Kundu decision tree
+P004   rejected         Rho equals fmax                  Kundu decision tree
+P101   rejected         Cumulative variance explained    Kundu decision tree
```
emdupre: If we're getting rid of the algorithm column, maybe we could just note in the description something like `(only in stabilized PCA decision tree)`

docs/outputs.rst
tsalo: That makes sense. You made the v3.2 gist, right? Would you mind making this one too, so that they're easier to find together?

.circleci/config.yml
tsalo: I'll revert it now.

docs/outputs.rst
tsalo: Changing now.

tedana/decomposition/eigendecomp.py
tsalo: In the interest of getting this PR ready to merge so we can work on validation analyses, I was thinking of just running the regular PCA in `tedpca` and calling the MLE-specific function only if that's requested. We could circle back to the organization later. WDYT?

docs/outputs.rst
emdupre: Created ! https://gist.github.com/emdupre/f65999c4952f1e0b499eee9e0e61ef0e

tedana/decomposition/eigendecomp.py
emdupre: Created: https://github.com/ME-ICA/tedana/issues/166

emdupre:
I think this is ready !

PR #167, opened 2018-12-06
[DOC] Improve logging

tsalo:
Does not close any issues, but makes some improvements to logs that will be necessary for validation analyses we're planning in [tedana-comparison](https://github.com/ME-ICA/tedana-comparison).

- Replace `label` argument with `out_dir`. This lets us output the log file to the output directory without having to perform operations we want to log (e.g., loading data) beforehand.
- Ignore the unnecessary LAPACK driver warning raised by `sklearn`.
- Catch and log `sklearn` warning when FastICA fails to converge.
- Move logging configuration from `tedana.workflows.tedana._main` to `tedana.workflows.tedana.tedana_workflow`.


tsalo:
I am having trouble figuring out how to actually close the file handler, so that when we call `tedana_workflow` repeatedly (e.g., in a for loop) it will write out to separate files. Currently, it continues to write out to the same log file, although it creates empty files for the subsequent log files.

I've tried adding the following at the end of `tedana_workflow`, but it doesn't seem to be doing anything:

```
fh.close()
del fh
logging.shutdown()
```

tedana/workflows/tedana.py
```diff
@@ -125,11 +125,11 @@ def _get_parser():
                         help='Perform PCA on wavelet-transformed data',
                         action='store_true',
                         default=False)
-    parser.add_argument('--label',
-                        dest='label',
+    parser.add_argument('--out',
```
emdupre: Totally subjective, but could we call it `--out-dir` to match with the string value ?

tedana/workflows/tedana.py
```diff
@@ -223,6 +223,24 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
     """
+    out_dir = op.abspath(out_dir)
+    if not op.isdir(out_dir):
+        os.mkdir(out_dir)
+
+    if debug and not quiet:
+        formatter = logging.Formatter(
+                    '%(asctime)s\t%(name)-12s\t%(levelname)-8s\t%(message)s',
+                    datefmt='%Y-%m-%dT%H:%M:%S')
+        fh = logging.FileHandler(op.join(out_dir, 'runlog.tsv'))
```
emdupre: Looking at the artifacts on circle it looks like the logs are all correctly written to screen, but I don't see any saved out to a tsv ?

tedana/workflows/tedana.py
tsalo: Absolutely.

tedana/workflows/tedana.py
tsalo: The tsv file is currently only written out if `debug` is used, and I didn't change either of the tests to run on debug. Would you like me to change the behavior (e.g., generate it always) or the tests (i.e., add `--debug` to one of them)?

emdupre: Let's change one of the tests, if that's ok ! Sorry, I totally misread and thought we had that there already. Thanks :)

emdupre: This looks great ! Final question -- right now we just save it as `runlog.tsv` -- should we put the date and time in the filename, in case the same data is run multiple times ?

tsalo: I really don't know what's best practice for logging. Adding in the date and time makes it slightly harder to grab the relevant file. Although with glob or regular expressions it's not that much harder...

tsalo: How do you feel about the ISO format in the filename?
E.g., `runlog_2018-12-11T11:15:26.228355.tsv`.

emdupre: My only concern with having colons in the filename is that I ran into an issue with some other filenames with colons in them just recently -- tar thinks they're living on another machine and freaks out. Not necessarily a problem, here, but I might prefer converting to dots to be safe... WDYT ?

tsalo: That's fine by me. I'll push that now.

tsalo: Wait a second.... what about the CI? The date time for the expected outputs list won't match the  actual outputs.

tsalo: Separate logs files are a good idea, I think. I'll just revert the test.

emdupre:
This looks great ! Merging now :tada: :sparkles:

PR #172, opened 2018-12-14
[FIX] Reduce user-defined mask when there is no good signal

### LINKED ISSUE #171: Error when voxels in explicit mask do not have "good" signal in any echoes
Originally discovered in a [Neurostars post](https://neurostars.org/t/fmriprep-to-tedana/2685/14?u=tsalo).

When using an explicit mask, if there are voxels in the mask that do not have "good" signal in any of the echoes (as estimated by `utils.make_adaptive_mask`) the adaptive mask will have a value of 0 in those voxels. Then, T2* is estimated using zero echoes, resulting in values of 0 at those voxels in the T2* limited and full maps. When the weights for the optimal combination are determined using the formula `alpha = tes * np.exp(-tes / ft2s)`, there is a divide-by-zero warning, and the weights are a set of zeros. Those zeros raise an error in `numpy.average`, because they sum to 0.

When this happens, I think that we should take one of two approaches.
1. Set those bad-signal voxels to 1 in the adaptive mask (i.e., pretend that the first echo at least has good signal).
2. Remove these bad-signal voxels from the regular mask in addition to the adaptive mask. This may be slightly more difficult, as the resulting mask will be a mask of a mask, but I'm sure it's doable.
    - Update: Not difficult. Super easy to do.

tsalo:
Closes #171.

- When there are voxels in the user-defined mask with no good signal (as determined by `make_adaptive_mask`), they will now be removed from the mask used in `tedana`.


tsalo:
We're going to need to update `test_make_adaptive_mask`. Does anyone have any ideas for a good test?

tedana/utils.py
```diff
@@ -150,6 +153,12 @@ def make_adaptive_mask(data, mask=None, minimum=True, getsum=False):
         # if the user has supplied a binary mask
         mask = load_image(mask).astype(bool)
         masksum = masksum * mask
+        # reduce mask based on masksum
+        if np.any(masksum[mask] == 0):
+            n_bad_voxels = np.sum(masksum[mask] == 0)
+            LGR.warning('{0} voxels in user-defined mask do not have good '
+                        'signal. Removing voxels from mask.'.format(n_bad_voxels))
```
emdupre: Could we add a "#TODO" about asking them to visually inspect the mask ? This is something we should include in the visual reports, eventually.

tedana/tests/test_utils.py
```diff
@@ -133,8 +133,7 @@ def test_make_adaptive_mask():
     mask, masksum = utils.make_adaptive_mask(data, mask=pjoin(datadir,
                                                               'mask.nii.gz'),
                                              minimum=False, getsum=True)
-    assert np.allclose(mask, nib.load(pjoin(datadir,
-                                            'mask.nii.gz')).get_data().flatten())
+    assert np.allclose(mask, masksum.astype(bool))
```
emdupre: Our test is failing, here, since we're now hitting this "bad voxel" condition, I suppose ? This is probably actually fine for now since it covers more lines of code, but we could add a "#TODO" here about testing with a mask that does not include bad voxels.

tedana/utils.py
tsalo: I can do that, although we aren't actually writing out the masks at the moment. I was planning to incorporate writing out the masks and other intermediate outputs into `verbose` in a future PR, but can at least write out the masks with `verbose` in this one.

tedana/tests/test_utils.py
tsalo: Yeah, a lot of voxels in the mask had bad data, so the mask file didn't match up with the adaptive mask. I'll add a TODO.

tedana/utils.py
tsalo: Oh sorry I misinterpreted your comment. I'll add a comment to the code.

tedana/workflows/tedana.py
```diff
@@ -285,6 +285,9 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
     mask, masksum = utils.make_adaptive_mask(catd, mask=mask,
                                              minimum=False, getsum=True)
     LGR.debug('Retaining {}/{} samples'.format(mask.sum(), n_samp))
+    if verbose:
+        io.filewrite(mask.astype(int), op.join(out_dir, 'mask.nii'), ref_img)
```
emdupre: Is there a benefit in writing out the mask ? Either it's user-provided or effectively just the adaptive mask, no ?

tsalo: Just the regular mask? There's probably no reason to write it out. I'll drop that one.

emdupre:
This is great ! Thanks @tsalo :sparkles: Merging !

PR #174, opened 2019-01-05
[ENH] Add verbose outputs for pipeline walkthrough

tsalo:
Per discussion in #163 (see this [comment](https://github.com/ME-ICA/tedana/pull/163#discussion_r238318982)), this PR adds in outputs necessary for #149.

I will need to update this PR after #152 is dealt with, but am keeping it separate to keep both PRs focused.

- Add `label`, `out_dir`, and `verbose` arguments to `fitmodels_direct`. Fitted models for TEDPCA and TEDICA can thus write out different files.
- Generate several files with `fitmodels_direct`, including *_R2_pred.nii, *_S0_pred.nii, *_betas_catd.nii, and *_metric_weights.nii.
    - `R2_pred` is a 5D array (X x Y x Z x E x C) with the predicted value at each echo and voxel, for each component, for the R2 model.
    - `S0_pred` is the same, but for the S0 model.
    - `metric_weights` is a 4D array (X x Y x Z x C) with the voxel- and component-wise weights for averaging the R2 and S0 F values to get component-specific Kappa and Rho values.
    - `betas_catd` is a 5D (X x Y x Z x E x C) with the beta for the raw data at each echo for each component. These values are compared to the predicted values for the R2 and S0 models in evaluating model fit.
- Add `verbose` argument to `tedpca`.
- Write out whitened (post-TEDPCA) optimally combined time series.


docs/_static/plot_optcom.ipynb
```diff
@@ -0,0 +1,488 @@
...
+    "```\n",
+    "tedana -d /Users/tsalo/Documents/tsalo/tedana_comparison/e5_data/p06.SBJ01_S09_Task11_e1.sm.nii.gz /Users/tsalo/Documents/tsalo/tedana_comparison/e5_data/p06.SBJ01_S09_Task11_e2.sm.nii.gz /Users/tsalo/Documents/tsalo/tedana_comparison/e5_data/p06.SBJ01_S09_Task11_e3.sm.nii.gz /Users/tsalo/Documents/tsalo/tedana_comparison/e5_data/p06.SBJ01_S09_Task11_e4.sm.nii.gz /Users/tsalo/Documents/tsalo/tedana_comparison/e5_data/p06.SBJ01_S09_Task11_e5.sm.nii.gz -e 15.4 29.7 44.0 58.3 72.6 --label mlepca --no_gscontrol\n",
+    "```\n",
+    "\n",
+    "MLEPCA was *not* used for PCA dimensionality estimation (# components retained with MLEPCA = 159, the max number, so no whitening occurs). Instead, SVD + the decision tree was used. Scikit-learn's FastICA was used for ICA, and the Kundu v3.2 algorithm was used to classify ICA components."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
```
emdupre: Are we showing these notebooks in the docs ? If not, can we upload unexecuted notebooks, for easier diff-ing ?

tedana/model/fit.py
```diff
@@ -208,6 +217,19 @@ def fitmodels_direct(catd, mmix, mask, t2s, t2s_full, tes, combmode, ref_img,
     else:
         mmix_new = mmix
 
+    if verbose:
```
emdupre: It looks like some of these `filewrite` calls use `mask`, while others use `t2s != 0`. Is there a reason to use both, here ?

tedana/workflows/tedana.py
```diff
@@ -302,20 +302,25 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
+                                                kdaw=10., rdaw=1., wvpca=wvpca,
+                                                verbose=verbose)
         mmix_orig, fixed_seed = decomposition.tedica(n_components, dd,
                                                      fixed_seed)
 
         if verbose:
             np.savetxt(op.join(out_dir, '__meica_mix.1D'), mmix_orig)
+            if ste == -1:
+                io.filewrite(utils.unmask(dd, mask),
+                             op.join(out_dir, 'tsoc_whitened.nii'), ref_img)
```
emdupre: Can we make this consistent with the `tsoc` filename (i.e., either `ts_OC_whitened.nii` or `tsoc.nii`) ?

tedana/model/fit.py
tsalo: Mask is overwritten with `t2s != 0` when we moved from selcomps v3 to v2, but I have raised an issue about it in #175. I can do them all here with `mask`, though, for consistency.

docs/_static/plot_optcom.ipynb
tsalo: We're not, but the figures generated are used. I'll empty the outputs.

tedana/workflows/tedana.py
tsalo: Can do.

Issue #177, opened 2019-01-06
Logging for eimask is somewhat incorrect


tsalo:
The [logging message](https://github.com/ME-ICA/tedana/blob/1bc32e46445f1d8c5b69ba2000280e2f8f591321/tedana/decomposition/_utils.py#L34) for tedana.decomposition._utils.eimask only works when the mask is generated from the raw multi-echo data. When TEDPCA is run on the optimally combined data (the most common approach), the logger will give "Creating eimask for echo 0", which is incorrect. We should make sure that the `eimask` logging messages reflect what data are used.

tsalo:
For some reason merging #184 didn't close this, so I'm closing it manually now.

PR #184, opened 2019-01-17
[DOC] Fix eimask logging, ste definitions in eigendecomp

### LINKED ISSUE #183: ste parameter typo in eigendecomp?
I was looking at Issue #177 to try and get more involved in the process, but noticed something. 

In eigendecomp.py, we see the following
https://github.com/ME-ICA/tedana/blob/a961cfcbff3f75155470edf37212fcbf3e61d0a6/tedana/decomposition/eigendecomp.py#L266-L271

This matches where the parameters are selected by the user.  
https://github.com/ME-ICA/tedana/blob/a961cfcbff3f75155470edf37212fcbf3e61d0a6/tedana/workflows/tedana.py#L181-L183

However, the intro docs for eigendecomp.py don't match:
https://github.com/ME-ICA/tedana/blob/a961cfcbff3f75155470edf37212fcbf3e61d0a6/tedana/decomposition/eigendecomp.py#L196-L200

Here 0 is set to Optimal Combine, rather than -1. I assume it is just a typo in eigendecomp - but wanted to raise this issue in case it had propagated through the code.

dowdlelt:
Closes #183 and #177  .

-change the tedpca ste default to -1, for optCom, matching tedana help
-update the comments at beginning, also to match
-update eimask to check len(ste), using info for correct logging


tsalo:
The changes look good to me. You just need to fix the trailing whitespace issue for Travis to pass.

tedana/decomposition/_utils.py
```diff
@@ -31,7 +37,10 @@ def eimask(dd, ees=None):
         ees = range(dd.shape[1])
     imask = np.zeros((dd.shape[0], len(ees)), dtype=bool)
     for ee in ees:
```
tsalo: What do you think about checking the size of `ees`. If `len(ees) < dd.shape[1]`, then it's either optcom or a subset of echoes. If `len(ees) == 1`, it's optcom, and if it's greater than 1 but less than the shape of the data, then it's a subset.

dowdlelt: Should we differentiate between three categories in logging, that is: Optcom, all echoes, or subset? I can see the logging message getting a little confusing - in that echo number could refer to the actual eche number or just indicate how many echoes it performed the function on. 

Maybe that is accounted for. This is my first real pull request so I'm still sorting things out.

tsalo: It is not accounted for yet. Perhaps a good simplified solution would be:

```python
if dd.shape[1] == 1:
    LGR.debug('Creating eimask for optimal combination')
else:
    LGR.debug('Creating eimask for echo {}'.format(ee))
```

This way we can keep the inputs to this function simple (i.e., not include `ste`).

tsalo:
There are still some whitespace issues, but other than that this looks good to me. Regarding those whitespace issues, they would probably be removed automatically by a good text editor. What are you using?

Can you update the list of proposed changes and the pull requests closed (this should close #177 as well)?

dowdlelt:
Sadly, at the moment I'm using github online - but normally I'd use atom. Was just easier for me to understand what was going on in the process by doing it online. Open to other suggestions.

tsalo:
No that's fine. You just have a couple more to fix so there's no reason to change your editor at this point.

dowdlelt:
whitespace was hiding in plain sight, think that is resolved now. thanks for the review.

tsalo:
No problem. Thank _you_ for catching the problem in the documentation and defaults.

I'm just going to add the DOC tag to the PR name and then I'll merge.

emdupre:
Congrats on the PR, @dowdlelt ! Go team :tada:

PR #189, opened 2019-01-21
[ENH] Adds issue templates for bugs and discussions

jbteves:
Partially closes #186

Changes proposed in this pull request:

- Adds bug template, discussion template

PR #191, opened 2019-01-21
[ENH] Normalize all the line endings

### LINKED ISSUE #190: Cross-Platform Line Endings
### Summary
In the current repository, a windows user will experience git diff issues when trying to modify the repository locally. 

### Suspected Cause
autoclrf was not set to true in the tedana repository

### Warning/Error Message
No warnings/errors, but git will not properly update or discard changes

### Additional
See here: https://help.github.com/articles/dealing-with-line-endings/

jbteves:
Closes #190 .

- normalizes line endings for Windows compatibility

Note: tested on MacOSX and Windows for cross-compatibility.


emdupre:
Thanks, @jbteves ! It looks like this is modifying the binary nifti files, though ? Was that intentional ?

jbteves:
No, it's just stating that any .nii files should be handled like binary files-- regardless of platform, the contents should be precisely the same.

jbteves:
I see now-- I'm not sure what that mode change means, but according to git the actual contents of the files shouldn't be different. Maybe we can check by loading up the contents on two systems and subtracting the images?

jbteves:
Okay, I researched the problem a bit more. Apparently the directory structure retains file permissions data for each file. So the files themselves are not modified, but their permissions are. We can either revert that part of the change or we can add 
```
core.filemode false
```
to the git config. More info [here](https://stackoverflow.com/questions/1257592/how-do-i-remove-files-saying-old-mode-100755-new-mode-100644-from-unstaged-cha). I'm not sure why but the issue appears to only be introduced if you clone into Windows.

emdupre:
OK, now that #189 is merged I'd like to tackle this one !

So it sounds like the only remaining question is whether we should update the permissions on the nifti files. Could you expand on the benefit to having those permissions updated ?

jbteves:
Hm, so I've simmered on this a bit. I'm not sure if `tedana` formally offers Windows support-- if not then there's no reason to introduce these changes. As far as permissions go, all it does is make the files non-executable. I assume that this is okay because there's no reason to execute a .nii file, and view that as an improvement. For some reason on my Windows machine this permission is coerced-- it may be because I don't have a default .nii reader available and the OS is protecting me.

emdupre:
We only require a python environment, at this point, so we should be cross-platform. I'm not sure why Windows would require that permissions update, but I just wanted to be sure it was necessary for our n=1 Windows sample :smile: 

I'm ok to merge this, then. @tsalo @rmarkello, please voice if you have any objection !

jbteves:
Hm, let me amend the commit message to include the file mode changes. Otherwise on review that might be confusing.

jbteves:
Scratch that, I forgot that amending commits that are already referenced remotely is typically bad practice-- we can just include this and read the context discussion out of the PR. Should be easy enough since it's just the one.

emdupre:
I'm going to go ahead and merge this so we can cut our next release. If it causes an issue in the wild, we'll issue a hotfix (but I think it should be ok !). Thanks, @jbteves :sparkles:

PR #204, opened 2019-01-27
[FIX] Do not use minimum mask for OC data in tedpca

### LINKED ISSUE #168: Error in tedpca with minimally preprocessed data from ds001491
I'm trying to run tedana on some minimally preprocessed (slice timing and motion correction only) data from [ds001491](https://openneuro.org/datasets/ds001491) (sub-01, images task). The preprocessing was somewhat hastily cobbled together, but I don't see any obvious problems in the [pipeline](https://github.com/tsalo/tedana-comparison/blob/aba5885df8208b689433ef398058db30ff88ccce/code/afni_preproc_step1.py) that could cause the following error.

Has anyone else come across this before?

```
tedana -d func/sub-01_task-images_echo-1_desc-realign_bold.nii.gz \
    func/sub-01_task-images_echo-2_desc-realign_bold.nii.gz \
    func/sub-01_task-images_echo-3_desc-realign_bold.nii.gz \
    -e 8.02 22.03 36.03 --out ds001491 --debug

INFO:tedana.workflows.tedana:Using output directory: /Users/tsalo/Documents/tsalo/tedana-comparison/sandbox/ds001491
INFO:tedana.workflows.tedana:Loading input data: ['func/sub-01_task-images_echo-1_desc-realign_bold.nii.gz', 'func/sub-01_task-images_echo-2_desc-realign_bold.nii.gz', 'func/sub-01_task-images_echo-3_desc-realign_bold.nii.gz']
DEBUG:tedana.workflows.tedana:Resulting data shape: (184320, 3, 493)
INFO:tedana.workflows.tedana:Computing adaptive mask
DEBUG:tedana.workflows.tedana:Retaining 134925/184320 samples
INFO:tedana.workflows.tedana:Computing T2* map
/Users/tsalo/anaconda/envs/python3/lib/python3.6/site-packages/scipy/stats/stats.py:1706: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return np.add.reduce(sorted[indexer] * weights, axis=axis) / sumval
DEBUG:tedana.workflows.tedana:Setting cap on T2* map at 3406.81902
INFO:tedana.combine:Optimally combining data with voxel-wise T2 estimates
INFO:tedana.decomposition.eigendecomp:Computing PCA of optimally combined multi-echo data
DEBUG:tedana.decomposition._utils:Creating eimask for echo 0
DEBUG:tedana.decomposition._utils:Eimask threshold boundaries: 1023.951 5119757.142

  File "/Users/tsalo/anaconda/envs/python3/bin/tedana", line 11, in <module>
  File "/Users/tsalo/Documents/tsalo/tedana/tedana/workflows/tedana.py", line 379, in _main
    tedana_workflow(**vars(options))
  File "/Users/tsalo/Documents/tsalo/tedana/tedana/workflows/tedana.py", line 318, in tedana_workflow
    ste=ste, wvpca=wvpca)
  File "/Users/tsalo/Documents/tsalo/tedana/tedana/decomposition/eigendecomp.py", line 200, in tedpca
    o[mask] = eimum
ValueError: NumPy boolean array indexing assignment cannot assign 111986 input values to the 134925 output values where the mask is true
```

tsalo:
Closes #168.

- Use inputted mask for masking optimally combined data in TEDPCA (as is done for concatenated data [--sourceTEs = 0] or indexed echoes [--sourceTEs = list]) instead of minimum mask.


tedana/decomposition/eigendecomp.py
```diff
@@ -266,16 +266,17 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG,
...
-        d = catd[mask].astype('float64')
+        d = catd[mask, ...].astype('float64')
     else:
         LGR.info('Computing PCA of echo #%s' % ','.join([str(ee) for ee in ste]))
-        d = np.stack([catd[mask, ee] for ee in ste - 1], axis=1).astype('float64')
+        d = np.stack([catd[mask, ee, :] for ee in ste - 1], axis=1).astype('float64')
 
     eim = np.squeeze(eimask(d))
     d = np.squeeze(d[eim])
+    d = np.squeeze(d)
```
emdupre: Sorry, why do we need this second `np.squeeze` now ?

tsalo: Whoops. That's a holdover from when I was looking at the eimask as the potential problem. Removed now.

tedana/decomposition/eigendecomp.py
```diff
@@ -266,13 +266,13 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG,
 
     if len(ste) == 1 and ste[0] == -1:
         LGR.info('Computing PCA of optimally combined multi-echo data')
-        d = OCcatd[utils.make_min_mask(OCcatd[:, np.newaxis, :])][:, np.newaxis, :]
+        d = OCcatd[mask, :][:, np.newaxis, :]
```
emdupre: One more question: for the two other options, we explicitly cast to float64. Are we sure `OCcatd` is float64 already ?

tsalo: I don't see anywhere where the type for the optimally combined data is explicitly set. Honestly, I think casting to float64 is unnecessary for any of the data, but we can do it to the optimally combined data too just to be consistent.

emdupre: I'm fine with either path forward, I'm just pro-consistency unless otherwise specified :sweat_smile:

emdupre: Maybe we can open this as another issue, because I'm not sure how much it overlaps here, but I'm actually pretty sure your initial take (that casting to float64 is unnecessary) is correct. My reason for bringing it up now is because it's an easy chance to change it and I do think we're increasing memory usage with the re-casting....

emdupre:
@jbteves or @dowdlelt: if you run this branch on the data you mentioned in in https://github.com/ME-ICA/tedana/issues/168#issuecomment-457847478 with `--sourceTEs=0` do you get equivalent results ? Trying to confirm that dropping float64 casting won't introduce other concerns !

dowdlelt:
@emdupre kicking off a test run now. Hope to have an answer soon.

jbteves:
@dowdlelt are you running this on our shared server? If so I don't see a point in replicating efforts, if not I'll run a test on it so we have at least two OS's.

dowdlelt:
negative @jbteves and I managed to lock my local machine because I got greedy with resources. Spin it up on the shared server, and I'll try another run tomorrow am.

jbteves:
Okay, @emdupre @tsalo here is how I tested:
1. Existing directory contained results with the code and --sourceTEs 0, supplied "user"-generated mask (from FSL BET) pre-fix
1. Cloned @tsalo's branch into my local repo, which the environment was pointed at. 
1. Re-ran on the same NifTI files (including original mask)
1. Reshaped the matrices to single vectors and subtracted
From this I got all 0's in the resulting subtraction, which indicates the results should be the same, for the two runs that have already finished (still waiting on several). Let me know if you want additional tests done-- I can try to knock them out. If you catch me in the next half hour I'll run them overnight.

emdupre:
That's reassuring ! Just to make sure I understand: the matrix you re-shaped was which output file ? 

Thanks so much for doing that !! :sparkles:

jbteves:
`dn_ts_OC.nii`

emdupre:
Great ! The only other way I can think to test it would be to directly compare the files, with something like

```python
import numpy as np
import nibabel as nb

file1 = nb.load('filename1.nii').get_data()
file2 = nb.load('filename2.nii').get_data()

np.allclose(file1, file2)
```

But that should give you just a `True` instead of a 0.

Unless @tsalo has any other concerns I'll go ahead and get this merged tomorrow !

tsalo:
Sounds good to me. I also think that, since this fix has the potential to impact a fair amount of data, we should probably draft a minor release after it's merged. Well this PR and the other things we've managed to do since the last release (e.g., tedort, the tedpca split, changes to the CLI, and vast improvements to the documentation).

emdupre:
OK, I'll go ahead and get this merged ! Thanks to @tsalo for the fix, and to @dowdlelt and @jbteves for the test on real data !

PR #207, opened 2019-01-30
[FIX] Propagate TR to ref_image header

### LINKED ISSUE #206: TR isn't set correctly in output
When running the current version of tedana, it appears that the TR of the output data is set to 1.00. 

It looks like this occurs around line 638 in io.py, where the ref image is created. Specifically the variable img had a complete header, with the appropriate TR available via the get_zooms function (http://nipy.org/nibabel/nibabel_images.html):
```
good_header = img.header
good_header.get_zooms()[-1]
```
However, the TR information has been removed/replaced(?) with the ref image, such that get_zooms only has the voxel dimensions.

dowdlelt:
Closes #206  .

- in load_data, the ref image now has a 4th dimension, TR in output is now correct
-

### DIFF
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -544,7 +544,7 @@ def load_data(data, n_echos=None):
     fdata = utils.load_image(img.get_data().reshape(nx, ny, nz, n_echos, -1, order='F'))
 
     # create reference image
-    ref_img = img.__class__(np.zeros((nx, ny, nz)), affine=img.affine,
+    ref_img = img.__class__(np.zeros((nx, ny, nz, 1)), affine=img.affine,
                             header=img.header, extra=img.extra)
     ref_img.header.extensions = []
     ref_img.header.set_sform(ref_img.header.get_sform(), code=1)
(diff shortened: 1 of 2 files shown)


dowdlelt:
Travis is unhappy with that extra '1', unsurprisingly. Though I will admit this is the first time I have ever encountered continuous integration. 

```
 assert ref.shape == (39, 50, 11)
E       assert (39, 50, 11, 1) == (39, 50, 11)
E         Left contains more items, first extra item: 1
E         Use -v to get the full diff
```

tsalo:
This looks good. Thank you!

It looks like you'll just need to update the test. You just need to change `assert ref.shape == (39, 50, 11)` to `assert ref.shape == (39, 50, 11, 1)` in the test function.

Could you also check to see if the 3D images being written out are still 3D? I just want to make sure that they aren't written out with a singleton fourth dimension.

Lastly, could you add the [FIX] prefix to this PR title and remove the [bug] one from the corresponding issue?

dowdlelt:
3dinfo reports that dn_ts_OC.nii has a time dimension, with the correct tr, whereas s0v.nii has no time dimension. Opens fine in AFNI as well. Let me know if you want another check of that.

tsalo:
If it has no time dimension then I think that should be fine. Thank you for checking!

emdupre:
I'll go ahead and merge, then ! Thanks @dowdlelt :sparkles:

Issue #188, opened 2019-01-18
Programmatically Control Thread Count


jbteves:
### Summary
Currently tedana uses all available cores or just one (#215). Users should be allowed to control more precisely the number of threads. 

### Goals
- [x] Determine a reasonable number of default threads
- [x] Implement changes to set that as a default
- [x] Determine whether users should be able to set a number of threads
- [ ] Make number of threads a user-accessible option in CLI

### Strategy
1. Investigate methods of importing user arguments before NumPy is imported. This appears to be surprisingly difficult. 

### Additional/Summary
Discovered this problem when I redlined my image processing server, triggering angry e-mails from IT. PR #215 prevents tedana from blowing up clusters but uses a hardcoded one-core limit via configuration file. (Thanks @tsalo!)

tsalo:
Optimally, the default should be to use all available cores, and users should be able to control it with an `n_threads` argument. That said, I've been having a fair amount of trouble controlling this within `tedana`.

We never explicitly control multithreading in `tedana`, so I believe that the best approach is to control the environment variables. This [SO post](https://stackoverflow.com/a/31622299/2589328) provides a list of variables we should set.

Setting the following in the terminal before calling `tedana` successfully limits the number of CPUs used on my Mac:
```
export MKL_NUM_THREADS=1
export NUMEXPR_NUM_THREADS=1
export OMP_NUM_THREADS=1
export VECLIB_MAXIMUM_THREADS=1
```
Unfortunately, setting these values within the workflow (with `os.environ`) so far _only_ works if at the top of the file, which means we couldn't set them as an argument. There might be a way around that, but supposedly it should only work if the modules that use these variables (esp. `numpy`) are imported _after_ setting the variables. However, even when I move imports of all non-builtin modules (except `argparse`) to below where the environment variables are set it doesn't work.

One interesting note though- on one core the ICA step takes a lot less time than using all available cores, although the results are somewhat different. We'll need to take a look at results to make sure they're still valid, but it's promising.

jbteves:
@tsalo I was re-reading the issues and I noticed that we have another anecdote of slow performance on HPC systems [here](https://github.com/ME-ICA/tedana/issues/141#issuecomment-430756537) in #141 
This makes me wonder if we should include figuring out what's going on with the multithreaded ICA as part of reliability analysis, thoughts?

tsalo:
[...]

PR #215, opened 2019-02-09
[ENH] Limit tedana to one core

tsalo:
References #188.

- Set all relevant environmental variables to limit the tedana workflow to one core.

#### Additional information
This PR is more of a temporary fix than a permanent solution. Ultimately, we want to be able to set the number of cores with an argument. Unfortunately, every attempt I've made to do that has failed. On the bright side, there is anecdotal evidence that limiting to one core will speed the pipeline up a fair bit, although we'll want to test that to make sure it's reproducible.

This will probably upset the linter because the environmental variables have to be set _before_ numpy and other multithreading packages are imported, which means that imports will be split at the top of the file. I'm not sure if we'll want to ignore this or to update the linter preferences to explicitly allow for it.


rmarkello:
This looks great, @tsalo! I agree it's a bit of a band-aid solution for now, but since most users will (ideally) be running `tedana` via the CLI this should hopefully avoid a lot of the issues of "can't set cores if numpy is imported first."

If you'd like to add:

```
[flake8]
ignore = E402
```

to the `setup.cfg` in the main directory then I think we can get rid of those linter warnings and this should be good to go.

rmarkello:
Interesting! I'm not totally sure why this one minor change suddenly caused _all_ the other linter errors... I did notice that we had previously been using `tox.ini` for `flake8` configuration, so it's possible that we overrode some settings? What do you think of the following approach:

1. Open a new PR migrating the `flake8` config stuff out of `tox.ini` and into `setup.cfg` (making `tox.ini` obsolete);
2. Address the "new" linter errors and, assuming that passes CI, merge into master; and
3. Rebase this onto master and hopefully have no problems!

Let me know, and apologies for making this more complicated than it should have been...

tsalo:
I'm happy to do that, but I've been testing locally and everything I do seems to disable the `putty-ignore` setting. I can't figure out how to make the setup.cfg (or tox.ini) ignore all of the things we want it to.

tsalo:
I've opened a new PR with those changes (#215) so we can discuss that there.

emdupre:
Out of curiosity, I checked the build time on Circle for these and we're not seeing any appreciable speed-boost. I wonder if that is also the case locally...

Should we make an issue about setting the number of cores programmatically, so we remember this is a (beautiful) band-aid ?

tsalo:
It's weird that there's no difference, although maybe it's because we only have access to two cores in CircleCI? Or the OS? Either way, I would definitely like to see if the duration changes predictably over a larger range of situations, but at least it will prevent `tedana` jobs from blowing up HPCs.

I think that #188 is meant to cover that, although maybe @jbteves could update that issue to reflect the direction we've chosen to go in (i.e., making control explicit) or we could close that one and open a new issue specifically for the `n_cores` option.

jbteves:
My kneejerk reaction is to leave it in that issue and update my comment so that the discussion history doesn't scatter; thoughts @tsalo and @emdupre?

emdupre:
:+1: to updating #188 ! Unless any other concerns, then, I'll merge this :rocket:

tsalo:
That works for me. I would recommend changing the name to something like "Add argument to set number of cores" and I can switch the label from `discussion` to `enhancement`. Thanks all!

PR #216, opened 2019-02-11
[STY] Consolidate linter settings and ignore some style warnings

tsalo:
Closes no issues. References #215.

- Move contents of `tox.ini` into `setup.cfg`
- Explicitly ignore `tests/` subfolder
- Ignore E126 (hanging indent does not match) because we explicitly want this for some multiline function calls
- Ignore E402 (import not at top of file) because we currently need to split imports in order to limit `numpy` and `sklearn` to one core
- ~~Resolve E226 warnings (no whitespace around arithmetic operator) that were previously ignored~~ You know what? I think we should ignore these. Removing the whitespace around string concatenation plus signs in function calls seems like best practice to me. Does anyone disagree?


tsalo:
@rmarkello I guess that the `putty-ignore` issue I was seeing locally does not reproduce in Travis, so maybe it's not a problem. I think we just need to make a decision about how we want to handle whitespace-around-operator errors. I vote to ignore them.

emdupre:
Sorry, I'm not sure why we'd ignore the whitespace around operator errors (also, not sure why they weren't getting flagged before ?!). Do you mean if we do:

```
concat = string1+string2
```

In that case, I'd really rather we switch to updated `.format()` calls. How many errors are we seeing ?

tsalo:
Sorry, I think I misrepresented what was actually causing most of these. A fair number are from actual math (e.g., `f05 = stats.f.ppf(q=1-0.05, dfn=1, dfd=n_echos-1)`). In that case it's getting mad about `1-0.05` and `n_echos-1`, but `f05 = stats.f.ppf(q=1 - 0.05, dfn=1, dfd=n_echos - 1)` is harder to read, IMO. Some instances probably should be fixed though. I'm happy to fix those, but still think we should ignore the error.

For the ones that are raised by string concatenation, there are only about 4 (all from my verbose additions). I can fix these as well.

emdupre:
Ah I see. Thanks for clarifying !

I'm hesitant to just ignore the error, since it sounds like some times we do want it and sometimes we don't. Two options I see for moving forward:

1. Use `#noqa` on the lines where we do want to ignore the error, or
2. Make those expressions easier to read, likely by enclosing them in parans. 

Personally I'd push for (2), since it seems possible that those lines could have other linting errors we'd ignore with `#noqa`. But WDYT ?

tsalo:
Okay, I've gone with option 2 when possible and simply added spaces when not. It should pass now.

emdupre:
LGTM ! :sparkler:

Issue #218, opened 2019-02-12
Incorporate features from MELODIC into PCA/ICA


tsalo:
### Summary
While we may not want to use MELODIC for our dimensionality reduction (i.e., PCA) and decomposition (ICA) steps since it would require non-Python dependencies and because a lot of its options aren't necessary for `tedana`, there are several options that we may want to integrate into our workflow. Here is a list:
- `--dimest`: use specific dim. estimation technique: lap, bic, mdl, aic, mean (default: lap)
    - These options may be better able than our currently implemented MLE to dimensionally reduce the data prior to ICA. I would like to see how many components each method generally estimates compared to our method, and to see if we can implement them in Python.
- `--nl`: nonlinearity: gauss, tanh, pow3, pow4
    - I don't actually know what this refers to. The documentation isn't exactly helpful. Still, though, it might be worth figuring out and integrating into `tedana`.
- `--maxit`: maximum number of iterations before restart
    - We may want to provide users with the option to override our default of 5000 iterations.
- `--maxrestart`: maximum number of restarts
    - We only try running ICA once, and continue with a warning if there is convergence failure. I definitely think this is better than raising an error, but what if we allowed it to change the random seed and re-run a few times by default? Preliminary tests from the reliability analysis indicate that some data fails every time, but other data only fails sometimes (and some doesn't seem to fail ever). This could improve the quality of results for those data that fail sometimes.

handwerkerd:
I think this is part of a bigger discussion on the correct way to do dimensionality reduction for multi-echo denoising and which options should be exposed to users. I agree that we don't want to use MELODIC itself because of unnecessary dependencies. Also I've experience some FSL functions being a bit brittle and having unknown consequences when using them in a way that differs from their originally intended purpose. Sticking with established general algorithms and figuring out which options to prioritize makes sense to me.

tsalo:
It's good to know that MELODIC has some issues. I think, at minimum, we should implement `maxit` and `maxrestart` (as indirectly requested [here](https://neurostars.org/t/questions-about-the-fmri-denoising-algorithm-tedana-convergence-warnings-sensitivity-to-rounding-of-echo-times-and-differences-between-released-and-unreleased-versions/3458/3?u=tsalo)) and we should look into more aggressive dimensionality estimation methods when possible. I think I can pretty easily implement the former.

tsalo:
[...]

PR #224, opened 2019-02-20
[ENH] Add options to control ICA attempts

tsalo:
References #218.

- Add options for the maximum number of iterations and restarts in ICA.
- Remove `fixed_seed` argument for `io.writeresults` and stop returning it in `tedica`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,7 +198,7 @@ jobs:
             source activate venv
             cd /tmp/data/five-echo/
             tedana -d p06.SBJ01_S09_Task11_e[1,2,3,4,5].sm.nii.gz \
-                -e 15.4 29.7 44.0 58.3 72.6 --verbose --maxit 500 --maxrestart 10 \
+                -e 15.4 29.7 44.0 58.3 72.6 --verbose \
                 --out-dir /tmp/data/five-echo/TED.five-echo/
       - run:
           name: Checking outputs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,7 +198,7 @@ jobs:
             source activate venv
             cd /tmp/data/five-echo/
             tedana -d p06.SBJ01_S09_Task11_e[1,2,3,4,5].sm.nii.gz \
-                -e 15.4 29.7 44.0 58.3 72.6 --verbose \
+                -e 15.4 29.7 44.0 58.3 72.6 --verbose --maxit 500 --maxrestart 10 \
                 --out-dir /tmp/data/five-echo/TED.five-echo/
       - run:
           name: Checking outputs
--- a/tedana/decomposition/eigendecomp.py
+++ b/tedana/decomposition/eigendecomp.py
@@ -392,7 +392,7 @@ def tedpca(catd, OCcatd, combmode, mask, t2s, t2sG,
     return n_components, kept_data
 
 
-def tedica(n_components, dd, fixed_seed):
+def tedica(n_components, dd, fixed_seed, maxit=5000, maxrestart=5):
     """
     Performs ICA on `dd` and returns mixing matrix
 
@@ -403,8 +403,14 @@ def tedica(n_components, dd, fixed_seed):
     dd : (S x T) :obj:`numpy.ndarray`
         Dimensionally reduced optimally combined functional data, where `S` is
         samples and `T` is time
-    fixed_seed : int
+    fixed_seed : :obj:`int`
         Seed for ensuring reproducibility of ICA results
+    maxit : :obj:`int`, optional
+        Maximum number of iterations for ICA. Default is 500.
+    maxrestart : :obj:`int`, optional
+        Maximum number of attempted decompositions to perform with different
+        random seeds. ICA will stop running if there is convergence prior to
+        reaching this limit. Default is 5.
 
     Returns
     -------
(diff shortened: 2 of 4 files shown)


emdupre:
This is amazing, thank you ! One question: Should `maxit` be 500 or 5000 ?

tsalo:
I'd like to drop it down to 500 (although I see now that I accidentally left it as 5000 in the tedica docstring).
Once we've merged this, though, I am planning to expand the reliability analysis so that we also look at how many iterations it generally takes to converge, as well as how many restarts we'll generally need (in conjunction with a reasonable number of iterations per restart) to have a strong chance of reaching convergence. I think we can come up with better defaults than 500 and 5 from that analysis.

emdupre:
Having this as a point for the reliability analysis seems like a great idea :+1: 

Just looking through our existing integration tests, it looks like the seed-updating works nicely, but isn't capped at 5 re-starts.

Three-echo:
```
INFO:tedana.decomposition.eigendecomp:ICA attempt 1 converged in 96 iterations
```

Five-echo
```
WARNING:tedana.decomposition.eigendecomp:ICA attempt 1 failed to converge after 500 iterations
WARNING:tedana.decomposition.eigendecomp:Random seed updated to 43
WARNING:tedana.decomposition.eigendecomp:ICA attempt 2 failed to converge after 500 iterations
WARNING:tedana.decomposition.eigendecomp:Random seed updated to 44
WARNING:tedana.decomposition.eigendecomp:ICA attempt 3 failed to converge after 500 iterations
WARNING:tedana.decomposition.eigendecomp:Random seed updated to 45
WARNING:tedana.decomposition.eigendecomp:ICA attempt 4 failed to converge after 500 iterations
WARNING:tedana.decomposition.eigendecomp:Random seed updated to 46
WARNING:tedana.decomposition.eigendecomp:ICA attempt 5 failed to converge after 500 iterations
WARNING:tedana.decomposition.eigendecomp:Random seed updated to 47
WARNING:tedana.decomposition.eigendecomp:ICA attempt 6 failed to converge after 500 iterations
WARNING:tedana.decomposition.eigendecomp:Random seed updated to 48
INFO:tedana.decomposition.eigendecomp:ICA attempt 7 converged in 146 iterations
```

tedana/decomposition/eigendecomp.py
```diff
@@ -423,20 +429,28 @@ def tedica(n_components, dd, fixed_seed):
...
+                LGR.warning('Random seed updated to {0}'.format(fixed_seed))
+            else:
+                LGR.info('ICA attempt {0} converged in {1} '
+                         'iterations'.format(i_attempt + 1, ica.n_iter_))
+                break
 
     mmix = ica.mixing_
     mmix = stats.zscore(mmix, axis=0)
-    return mmix, fixed_seed
+    return mmix
```
emdupre: Is there a reason to not return the `fixed_seed` value, now ? I think it'd still be nice to have to keep provenance, but maybe there's a reason to remove it that I'm missing....

tsalo: I think we must have originally used it in `io.writeresults` for something (filenames maybe?), but that function no longer uses the seed so it's not actually used anywhere. Plus, the log file will now have that info, along with number of iterations required for convergence, for those who might want it.

emdupre:
Ah, I'm an idiot, you have `maxrestart 10` in the test call. Should we just set that as the default ? Or was there a particular reason you used it there ?

tsalo:
I went with 5 for a default because I thought that was what MELODIC had (turns out it's 6), but would be happy to change to 10 to match the tests. I just chose 10 in the five-echo test because I knew convergence wasn't happening with 5 unless I chose a specific seed. Would you prefer 10 as our default? There's a nice symmetry to it since with the default `maxit` being 500 that would be equal to a cap of 5000 iterations (our old default).

tedana/decomposition/eigendecomp.py
emdupre: We did use it, I think to go in the summary file. But if we're ok that it's not just in the log file, that sounds fine by me ! :+1:

emdupre:
I think I'd prefer 10 as a default, since then at least all of our test datasets pass under our default params _and_ it has that symmetry :sparkles: If that sounds reasonable to you !

tsalo:
👍 That works for me. Hopefully we'll be able to come up with more robust defaults from the reliability analysis, but I'm quite happy with 500 and 10.

.circleci/config.yml
```diff
@@ -198,7 +198,7 @@ jobs:
             source activate venv
             cd /tmp/data/five-echo/
             tedana -d p06.SBJ01_S09_Task11_e[1,2,3,4,5].sm.nii.gz \
-                -e 15.4 29.7 44.0 58.3 72.6 --verbose \
+                -e 15.4 29.7 44.0 58.3 72.6 --verbose --maxit 500 --maxrestart 10 \
```
emdupre: Can we remove these extra arguments since they're the defaults ?

emdupre:
One small question, but this looks great ! Excited to have this merged !! :rocket:

.circleci/config.yml
tsalo: 👍

emdupre:
LGTM ! Merging :sparkles: :rocket:

Issue #113, opened 2018-08-18
Revisit make_adaptive_mask


tsalo:
Per discussion in #102, we want to take a closer look at how `make_adaptive_mask` determines which voxels have sufficient signal at each TE. The threshold applied below appears to be arbitrary:

https://github.com/ME-ICA/tedana/blob/0c5f93dc06e5fbbbe49811fa5eb717f7ea87ec22/tedana/utils/utils.py#L195-L210

As mentioned in #102, @rmarkello also thinks that the thresholds are hard-coded based on a three-echo dataset, so we'll want to look into that as well.

We have access to a five-echo dataset from @handwerkerd, so we can use that to investigate the method by which the signal quality threshold is determined and see if there's a better method out there. We can also use the quality of the log-linear fit to the decay model, and perhaps calculate TSNR for each echo's time series.

Does anyone have any thoughts on how we should do this?

BTW I've started looking into using TSNR to make an adaptive mask [here](https://github.com/tsalo/tedana/blob/7cd2abff339fd25a4d69ace95d295f52035053be/tsnr_mask.ipynb).

tsalo:
Given a [recent problem case](https://neurostars.org/t/questions-about-the-fmri-denoising-algorithm-tedana-convergence-warnings-sensitivity-to-rounding-of-echo-times-and-differences-between-released-and-unreleased-versions/3458) that appears to have at least partially caused by poor automatic masking, I think we should make sure to circle back to this issue. In the linked case, `make_adaptive_mask` identified almost the whole bounding box, while using `nilearn.masking.compute_epi_mask` (run on the first echo) to generate an explicit mask returned reasonable results.

tsalo:
Does anyone have any concerns about updating the workflow to use `nilearn.masking.compute_epi_mask` to create an initial mask (when no explicit mask is provided) prior to running `make_adaptive_mask`?

dowdlelt:
It seems that having a standardized automated way to get things started is very useful. As long as the masking isn't removing too much, such that the denoised output data isn't useful, it seems fine to have something that prevents the 'whole bounding-box' problem.

tsalo:
Sounds good to me. I've opened a new PR, #226, with the change. It doesn't solve all of the issues with `make_adaptive_mask`, but it's a start.

tsalo:
[...]

PR #226, opened 2019-02-27
[ENH] Automatically use Nilearn's EPI mask when no explicit mask is provided

tsalo:
References #113.

- When an explicit mask is not provided, use [nilearn.masking.compute_epi_mask](https://nilearn.github.io/modules/generated/nilearn.masking.compute_epi_mask.html) on the first echo to define the initial mask, which will be adjusted by `make_adaptive_mask`. The first echo should have the least dropout, and thus should produce the largest mask, but will avoid problems we've been seeing where the adaptive mask includes almost all of the voxels in the bounding box.

### DIFF
--- a/docs/approach.rst
+++ b/docs/approach.rst
@@ -45,6 +45,16 @@ number of echoes with "good" signal. When :math:`T_{2}^*` and :math:`S_{0}` are
 calculated below, each voxel's values are only calculated from the first :math:`n`
 echoes, where :math:`n` is the value for that voxel in the adaptive mask.
 
+.. note::
+    ``tedana`` allows users to provide their own mask. The adaptive mask will
+    be computed on this explicit mask, and may reduce it further based on the
+    data.
+    If a mask is not provided, ``tedana`` runs `nilearn.masking.compute_epi_mask`_
+    on the first echo's data to derive a mask prior to adaptive masking.
+    The workflow does this because the adaptive mask generation function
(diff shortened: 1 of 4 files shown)


emdupre:
I compared the artifacts for the five-echo dataset for this PR and current master -- it looks like (unsurprisingly) the mask differences are rippling down to the component level.

For example, in the [most recent master circle build](https://1852-110845855-gh.circle-artifacts.com/0/tmp/data/five-echo/TED.five-echo/comp_table_ica.txt) the first component is rejected for mid-kappa thresholding, while [with this PR](https://1893-110845855-gh.circle-artifacts.com/0/tmp/data/five-echo/TED.five-echo/comp_table_ica.txt) we see that the first component is now accepted.

Just to say explicitly: I am in favor of this change (!), but it might be something we're ready with our rationale for, either noted here or in the docs.

tsalo:
I can definitely update the documentation in the Processing pipeline details page, the tedana_workflow documentation, and the CLI documentation.

Also, the CI run where the first component (i.e., the one with the highest Kappa) is classified as mid-Kappa seems to be from #228. That shouldn't be possible, so I'm going to dig into that over in that PR.

emdupre:
Do you want to merge this, or do you want to wait pending investigations in #229 ?

tsalo:
I don't think anything in #229 will tie directly into the mask, so I'd be happy to merge this now.

emdupre:
Merging then :tada:

PR #227, opened 2019-02-27
[REF] Clean up outdated/unused functions

tsalo:
Closes None. This PR will clean up the package a little bit. If we ever plan to re-integrate v3.2 into `tedana`, we can pull the removed functions from old releases. 

- Remove the following functions tied to v3.2 of the selection criteria, which we no longer support:
    - `selection._utils.do_svm`
    - `utils.moments`
    - `utils.gaussian`
    - `utils.fitgaussian`
- Remove the following functions which are outdated due to package-wide changes:
    - `utils.get_dtype`: We dropped support for surfaces, so I think this is no longer necessary. It's not used anywhere in the package.
    - `io.ctabsel`: We use csv/tsv files, and will use jsons, for the component tables. The old text files are no longer used.
- Remove the `minimum` argument for `utils.make_adaptive_mask`. This argument set to `True` by default, but is actually never used.
- Remove `utils.make_min_mask`. This goes with removing the `minimum` argument for `utils.make_adaptive_mask`.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -145,14 +145,11 @@ API
    :template: function.rst
 
    tedana.io.split_ts
-
-   tedana.io.ctabsel
    tedana.io.filewrite
    tedana.io.gscontrol_mmix
    tedana.io.load_data
    tedana.io.new_nii_like
    tedana.io.write_split_ts
-   tedana.io.writect
    tedana.io.writefeats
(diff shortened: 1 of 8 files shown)


emdupre:
:tada: :tada: :tada: 

Love this. Merge when you're ready !

Issue #223, opened 2019-02-19
Concerns regarding TE-(in)dependence metric calculation


tsalo:
### Summary
I would like to better document and understand how model metrics are computed. I have some concerns about some of the steps- especially how `WTS` is calculated and how regressions with `get_coeffs` and `computefeats2` are performed.

### Additional Detail

1.  Compute optimal combination from T2* and echo data.
2.  Compute un-normalized weights for each component for optimally combined data (`WTS`) with `computefeats2` and normalized mixing matrix (when provided, otherwise regular mixing matrix).
    1. Use signs of skew values for weights (per component) to flip components and weights so that component time series positively align with the data.
        -  **PROBLEM**: The global signal trend for the 5-echo data is consistently being rendered as going _down_ in the ICA step (although it goes up, as expected, in the PCA step). Nothing I've tried, including using betas, z-statistics, mean, and skew, has fixed this, but it makes me suspicious of the sign determination method.
            - **NOTE**: This can be discussed in #316.
    2. Estimate normalized total variance explained per component from weights.
    -  **PROBLEM**: While the comments say "un-normalized weights", the function `computefeats2` does not return un-normalized weights. It computes the weights using the data after it has been z-scored over time. It then caps the weights at +/- 0.999 and z-transforms them using Fisher's transform as if they were correlation coefficients.
        - **SOLUTION**: If we're interested in using z-values for each of the components to the optimally combined data, we can run a regression and compute those z-scores (as in [this post](https://stackoverflow.com/a/42677750/2589328)).
        - **NOTE**: This can be discussed in #178.
    - ~~**PROBLEM**: I'm pretty sure that the normalized mixing matrix, which is only used with PCA, is actually z-scored along the wrong axis.~~ FIXED WITH #228
        - **SOLUTION**: We can just normalize each component's time series over time (instead of each volume's values over components).
3.  Compute un-normalized weights for each component for *demeaned* optimally combined data (`tsoc_B`) with `get_coeffs` and unnormalized mixing matrix.
    1. Convert weights to percent signal change units.
    2. Estimate total variance explained per component from weights.
    -  **PROBLEM**: Demeaning data should only impact weights if mixing matrix is not demeaned, which it _should_ be, but sometimes isn't.
        - **SOLUTION**: Mean-center the mixing matrix/X _within_ `get_coeffs` and drop the `add_const` argument (per #179).
4.  Compute un-normalized weights for each component for each echo with `get_coeffs`.
[...]

PR #208, opened 2019-01-30
[ENH] Adding simple figure output

dowdlelt:
Closes no open issues :( .

Changes proposed in this pull request:

* added optional --png argument
* adds a new module, viz.py, containing functions for figure creation
* If --png is called, a few plots are created:
* component time courses, weight maps, and fft
* Title includes variance, kappa & rho scores
* Kappa vs Rho scatter, size scales with variance explained
* Summary Figure
* Timecourses are color-coded for accepted/rejected, etc status
* variance is specified in plot title
* Color map update, color bar added

This is fairly ugly code, I think, but I am strongly of the opinion that having a relatively fast, built in and easy/dumb way to look at the output is essential. Also, can't ever get feedback on terrible coding if I keep it all hidden. 

This creates 3 types of figures:

# Component plot
One for each component in the ICA. Has timeseries, beta weight maps and fft. Zeros are masked.
image

# Kappa vs Rho
Scatter plot showing kappa vs rho values, with shape/color showing classification and size ~variance explained. 
image

# Summary Figure
Shows total variance explained by classification, and number of comps. 
image

Issue #179, opened 2019-01-09
get_coeffs run on mean-centered data for no reason


tsalo:
The function `get_coeffs` performs a least-squares fit of parameter `X` (generally a mixing matrix) on parameter `data` (generally optcom or other data). It is used in the following locations:
- [writeresults](https://github.com/ME-ICA/tedana/blob/49f70ac2e8af2b98b1a67465b55951909b5e6edb/tedana/io.py#L339): `data` is not demeaned. `add_const` is not provided, so it defaults to False.
- [split_ts](https://github.com/ME-ICA/tedana/blob/49f70ac2e8af2b98b1a67465b55951909b5e6edb/tedana/io.py#L140): `X` is ICA mixing matrix and `data` is demeaned data. `add_const` is not provided, so it defaults to False.
- [write_split_ts](https://github.com/ME-ICA/tedana/blob/49f70ac2e8af2b98b1a67465b55951909b5e6edb/tedana/io.py#L201): `X` is ICA mixing matrix and `data` is demeaned data. `add_const` is not provided, so it defaults to False.
- [fitmodels_direct](https://github.com/ME-ICA/tedana/blob/b048f491fae278fd1d8aa2baec11773d38dfef55/tedana/model/fit.py#L106): `X` is ICA or PCA mixing matrix and `data` is demeaned optimally combined data. `add_const` is not provided, so it defaults to False.
- [fitmodels_direct](https://github.com/ME-ICA/tedana/blob/b048f491fae278fd1d8aa2baec11773d38dfef55/tedana/model/fit.py#L122) (again): `X` is ICA or PCA mixing matrix and `data` is concatenated data (`catd`; not demeaned, afaik). `add_const` is not provided, so it defaults to False.
- [computefeats2](https://github.com/ME-ICA/tedana/blob/b048f491fae278fd1d8aa2baec11773d38dfef55/tedana/model/fit.py#L311): `X` is ICA or PCA mixing matrix and `data` is demeaned. `add_const` is not provided, so it defaults to False.

As far as I can tell, `add_const` isn't used anywhere in the package. That said, I also realized that demeaning `data` along axis 1 (which is how it is done when it is done) does not affect the results, so I'm not sure why we do it.

A really weird behavior that I don't understand at all is, when Y is not mean-centered, X gives the correct beta (with no need for an intercept) if X is mean-centered, but not if it's not. Does anyone know why that is?

[...]

Issue #178, opened 2019-01-07
Calculating coefficients in computefeats2


tsalo:
I believe that `computefeats2` tries to calculate correlation coefficients between the input data and the mixing matrix (see below) by variance normalizing the data and using `np.linalg.lstsq`. It then crops extreme values (<-0.999 and >0.999) and converts the correlation coefficients to z-values. I assumed that the cropping was to prevent large z-values with correlation coefficients approaching 1 and -1. However, it doesn't look like normalization is doing what we want, and the "correlation coefficients" can in fact end up quite large. I added in a couple of lines to print out the max and min values of `data_R` before cropping, and got values as high as +/- 12 with our five-echo test dataset.

If I'm right, then this is a bug, although I don't recall if `computefeats2` is used for anything beyond generating component weight maps.

https://github.com/ME-ICA/tedana/blob/1bc32e46445f1d8c5b69ba2000280e2f8f591321/tedana/model/fit.py#L310-L318

However, before I try fixing this, I want to make sure that I'm interpreting the function correctly. Is anyone familiar enough with this function to take a look?

BTW, to fix the issue (and get betas equivalent to correlation coefficients), I would do the following:
```python
data_z = stats.zscore(data[mask], axis=-1)
mmix_z = stats.zscore(mmix, axis=0)

# get betas of `data`~`mmix` and limit to range [-0.999, 0.999]
data_R = get_coeffs(data_z, mmix_z, mask=None)
```

tsalo:
If we're interested in getting z-scores for the betas, then I think the following is an appropriate replacement for `computefeats2`:

```
def t_to_z(t_values, dof):
    """
    From Vanessa Sochat's TtoZ package.
    """
    # Select just the nonzero voxels
    nonzero = t_values[t_values != 0]

    # We will store our results here
    z_values = np.zeros(len(nonzero))

    # Select values less than or == 0, and greater than zero
    c = np.zeros(len(nonzero))
    k1 = (nonzero <= c)
    k2 = (nonzero > c)

    # Subset the data into two sets
    t1 = nonzero[k1]
    t2 = nonzero[k2]

    # Calculate p values for <=0
    p_values_t1 = stats.t.cdf(t1, df=dof)
    z_values_t1 = stats.norm.ppf(p_values_t1)
    z_values_t1[np.isinf(z_values_t1)] = stats.norm.ppf(1e-16)

    # Calculate p values for > 0
[...]

PR #228, opened 2019-02-28
[FIX] Normalize PCA mixing matrix over time, not component

tsalo:
References #223.  One of the concerns I brought up in #223 is that the normalized PCA mixing matrix, which is only used to calculate the weight maps (`WTS`) within `fitmodels_direct`, is normalized over component, rather than over time. This strikes me as invalid, though I could be misinterpreting the purpose of the normalization. This will not impact the MLE dimensionality estimation, but should improve the validity of the Kundu PCA decision tree.

- Z-score the PCA mixing matrix over time (per component), rather than over components (per TR).


tsalo:
The first echo in the five-echo dataset (i.e., the one with the highest kappa) is being rejected as a mid-Kappa component. This makes no sense for a couple of reasons:
1. It's the highest Kappa component. It really can't be a mid-Kappa component if it's the highest Kappa component.
2. The proposed change only affects TEDPCA metric calculation. Since this test uses MLE dimensionality estimation, those metrics _aren't even used_!

tsalo:
My bad. This issue is definitely showing up in `master`, so it's not linked to the proposed changes.

emdupre:
To the best of my understanding, the purpose of normalization was so that each components would be on the same scale when passed on to the PCA decision tree (and therefore could use e.g. the same thresholds). If we normalize by time, that assumption no longer holds. 

Could you explain a little more why you were thinking it would make sense to normalize over time ?

tsalo:
The old way was z-scoring each timepoint across components, while this fix z-scores each component across timepoints. The former approach doesn't just rescale the time series, but also changes them. Granted, the difference is small (the correlation for a random array with 80 components and 160 timepoints before vs. after z-scoring is ~0.99), but it should also not be a valid change to make, as far as I can tell.

emdupre:
Ah, sorry, I think I misunderstood ! So to clarify (for me):

What is the correlation between the original components and this new normalization ? The 0.99 is with the old normalization, correct ?

I think we should be checking the PCA selection tree in the integration tests. This seems like as good a time as any, since I want to know how this is impacting the PCA selection. WDYT ? Should we add it to the three echo dataset ?

tsalo:
The time series correlate perfectly after z-scoring the new way. Yeah, the old way gets 0.999.

That's a good idea. I can change the three-echo integration test in this PR.

emdupre:
If this is correlating at 1.0 then I'm wondering if it's really a necessary normalization -- obviously removing the old one seems to be !

If we can fix the merge conflict (sorry, I think I pulled it in with #208 ) then this LGTM !

tsalo:
It's used to generate the normalized version of the mixing matrix, which is used in `fitmodels_direct`. Hypothetically, the scale of the time series should impact the parameter estimates, and I believe that the goal of the z-scoring is to make the parameter estimates from `computefeats2` equivalent to betas. This is definitely something worth discussing in the larger context of how metric calculation should be performed (e.g., in #178, #179, and #223), but if we work under the assumption that the current version of `fitmodels_direct` is generally correct, then this is still a bug that needs to be fixed.

emdupre:
I think this is good to merge. It sounds like we need to do some issue clean-up around metric calculation -- we can deal with that after this is in :)

PR #229, opened 2019-03-01
[NO-MRG] Debug mid-Kappa classification

tsalo:
References https://github.com/ME-ICA/tedana/pull/226#issuecomment-468533605. The top component is being classified as mid-Kappa. This isn't happening in #226 or locally, so it is definitely fragile.


tsalo:
I don't think I'm going to be able to figure out much about issues with component selection at this point. The rationale behind each step is just missing. While I can see the value of most of the selection steps, I can't determine why each metric was chosen specifically, what alternatives were originally considered, or why they're in the order they're in. I think the best next step is to close this PR and to focus on making sure that the metrics used by the component selection algorithm are properly calculated. We have a few issues open that touch on that.

emdupre:
Yes, I think component selection is the next big hurdle. Would you want to talk about that in the next developers's call, @tsalo ?

Thanks for looking into this !!

PR #236, opened 2019-03-18
[ENH] Update Figure Generation Code

### LINKED ISSUE #232: Improve generated figures
_Suggestions from @handwerkerd for improving generated figures. Transferred from comment on #208._ 

  * Colormap options for the components. The red/blue color maps makes the extremes pop out while artifacts in ICA components are sometimes clearer with grayscale. Even if it's out a parameter, having a way to play with the colorscale without making deep edits to the code would be useful.
  * Each component screen can also include numerical information besides variance. I suggest adding at least Kappa & Rho values.
  * Whether a component was accepted/rejected/ignored should be visible with each component output. Ideally the part of the decision tree that resulted in that classification can also be included
  * I'm not sure a bar graph of accepted/rejected/ignored variance has added value over just displaying the numbers. If anything, this might be one of the few cases where a pie chart is visually informative.

dowdlelt:
Closes #232 .

- Adds more details to title of plots, including clearly labeling component decision and rationale(s), if rejected or ignored
- trims away all zero slices prior to plotting brains, so that every figure slice contains data, rather than producing completely empty blank spaces
- Replaces bar plot with more informative pie chart
- pie chart shows total variance (outer ring) as well as individual component variance (inner ring) along with number of components in each category. 
- if some variance is unexplained, pie chart will show as well, though calculation is not perfectly accurate at this time
- a number of fixes regarding matplotlib debugging/usage mentioned by @tsalo added
- Time series and fft plots no longer have yticks - not that useful
- fft now completely fills plot
- custom color map can be requested, ex. --png-cmap 'bone. If not used or specified incorrectly, code will use use default
- 'figure' folder creation moved to tedana workflow, figures now written using out_dir variable
- get_spectrum moved to utils

Example of pie chart: 
image

Example of new timeseries plots, with titles.
image

### DIFF
--- a/tedana/viz.py
+++ b/tedana/viz.py
@@ -77,7 +77,7 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
     """
     # Check that colormap provided exists
     if png_cmap not in plt.colormaps():
-        LGR.info('Provided colormap is not valid, proceeding with coolwarm')
+        LGR.warning('Provided colormap is not valid, proceeding with coolwarm')
         png_cmap = 'coolwarm'
     # regenerate the beta images
     ts_B = model.get_coeffs(ts, mmix, mask)
@@ -101,12 +101,12 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
             expl_text = 'accepted'
         elif compnum in rej:
             line_color = 'r'
-            expl_text = 'rejection reason: ' + comptable.iloc[compnum]["rationale"]
+            expl_text = 'rejection reason(s): ' + comptable.iloc[compnum]["rationale"]
         elif compnum in midk:
             line_color = 'm'
         else:
             line_color = 'k'
-            expl_text = 'ignored'
+            expl_text = 'ignored reason(s): ' + comptable.iloc[compnum]["rationale"]
 
         allplot = plt.figure(figsize=(10, 9))
         ax_ts = plt.subplot2grid((5, 6), (0, 0),
@@ -186,7 +186,7 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
         # Fix spacing so TR label does overlap with other plots
         allplot.subplots_adjust(hspace=0.4)
         plot_name = 'comp_{}.png'.format(str(compnum).zfill(3))
-        compplot_name = os.path.join(out_dir, 'figures', plot_name)
+        compplot_name = os.path.join(out_dir, plot_name)
         plt.savefig(compplot_name)
         plt.close()
 
@@ -233,7 +233,7 @@ def write_kappa_scatter(comptable, out_dir):
     ax_scatter.xaxis.label.set_fontsize(20)
     ax_scatter.yaxis.label.set_fontsize(20)
     ax_scatter.title.set_fontsize(25)
-    scatter_title = os.path.join(out_dir, 'figures', 'Kappa_vs_Rho_Scatter.png')
+    scatter_title = os.path.join(out_dir, 'Kappa_vs_Rho_Scatter.png')
     plt.savefig(scatter_title)
 
     plt.close()
@@ -303,5 +303,5 @@ def write_summary_fig(comptable, out_dir):
     ax.set_title('Variance Explained By Classification', fontdict={'fontsize': 28})
     if unexpl_var < [0.001]:
         plt.text(1, -1, '*Unexplained Variance less than 0.001', fontdict={'fontsize': 12})
-    sumfig_title = os.path.join(out_dir, 'figures', 'Component_Overview.png')
+    sumfig_title = os.path.join(out_dir, 'Component_Overview.png')
     plt.savefig(sumfig_title)
--- a/tedana/viz.py
+++ b/tedana/viz.py
@@ -12,6 +12,23 @@ from tedana import model
 LGR = logging.getLogger(__name__)
 
 
+def trim_edge_zeros(arr):
+    mask = arr!=0
+    bounding_box = tuple(
+                         slice(np.min(indexes), np.max(indexes) + 1)
+                         for indexes in np.where(mask))
+    return arr[bounding_box]
+    """
+    Trims away the zero-filled slices that surround many 3/4D arrays
+
+    Parameters
+    ----------
+    ndarray: (S x T) array like
+        the array on which the trimming should occur
+    """
+
+
+
 def write_comp_figs(ts, mask, comptable, mmix, n_vols,
                     acc, rej, midk, empty, ref_img):
     """
@@ -50,7 +67,10 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
     # regenerate the beta images
     ts_B = model.get_coeffs(ts, mmix, mask)
     ts_B = ts_B.reshape(ref_img.shape[:3] + ts_B.shape[1:])
-    # Mask out zeros
+    # trim edges from ts_B array
+    ts_B = trim_edge_zeros(ts_B)
+
+    # Mask out remaining zeros
     ts_B = np.ma.masked_where(ts_B == 0, ts_B)
 
     # Get repetition time from ref_img
--- a/tedana/viz.py
+++ b/tedana/viz.py
@@ -41,7 +41,6 @@ def trim_edge_zeros(arr):
     return arr[bounding_box]
 
 
-
 def write_comp_figs(ts, mask, comptable, mmix, ref_img, out_dir,
                     png_cmap):
     """
--- a/tedana/viz.py
+++ b/tedana/viz.py
@@ -40,8 +40,7 @@ def trim_edge_zeros(arr):
     """
 
 
-def write_comp_figs(ts, mask, comptable, mmix, n_vols,
-                    acc, rej, midk, empty, ref_img, out_dir,
+def write_comp_figs(ts, mask, comptable, mmix, ref_img, out_dir,
                     png_cmap):
     """
     Creates static figures that highlight certain aspects of tedana processing
@@ -61,20 +60,13 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
     mmix : (C x T) array_like
         Mixing matrix for converting input data to component space, where `C`
         is components and `T` is the same as in `data`
-    n_vols : :obj:`int`
-        Number of volumes in original time series
-    acc : :obj:`list`
-        Indices of accepted (BOLD) components in `mmix`
-    rej : :obj:`list`
-        Indices of rejected (non-BOLD) components in `mmix`
-    midk : :obj:`list`
-        Indices of mid-K (questionable) components in `mmix`
-    empty : :obj:`list`
-        Indices of ignored components in `mmix`
     ref_img : :obj:`str` or img_like
         Reference image to dictate how outputs are saved to disk
 
     """
+    # Get the lenght of the timeseries
+    n_vols = len(mmix)
+
     # Check that colormap provided exists
     if png_cmap not in plt.colormaps():
         LGR.warning('Provided colormap is not valid, proceeding with coolwarm')
--- a/tedana/viz.py
+++ b/tedana/viz.py
@@ -289,7 +289,7 @@ def write_summary_fig(comptable, out_dir):
 
     labels = counts.values()
 
-    fig, ax = plt.subplots(figsize=(20, 10))
+    fig, ax = plt.subplots(figsize=(16, 10))
     size = 0.3
     # Build outer, overall pie chart, and then inner individual comp pie
     ax.pie(var_expl, radius=1, colors=outer_colors, labels=labels,
--- a/tedana/viz.py
+++ b/tedana/viz.py
@@ -231,8 +231,10 @@ def write_kappa_scatter(comptable):
 
 def write_summary_fig(comptable):
     """
-    Creates a bar graph showing total variance explained by each component
-    as well as the number of components identified for each category.
+    Creates a pie chart showing 1) The total variance explained by each 
+    component in the outer ring, 2) the variance explained by each 
+    individual component in the inner ring, 3) counts of each classification 
+    and 4) the amount of unexplained variance.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


tedana/workflows/tedana.py
```diff
@@ -422,15 +430,20 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
     if png:
         LGR.info('Making figures folder with static component maps and '
                  'timecourse plots.')
+        # make figure folder first
+        if not os.path.isdir('figures'):
+            os.mkdir('figures')
```
```suggestion
        if not op.isdir(op.join(out_dir, 'figures')):
            os.mkdir(op.join(out_dir, 'figures'))
```

tedana/workflows/tedana.py
```diff
@@ -221,6 +226,9 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
         Generate intermediate and additional files. Default is False.
     png : obj:'bool', optional
         Generate simple plots and figures. Default is false.
+    png-cmap : obj:'str', optional
```
tsalo: png_cmap

tedana/viz.py
```diff
@@ -5,15 +5,45 @@
...
+    ndarray: (S x T) array_like
+        an array with reduced dimensions, such that the array contains only
+        non_zero values from edge to edge.
+    """
+
 
 
 def write_comp_figs(ts, mask, comptable, mmix, n_vols,
-                    acc, rej, midk, empty, ref_img):
+                    acc, rej, midk, empty, ref_img, out_dir,
```
tsalo: This function shouldn't need `acc`, `rej`, `midk`, or `empty` since it already gets `comptable`.

tedana/viz.py
```diff
@@ -46,41 +76,47 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
         Reference image to dictate how outputs are saved to disk
 
     """
-
+    # Check that colormap provided exists
+    if png_cmap not in plt.colormaps():
+        LGR.info('Provided colormap is not valid, proceeding with coolwarm')
```
tsalo: I think this should be a warning.

tedana/viz.py
```diff
@@ -46,41 +76,47 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
...
-
+    expl_text = ''
     for compnum in range(0, mmix.shape[1], 1):
 
         if compnum in acc:
             line_color = 'g'
+            expl_text = 'accepted'
         elif compnum in rej:
             line_color = 'r'
+            expl_text = 'rejection reason: ' + comptable.iloc[compnum]["rationale"]
```
tsalo: reason(s) or rationale(s). There can be more than one reason for rejection.

tedana/viz.py
```diff
@@ -46,41 +76,47 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
...
         elif compnum in midk:
             line_color = 'm'
         else:
             line_color = 'k'
+            expl_text = 'ignored'
```
tsalo: There are also rationales associated with ignored components.

tedana/viz.py
```diff
@@ -95,18 +131,19 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
...
 
         ax_ts.plot(mmix[:, compnum], color=line_color)
 
         # Title will include variance from comptable
-        comp_var = "{0:.2f}".format(comptable.iloc[compnum][3])
-        comp_kappa = "{0:.2f}".format(comptable.iloc[compnum][1])
-        comp_rho = "{0:.2f}".format(comptable.iloc[compnum][2])
-        plt_title = 'Comp. {}: variance: {}%, kappa: {}, rho: {}'.format(compnum, comp_var,
-                                                                         comp_kappa, comp_rho)
+        comp_var = "{0:.2f}".format(comptable.iloc[compnum]["variance explained"])
```
tsalo: Should we use "variance explained" or "normalized variance explained" here and elsewhere?
@emdupre Thoughts?

tedana/viz.py
tsalo: Also we can infer `n_vols` from `ts` or `mmix`.

tsalo:
> @tsalo just noticed that I've duplicated your work - I looked through your viz_update branch and it seems like I wasn't too far off the mark on my fixes but would still appreciate a close look.

I did play around a bit in via-update, but went a little too far with refactoring your code, which is why I don't plan to open a PR or anything. It did help me understand what was happening though.

tedana/viz.py
tsalo: You'll also need to strip off trailing semicolons.

tedana/workflows/tedana.py
```diff
@@ -422,15 +430,21 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
                  'timecourse plots.')
+        # make figure folder first
+        if not os.path.isdir('figures'):
+        if not op.isdir(op.join(out_dir, 'figures')):
+            os.mkdir(op.join(out_dir, 'figures'))
+
         viz.write_comp_figs(data_oc, mask=mask, comptable=comptable, mmix=mmix,
                             n_vols=n_vols, acc=acc, rej=rej, midk=midk,
-                            empty=ign, ref_img=ref_img)
+                            empty=ign, ref_img=ref_img, out_dir=out_dir,
```
tsalo: Perhaps `out_dir` for these functions should be `op.join(out_dir, 'figures')` (i.e., `ref_img=ref_img, out_dir=op.join(out_dir, 'figures'), ...`

tedana/workflows/tedana.py
```diff
@@ -422,15 +430,21 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
     if png:
         LGR.info('Making figures folder with static component maps and '
                  'timecourse plots.')
+        # make figure folder first
+        if not os.path.isdir('figures'):
```
tsalo: I think I did a bad job using the suggestion function in GitHub. You need to delete this line now.

tedana/workflows/tedana.py
dowdlelt: I was matching --out-dir with that choice. I do prefer the underscore however.

tsalo: Sorry, I just mean for the function's docstring. The CLI argument is `--png-cmap`, but the Python function's argument is `png_cmap`, so the docstring for the function should have `png_cmap`.

tedana/workflows/tedana.py
dowdlelt: gotcha

tedana/viz.py
dowdlelt: I see, added now

tedana/viz.py
dowdlelt: does that that feeling about it. changed

tedana/viz.py
dowdlelt: I use each of those to color the lines, rather than redoing the work to pull it out of the comp table again. Unless those are going to be removed in the future?

tedana/workflows/tedana.py
dowdlelt: I like this, cleans up figure writing a bit

tedana/viz.py
dowdlelt: Good point. I'll fiddle with that in a bit.

tedana/workflows/tedana.py
dowdlelt: ah, i missed that. I'll go ahead and update that

tedana/viz.py
dowdlelt: should work well now, else statement now will deal with a new classification somewhat gracefully by plotting a grey line (I think, testing now)

tedana/viz.py
dowdlelt: edited, but have not changed the semicolons, not quite sure how to do that (without removing the middle semicolons).

tsalo: It turns out `rstrip` accepts characters, so you can do `comptable.iloc[compnum]["rationale"].rstrip(';')`. Alternatively, it might be worth it to do that in selcomps so that we don't have to deal with this separately in each location where we reference rationales. Something like `comptable['rationale'] = comptable['rationale'].str.rstrip(';')` at the end of selcomps would probably work.

dowdlelt:
Travis really doesn't like 3 long lines, so I'll think on that. Otherwise, I've dealt with everything but the trailing semicolons.

tedana/viz.py
dowdlelt: Trying rstrip now.

dowdlelt: I remove the trailing ';'  with str.rstrip prior to the loops, so it is all done at once, rather than over and over. Seems to work well, and doesn't interfere with multiple rationales.

tedana/viz.py
dowdlelt: Thinking more on this, both of these sum up to be close to 100, even in cases where the PCs entered into in the ICA step explain <100% of the variance, right, because ICs are not orthogonal? I'm looking at an older run of the meica tedana, where the comp_table says 76.7% of the variance was explained, but summing the Variance and Variance norm columns gives me 100.03 and 96.78% respectively. 

Means that `100 - sum(variance_expl)` that calculates the unexplained variance is never going to work quite right. 

Nor would calculating the variance explained (as is done in io.py, around line 202) and adding that in give the right number - because then the pie chart would be relative to the total (ie ~100 variance from ICA + unexplained = some_number >100). Perhaps some scaling would be needed for the values, prior to giving them to that part of figure creation.

dowdlelt:
This now seems to work great without complaints. Pending further review and general acceptance of this pull request, I am curious if we would want to save the variance explained question for another round...

tsalo:
It looks great to me. I'm happy to hold off on discussing variance explained. In fact, my concerns regarding the variance explained calculated within `fitmodels_direct` are already outlined in #223, and we can circle back to the plots once that's figured out.

dowdlelt:
I've been meaning to dig into #223 more - guess we all have a bit more time with the call delayed. Thanks for all the help with this!

tedana/viz.py
```diff
@@ -32,55 +60,60 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
...
     ref_img : :obj:`str` or img_like
         Reference image to dictate how outputs are saved to disk
 
     """
+    # Get the lenght of the timeseries
+    n_vols = len(mmix)
 
+    # Check that colormap provided exists
+    if png_cmap not in plt.colormaps():
+        LGR.warning('Provided colormap is not valid, proceeding with coolwarm')
```
emdupre: Very light stylistic preference -- feel free to ignore !

```suggestion
        LGR.warning('Provided colormap is not recognized, proceeding with default')
```

tedana/viz.py
```diff
@@ -32,55 +60,60 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
...
-        Number of volumes in original time series
-    acc : :obj:`list`
-        Indices of accepted (BOLD) components in `mmix`
-    rej : :obj:`list`
-        Indices of rejected (non-BOLD) components in `mmix`
-    midk : :obj:`list`
-        Indices of mid-K (questionable) components in `mmix`
-    empty : :obj:`list`
-        Indices of ignored components in `mmix`
     ref_img : :obj:`str` or img_like
```
emdupre: Can we update the docstring to explain `png_cmap` ?

tedana/viz.py
```diff
@@ -32,55 +60,60 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
...
-        else:
+            expl_text = 'rejection reason(s): ' + comptable.iloc[compnum]["rationale"]
+        elif comptable.iloc[compnum]["classification"] == 'ignored':
             line_color = 'k'
+            expl_text = 'ignored reason(s): ' + comptable.iloc[compnum]["rationale"]
+        else:
+            # Classification not added
+            # If new, this will keep code running
+            line_color = '0.75'
+            expl_text = 'unknown classification'
```
emdupre: Instead of "Unknown" could we say "Other" ? I ask since we likely will _know_ what new classifications are, even if they don't fit into these bins.

tedana/viz.py
emdupre: :+1: for postponing the discussion until #223  !

tedana/viz.py
```diff
@@ -144,17 +178,18 @@ def write_comp_figs(ts, mask, comptable, mmix, n_vols,
...
         allplot.subplots_adjust(hspace=0.4)
         plot_name = 'comp_{}.png'.format(str(compnum).zfill(3))
-        compplot_name = os.path.join('figures', plot_name)
+        compplot_name = os.path.join(out_dir, plot_name)
         plt.savefig(compplot_name)
         plt.close()
 
 
-def write_kappa_scatter(comptable):
+def write_kappa_scatter(comptable, out_dir):
```
emdupre: Can we add `out_dir` to the docstring ?

tedana/viz.py
```diff
@@ -196,16 +231,18 @@ def write_kappa_scatter(comptable):
...
     ax_scatter.title.set_fontsize(25)
-    scatter_title = os.path.join('figures', 'Kappa_vs_Rho_Scatter.png')
+    scatter_title = os.path.join(out_dir, 'Kappa_vs_Rho_Scatter.png')
     plt.savefig(scatter_title)
 
     plt.close()
 
 
-def write_summary_fig(comptable):
+def write_summary_fig(comptable, out_dir):
```
emdupre: Documenting `out_dir` again

tedana/viz.py
```diff
@@ -5,15 +5,43 @@
...
 
-def write_comp_figs(ts, mask, comptable, mmix, n_vols,
-                    acc, rej, midk, empty, ref_img):
+def trim_edge_zeros(arr):
+    mask = arr != 0
+    bounding_box = tuple(
+                         slice(np.min(indexes), np.max(indexes) + 1)
+                         for indexes in np.where(mask))
+    return arr[bounding_box]
+    """
```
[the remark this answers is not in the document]
dowdlelt: docstring now in its home

tedana/viz.py
dowdlelt: yup

tedana/viz.py
dowdlelt: I'm good with other

tedana/viz.py
dowdlelt: done

tedana/viz.py
dowdlelt: and done

dowdlelt:
Awesome Awesome. Thanks for all the feedback folks

tsalo:
Since everyone's happy with the current version, are we ready to merge?

emdupre:
Push the button whenever you're ready, @tsalo :+1:

PR #238, opened 2019-03-20
[FIX, DOC] Use countnoise in decision table within selcomps

tsalo:
Closes None. In the process of documenting `selcomps`, I realized that I must have made a mistake when I added in table support to `selcomps`. Namely, the variable `countnoise` was not being incorporated properly in the decision tables used for certain component selection steps. `countnoise` _was_ used properly in other steps, so I think the impact is minimal.

- Remove `countnoise` variable from `selcomps`. When we switched over to using tables, I forgot to remove this variable, so it was being initialized as an array of zeros but wasn't filled out (because the associated column of `comptable` was). `countnoise` was only used in the decision table steps, so it was essentially being ignored as a metric there.
- Add some documentation about the different steps of the component selection decision tree. The goal here is not just to determine how each metric is used, but also why that's the best way to evaluate what that metric is meant to evaluate.
- Split rationale I002 (Rho greater than Kappa or more significant voxels in S0 model than R2 model) into I002 (Rho greater than Kappa) and I003 (more significant voxels in S0 model than R2 model), and shift all other rationales accordingly. While this step was originally consolidated, there is an "or" in the step, so there's no reason to keep these two rationales together.
- Update candartA and candartB rationales. Both are considered kinds of mid-Kappa, so I changed the rationales from "Artifact candidate type [A|B]" to "Mid-Kappa artifact type [A|B]".
- Replace anomalous `ct_df` variable in `tedpca` with `comptable` to match package convention.

### DIFF
--- a/tedana/selection/select_comps.py
+++ b/tedana/selection/select_comps.py
@@ -95,17 +95,23 @@ def selcomps(seldict, comptable, mmix, manacc, n_echos):
         # Move decision columns to end
         comptable = comptable[[c for c in comptable if c not in cols_at_end] +
                               [c for c in cols_at_end if c in comptable]]
+        comptable['rationale'] = comptable['rationale'].str.rstrip(';')
         return comptable
 
     """
-    Do some tallies for no. of significant voxels
+    Tally number of significant voxels for cluster-extent thresholded R2 and S0
+    model F-statistic maps.
     """
     countnoise = np.zeros(n_comps)
     comptable['countsigFR2'] = F_R2_clmaps.sum(axis=0)
     comptable['countsigFS0'] = F_S0_clmaps.sum(axis=0)
 
     """
-    Make table of dice values
+    Generate Dice values for R2 and S0 models
+    - dice_FR2: Dice value of cluster-extent thresholded maps of R2-model betas
+      and F-statistics.
+    - dice_FS0: Dice value of cluster-extent thresholded maps of S0-model betas
+      and F-statistics.
     """
     comptable['dice_FR2'] = np.zeros(all_comps.shape[0])
     comptable['dice_FS0'] = np.zeros(all_comps.shape[0])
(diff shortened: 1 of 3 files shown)


tedana/selection/select_comps.py
```diff
@@ -188,87 +223,99 @@ def selcomps(seldict, comptable, mmix, manacc, n_echos):
...
                           getelbow(comptable['kappa'], return_val=True)))
     rho_elbow = np.mean((getelbow(comptable.loc[ncls, 'rho'], return_val=True),
                          getelbow(comptable['rho'], return_val=True),
                          utils.getfbounds(n_echos)[0]))
 
-    # Initial guess of good components based on Kappa and Rho elbows
-    good_guess = ncls[(comptable.loc[ncls, 'kappa'] >= kappa_elbow) &
-                      (comptable.loc[ncls, 'rho'] < rho_elbow)]
+    # Initial guess of accepted components based on Kappa and Rho elbows
+    acc_poss = ncls[(comptable.loc[ncls, 'kappa'] >= kappa_elbow) &
```
emdupre: Is the `poss` in this variable name short for `possible` ? I stumbled on this at first. WDYT about `acc_prov` for `provisionally accepted` ? Either way, can we update the comment immediately above to spell this out ?

tedana/selection/select_comps.py
```diff
@@ -310,19 +357,20 @@ def selcomps(seldict, comptable, mmix, manacc, n_echos):
...
         ign_add1 = np.intersect1d(
             acc[comptable.loc[acc, 'kappa'] <= kappa_elbow],
             acc[comptable.loc[acc, 'variance explained'] > new_varex_lower])
         ign_add1 = np.setdiff1d(ign_add1, midk)
         comptable.loc[ign_add1, 'classification'] = 'ignored'
-        comptable.loc[ign_add1, 'rationale'] += 'I011;'
-        ign = np.union1d(ign, ign_add1)
-        acc = np.setdiff1d(acc, np.union1d(midk, ign))
+        comptable.loc[ign_add1, 'rationale'] += 'I012;'
+        ign = np.union1d(ign, ign_add1)  # unnecessary
```
emdupre: Should we just remove these two lines if they're unnecessary ?

tedana/selection/select_comps.py
tsalo: That's totally cool by me.

tedana/selection/select_comps.py
tsalo: Yup, definitely. Will do.

PR #240, opened 2019-03-22
[REF] Add gscontrol module

tsalo:
Closes None.

- Move gscontrol_raw and gscontrol_mmix into new gscontrol module. They are currently in model.fit and io, respectively, which are not good places for them.


tsalo:
Merging now!

PR #241, opened 2019-03-22
[FIX] center component map at zero

dowdlelt:
Closes no issue

- Center Color bar at zero, per dev call request.
Before:
image

After:
image

### DIFF
--- a/tedana/viz.py
+++ b/tedana/viz.py
@@ -150,9 +150,11 @@ def write_comp_figs(ts, mask, comptable, mmix, ref_img, out_dir,
         title = ax_ts.set_title(plt_title)
         title.set_y(1.5)
 
-        # Set range to ~1/10th of max beta
-        imgmax = ts_B[:, :, :, compnum].max() * .1
-        imgmin = ts_B[:, :, :, compnum].min() * .1
+        # Set range to ~1/10th of max positive or negative beta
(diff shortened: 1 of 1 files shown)


tedana/viz.py
```diff
@@ -150,9 +150,11 @@ def write_comp_figs(ts, mask, comptable, mmix, ref_img, out_dir,
         title = ax_ts.set_title(plt_title)
         title.set_y(1.5)
 
-        # Set range to ~1/10th of max beta
-        imgmax = ts_B[:, :, :, compnum].max() * .1
-        imgmin = ts_B[:, :, :, compnum].min() * .1
+        # Set range to ~1/10th of max positive or negative beta
+        temp_max = np.abs(ts_B[:, :, :, compnum].max())
+        temp_min = np.abs(ts_B[:, :, :, compnum].min())
+        imgmax = np.max([temp_max, temp_min]) * .1
```
tsalo: I think this could be simplified to 
```suggestion
        imgmax = 0.1 * np.abs(ts_B[:, :, :, compnum]).max()
```

tsalo:
The figure doesn't look quite as good, but I think it's more informative this way. Just one comment that could simplify the code, and then I think it'll be ready to merge. Thanks!

PR #245, opened 2019-03-30
Added binder example#149

sushmaunnibhavi:
Closes # .

-added binder link
-updated info.py


tedana/info.py
```diff
@@ -35,7 +34,9 @@
     'scipy',
     'versioneer',
     'pandas',
-    'matplotlib'
+    'matplotlib',
```
jbteves: I'm sorry to nitpick like this, but could you add `seaborn` and `nilearn` before matplotlib so that the diffs make it clear that dependencies are only being added?

PR #260, opened 2019-04-20
[FIX] Fix use of d_table_score

tsalo:
Closes None. This was originally part of #247, but has been split off because it is self-contained and should be merged separately.

At one point, the d_table_score metric is used when d_table_score_scrub should have been used. This is a minor bug.

- Replace `d_table_score` with `d_table_score_scrub` in one instance.


tsalo:
This is a minor, but clear, bug, so I'd like to merge it now. I don't think it requires review.

PR #261, opened 2019-04-20
[FIX] Sort comptable by varex before identifying outlier components

### LINKED ISSUE #176: Determine rationale for for loop in Step 2 of selcomps (v2.5)
@emdupre and I noticed an odd for loop in the component selection function in emdupre/tedana#16, wherein a copy of the set of accepted components (at that stage) is reduced a total of three times and is then used later in the function. We included a note to look into this later, so I thought I would open an issue for it.

First and foremost, we want to make sure that this isn't a hardcoded reference to the number of echoes. If it isn't, then it would be nice to know what it _is_ supposed to be.

https://github.com/ME-ICA/tedana/blob/1bc32e46445f1d8c5b69ba2000280e2f8f591321/tedana/selection/select_comps.py#L198-L204

tsalo:
Closes #176. While this does not clarify everything about this step (e.g., why it is run three times specifically, or why the outlier components are excluded when computing elbows but not when selecting components later), it does resolve the most pressing issue of explaining why the step exists.

This was originally part of #247, but has been split off because it is self-contained and should be merged separately.

- Sort component table when identifying `ncls` (components presumably without outlier variance explained) by descending variance explained. Before, the component table was sorted by descending Kappa values when identifying outliers. Per discussions with @handwerkerd, this was identified as a bug.


tsalo:
Awesome! I'll wait until we merge #266, handle the merge conflicts, and then merge this in.

tsalo:
Unfortunately the merge conflicts are being terrible, so I'm going to just close this and open a new PR from a fresh branch.

jbteves:
Okay. I think that you should be able to fix them by rebasing this branch on top of master, resolving conflicts using "theirs" and then force-pushing here. But I'm not sure what exactly the conflicts are so I could be wrong.

PR #262, opened 2019-04-20
[REF, DOC] Document and refactor selcomps

tsalo:
Closes None. References #84.

This was originally part of #247, but has been split off because it is self-contained and should be merged separately.

- Improve docstrings
- Eliminate unnecessary ordinal ranking
- Replace “acc” with “unclf” (unclassified)
- Eliminate unnecessary intersections between arrays
    - e.g., candartA, candartB, ign_add0
- Clean up unnecessarily complicated indexing with .loc
- Use mean of ranks instead of sum for decision table scores, which lets us drop the `n_decision_metrics` variable
- Add kappa ratio score to component table


tsalo:
These changes should not affect behavior (all related changes that do are in separate PRs). I think this is ready for review.

tedana/selection/select_comps.py
```diff
@@ -25,9 +25,8 @@ def selcomps(seldict, comptable, mmix, manacc, n_echos):
     seldict : :obj:`dict`
         A dictionary with component-specific features used for classification.
         As output from `fitmodels_direct`
-    comptable : (C x X) :obj:`pandas.DataFrame`
-        Component metric table. One row for each component, with a column for
-        each metric. The index should be the component number.
+    comptable : (C x 5) :obj:`pandas.DataFrame`
+        Component metric table
```
emdupre: Should the docstring explain the shape, still ?

tedana/selection/select_comps.py
```diff
@@ -80,12 +79,12 @@ def selcomps(seldict, comptable, mmix, manacc, n_echos):
...
-    # List of components
-    midk = []
-    ign = []
+    # Lists of components
     all_comps = np.arange(comptable.shape[0])
-    # acc remains a full list that is whittled down over criteria
-    acc = np.arange(comptable.shape[0])
+    # unclf is a full list that is whittled down over criteria
+    # since the default classification is "accepted", at the end of the tree
+    # the remaining elements in unclf match up to the accepted components
```
emdupre:
```suggestion
    # the remaining elements in unclf are classified as accepted
```

tedana/selection/select_comps.py
```diff
@@ -80,12 +79,12 @@ def selcomps(seldict, comptable, mmix, manacc, n_echos):
...
+    unclf = np.arange(comptable.shape[0])
```
emdupre: Why have both `all_comps` and `unclf` defined, here ?

tedana/selection/select_comps.py
```diff
@@ -284,90 +281,97 @@ def selcomps(seldict, comptable, mmix, manacc, n_echos):
...
         ign_add1 = np.intersect1d(
-            acc[comptable.loc[acc, 'kappa'] <= kappa_elbow],
-            acc[comptable.loc[acc, 'variance explained'] > new_varex_lower])
+            unclf[comptable.loc[unclf, 'kappa'] <= kappa_elbow],
+            unclf[comptable.loc[unclf, 'variance explained'] > new_varex_lower])
         ign_add1 = np.setdiff1d(ign_add1, midk)
         comptable.loc[ign_add1, 'classification'] = 'ignored'
         comptable.loc[ign_add1, 'rationale'] += 'I012;'
 
+    # at this point, unclf is equivalent to accepted
```
emdupre: Not sure I understand having this comment here, but happy to defer to you ! Just pointing it out as potentially confusing.

tedana/selection/select_comps.py
tsalo: They are both referenced independently throughout the function, and `unclf` is edited while `all_comps` is not. Is that kind of what you were asking about?

emdupre: Yes, sorry, I thought it was just confusing that they were both actually defined. It would have made more sense to me if it were something like
```python
all_comps = np.arange(comptable.shape[0])
unclf = all_comps.deepcopy()
```

tedana/selection/select_comps.py
tsalo: `unclf` isn't used to define accepted components. All components are initially classified as "accepted" and then the function builds up the ignored and rejected classifications. The remaining list of "unclassified" components ends up being the list of components remaining with the "accepted" classification in the component table. I'm not sure how best to convey that.

tedana/selection/select_comps.py
emdupre: (Realizing "defined" is a really bad term, sorry !)

tedana/selection/select_comps.py
tsalo: I'm not sure how this was changed. Reverting it back now.

tedana/selection/select_comps.py
emdupre: I think I was just confused with the comment being there since `unclf` is not actually called again at this point. I get what you're trying to say, but my first reaction was sort of like "why is it here?" One option would be to do something like `del unclf`, though I'm not sure that actually clarifies anything.

Very light opinion, so feel free to just ignore this point entirely !

tedana/selection/select_comps.py
tsalo: I can do that.

PR #263, opened 2019-04-20
[ENH] Improve manual component selection

tsalo:
Closes None. This was originally part of #247, but has been split off because it is self-contained and should be merged separately.

- Allow users to re-run tedana on the same directory without raising copyfile errors.
- Retain original classification and rationale when manually overriding classification.
- Allow `manacc` to be a list to make it easier to run within Python.
- Clean up logic around checking mixm, ctab, and manacc arguments.

### DIFF
--- a/docs/outputs.rst
+++ b/docs/outputs.rst
@@ -129,24 +129,24 @@ P007   rejected         Rho below fmin (only in stabilized PCA decision tree)
 
 TEDICA codes
 ````````````
-=====  ===============  ========================================================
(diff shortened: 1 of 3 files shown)


tsalo:
I have tested the various combinations of ctab, manacc, and mix locally, and everything seems to work fine. I think this is ready for review.

tsalo:
I think that's a good idea. I just made a commit with that change to the documentation, although the actual change to the manual classification will have to go in #266.

Issue #214, opened 2019-02-09
Drop PAID support from tedana workflow


tsalo:
### Summary
In #210, I assumed that we wouldn't support the PAID combination method in the denoising workflow because I believe it is only valid on smoothed data and we don't want to perform denoising on smoothed or warped data. However, we do currently support PAID, which is why I'd like to discuss the possibility of removing that support. 

### Additional Detail
PAID assumes that "noise is isotropic and homogeneous throughout the image", which is only the case after smoothing. Since we don't want to perform multi-echo denoising (i.e., the `tedana` workflow) on smoothed data, I don't think that PAID should be an option for `tedana_workflow`, even though we should retain it for `t2smap_workflow`.

emdupre:
To expand on this a little more: The idea, then, would be that T2* map generation could be used in any workflow, and in those with smoothed input data, the PAID method would be reasonable to use. But, because the `tedana_workflow` includes decomposition and denoising, we want to make sure that's performed on unsmoothed data. Is that right, @tsalo ?

If so, I think @handwerkerd and @javiergcas will have the most context for this, but I'm inclined to agree with your suggestion.

To confirm, then, concrete steps would be to remove the `ste` flag from [here](https://github.com/ME-ICA/tedana/blob/16cfa720f6202298d7587e72976c6e5e78eb904e/tedana/workflows/tedana.py#L93) (and subsequent workflow steps) but leave it in place [here](https://github.com/ME-ICA/tedana/blob/16cfa720f6202298d7587e72976c6e5e78eb904e/tedana/workflows/t2smap.py#L67) ? We would also need to update the documentation for that option in `t2smap_workflow` to explain its relative utility (and provide a reference, as we do for `t2s`).

PR #264, opened 2019-04-20
[REF, DOC] Document PAID combination method

### LINKED ISSUE #210: Improve documentation of PAID combination method
### Summary
We support the parallel-acquired inhomogeneity desensitized (PAID) ME-fMRI combination method from [Poser et al. (2006)](https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.20900) through `--combmode ste`. Per [Kundu et al. (2017)](https://www.sciencedirect.com/science/article/pii/S1053811917302410):
> The combination of TE signals in the manner of Eq. (3), referred to as parallel-acquired inhomogeneity desensitized (PAID) ME-fMRI, is comparable to that of Eq. (2) [`tedana`'s default combination method] **given noise is isotropic and homogeneous throughout the image**. However, given the high levels of artifact typical in most fMRI data, which adds to variance and thus reduces signal-to-noise ratio, the assumption of isotropic noise may not [be] suitable. In other words, depending on data quality and method of estimation, SNR may not reflect BOLD sensitivity and thus may not lead to optimal BOLD contrast. This condition makes the approach in Eq. (2) an appropriate choice if a T2* map is available. **Given the performance of modern computers and numerical computation software, computing per-voxel estimates of T2\* and T2\*-based weights is not prohibitive after acquiring ME-fMRI data.**

@emdupre and I discussed this a bit, and we interpret the above as indicating that one should only use the PAID method on smoothed data. Additionally, I believe that our choice of terminology (i.e., referring to PAID combination as `ste`, is somewhat confusing, especially since we often code the sourceTEs argument as `ste` as well.

### Next Steps
To resolve this, we should update the documentation in the following places to include the above caveats regarding combination method (i.e., not preferred and only on smoothed data):
- The argument documentation for the `tedana` and `t2smap` CLI.
- The docstrings for the `tedana` and `t2smap` workflow functions.
- The docstrings for any relevant functions in `combine.py`.

Pending #214, we should also add a note box to the [Processing pipeline details page](https://tedana.readthedocs.io/en/latest/approach.html#optimal-combination) to say that we support PAID in `t2smap` (a workflow for optimal combination), but not in `tedana` (i.e., denoising), and why that is (i.e., because PAID is only appropriate for smoothed data and denoising shouldn't be applied to smoothed data).

We should also change how we refer to the PAID combination method as `paid`, instead of `ste`, in order to reduce confusion.

tsalo:
Closes #210, #214. This was originally part of #247, but has been split off because it is self-contained and should be merged separately.

- Rename `combmode` value “ste” to “paid”
- Rename variable `ste` to `source_tes` for clarity
- Drop PAID support from tedana (because PAID requires smoothed data and multi-echo denoising does not work with smoothed data) but keep it in t2smap workflow
- Add note box about PAID to documentation

### DIFF
--- a/docs/approach.rst
+++ b/docs/approach.rst
@@ -142,7 +142,7 @@ of the other echoes (which it is).
 
 .. note::
     An alternative method for optimal combination that
-    does not use :math:`T_{2}^*`, known as the parallel-acquired inhomogeneity
+    does not use :math:`T_{2}^*`, is the parallel-acquired inhomogeneity
     desensitized (PAID) ME-fMRI combination method (`Poser et al., 2006`_).
     This method is not supported in the tedana workflow, however, because
     it assumes that noise is "isotropic and homogeneous throughout the image",
--- a/docs/approach.rst
+++ b/docs/approach.rst
@@ -140,6 +140,15 @@ of the other echoes (which it is).
 .. image:: /_static/10_optimal_combination_timeseries.png
   :align: center
 
+.. note::
+    The tedana package also supports another method for optimal combination that
+    does not use :math:`T_{2}^*`, known as the parallel-acquired inhomogeneity
+    desensitized (PAID) ME-fMRI combination method (`Poser et al., 2006`_).
+    This method is not supported in the tedana workflow, however, because
+    it assumes that noise is "isotropic and homogeneous throughout the image",
+    and so can only be run on smoothed data. This is not compatible with the
+    multi-echo denoising procedure, but can be used in the t2smap workflow.
+
 TEDPCA
 ``````
 The next step is to identify and temporarily remove Gaussian (thermal) noise
@@ -223,3 +232,4 @@ robust PCA.
 
 .. _nilearn.masking.compute_epi_mask: https://nilearn.github.io/modules/generated/nilearn.masking.compute_epi_mask.html
 .. _Power et al. (2018): http://www.pnas.org/content/early/2018/02/07/1720985115.short
+.. _Poser et al., 2006: https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.20900
--- a/docs/approach.rst
+++ b/docs/approach.rst
@@ -141,7 +141,7 @@ of the other echoes (which it is).
   :align: center
 
 .. note::
-    The tedana package also supports another method for optimal combination that
+    An alternative method for optimal combination that
     does not use :math:`T_{2}^*`, known as the parallel-acquired inhomogeneity
     desensitized (PAID) ME-fMRI combination method (`Poser et al., 2006`_).
     This method is not supported in the tedana workflow, however, because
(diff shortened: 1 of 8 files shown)


docs/approach.rst
```diff
@@ -140,6 +140,15 @@ of the other echoes (which it is).
 .. image:: /_static/10_optimal_combination_timeseries.png
   :align: center
 
+.. note::
+    The tedana package also supports another method for optimal combination that
```
emdupre:
```suggestion
    An alternative method for optimal combination that
```

docs/approach.rst
```diff
@@ -140,6 +140,15 @@ of the other echoes (which it is).
...
+    does not use :math:`T_{2}^*`, known as the parallel-acquired inhomogeneity
```
emdupre:
```suggestion
    does not use :math:`T_{2}^*`, is the parallel-acquired inhomogeneity
```

docs/approach.rst
```diff
@@ -140,6 +140,15 @@ of the other echoes (which it is).
...
+    desensitized (PAID) ME-fMRI combination method (`Poser et al., 2006`_).
+    This method is not supported in the tedana workflow, however, because
```
emdupre: Since the t2smap workflow is called as part of the tedana workflow, this is a little confusing, to me.

docs/approach.rst
```diff
@@ -140,6 +140,15 @@ of the other echoes (which it is).
 .. image:: /_static/10_optimal_combination_timeseries.png
   :align: center
 
+.. note::
```
emdupre: Can we have each sentence on a newline ? Sorry !

docs/approach.rst
tsalo: We could hypothetically allow users to run tedana without denoising (e.g., `--disable-denoising`?), which would be the same as the t2smap workflow, but that would mean managing how the combmode and denoising arguments interact (i.e., using paid and denoising should raise a warning and switch paid to t2s automatically).

docs/approach.rst
tsalo: Done.

docs/approach.rst
emdupre: Right, sorry, I wasn't thinking quite like that ! I just meant that it is technically in the tedana workflow, but it's not accessible, you're right.

Maybe we could just say something like 
```
This method specifically assumes that noise in the acquired echoes is "isotopic and 
homogeneous throughout the image," meaning it should be used on smoothed data. 
As we do not recommend performing tedana denoising  on smoothed data, 
we discourage using PAID within the tedana workflow.
We do, however, make it accessible as an alternative combination method 
in the t2smap workflow.
```

tsalo:
Not sure how I "dismissed" your review, but I just added a commit with the requested change. I also think using [skip ci] might have made a mess of things somehow?

KirstieJane:
@tsalo @emdupre - is the protected branch setup to dismiss reviews when there are new commits? That might explain @tsalo question above (https://github.com/ME-ICA/tedana/pull/264#issuecomment-485607201)!

emdupre:
Yes, sorry, that's how it's set up ! Happy to think about whether that makes sense, but I went ahead and re-approved, for now :smile_cat:

Issue #166, opened 2018-12-06
Reorganize tedpca to better mirror selcomps


emdupre:
Since `tedpca` is now an optional decision tree, it would make sense to move to the `selection` module with `selcomps`. However, putting those two together reveals a few inconsistencies that should be smoothed out.

 As an example: why do we use an `eimask`-generated mask when computing Kappa and Rho values following `tedpca`, but not for `tedica` ?

tsalo:
As part of this, I think the following would be good ideas:
- Move `kundu_tedpca` into `tedana.selection`.
- Fit models to PCA components and select components in `tedana_workflow`, instead of `tedpca`.
- Remove `pcastate` dictionary and pickle file. I believe that all of the relevant information is available between the mixing matrix (saved to `mepca_mix.1D`), the component maps (saved to `mepca_OC_components.nii`), and the component table. The only thing not currently tracked is the PCA variance explained, since the component table contains a "variance explained" estimated in `fitmodels_direct`. The two are definitely not the same.

jbteves:
@tsalo @emdupre Could I update the top-level comment to reflect @tsalo's list above?

PR #265, opened 2019-04-21
[REF] Split eigendecomp into ICA and PCA files

tsalo:
Closes None. References #166. This was originally part of #247, but has been split off because it is self-contained and should be merged separately.

- Split eigendecomp into ICA and PCA files.
- Move kundu_tedpca into selection module.
- Add out_dir argument to tedpca.
- Remove pcastate file. PCA doesn't take very long.
- Transpose PCA mixing matrix internally so it's in the same format as the ICA mixing matrix.


emdupre:
Sorry, why did we want to put ICA and PCA in their own modules ?

tsalo:
The other changes are perhaps more important (e.g., moving kundu_tedpca into the selection module and removing the pcastate file), but my thinking is that separating ICA and PCA will work better when we add new decomposition methods (one of our future goals). Plus, if we want to support low-memory incremental PCA, then that may merit an additional function within pca.py (like run_mlepca). If we end up expanding the ICA- and PCA-related functions, then it makes sense to group them within their own submodules.

tedana/decomposition/ica.py
```diff
@@ -0,0 +1,76 @@
+"""
+Signal decomposition methods for tedana
```
emdupre:
```suggestion
ICA and related signal decomposition methods for tedana
```

tedana/decomposition/ica.py
```diff
@@ -0,0 +1,76 @@
...
+    maxit : :obj:`int`, optional
+        Maximum number of iterations for ICA. Default is 500.
+    maxrestart : :obj:`int`, optional
+        Maximum number of attempted decompositions to perform with different
+        random seeds. ICA will stop running if there is convergence prior to
+        reaching this limit. Default is 10.
+
+    Returns
+    -------
+    mmix : (T x C) :obj:`numpy.ndarray`
```
emdupre: Should the docstring include that this is z-scored ?

tedana/decomposition/pca.py
```diff
@@ -0,0 +1,242 @@
+"""
+Signal decomposition methods for tedana
```
emdupre:
```suggestion
PCA and related signal decomposition methods for tedana
```

tedana/decomposition/pca.py
```diff
@@ -0,0 +1,242 @@
...
+      booktitle={Advances in neural information processing systems},
+      pages={598--604},
+      year={2001}
+    }
+    """),
+    description='Introduces method for choosing PCA dimensionality '
+                'automatically')
+def run_mlepca(data):
+    """
+    Run Singular Value Decomposition (SVD) on input data.
```
emdupre:
```suggestion
    Run Singular Value Decomposition (SVD) on input data,
    automatically select components on MLE variance cut-off.
```

tedana/decomposition/pca.py
```diff
@@ -0,0 +1,242 @@
...
+        Dimensionality augmentation weight for Rho calculations. Must be a
+        non-negative float, or -1 (a special value). Default is 1.
+    out_dir : :obj:`str`, optional
+        Output directory.
+    verbose : :obj:`bool`, optional
+        Whether to output files from fitmodels_direct or not. Default: False
+
+    Returns
+    -------
+    n_components : :obj:`int`
```
emdupre: Can we flip these to match the output ordering ?

PR #266, opened 2019-04-21
[REF] Reorganize selcomps and fitmodels_direct

### LINKED ISSUE #16: Break `selcomps` into functions
_From @emdupre on November 15, 2017 16:31_

Currently, the function for selecting components contains (at least) 7 steps. It would be better to demarcate these into individual functions for interpretability / testing purposes. A good starting point for structuring these would be the code comments in `selcomps` itself.

_Copied from original issue: emdupre/tedana#4_

tsalo:
Closes #16. This was originally part of #247, but has been split off because it is self-contained and should be merged separately.

- Rename fitmodels_direct to dependence_metrics
- Split sel_comps into manual_selection, kundu_selection_v2, and kundu_metrics
- Move kundu_metrics into model.fit
- Rename selcomps to tedica
- Rename PCA-derived variance explained column in PCA component table. This will no longer overwrite the one from fitmodels_direct
- Add reorder_dataframe function to selection._utils
- Update website documentation to match API changes
- Add relevant tests


docs/api.rst
```diff
@@ -20,22 +20,37 @@ API
 .. _calibration_ref:
 
 
-:mod:`tedana.model`: Modeling TE-dependence
---------------------------------------------------
+:mod:`tedana.decay`: Modeling signal decay across echoes
```
emdupre: I think we should keep the name `model`, here, as it's relatively agnostic to the kind of TE-dependence we're modelling. For example, for thermal noise we'd like to confirm that it stays stable across all TEs -- I'm not sure it makes sense to call that `decay`.

docs/api.rst
```diff
@@ -65,67 +80,65 @@ API
 .. _calibration_ref:
 
 
-:mod:`tedana.combine`: Combine time series
---------------------------------------------------
+:mod:`tedana.model`: Computing TE-dependence metrics
```
emdupre:
```suggestion
:mod:`tedana.metrics`: Computing TE-dependence metrics
```

docs/api.rst
```diff
@@ -20,22 +20,37 @@ API
...
+   tedana.decay.fit_decay_ts
 
-   :template: module.rst
+.. currentmodule:: tedana
+
+.. _calibration_ref:
 
-   tedana.model.fit
+
+:mod:`tedana.combine`: Combining time series across echoes
```
emdupre: Here you have `combine`, but below you've replaced it with `model`. We should harmonize.

docs/api.rst
```diff
@@ -65,67 +80,65 @@ API
...
    :template: function.rst
 
-   tedana.combine.make_optcom
+   tedana.model.dependence_metrics
+   tedana.model.kundu_metrics
 
    :template: module.rst
 
-   tedana.combine
+   tedana.model.fit
```
emdupre: I'm not sure that we want to call this a `.fit`, since it could be easily confused with the sklearn convention.

tedana/decomposition/pca.py
```diff
@@ -200,10 +200,11 @@ def tedpca(data_cat, data_oc, combmode, mask, t2s, t2sG,
 
     # Normalize each component's time series
     vTmixN = stats.zscore(comp_ts, axis=0)
-    _, comptable, _, _ = model.fitmodels_direct(
-                data_cat, comp_ts, eimum, t2s, t2sG, tes, combmode, ref_img,
-                reindex=False, mmixN=vTmixN, full_sel=False,
+    comptable, _, _, _ = model.dependence_metrics(
+                data_cat, data_oc, comp_ts, eimum, t2s, tes, ref_img,
+                reindex=False, mmixN=vTmixN, method=None,
```
emdupre: `method` is a little too ambiguous of an argument, here.

tedana/model/fit.py
```diff
@@ -8,90 +8,87 @@
...
-        Default: False
-    mmixN : array_like, optional
-        Default: None
-    full_sel : bool, optional
-        Whether to perform selection of components based on Rho/Kappa scores.
-        Default: True
+        Whether to sort components in descending order by Kappa. Default: False
+    mmixN : (T x C) array_like, optional
+        Z-scored mixing matrix. Default: None
+    method : {'kundu_v2', 'kundu_v3', None}, optional
```
emdupre: Maybe we could call this `decision` ?

emdupre: Does `None` make sense as the default argument ? What behavior do we expect when we return an empty `seldict` ?

emdupre: I'm also still really unsure that we should have `kundu_v3` here, since it's not reintegrated, and we don't know when it will be.

tedana/tests/test_model_fit_dependence_metrics.py
```diff
@@ -0,0 +1,92 @@
...
+Tests for tedana.model.fit
+"""
+
+import numpy as np
+import pytest
+
+from tedana.model import fit
+
+
+def test_break_dependence_metrics():
```
emdupre: :heart_eyes:

docs/api.rst
tsalo: The reason I put model below decay is that decay comes first in the workflow, so I think there's a sort of flow there, but I didn't change the names of the modules.

docs/api.rst
tsalo: I just reordered them based on when they occur in the workflow. It's possible that I missed some changes though, so I apologize if that's adding to the confusion.

docs/api.rst
tsalo: I've never liked `fit` either. We can change `tedana.model` to `tedana.metrics` (as you suggested above), but we'll still need a file below that in order to keep model/metrics as a subfolder-organized submodule. I fully expect us to expand our sets of metrics (e.g., with AROMA-like motion- and artifact-based ones), which is why I don't think we should just put it in one top-level file (i.e., `tedana/metrics.py`), but that still leaves us with the issue of what to name the file.

docs/api.rst
emdupre: Ah, ok, sorry ! Wasn't clear from the diff.

docs/api.rst
emdupre: Since we're not quite there yet (e.g., adding motion based metrics), though, I think we should just keep it in a top-level file for now -- we can expand when we add more. WDYT ?

tedana/model/fit.py
tsalo: `None` is the PCA decision tree-compatible version, which doesn't return metric maps and only gives back the component table.

`kundu_v3` requires certain metric maps in `seldict` (e.g., `PSC`, `WTS`, and `tsoc_B`) that `kundu_v2` does not. I didn't want to remove those metric maps from the overall function, because it would be annoying to re-implement them later, but also didn't want to output an oversized version of `seldict` when it's not necessary. I can drop it from the arguments and docstring, but I'd like to keep it in the if/else statement below.

tsalo: I don't think `decision` is right, since that implies that actual decisions are made in this function. It really just determines which outputs will be generated. PCA requires just the component table, while Kundu ICA v2 requires that and several metric maps, and Kundu ICA v3 requires the component table and even more metric maps.

What about `algorithm`?

docs/api.rst
tsalo: Also, #273 moves the two remaining functions (`get_coeffs` and `computefeats2`) in `model.fit` into a new stats module, so the only things in `model.fit` will be the metrics (`dependence_metrics` and `kundu_metrics`).

tsalo: I can move it there then. So, `tedana/model/fit.py` will become `tedana/model.py` and the `tedana/model/` folder will be removed. Does that sound good?

tedana/selection/_utils.py
```diff
@@ -8,6 +8,17 @@
 LGR = logging.getLogger(__name__)
 
 
+def reorder_dataframe(comptable):
+    """
+    Reorder columns in component table and clean up rationale column.
```
jbteves: Could you elaborate slightly more for this docstring? It looks from the code that you're keeping only the classification and rationale columns, while stripping out semicolons from the rationale, same as the original larger module. It may seem straightforward but I think it's still helpful to note this in the docstring. It's true that this level of comment detail was absent in the original, but with this as an independent function I think it would be very useful.

tedana/selection/_utils.py
```diff
@@ -8,6 +8,17 @@
 LGR = logging.getLogger(__name__)
 
 
+def reorder_dataframe(comptable):
```
jbteves: This may be pedantic, and therefore I don't hold this opinion strongly, but because we're also stripping out the semicolons here, might `clean_dataframe` be a more appropriate name?

tedana/selection/tedica.py
```diff
@@ -260,15 +209,14 @@ def selcomps(seldict, comptable, mmix, manacc, n_echos):
                     (comptable.loc[ncls, 'rho'] < rho_elbow)]
 
     if len(acc_prov) == 0:
-        LGR.warning('No BOLD-like components detected')
+        LGR.warning('No BOLD-like components detected. Ignoring all remaining '
```
jbteves: I think this is a very good tweak to the logger warning; inexperienced people may be confused when they see "ignored" components in the absence of it after a failure to detect BOLD-like components (I was, anyway). Nice job.

tedana/tests/test_model_fit_dependence_metrics.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    Ensure that dependence_metrics fails when input data do not have the right
+    shapes.
```
jbteves: I'm not sure if we have formatting guidelines for tests; I don't recall seeing any in the linked NumPy style guide either, so I could be off base. 
Could we put brief comments indicating what these tests are checking for? The formatting makes it somewhat difficult to me (and it might just be me) to tell what is actually being tested at each assert.

tedana/tests/test_model_kundu_metrics.py
```diff
@@ -0,0 +1,37 @@
...
+import pytest
+import numpy as np
+import pandas as pd
+
+from tedana.model import fit
+
+
+def test_smoke_kundu_metrics():
+    """
+    Smoke test for kundu metrics function
```
jbteves: To be clear, this is just to make sure that nothing is so wrong as to be on fire, not a test specific to the multi-echo or imaging domain, right? If that's the case I have a slight objection to calling it smoke test, in that at some point we would really like the function to just be a full test, so that later we'll have to rename it as it's filled out.

jbteves: Additionally, could we have a one-sentence description of what the test is doing?

tedana/tests/test_selection.py
```diff
@@ -0,0 +1,31 @@
...
+import os.path as op
+
+import pytest
+import numpy as np
+import pandas as pd
+
+from tedana import selection
+
+
+def test_manual_selection():
```
jbteves: It can be boilerplate, but can we get a docstring for this function?

tedana/workflows/tedana.py
```diff
@@ -389,27 +389,27 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
 
-        comptable = selection.selcomps(seldict, comptable, mmix, manacc, n_echos)
-    elif ctab is not None and manacc is not None:
-        LGR.info('Using supplied ICA mixing matrix, component table, and '
-                 'accepted components')
-        mmix = np.loadtxt(op.join(out_dir, 'meica_mix.1D'))
-        comptable = pd.read_csv(ctab, sep='\t', index_col='component')
-        comptable = selection.selcomps({}, comptable, mmix, manacc, n_echos)
+        comptable = model.kundu_metrics(comptable, metric_maps)
+        comptable = selection.kundu_selection_v2(comptable, n_echos, n_vols)
```
jbteves: This is SO MUCH MORE LEGIBLE, awesome!

tedana/tests/test_model_kundu_metrics.py
tsalo: We won't rename this test- we'll either delete it or add a separate one for meaningful testing.

jbteves: Sorry, didn't realize that was the plan. That's cool, then, and most of my other comments don't apply.

tsalo:
@jbteves Thanks! Would you mind changing your review status?

jbteves:
@tsalo done! Awesome.

tsalo:
Yay, merging now.

PR #273, opened 2019-04-24
[REF] Create new stats module

### LINKED ISSUE #249: Move statistical functions into new stats module
### Summary
We have a couple of statistical functions that I think should be moved into their own module. Currently, most of these functions are used in multiple modules, but are kept in `module._utils` submodules, which doesn't make a lot of sense.

### Additional Detail
The functions that can be moved into `stats` are:
1. `model.fit.get_coeffs`: Uses `numpy.linalg.lstsq` to compute parameter estimates. Used in `model.fit.fitmodels_direct`, `io.split_ts`, `io.write_split_ts`, `io.writeresults`, and `viz.write_comp_figs`.
2. `model.fit. computefeats2`: Uses `get_coeffs` to compute parameter estimates and pseudo-z-statistics. Used in `model.fit.fitmodels_direct`, `decomposition.eigendecomp.tedpca`, and `io.writefeats`.
3. `utils.getfbounds`: Just gets the F-statistic for three p-values (0.05, 0.025, 0.01) based on degrees of freedom (from n_echos). Used in `selection.select_comps.selcomps`, `eigendecomp.decomposition.kundu_tedpca`, and `model.fit.fitmodels_direct`.

### Next Steps
1. Create stats file (`stats.py`).
2. Move functions from their current positions to new file.
3. Update imports throughout package.
4. Update tests.
5. Update API organization in docs.

tsalo:
Closes #249.

- `tedana.model.fit.get_coeffs` --> `tedana.stats.get_coeffs`
- `tedana.model.fit.computefeats2` --> `tedana.stats.computefeats2`
- `tedana.utils.getfbounds` --> `tedana.stats.getfbounds`
- Update corresponding tests and API documentation.


tedana/stats.py
```diff
@@ -0,0 +1,149 @@
...
+        raise ValueError('Second dimensions (number of volumes) of data ({0}) '
+                         'and mmix ({1}) do not match.'.format(data.shape[0],
+                                                               mmix.shape[0]))
+
+    # demean masked data
+    data_vn = stats.zscore(data[mask], axis=-1)
+
+    # get betas of `data`~`mmix` and limit to range [-0.999, 0.999]
+    data_R = get_coeffs(data_vn, mmix, mask=None)
+    data_R[data_R < -0.999] = -0.999
```
jbteves: This is somewhat pedantic, but here is an example of a "mysterious number." The comment above helps, but if for some reason we were to change those values, we would have to manually catch them and not make typos. Review helps for that, but I am biased towards less work for reviewers. ; - )

tedana/tests/test_stats_getfbounds.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+
+import numpy as np
+import pytest
+
+from tedana.stats import getfbounds
+
+
+def test_getfbounds():
+    good_inputs = range(1, 12)
```
jbteves: In the absence of a comment or named number, it's not clear why the range is 1,12, as another example of a mysterious number.

tedana/stats.py
tsalo: I have concerns about the calculation of the beta-values and their conversion to z-values as implemented here (see #178 and #223). My hope is that we will fix those issues (and thus eliminate the need for this weird cropping), which will also involve removing these hard-coded values.

tedana/tests/test_stats_getfbounds.py
tsalo: I believe that those values correspond to reasonable numbers of echoes. I can add a comment to clarify.

tedana/stats.py
jbteves: Oh, I hope I was clear, I don't think that this particular case should stop us from merging, and I know there are discussions about it, but I wanted to give a more concrete example for my comment. In fact, I DON'T think this should be adjusted in the PR because that's independent to the refactor.

tedana/tests/test_stats_getfbounds.py
jbteves: That would be fantastic, thank you!

jbteves: That said, I don't think it has to get fixed before merging this because that was already hardcoded; you haven't actually changed the test itself. Sorry if that wasn't clear!

PR #276, opened 2019-04-26
Add test for breaking tedpca

souravsingh:
The PR adds a test to break `tedpca` function to check if the shape being returned is correct. More tests need to be added to check for different parameters.

@KirstieJane @tsalo


souravsingh:
@tsalo Can yu reivew the PR?

tsalo:
It doesn’t look like this PR is ready for review. The test is not passing, and the WIP tag indicates that it is a work-in-progress. I can help you figure out why the test is not passing, if you’d like.

souravsingh:
It looks like the error is occuring due to wrong size for one of the parameters. Maybe a wrong size was passed to `data_oc`?

souravsingh:
Closing the PR since no support is being given

PR #280, opened 2019-05-02
[ENH] Adds static logging filename

### LINKED ISSUE #169: Update tests to include --debug
In #167 we added log files that are written out if `--debug` is called; however, we are not yet testing that functionality since the exact log filename will depend on the time the job is executed. It would be great to update our test to call `--debug`, and to check for a logfile matching the general pattern !

jbteves:
Closes #169 

- Makes log name tedana_run so that it is static
- Detects name collisions and appends _old until there is no collision

### DIFF
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -10,6 +10,7 @@ os.environ['VECLIB_MAXIMUM_THREADS'] = '1'
 os.environ['OPENBLAS_NUM_THREADS'] = '1'
 
 import os.path as op
+import glob
 import shutil
 import logging
 
@@ -264,12 +265,16 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
 
     if debug and not quiet:
         # ensure old logs aren't over-written
-        logname = op.join(
-                out_dir,
-                'tedana_run')
-
-        while op.isfile(logname):
-            logname += '_old'
+        basename = 'tedana_run'
+        extension = 'txt'
+        logname = op.join(out_dir,(basename + '.' + extension))
+        logex = op.join(out_dir,(basename + '*'))
+        previouslogs = glob.glob(logex)
+        previouslogs.sort(reverse=True)
+        for f in previouslogs:
+            previousparts = op.splitext(f)
+            newname = previousparts[0] + '_old' + previousparts[1]
+            os.rename(f,newname)
 
         # set logging format
         formatter = logging.Formatter(
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -267,14 +267,14 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
         # ensure old logs aren't over-written
         basename = 'tedana_run'
         extension = 'txt'
-        logname = op.join(out_dir,(basename + '.' + extension))
-        logex = op.join(out_dir,(basename + '*'))
+        logname = op.join(out_dir, (basename + '.' + extension))
+        logex = op.join(out_dir, (basename + '*'))
         previouslogs = glob.glob(logex)
         previouslogs.sort(reverse=True)
         for f in previouslogs:
             previousparts = op.splitext(f)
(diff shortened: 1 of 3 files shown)


jbteves:
I don't have strong opinions about the filename but I realized that this is a very useful feature as we start building out more tests. Please let me know if you'd like me to modify it.

tedana/workflows/tedana.py
```diff
@@ -264,12 +263,21 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
         os.mkdir(out_dir)
 
     if debug and not quiet:
+        # ensure old logs aren't over-written
+        logname = op.join(
+                out_dir,
+                'tedana_run')
+
+        while op.isfile(logname):
+            logname += '_old'
```
tsalo: This is clever, but I think it will not do what we want. The newest log file will end up being `tedana_run_old_old_old.txt` (or whatever). It would take a bit more effort, but I think a preferable approach would be to identify all of the run files and to rename each existing file with an additional `_old` in the name. So
`tedana_run.txt` --> `tedana_run_old.txt`
`tedana_run_old.txt` --> `tedana_run_old_old.txt`
and the new file is `tedana_run.txt`

That said, I don't know if it's worth the effort, since the naming for the log file will almost certainly change to match the BIDS Derivatives specification. I have no clue what that file will end up looking like.

I would recommend trying to see if you can implement the above fairly easily. If you can't, then maybe only retain two (`tedana_run.txt` as the new one and `tedana_run_old.txt` for any existing one).

tsalo:
Thank you for doing this. I wish there was some way for us to work around the date-timed file naming convention, but I get why it's not possible (especially once we shift over to BIDS Derivatives). Could you also update one of the CircleCI tests to use `debug` as well (and the corresponding outputs.txt file)?

jbteves:
Hm, unfortunately I don't know how to use CircleCI that well. I also suspect there will be a problem because there are timestamps inside the logfile itsef; I wonder, what would you think of separating the timestamps on separate lines? Then we could only compare even lines, ignoring odd lines, and make it really straightforward to compare outputs.

tedana/workflows/tedana.py
jbteves: Well... it's hacky, but I could change old->new... I think if people are actually going to re-run tedana on the same folder over and over we may want something generally more sophisticated, but I figured that this pass would at least make it possible to work on integrating `--debug` to CircleCI

tsalo:
The contents of the files won't be a problem, since we're currently just checking that the files exist. As far as CircleCI goes, all you need to do is add `--debug` to the tedana call either [here](https://github.com/ME-ICA/tedana/blob/a11275ddadb9276d36339b0d296d521f5fd6c654/.circleci/config.yml#L143-L145) (for the three-echo test dataset) or [here](https://github.com/ME-ICA/tedana/blob/a11275ddadb9276d36339b0d296d521f5fd6c654/.circleci/config.yml#L201-L203) (for the five-echo test dataset). The corresponding outputs files are [here](https://github.com/ME-ICA/tedana/blob/a11275ddadb9276d36339b0d296d521f5fd6c654/.circleci/tedana_outputs.txt) for the three-echo and [here](https://github.com/ME-ICA/tedana/blob/a11275ddadb9276d36339b0d296d521f5fd6c654/.circleci/tedana_outputs_verbose.txt) for the five-echo.

tedana/workflows/tedana.py
jbteves: After further reflection, it's not that bad to incorporate this suggestion. Let me fiddle and test before pushing a corrective commit.

jbteves:
Okay, I added more LOC so it will also be easier later to modify how it operates when the BIDS spec is made. Probably needs closer review since it's more complicated.

tedana/workflows/tedana.py
jbteves: This should be fixed in 1b3c52c

jbteves:
@tsalo @emdupre just a quick re-request for review. In particular @tsalo I think I addressed your original concern with the latest commit.

PR #281, opened 2019-05-02
[DOC] Fixes Random Seed Help Text

### LINKED ISSUE #193: Random Seed default does not match help text
Currently, we see from the [help text](https://github.com/ME-ICA/tedana/blob/0b4e0e3a13644a1fc14b26c2f72b8066cebb3e7f/tedana/workflows/tedana.py#L196-L199) that the default seed is -1, however it looks like the default value if none is passed is [actually 42](https://github.com/ME-ICA/tedana/blob/0b4e0e3a13644a1fc14b26c2f72b8066cebb3e7f/tedana/workflows/tedana.py#L126-L132).

jbteves:
Closes #193 .

- Fixes random seed help text

### DIFF
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -138,9 +138,10 @@ def _get_parser():
     optional.add_argument('--seed',
                           dest='fixed_seed',
                           type=int,
-                          help=('Value passed to repr(mdp.numx_rand.seed()) '
-                                'Set to an integer value for reproducible ICA results; '
-                                'otherwise, set to -1 for varying results across calls.'),
+                          help=('Value passed to repr(mdp.numx_rand.seed()). '
+                                'Set to an integer value for reproducible ICA results. '
+                                'Set to -1 for varying results across ICA calls. '
+                                'Default=42.'),
                           default=42)
     optional.add_argument('--png',
                           dest='png',


jbteves:
I know we like to leave good first issues open, but that one's been open for a while at this point. Can close if maintainers disagree.

tedana/workflows/tedana.py
```diff
@@ -140,7 +140,7 @@ def _get_parser():
                           type=int,
                           help=('Value passed to repr(mdp.numx_rand.seed()) '
                                 'Set to an integer value for reproducible ICA results; '
-                                'otherwise, set to -1 for varying results across calls.'),
+                                'otherwise, set to 42 for varying results across calls.'),
```
emdupre: I think we actually want to say something like this

```suggestion
                                'otherwise, set to -1 for varying results across calls. '
                                'Default value is 42.'),
```

emdupre:
Thanks for addressing this ! One requested change and then I think this is good to merge :+1:

jbteves:
Hm, I decided to read through the other help texts to make sure it conforms with them; most others don't list a default. It shouldn't go on this PR, but should I open an issue to go ahead and add defaults to the other options as well?

emdupre:
I don't think we strongly need to list our defaults in the help text, since folks can usually access them from the function signature. But this is a bit of a special case, in my mind, since we've (already) confused folks as to what seed is generally used !

Happy to be argued that all arguments need defaults, but agree that's for another PR :smile_cat:

tedana/workflows/tedana.py
```diff
@@ -140,7 +140,7 @@ def _get_parser():
                           type=int,
                           help=('Value passed to repr(mdp.numx_rand.seed()) '
                                 'Set to an integer value for reproducible ICA results; '
-                                'otherwise, set to -1 for varying results across calls.'),
```
emdupre: So sorry, I meant that the default should be given in addition to this note about the special meaning of -1. I think we need both pieces of information, here !

jbteves: Oh, duh, that makes sense. Let me make Travis work some more.

PR #282, opened 2019-05-03
Update multi-echo.rst

handwerkerd:
Progress on #83 .

- Adds several paragraphs on acquisition recommendations to the documentation
- The part of #83 that is still missing is filling out the "Sequences" section of the documentation to say the status/location of multi-echo sequences for each of the main vendors.
- I'm hesitant to add specific scanning parameters here because I don't want to recommend anything overly-specific that users might thoughtlessly copy and use, unless those specific parameter sets have been validated in some way (or at least could be linked to specific publications)

### DIFF
--- a/docs/multi-echo.rst
+++ b/docs/multi-echo.rst
@@ -48,6 +48,16 @@ Sequences
 *********
 * Multi-echo sequences: who has them and how to get them.
 
+Acquisition Parameter Recommendations
+************************************
+There is no empirically tested best parameter set for multi-echo acquisition. The guidelines for optimizing parameters are similar to single-echo fMRI. Parameter choices require balancing the needs of specific studies with regards to spatial resolution and coverage, sampling rate, signal-to-noise ratio, signal drop-out, distortion, and artifacts. Optimize these parameters based on whatever a study's priorities are. The compromise for transitioning to multi-echo is either losing a few slices, increasing the TR, or increasing within-slice or multi-slice acceleration. A rough rule of thumb is that collecting 3 echoes with a similar parameter set to single-echo will involve a loss of 10% of slices (36 instead of 40 slices), a 10% increase in TR (2.2sec instead of 2s) or an increase in acceleration. If one increases acceleration to avoid compromises in spatial coverage or time, it is worth doing an empirical comparison to make sure there isn't a non-trivial loss in SNR or an increase of artifacts.
+
+A minimum of 3 echoes is recommended for running TE-dependent denoising. More echoes may be useful, because that would allow for more accurate estimates of BOLD and non-BOLD weighted fluctuations, but more echoes would require additional compromises in spatiotemporal coverage or more acceleration. Whether the benefits of more echoes balance out the additional acquisition compromises is an open research question. While there are successful studies that don’t follow this rule, it may be useful to have at least one echo that is earlier and one echo that is later than TE one would use for single-echo T2* weighted fMRI. 
+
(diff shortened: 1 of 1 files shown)


docs/multi-echo.rst
```diff
@@ -48,6 +48,16 @@ Sequences
 *********
 * Multi-echo sequences: who has them and how to get them.
 
+Acquisition Parameter Recommendations
+************************************
+There is no empirically tested best parameter set for multi-echo acquisition. The guidelines for optimizing parameters are similar to single-echo fMRI. Parameter choices require balancing the needs of specific studies with regards to spatial resolution and coverage, sampling rate, signal-to-noise ratio, signal drop-out, distortion, and artifacts. Optimize these parameters based on whatever a study's priorities are. The compromise for transitioning to multi-echo is either losing a few slices, increasing the TR, or increasing within-slice or multi-slice acceleration. A rough rule of thumb is that collecting 3 echoes with a similar parameter set to single-echo will involve a loss of 10% of slices (36 instead of 40 slices), a 10% increase in TR (2.2sec instead of 2s) or an increase in acceleration. If one increases acceleration to avoid compromises in spatial coverage or time, it is worth doing an empirical comparison to make sure there isn't a non-trivial loss in SNR or an increase of artifacts.
```
emdupre: Should we say _why_ there is this compromise ? Or point out to a reference ?

docs/multi-echo.rst
```diff
@@ -48,6 +48,16 @@ Sequences
...
+
+A minimum of 3 echoes is recommended for running TE-dependent denoising. More echoes may be useful, because that would allow for more accurate estimates of BOLD and non-BOLD weighted fluctuations, but more echoes would require additional compromises in spatiotemporal coverage or more acceleration. Whether the benefits of more echoes balance out the additional acquisition compromises is an open research question. While there are successful studies that don’t follow this rule, it may be useful to have at least one echo that is earlier and one echo that is later than TE one would use for single-echo T2* weighted fMRI.
```
emdupre: Should we say the general recommendation for a single-echo T2* weighted fMRI is a TE of about 30ms ?

handwerkerd:
I re-wrote the text to address the above comments. I've added more line breaks. I added a more detailed explanation of the cost of multi-echo and how that relates to other parameter choices. As part of the explanation, I included an example TE of 30s. I don't want to mention 30ms out of an example because that is a 3T specific parameter and I didn't want to hunt for a good reference on the TE to T2* relationship across field strengths and tissue types right now. The edited version is on the PR at https://github.com/ME-ICA/tedana/blob/pr/282/docs/multi-echo.rst but I'm not sure why I can't seem to link to it directly on this PR page.

jbteves:
@handwerkerd did you push your commit to this branch? For some reason the PR only shows the original commit.

jbteves:
Also, @handwerkerd did you mean to fork from your master branch? If you do a squash merge into `tedana` that will make your history a bit hairy.

emdupre:
I think it was probably an accident, but it's one we can fix pretty easily ! One idea would be to just close this PR, and then open a new PR from the "pr/228" branch (https://github.com/ME-ICA/tedana/blob/pr/282/docs/multi-echo.rst).

To "fix" your master branch, then just `git fetch upstream` and `git merge master` after we merge that PR in.

Let me know if that seems ok !

emdupre:
PS: I _really_ like the updated text, @handwerkerd ! Thanks again :sparkles:

PR #288, opened 2019-05-13
[DOC] Adds information on why we use multi-echo

emdupre:
Closes None :disappointed: 

- Updates multi-echo.rst to explain why researchers might use multi-echo fMRI
- Updates multi-echo.rst to include the Cam-CAN dataset.


emdupre:
Thanks for the review, @tsalo and @jbteves ! Updated to address comments -- LMK what you think :rocket:

PR #314, opened 2019-05-27
[DOC] Addition to the multi-echo fMRI section to include more background

### LINKED ISSUE #290: Update docs to make it more beginner friendly
### Summary
The documentation as of now has a couple of parts that could be updated to become friendlier for contributors that are just starting out with tedana. In particular, some terms and explanations could be flushed out a bit more (definition, more information, etc). And the first page could redirect the users to pages that are relevant to the information they are looking for.

### Additional Detail
In the homepage, I think it is helpful to refer users to either the 'Multi-echo fMRI' or 'Processing pipeline details' pages depending on what they want to learn more about tedana. In particular, if they want to learn more about the multi-echo fMRI technology, they should visit 'Multi-echo fMRI' and if they want to learn more about how tedana is analyzing the data, they should visit 'Processing pipeline details.' Right now there is a 'Content' section, but I think that the above information could go well in the 'About' section.

In both 'Multi-echo fMRI' and 'Processing pipeline details,' there are a few acronyms that are introduced early on, but not explained until either later or not at all. I think it would be good to expect the people who refer to that docs to refer to them out of order, so each page should expand on the acronym without assuming that the reader already knows the acronym from another page. For example, on 'Processing pipeline details,' the BOLD acronym is used, but it was explained in 'Multi-echo fMRI.'

For the 'Processing pipeline details' page, PCA and ICA are referred to in the beginning, but not explained until the later bullet points. Even then, it would be good to have a short 1 sentence summary about the PCA and ICA steps before big sections TEDPCA and TEDICA to give the readers a sense of where everything is going. 

In addition, the 'Processing pipeline details' could use some restructuring, in making the image and the section match (though it is quite organized right now). While making these changes, I will also try to add links to the page to prepare for the upcoming Hackathon.

This is just my perspective as someone who is just starting to read the docs for tedana! I will work on this for the next week, please assign me to this issue 😸 

### Next Steps
- [x] Edit the homepage to redirect users
- [x] Make additions to some sections in 'Multi-echo fMRI' (Make additions to ME MRI, why ME, and Sequences sections, which will add clarity)
- [ ] Make changes to 'Processing pipeline details' to explain all acronym, make sure that the page is readable independently 
- [ ] Make changes to 'Processing pipeline details' by adding more details to steps before sections to ensure a better logical flow
- [ ] Link sections to code or relevant API page

monicayao:
Closes part of #290 

- In the multi-echo fMRI sections, I have just included more information about what regular fMRI EPI give
- I had planned to add more to the "Why use multi-echo?" but it was already addressed in a previous PR! 
- Link users to pipeline>optimal combinations, and fixed link to the paper in "Why use multi-echo?" section

### DIFF
--- a/docs/multi-echo.rst
+++ b/docs/multi-echo.rst
@@ -1,7 +1,11 @@
 Multi-echo fMRI
 ===============
-In multi-echo (ME) fMRI, data are acquired for multiple echo times, resulting in
-multiple time series for each voxel.
+Functional MRI (fMRI) uses the change in blood flow, 
+and specifically blood oxygen level, in the brain to measure brain activity. 
+Only one unit of data, in units of voxel, 
+is collected per radio frequency (RF) pulse for most fMRI echo-planar imaging (EPI). 
+In contrast, in multi-echo (ME) fMRI, data are acquired for multiple echo times, 
+resulting in multiple time series for each voxel.
 
 The physics of multi-echo fMRI
 ------------------------------
@@ -62,7 +66,7 @@ spatial coverage, sample rate, signal-to-noise ratio, signal drop-out, distortio
 
 The one difference with multi-echo is a slight time cost.
 For multi-echo fMRI, the shortest echo time (TE) is essentially free since it is collected in the
-gap between the radio frequency (RF) pulse and the single-echo acquisition.
+gap between the RF pulse and the single-echo acquisition.
 The second echo tends to roughly match the single-echo TE.
 Additional echoes require more time.
 For example, on a 3T MRI, if the T2* weighted TE is 30ms for single echo fMRI,
--- a/docs/multi-echo.rst
+++ b/docs/multi-echo.rst
@@ -39,7 +39,7 @@ This allows researchers to benchmark their results.
 
 For optimally combined: Rather than analyzing single-echo time series separately,
 we can combine them into a "optimally combined time series".
-For more information on this combination, see :ref:`approach`.
+For more information on this combination, see `processing pipeline details`_.
 Optimally combined data exhibits higher SNR and improves statistical power of analyses in regions
 traditionally affected by drop-out.
 
@@ -53,7 +53,8 @@ Collecting multi-echo EPI allows us to leverage all of the information available
 as well as additional information only available when looking at signal decay across multiple TEs.
 We can use this information to denoise the optimally combined time series.
 
-.. _Pruim et al., 2015: https://www.sciencedirect.com/science/article/pii/S1053811915001822
+.. _processing pipeline details: https://tedana.readthedocs.io/en/latest/approach.html#optimal-combination
+.. _Pruim et al. (2015): https://www.sciencedirect.com/science/article/pii/S1053811915001822
 
 Acquisition Parameter Recommendations
 -------------------------------------
--- a/docs/multi-echo.rst
+++ b/docs/multi-echo.rst
@@ -1,11 +1,9 @@
 Multi-echo fMRI
 ===============
-Functional MRI (fMRI) uses the change in blood flow, 
-and specifically blood oxygen level, in the brain to measure brain activity. 
-Only one unit of data, in units of voxel, 
-is collected per radio frequency (RF) pulse for most fMRI echo-planar imaging (EPI). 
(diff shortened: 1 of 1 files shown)


docs/multi-echo.rst
```diff
@@ -49,7 +53,8 @@ Collecting multi-echo EPI allows us to leverage all of the information available
 as well as additional information only available when looking at signal decay across multiple TEs.
 We can use this information to denoise the optimally combined time series.
 
-.. _Pruim et al., 2015: https://www.sciencedirect.com/science/article/pii/S1053811915001822
+.. _processing pipeline details: https://tedana.readthedocs.io/en/latest/approach.html#optimal-combination
+.. _Pruim et al. (2015): https://www.sciencedirect.com/science/article/pii/S1053811915001822
```
jbteves: Here you make it look like the line was deleted; can you just add the processing pipeline details line *below* the Prium et al reference?

docs/multi-echo.rst
```diff
@@ -1,7 +1,11 @@
 Multi-echo fMRI
 ===============
-In multi-echo (ME) fMRI, data are acquired for multiple echo times, resulting in
-multiple time series for each voxel.
+Functional MRI (fMRI) uses the change in blood flow, 
+and specifically blood oxygen level, in the brain to measure brain activity. 
+Only one unit of data, in units of voxel,
```
jbteves: voxel is not a unit; I don't think this line says what you mean.

tsalo: Possible edit:
> In most echo-planar image (EPI) fMRI sequences, only one brain image is acquired at each repetition time.
> In contrast, in multi-echo (ME) fMRI, data are acquired for multiple echo times, resulting in multiple volumes with varying levels of contrast acquired at each repetition time.

docs/multi-echo.rst
tsalo: I think it's easier to interpret if the links appear in the same order as they occur in the text. In this case, it looks like the processing pipeline details link is before the Pruim reference.

jbteves: True; disregard my comment

docs/multi-echo.rst
monicayao: > voxel is not a unit; I don't think this line says what you mean.

@jbteves I was under the impression that voxel is a unit of 3D images as pixel is a unit for a 2D images? Could you elaborate a little?

monicayao: > Possible edit:
> 
> > In most echo-planar image (EPI) fMRI sequences, only one brain image is acquired at each repetition time.
> > In contrast, in multi-echo (ME) fMRI, data are acquired for multiple echo times, resulting in multiple volumes with varying levels of contrast acquired at each repetition time.

@tsalo I wanted to mention radio frequency, as it is used in the other sections on that page. I appended it to what you wrote an made another commit, please see if it makes sense!

jbteves: So what I said was wrong, because what I said was overly restrictive for the definition of unit. I think what @tsalo wrote was a better and more helpful suggestion, but I will attempt to unpack what I meant so I don't mislead you. I tend to think of units as things that describe a set quantity, like meters or grams, but a voxel is just something to be counted, which is why my kneejerk reaction was to type "not a unit" (which is wrong). However, the real problem that I failed to identify is that you state that only one unit of data is taken, which indicates that basically you're grabbing one number to represent one quantity at one point in time and space, and then you state that the unit of data is voxel, which is a unit in space only. So those two statements are actually subtly incompatible. What I *should have said* is that it would be more illuminating to say that in most scans, you acquire *one sample* at each point in time and space, whereas in multi-echo you collect *multiple samples* in time and space. Because the voxels are actually basically the same between echoes, it's the samples that are different because each sample has a different intensity. Voxel *is* a unit, but it's not the unit that's pertinent to multi-echo. So I hope that communicates what I originally meant, which was not to tersely state something incorrect. Please let me know if you're more, less, or the same amount of confused. Sorry about all that.

monicayao: @jbteves Yeah I got it! I was saying 

> one "unit" of data, in "units" of voxel

which is quite confusing now that I look back on it too. Looking back at it, it isn't necessary so I replaced it with @tsalo's version.

PR #325, opened 2019-06-03
[REF] Changes model module -> metrics module

### LINKED ISSUE #277: Rename tedana.model to tedana.metrics
### Summary
With #273 we'll be moving statistics-related functions out of `tedana.model.fit`, and so that module will become purely about metric calculation. This rename was proposed by @emdupre in her review of #266, but will be done separately to reduce PR bloat.

### Next Steps
1. Rename `tedana/model/` to `tedana/metrics/`. 
2. Update all imports throughout package, including tests. 
3. Update API documentation.

jbteves:
Closes #277 .

- Renames `tedana.model` to `tedana.metrics`


tsalo:
@jbteves @emdupre what do you think about renaming `fit.py` to something more in line with our plans for expansion? I was thinking maybe `meica.py`, because these metrics are specific to the ME-ICA pipeline.

jbteves:
@tsalo I think maybe `kundu_fit` since that's what we're naming what I would consider "original pipeline" code. We already have `kundu_selection`, for example. This seems like a natural extension. Plus I think it might be confusing to call something `meica` if we're deprecating it...

emdupre:
Agree that `meica` is likely to cause confusion, but I like the idea of renaming to make it clear that this was the original fitting method ! `kundu_fit` sounds just right to me :smile_cat:

jbteves:
Okay, 4d13bc6 should fix that.

PR #326, opened 2019-06-03
[REF] Gets rid of mask argument in tedana.fit.dependence_metrics

### LINKED ISSUE #175: Eliminate Mask Argument in fitmodels_direct
### Summary
Per this [comment](https://github.com/ME-ICA/tedana/pull/119#discussion_r233692095) in #119, we are currently overriding the `mask` argument with nonzero voxels in the `t2s` array. This is because, in the "old" version of selcomps (v3.2), masking and unmasking were applied repeatedly to coerce arrays into the right shape, while in v2.5 we've taken a simpler approach wherein we expect the arrays to be the same shape to begin with.

For example, we have `F_R2_clmaps`, which is `n_data_voxels` (`t2s != 0`) x `n_components`, and `Br_R2_clmaps`, which is `n_voxels` (from `mask`) x `n_components`. However, in either version of the component selection function, `Br_R2_clmaps` is masked to match the shape of `F_R2_clmaps`, so I think it's fine to skip a step and just construct `Br_R2_clmaps` in the same shape as `F_R2_clmaps`.

### Next Steps
- [ ] Eliminate the `mask` argument entirely and switch to explicitly using `t2s != 0` in `fitmodels_direct`.

jbteves:
Closes #175  .

- Gets rid of mask argument in tedana.fit.dependence_metrics
-


jbteves:
Note: this will cause conflicts with #325, so let's try and hash out any review over there before moving here so I can just push a rebase on master once that's merged.

jbteves:
I changed my mind and did the proper thing of fixing the conflicts; so this is ready for review @tsalo @emdupre

tedana/metrics/kundu_fit.py
```diff
@@ -68,9 +66,11 @@ def dependence_metrics(catd, tsoc, mmix, mask, t2s, tes, ref_img,
...
     mmix_new : :obj:`numpy.ndarray`
     """
+    # Use t2s as mask
+    mask = t2s != 0
     if not (catd.shape[0] == t2s.shape[0] == mask.shape[0] == tsoc.shape[0]):
         raise ValueError('First dimensions (number of samples) of catd ({0}), '
-                         'tsoc ({1}), t2s ({2}), and mask ({3}) do not '
+                         'tsoc ({1}), and t2s ({2}) do not '
                          'match'.format(catd.shape[0], tsoc.shape[0],
                                         t2s.shape[0], mask.shape[0]))
```
tsalo: Need to drop mask from string formatting.

jbteves:
@tsalo should be corrected.

jbteves:
@tsalo now that I haven't introduced a syntax error I think your review is addressed.

PR #331, opened 2019-06-07
[FIX] Make bi-sided clustering default

tsalo:
Closes None.

- Make the default thresholding method for cluster-extent thresholding "bi-sided" instead of "two-sided". This will cluster positively and negatively weighted voxels separately. Currently, they are merged into super-clusters.


jbteves:
@tsalo is this a WIP?

tsalo:
No, I just forgot to switch branches when I made that commit.

jbteves:
I hate it when I do that. Re-approved.

tsalo:
@emdupre @handwerkerd any conceptual concerns about this change?

emdupre:
What is the effect here on the outcomes ?

tsalo:
Any impact is minimal. In practice (using the five-echo test dataset), I get the same values for most metrics, except for small differences in `countnoise` (higher with the proposed changes, as some mixed "clusters" are no longer counted as clusters).

dowdlelt:
Conceptually this makes sense to me - in that if there is a alternating pattern of positive and negative betas on a map this would group that into a super cluster, when it is more likely to be an artifact from motion/GRAPPA, that is, reflective of bad structure in the data. So I am pro this.

jbteves:
@dowdlelt do you mind reading over the code and formally reviewing it?

tedana/utils.py
```diff
@@ -250,10 +250,10 @@ def threshold_map(img, min_cluster_size, threshold=None, mask=None,
         Boolean array for masking resultant data array. Default is None.
     binarize : bool, optional
         Default is True.
-    sided : {'two', 'one', 'bi'}, optional
+    sided : {'bi', 'two', 'one'}, optional
         How to apply thresholding. One-sided thresholds on the positive side.
         Two-sided thresholds positive and negative values together. Bi-sided
-        thresholds positive and negative values separately. Default is 'two'.
+        thresholds positive and negative values separately. Default is 'bi'.
     """
```
dowdlelt: Should this section specify the details of what is returned?

tsalo: Added.

jbteves:
@all-contributors add @dowdlelt for review

tsalo:
Thanks @dowdlelt and @jbteves! Merging now.

PR #333, opened 2019-06-07
[FIX] Add TR checking and user option

### LINKED ISSUE #332: Figures fail when TR is missing from header
### Summary
If the TR is missing from the dataset header, the `--png` figure creation process will fail. This is because the sampling rate is needed for the FFT - and a divide by zero error will occur. 

### Additional Detail
A user ran into this problem here:
https://neurostars.org/t/tedana-denoising-package-axis-limit-issue-with-the-new-plotting-feature/4424
And it looks to be caused by using SPM to process the data, but it is unclear what specific step led to the problem. User reports that the BIDS validator also complained.

### Next Steps
- [ ] Check TR on data input to confirm header TR valid
- [ ] Create optional flag (`--TR`?) so user can provide TR values directly
- [ ] Warn user if provided TR is doesn't match

jbteves:
Closes #332  .

- Checks for a TR of 0 and throws error if it is
- Allows user override of a TR of 0 via `--TR` option
- Throws warning if user overrides nonzero TR


tedana/workflows/tedana.py
```diff
@@ -332,7 +332,7 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
     elif user_tr != 0:
         # Coerce TR to be user-supplied value
         zooms = ref_img.header.get_zooms()
-        new_zooms = (zooms[0],zooms[1],zooms[2],user_tr)
+        new_zooms = (zooms[0], zooms[1], zooms[2], user_tr)
         ref_img.header.set_zooms(new_zooms)
 
         if tr != 0:
```
dowdlelt: Shouldn't this be 
`if tr != user_tr:`

I can imagine a case where the user is providing a TR, copies that code to another run, and forgets to edit. Thus, the dataset TR is correct (non-zero), but the user is overriding it with an incorrect value.

jbteves: Hm. My logic was that you should throw a warning anytime the TR is nonzero and the user is supplying one, but I guess that your change makes sense because what you *really* care about is a TR modification. I'll modify that. 

This is all predicated though on @tsalo and @emdupre agreeing with adding that option.

jbteves: 60f9858 should close this

dowdlelt: Right, right. I'm good with the warning at 328 - if the header is zero, the user should know - so that takes care of that. But if the header is some_value, and the user provides a different value, then the second warning (line 339) will show up, letting them know that they are forcing the TR to be different - which they may or may not want. 

If the user provides a tr that matches the header, then there is no warning - that seems right to me.

tedana/workflows/tedana.py
```diff
@@ -169,6 +169,15 @@ def _get_parser():
                                 'convergence is achieved before maxrestart '
                                 'attempts, ICA will finish early.'),
                           default=10)
+    optional.add_argument('--TR',
```
tsalo: I'd prefer to keep this lowercase for the sake of consistency.

tedana/workflows/tedana.py
```diff
@@ -187,7 +196,9 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
                     tedort=False, gscontrol=None, tedpca='mle',
                     source_tes=-1, combmode='t2s', verbose=False, stabilize=False,
                     out_dir='.', fixed_seed=42, maxit=500, maxrestart=10,
-                    debug=False, quiet=False, png=False, png_cmap='coolwarm'):
+                    debug=False, quiet=False,
+                    png=False, png_cmap='coolwarm', user_tr=0.0
```
tsalo: I like `t_r` personally. I don't think we need the `user_`. I'd also prefer to default to `None` instead of 0.

tedana/workflows/tedana.py
```diff
@@ -311,6 +322,23 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
+                      ' or fix your file header.')
+    elif user_tr != 0:
+        # Coerce TR to be user-supplied value
+        zooms = ref_img.header.get_zooms()
+        new_zooms = (zooms[0], zooms[1], zooms[2], user_tr)
+        ref_img.header.set_zooms(new_zooms)
+
+        if tr != user_tr:
+            LGR.warning('Mismatch in header TR and user-supplied TR,'
+                        ' please verify. Proceeding anyway.')
```
tsalo: It would be nice to say that it's specifically proceeding with user-supplied TR here.

tedana/workflows/tedana.py
```diff
@@ -311,6 +322,23 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
     n_samp, n_echos, n_vols = catd.shape
     LGR.debug('Resulting data shape: {}'.format(catd.shape))
 
+    # check if TR is 0
+    tr = ref_img.header.get_zooms()[-1]
+    if tr == 0 and user_tr == 0:
+        raise IOError(' Dataset has a TR of 0. This indicates incorrect'
```
tsalo: Extraneous space at the beginning of this.

tedana/workflows/tedana.py
tsalo: I would also make this an `info` and drop the warning elements of the message. E.g., just "Overriding header TR with user-supplied TR.").

tedana/workflows/tedana.py
tsalo: The `tr` from the data can be changed to something clearer, like `img_t_r` or something.

jbteves:
@tsalo I think e0891e8 addresses your review; let me know if not.

jbteves:
Wait, I lied, but c35141e should fix it actually.

tsalo:
Pretty much everything now looks great. I just don't think we need the "mismatch + please verify" bit. If you're feeding in a TR manually then I think we can assume you're doing it for a reason. Something shorter like `"Overriding image TR ({0}) with manual TR ({1}).".format(img_t_r, t_r)` (but better) should be sufficient.

jbteves:
I did that because @dowdlelt envisioned a scenario (which I think is quite likely) where someone has been running on a dataset with a bad TR in the header, then they copy code over to the next dataset and override a good TR accidentally.

dowdlelt:
In addition to that case, I can also imagine that some thinks they know the TR of some data, and they want to be explicit - but perhaps they are wrong? or the data is wrong? I think in both of theses cases, as well as mistaken code copying the user should at least be told that something strange is going on.

jbteves:
Okay @handwerkerd and @tsalo ; this is dramatically simplified compared to the original. Please re-review and see if you agree with my message and gist.

dowdlelt:
I thought we only wanted it to break in the png=True case? DayQuil has kicked in, so I might be misremembering...

jbteves:
@dowdlelt you're right; should be fixed in 2714dd6

Issue #336, opened 2019-06-17
Add documentation page with info linking to other multi-echo denoising software


handwerkerd:
At the OHBM 2019 Meeting, we received feedback that tedana's readthedocs info might be useful to all multi-echo users, but not all efforts to do multi-echo denoising will end up in tedana in the near future or ever. The suggest was to make a page our documentation where summaries of other available options and links to those options can be shared.

### Next Steps
1. Make a page where this info can be added.
2. Make stub paragraphs for currently available options
3. Invite others to contribute more information.

tsalo:
A partial list from the [meeting notes](https://docs.google.com/document/d/1wtOBKmKjQzmFpyo5V2a9n9ejtl2nCeHt3cQ63SdTHXc/edit):
- [3dMEPFM](https://afni.nimh.nih.gov/pub/dist/doc/program_help/3dMEPFM.html) (with multivariate version coming out next year)
- Bayesian approach uses the noise structure across time take advantage of priors and not use ICA. [OHBM 2019 poster M868](https://ww5.aievolution.com/hbm1901/index.cfm?do=abs.viewAbs&abs=5026)
- Group ICA applied to multi-echo data. [OHBM 2019 poster W337](https://ww5.aievolution.com/hbm1901/index.cfm?do=abs.viewAbs&abs=1286)
- Dual echo imaging by Molly Bright
- Beamforming by Pieter Buur: [paper](https://ieeexplore.ieee.org/document/4740318) and [thesis](https://repository.ubn.ru.nl/bitstream/handle/2066/91291/91291.pdf)
- MATLAB tool by Rasim Boyacioglu when at Donders
- Rasmus working with rapid imaging people to create a 3d radial acquisition to potentially create multiple echoes from parts of the k space that was acquired
- Martin Havlicek biexponential model taking information in the data to model the curves directly

jbteves:
Do we have contacts for these people? If so, how would you feel about asking them to just contribute a sentence for use on the page? That way each group gets to choose how it's represented on it.

emdupre:
That sounds like a great idea. If we have a stub of a page to point them to, I think that would help !

jbteves:
Please chime in on #343 if you have opinions about how we should request these contributions and how we should manage them.

emdupre:
Is this addressed in #419 ?

emdupre:
Pinging @dowdlelt @mvaziri !

emdupre:
Closed via #419.

PR #340, opened 2019-06-20
[doc] Add poster from OHBM 2019 meeting, fix RTD

### LINKED ISSUE #173: Upload tedana OHBM abstract somewhere in the docs
I'm not 100% sure where exactly we should put it, but I like the idea of having the abstract up in our documentation somewhere!

Let me know if you have any concerns about adding it, and/or bandwidth to upload it otherwise I'll try to have a bash next week ✨

emdupre:
Closes #173, #350 

- Adds OHBM 2019 poster to RTD site
- Corrects help text for random seed; confusion which came up during the software demo
- @rmarkello additions to fix RTD warnings

I know @rmarkello also had a discussion with Lauren Atlas about adding more information on when to use multi-echo to the docs -- could you confirm where that would be best placed ?


tedana/workflows/tedana.py
```diff
@@ -138,7 +138,7 @@ def _get_parser():
     optional.add_argument('--seed',
                           dest='fixed_seed',
                           type=int,
-                          help=('Value passed to repr(mdp.numx_rand.seed()). '
+                          help=('Value random initialization of ICA algorithm. '
```
```suggestion
                          help=('Value used for random initialization of ICA algorithm. '
```

emdupre:
I think listing in reverse-chronological is a great idea ! Will update :+1: 

I think for now I'd prefer to include them rather than link out, but definitely something to keep in mind as our community grows :smile_cat:

rmarkello:
I think that all the recommendations should ideally go in one section, together. It looks like currently the acquisition parameter recommendations are under the multi-echo section of the docs, so for the time being a section right before that (and after "Why use multi-echo?") with some general recommendations would work. Something like "If you are new to multi-echo and want to follow a traditional pre-processing pipeline with limited manual intervention and inspection, we recommend only using the optimal combination via `t2smap`. If you have the resources to devote to manual inspection of every run then we encourage you to use `tedana`."

It would be great if this sort of section explicitly was also linked from the landing page of the documentation where the links to Multi-echo fMRI and Processing pipeline details are. For researchers new to multi-echo having this sort of thing very visible would be awesome / very helpful.

jbteves:
I agree with you, @rmarkello, but maybe let's keep the scope of this PR small and continue discussion on @handwerkerd 's open issue #336 or opening a separate one with additional tedana recommendations. I think there's probably some tie-ins to both, since a new user will want both pages. Does that sound good?

rmarkello:
Yeah, you're welcome to incorporate those changes wherever; I was just responding to @emdupre's question in the initial PR commit!

jbteves:
Ah, sorry I confused the situation! 😬

rmarkello:
No worries! 

To recount the takeaways from the conversation so everyone is approximately on the same page, our discussion after the multi-echo meeting in the OSR focused on how researchers who have collected multi-echo data for one reason or another (the local MR physicist convinced them, a visiting researcher told them to do it, they inherited a dataset, etc.) can find multi-echo processing supremely overwhelming (even with all the advancements `tedana` has made in reducing that!). As such, given all the currently open issues on the denoising procedure and its stability / reliability, it might be nice to have a set of very **high-level** recommendations for people who have multi-echo data but just need to _get going_ with their analyses. 

For example, you can imagine a new tenure-track researcher with a small lab who just collected multi-echo fMRI task data for 40 subjects x 8 runs doesn't necessarily have the time to go through and manually inspect all the components for multi-echo ICA denoising (potentially >200 components for 320 independent runs) on TOP of all the normal visual inspection they have to do for the pre-processing / normalization of fMRI data. At the point, the researcher might just want to take the SNR boost from the optimal combination and move along&mdash;but how do they know that's an option? We discussed that `tedana` could do a lot of good by making some recommendations for these sorts of cases since the developers here are some of the ones at the cutting-edge of these denoising difficulties and know, more-or-less, what works and what doesn't. 

The docs have already started to work towards these sorts of things with acquisition recommendations, but the thought here was having something up-front to say: "Hey, we know that multi-echo denoising is a complicated procedure. And right now, there are still a lot of problems with it (<link to issues>). If you aren't interested in getting into the weeds with your data right now, we recommend using the optimal combination and moving forward with more traditional processing pipelines."

The work in #336 is definitely complementary (i.e., if you don't want to use `tedana`, here are some other options!), but definitely not part-and-parcel with these sort of recommendations.

jbteves:
@emdupre this actually closes #173; do you mind editing your comment?

emdupre:
> you dropped the submodules selection._utils, decomposition._utils, and metrics.kundu_fit

I'm not sure we should add the first two back in, since they're actually private modules (as denoted by the underscore). If you think we should make them public facing, maybe we should rename them ? WDYT ?

rmarkello:
Hey @tsalo! Just following up on a few things:

1. Totally agree that dropping the module names would be beneficial! We retained them simply so that the generated documentation would look approximately the same as the current documentation (sans warnings on compilation). Happy to see the prefixes removed, though.
2.What @emdupre said. Plus, the drop in `metrics.kundu_fit` was because the only two functions in the module are already documented on the main API page.

I'm not sure I totally understand what you mean by "write out each function and class to its own page." Is that not what's happening right now? For example, when I click `tedana.workflows.tedana_workflow` on the API page I get redirected [here](https://tedana.readthedocs.io/en/latest/generated/tedana.workflows.tedana_workflow.html#tedana.workflows.tedana_workflow). Could you elaborate a bit on what you were thinking?

tsalo:
@emdupre I guess it's not really a big deal. I was imagining that private functions/modules wouldn't show up in the main API page, but would still be accessible through module-specific pages. Unfortunately, we're not able to do that yet.

@rmarkello Only the functions and classes specified in api.rst get their own pages. If we put in a module, the module template will make it so that the module has its own page with a table at the top and all of the functions and classes below. So, using `kundu_fit` as an example, we have [a page for the function dependence_metrics](https://tedana.readthedocs.io/en/latest/generated/tedana.metrics.dependence_metrics.html#tedana.metrics.dependence_metrics) and [a page for the submodule kundu_fit](https://tedana.readthedocs.io/en/latest/generated/tedana.metrics.kundu_fit.html#module-tedana.metrics.kundu_fit). That module page has the nice table, but it also has dependence_metrics documented on the page (rather than linking to the separate page for the function). Optimally the module page would just have the table, and the links wouldn't resolve to sections on the module page but rather to function-specific pages. That way we wouldn't need to explicitly include every function in api.rst when there are submodules.

In a perfect world, all we would need is to list the modules in api.rst and it would create the tables we currently have. Any submodules would link to pages containing just tables that would also be automatically generated.

rmarkello:
Okay, so just to make sure we're on the same page... Going off the `kundu_fit` example, your ideal would be for the main API page for that module to look like:

### `tedana.metrics`: Computing TE-dependence metrics
| | |
| --- | --- |
| **`tedana.metrics.kundu_fit`** | Fit models |

(ignoring the blank header in the table because GitHub can't render header-less tables). That is, the table has _neither_ of the submodule functions listed because they belong to the submodule itself. Then, when you click on the submodule `tedana.metrics.kundu_fit`, you're taken to a page with a table like the one currently at the top of [this page](https://tedana.readthedocs.io/en/latest/generated/tedana.metrics.kundu_fit.html#module-tedana.metrics.kundu_fit), but _without_ the function documentation listed below the table. Then you can click on, e.g., `dependence_metrics()`, you're taken to a page [like this](https://tedana.readthedocs.io/en/latest/generated/tedana.metrics.kundu_fit.html#module-tedana.metrics.kundu_fit)?

Let me know if I'm misunderstanding! Sorry for the back-and-forth.

tsalo:
That's exactly it! Except we could have the functions from `tedana.metrics` that we really care about manually listed in api.rst as well. And it wouldn't give us duplicate entry warnings.

rmarkello:
:raised_hands: glad we've got that all sorted! Now on to debating the merits of this schema... :sweat_smile: 

I'm curious (genuinely!) as to what benefit you see in having API documentation for 4th-level functions, like `tedana.metrics.kundu_fit.dependence_metrics()`, when those functions are accessible via `tedana.metrics.dependence_metrics()`? In my mind the benefit of having to manually curate the api.rst page is that it forces thinking about which functions _should_ be documented. In my mind users shouldn't have to dive into the `kundu_fit` submodule at all (or the `~._utils` modules)&mdash;they'd just use the 3rd-level `tedana.metrics` functions directly.

Honestly, given the way `tedana` is designed to be used (i.e., calling either `tedana` or `t2smap` from the command line), I'm not completely sold on their being a public-facing API page at all! That said, that goes back to earlier discussions about whether `tedana` should be just a set of user-facing workflows or a more comprehensive set of functions for working with multi-echo data. If the latter then the API definitely makes sense, but as of right now I feel the workflows are the work-horses of the library...

tsalo:
I see those duplicates as more of an acceptable bug than a feature. There's no benefit to having them, but they'll need to exist if we want both manually specified pages on `api.rst` and automatically generated pages for the modules. There should be _some_ way to access the documentation for everything- including private submodules. They just don't need to be highlighted on the API page, I think. Perhaps it's just that I've been too aggressive about making functions and submodules private.

I think we definitely need the API page. Even if users don't use it, it's necessary as a reference for developers. Also, I believe that the current vision for tedana is to provide canonical workflows _and_ to act as a sandbox in which various multi-echo denoising approaches can be collected, so the API is useful in that sense as well.

handwerkerd:
For the discussion of what functions to highlight, the benefit I see for a future metrics module is, when each metric can be independently calculated in a separate function, it would be very useful to have a clear list of these metrics in one place (possibly in a curated order) that might exclude/segregate lower level functions that are used to connect the metrics, but 99% of developers wouldn't ever want to call directly. That said, organizing functions/modules in a way where this would happen automatically would be a lot more sustainable than relying on manual curation of api.rst.

Issue #269, opened 2019-04-22
Reduce memory usage for metric calculations


jbteves:
### Summary
RAM usage appears very high during metric calculations. In order to reduce memory usage, we should look for places where we can write images to disk and free the memory they were using. 

### Additional Detail
There are actually two places where memory usage is high per a quick test by @dowdlelt: 
1. PCA
1. Metric Calculation.

However, as @tsalo has pointed out in Gitter, PCA is implemented in sklearn and its memory usage may be stuck high. Since we have more control over the metric calculation, we should attempt to curb its memory usage where possible, especially since as of issue #254 there is at least one user who cannot run `tedana` to completion since its memory usage was so large. (See figure [here](https://github.com/ME-ICA/tedana/issues/254#issuecomment-485448886)). 

### Next Steps
- [ ] Identify where files can be written to disk and reloaded later
- [ ] Implement these changes in code
- [ ] Determine new memory requirements per these changes

tsalo:
We've had another user report a memory problem (see [here](https://neurostars.org/t/image-orientation-changed-after-running-tedana/4443/7?u=tsalo)), so I'd like to move this issue up in terms of priority. I've asked him to give some information about his dataset so that we can simulate some data that are large enough to replicate the issue. Once I have that, I'll dig into this.

dowdlelt:
I know tedana has masking and unmasking features - is it operating on a vector of values within the mask or is it still holding all the ~zero values on the outside of the mask/brain in memory? In the original MEICA the data was actually cropped around the edges to remove anything outside of the brain. It was annoying (output dims differed from input), but effective in reducing memory usage. 

I know that the clustering steps require that the brain is organized as 3D, but perhaps the heaviest lifting steps could be done on Vox X Time data, if they are not already.  I think the tradeoff for internally flipping the data back and forth would be worth it for less memory usage. 

We could always reproduce the original MEICA implementation of cutting away slabs of data that is outside of the brain mask and then adding zero filler slabs back at the end so that the data is the same dimension again.

tsalo:
I believe that it operates on the masked data already. The mask is then used to unmask the data before writing it out, so that shouldn't be a problem here (and we shouldn't have to crop the unmasked images).

jbteves:
[...]

PR #345, opened 2019-06-21
[ENH, REF] Reduce memory requirements for metric calculation and PCA

tsalo:
References #269. I very much doubt that these changes will solve the problem on their own, and I'm not even sure if they'll help much.

- Add `--lowmem` argument to trigger use of IncrementalPCA. Should also trigger other low-memory steps we might implement in the future.
- Clean up arrays in `dependence_metrics` when they're no longer used.
- Operate on masks arrays as much as possible within `dependence_metrics`.
- Make `mask` argument for `computefeats2` optional.

### DIFF
--- a/tedana/metrics/kundu_fit.py
+++ b/tedana/metrics/kundu_fit.py
@@ -209,8 +209,10 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
 
     if verbose:
         # Echo-specific weight maps for each of the ICA components.
-        io.filewrite(betas, op.join(out_dir, '{0}betas_catd.nii'.format(label)),
+        io.filewrite(utils.unmask(betas, mask),
+                     op.join(out_dir, '{0}betas_catd.nii'.format(label)),
                      ref_img)
+
         # Echo-specific maps of predicted values for R2 and S0 models for each
         # component.
         io.filewrite(utils.unmask(pred_R2_maps, mask),
--- a/tedana/metrics/kundu_fit.py
+++ b/tedana/metrics/kundu_fit.py
@@ -249,7 +249,7 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
                 np.squeeze(utils.unmask(F_R2_maps[:, i_comp], mask)))
             F_R2_clmaps[:, i_comp] = utils.threshold_map(
                 ccimg, min_cluster_size=csize, threshold=fmin, mask=mask,
-                binarize=True).astype(bool)
+                binarize=True)
             countsigFR2 = F_R2_clmaps[:, i_comp].sum()
 
             ccimg = io.new_nii_like(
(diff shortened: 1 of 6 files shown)


emdupre:
Can we also set `copy=False` as the default for regular PCA ?

tsalo:
Absolutely. Done!

tedana/metrics/kundu_fit.py
```diff
@@ -240,40 +246,41 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
             # Cluster-extent threshold and binarize F-maps
             ccimg = io.new_nii_like(
                 ref_img,
-                np.squeeze(utils.unmask(F_R2_maps[:, i_comp], t2s != 0)))
+                np.squeeze(utils.unmask(F_R2_maps[:, i_comp], mask)))
             F_R2_clmaps[:, i_comp] = utils.threshold_map(
                 ccimg, min_cluster_size=csize, threshold=fmin, mask=mask,
-                binarize=True)
+                binarize=True).astype(bool)
```
emdupre: Why do we need to explicitly cast to boolean after `binarize=True` ? Should that casting instead be performed inside `utils.threshold_map` ?

tedana/metrics/kundu_fit.py
```diff
@@ -113,59 +118,53 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
...
-    # mask arrays
-    mumask = mu[t2s != 0]
-    t2smask = t2s[t2s != 0]
-    betamask = betas[t2s != 0]
-
     # set up Xmats
-    X1 = mumask.T  # Model 1
-    X2 = np.tile(tes, (1, n_data_voxels)) * mumask.T / t2smask.T  # Model 2
+    X1 = mu.T  # Model 1
+    X2 = np.tile(tes, (1, n_voxels)) * mu.T / t2s.T  # Model 2
```
emdupre: What's the motivation for the name change here (i.e., `n_data_voxels` to `n_voxels`) ? Did we change the meaning or just the name ?

tsalo: `n_data_voxels` and `n_voxels` originally corresponded to `t2s != 0` and `mask`, respectively. With the replacement of `mask` with `t2s != 0`, they became duplicates of one another.

tedana/metrics/kundu_fit.py
tsalo: I'm not sure. It's true that binarized data should be boolean, but nibabel is super-mean and for some reason doesn't work with boolean arrays, so there's an argument to be made for using integers as the default for binarized arrays in `threshold_map`.

emdupre: Looking through I think the function is only used in these five instances you're updating, so I think we should just update the function, probably ! If we use it in other ways, we can re-consider the default. WDYT of that idea ?

tsalo: Sounds good. Should be good now.

tsalo:
I have no clue how much (or even if) these changes help. I don't know how to do proper memory profiling to find out either.

jbteves:
There aren't really builtins other than manually running `pdb` and noting the consumption, unfortunately. There are some packages to do it but there's no reason to run the profile before merging; we may want to take a while to think of a good approach before we publish memory consumption guidelines anyway.

PR #349, opened 2019-07-02
[DOC, ENH] Generate workflow description for each run

### LINKED ISSUE #330: Add boilerplate generation to tedana workflow
### Summary
It would be a great feature if we could output a citable boilerplate (a la fMRIPrep) when running tedana. This would be a full paragraph describing the inputs, steps taken, and including citations for relevant works.

### Additional Detail
I see a couple of ways forward on this. The harder (but potentially more extensible) approach would be to convert our functions to classes, as is done in fMRIPrep, and to incorporate some kind of attribute with the description of the step into each class. The paragraph can be built by compiling the classes' descriptions in sequence with the workflow. The other, definitely easier approach would be to have a long string that we can format based on inputs. If we want to incorporate run information (e.g., number of components), we can also include information from the run file. I don't think I'd recommend this, because it would mean having a different boilerplate for each run of data. That information can go in a table or something if people really want to report it.

### Next Steps
- [x] Write format-able workflow description string.
- [x] Format string with workflow inputs (after any conflicts are resolved).
- [x] Output formatted string to boilerplate file in tedana output directory.

tsalo:
Closes #330.

- Generate boilerplate when running the tedana workflow and save to `report.txt`.
- Add `report.txt` to CircleCI outputs.


emdupre:
This is really cool ! Can we render a sample report in the documentation ?

tsalo:
You mean like add the one from #330 to outputs.rst?

emdupre:
Yes, I think that would be really useful !

tsalo:
Added. I don't love how RST renders block quotes (it just tabs them out), but I think this will be very informative for users.

emdupre:
Thanks, @tsalo ! I agree.

One follow-up question is whether / how this interacts with DueCredit. Since I think we have a few citations there that are not in these reports !

tsalo:
Which ones?

emdupre:
I was thinking the PCA one https://github.com/ME-ICA/tedana/blob/1baed235d1d4549cc0ee95716e23721b289a1879/tedana/decomposition/pca.py#L19

and the two for Dice https://github.com/ME-ICA/tedana/blob/0c0d7047f6b39a86aa58a0ca8fd75dbb5bdb077d/tedana/utils.py#L131

tsalo:
Minka's citation should be in the boilerplate. It's just that the example uses Kundu TEDPCA.

I totally missed Dice (nice catch). I can add that one in.

tsalo:
I don't think we update our citations frequently enough to merit that. I think we'll probably have to update our contributing guidelines anyway once other multi-echo developers start adding their methods into `tedana`, and making sure they receive their due credit will probably be part of that. I think we can wait until that point to update our guidelines.

emdupre:
> I think we'll probably have to update our contributing guidelines anyway once other multi-echo developers start adding their methods into tedana, and making sure they receive their due credit will probably be part of that.

Agreed, but just from the recent RTD issues I think we generally need to do a better job of remembering to update the docs (duecredit, in my mind, is part of documentation). It's just easy to forget when the code runs ! If you want to wait to update guidelines around that, I'm fine with it, but I'd like to note it somewhere so we remember to include it later. Maybe this conversation is enough.

jbteves:
@emdupre and @tsalo can you throw the duecredit/placement of citations into an issue and link it from here?

tsalo:
@jbteves Done with #362. Merging this now.

jbteves:
@tsalo thanks!

Issue #350, opened 2019-07-02
ReadTheDocs compile warnings


handwerkerd:
### Summary
When I tried to compile the documentation with "make html" I saw a bunch of warnings. Some were minor, but many seem to be because files or functions were moved into different subdirectories from when api.rst was first made. I think this means that not all functions will automatically have documentation rendered.

### Additional Detail
Some warnings include:
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:11: WARNING: toctree references unknown document 'generated/tedana.workflows'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:30: WARNING: toctree references unknown document 'generated/tedana.decay'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:49: WARNING: toctree references unknown document 'generated/tedana.combine'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:67: WARNING: toctree references unknown document 'generated/tedana.decomposition'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:90: WARNING: toctree references unknown document 'generated/tedana.metrics'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:113: WARNING: toctree references unknown document 'generated/tedana.selection'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:136: WARNING: toctree references unknown document 'generated/tedana.gscontrol'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:155: WARNING: toctree references unknown document 'generated/tedana.io'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:181: WARNING: toctree references unknown document 'generated/tedana.stats'
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:201: WARNING: toctree references unknown document 'generated/tedana.utils'

There's also a repeated warning:
/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/api.rst:39: WARNING: Duplicate explicit target name: "calibration_ref".
because .. _calibration_ref: is repeated throughout api.rst rather than being changed for each section.

### Next Steps
I started to fix a couple of these, but it would go faster if someone who moved some of these function files around & know what was initially intended for each section can correct this.

[...]

PR #351, opened 2019-07-02
Multi-echo background documentation edits

handwerkerd:
Progress for #290

- Created a new page for the embedded publications spreadsheet rather than just having a link to a google doc buried in the middle of a page
- Put a few more background references and another educational talk into multi-echo.rst
- Corrected. a few quirks that were causing problems with how the documentation was being formatted for ReadTheDocs

Others should take a look at how the spreadsheet renders within ReadTheDocs to see if they think it's ok. It goes a bit beyond the edge of the standard frame, but I think this still looks better than just asking people to go to a separate webpage with the spreadsheet.

I've explained more about the ReadTheDocs warnings in #350 but I fixed some of the low handing fruit in this PR


handwerkerd:
This should be a fairly quick review. As @tsalo noted, just run "make html" in the docs directory & you'll be able to see that everything rendered correctly. It would be good if @dowdlelt checked to see if he likes how his spreadsheet of publications is rendering within the documentation. In relation to #350 it would be useful if @jbteves and others also ran "make html" to see what the underlying issue is.

PR #354, opened 2019-07-04
[DOC] Walk through TE-dependence in more detail

tsalo:
Closes None.

- Add notebook that simulates BOLD and non-BOLD fluctuations to show how we evaluate them in tedana.
- Add page to documentation about the calculation of TE-dependence and TE-independence metrics.
- Consolidate existing real-data example notebooks into one.
- Switch real-data example notebooks to use Logan's 8-echo data.


tsalo:
The range of values will depend on signal-to-noise ratios for both S0 and T2* fluctuations, as well as relative contributions from each. I'm not sure what counts as reasonable. If I were to take a guess, I'd say 0-100, but I'd prefer to have input from others before committing to that.

What would you prefer instead of "you may notice." I'm totally fine changing the language.

`make.bat` is the result of #351, which also diffed the file without making changes. I tried to discard my local changes to the file (since there were none), but GitHub Desktop was having none of it.

jbteves:
I personally prefer "note," which is also a bit shorter. It's super nitpicky. 
It may be asking a bit much to pin down a range, that's true. I'll go ahead and approve.

Issue #352, opened 2019-07-04
Question regarding number of echoes / max echo time


cjl2007:
Hi - I am experimenting a bit with some multi-echo acquisitions with a high number of echoes, 8 total. 

I understand that the majority of the signal of interest has been lost in most brain areas for these later echoes, but I wanted to test empirically at what point there is no benefit to adding echoes. 

I am surprised to see that tedana fails to identify any bold like components when using all 8 echoes, but does if I include the first ~5. 

Is there some limit within tedana, either in terms of the number of echoes or the latest echo time, that I should keep in mind? 

Thanks

tsalo:
It is much more likely that the decision tree is just very brittle, which is something we're trying to work on. If you look at the component table, are most of the components being rejected because of I007, I009, or I010? Those are "mid-kappa" rejections, which tend to be the most problematic.

cjl2007:
I006, actually (“ no good components found).

It worked properly when including 5 and 6 echoes total. Rerunning now with 7 total. The “no bold components” message was from my first attempt, when I included all 8.

tsalo:
Okay, that still fits though. I006 occurs after the "absolute" classifications (i.e., classifications that occur independently for each component and don't depend on metrics from other components) at the beginning of the "relative" classifications. These "relative" classifications use things like elbows across components in the metrics to identify good/bad components, and seem to be the most finicky.

smoia:
@cjl2007 , can I ask you what are your echo times? And also which metric are you using (mle, kundu, ...)?
I was running MEICA with 7 without any problem, but the last one was still in the 80-85 ms range.

@tsalo, shouldn't this be taken care of by the adaptive t2s mask? that's what I understood previously...

cjl2007:
Echoes = 13.40,31.11,48.82,66.53,84.24,101.95,119.66,137.77
I was using the Kundu option.

smoia:
And how many components do you have?

tsalo:
@smoia The adaptive mask isn't used in metric calculation. I believe that it's only used for T2*/S0 calculation, which are only used for optimal combination. Since optimal combination is a weighted average where no weights are ever zero, echoes affected by dropout still contribute to the optimally combined data.

Technically T2* _is_ used in fitting the TE-dependence model, but since it's a constant it actually has no impact. Plus, since the adaptive mask isn't used, all echoes are included for the metric calculation.

smoia:
I see @tsalo , now I understand, thank you!

[...]

Issue #312, opened 2019-05-26
Use overall signal decay in adaptive mask creation


tsalo:
### Summary
As discussed in #113, one possible improvement to the `make_adaptive_mask` function would be to flag echoes where the mean signal for that voxel fails to decrease from one echo to the next. This can be done in conjunction with the current method, rather than as a replacement.

### Additional Detail
There are probably a lot of ways in which we can improve this first pass (e.g., not using the mean, setting tolerances for the differences, etc.).

### Next Steps
- [ ] Discuss this method
- [ ] If we want to move forward with it, implement the method in `make_adaptive_mask`

dowdlelt:
I like this idea - in addition, if we were to use more advanced methods (ie more computationally demanding methods) for fitting decay it would be important to remove these bad voxels from consideration as well (or at least know about them). 

Would this idea propose excluding those voxels entirely? Such that there are holes in the brain? That might could cause some issues..

tsalo:
@dowdlelt Sorry, somehow I never saw your response. It does look like some voxels in CSF are flagged with only one good echo, but I think that's okay. I haven't seen any cases with holes in gray or even white matter.

jbteves:
Hm, I'd be interested in doing this as it also incorporates some of @smoia's requests RE: echo order. If this metric tells us all voxels do that, it probably means that the wrong echo order was supplied.

PR #358, opened 2019-07-10
[ENH] Only use good echoes for metric calculation/optimal combination

### LINKED ISSUE #255: Default settings for non-MLE PCA raise warning
### Summary
The default settings for sklearn's PCA (used for the Kundu tedpca decision tree, but not MLE) involve deriving as many components as there are volumes in the time series. The last component explains no variance and causes the following warning to be raised for both the three-echo rest and five-echo task test datasets:

```
tedana/tedana/model/fit.py:169: RuntimeWarning: divide by zero encountered in true_divide
  F_S0 = (alpha - SSE_S0) * (n_echos - 1) / (SSE_S0)
```

### Additional Detail
From what I recall, PCA should estimate `n_vols - 1` components, not `n_vols`, which would explain why the last component explains no variance and is causing this divide-by-zero warning. I think we should override the default number of components here with `n_vols - 1`. This will prevent the warning, but the components that matter will be the same.

However, I also noticed that changing the number of components will also impacts the decision tree, so I decided to raise an issue where we could discuss this rather that opening a small bug-fix PR.

### Next Steps
1. If we decide to override the default number of PCA components, then here:
https://github.com/ME-ICA/tedana/blob/82447c5e73133f5eeb566c56f27f396a78ab7d98/tedana/decomposition/eigendecomp.py#L301-L306
we can simply change `ppca = PCA()` to `ppca = PCA(n_components=(n_vols - 1))`

tsalo:
Closes #255. References #352.

- Use adaptive mask when calculating dependence metrics and for optimal combination
- Change references to betas into pes or param_estimates when appropriate
- Improve memory management slightly
- Limit default PCA to number of volumes - 1, which should fix the warning from #255


tsalo:
For this to work, we need to trust that the adaptive mask is accurate and restrict everything to voxels with at least 3 good echoes. I am fine with this, although we should take a look at #312 and #113 to make sure we're confident with `make_adaptive_mask`.

The only thing I'm a little concerned about is the F-statistics. The adaptive mask is pretty smooth, but if we assume that it's not then we're talking about adjacent voxels with different degrees of freedom.

emdupre:
Revisit pending discussions on #113 , hopefully during the tedana hackathon !

tsalo:
Per discussion at hackathon and coffee shop _after_ hackathon, we are going to move forward with this at a conceptual level. I will resolve conflicts and then we can do a review of the actual implementation.

tsalo:
The merge conflicts have been resolved and now this PR is ready for review!

handwerkerd:
Before taking a few days off, I focusing more on writing #491. @tsalo, Would you prefer that I review & approve this PR before #491? Should I try to add some of the bug fixes that 491 is supposed to address into here or keep this one as-is and then address those added issues in 491?

tedana/combine.py
```diff
@@ -107,8 +107,8 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
         Concatenated BOLD data.
     tes : (E,) :obj:`numpy.ndarray`
         Array of TEs, in seconds.
-    mask : (S,) :obj:`numpy.ndarray`
-        Brain mask in 3D array.
+    adaptive_mask : (S,) :obj:`numpy.ndarray`
+        Brain adaptive_mask in 3D array.
```
handwerkerd: adaptive_mask is just the number of good echoes and, if the number is less than the total number of echoes, there is an assumption that the latest echoes will be the ones excluded from the mask. This seems like a reasonable assumption, but any function that applies the adaptive_mask should note this assumption in the documentation. I suspect this will prevent confusion & headaches in the future.

tedana/metrics/kundu_fit.py
```diff
@@ -126,16 +123,17 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
...
     mu = catd.mean(axis=-1, dtype=float)
     tes = np.reshape(tes, (n_echos, 1))
     fmin, _, _ = getfbounds(n_echos)
 
-    # set up Xmats
-    X1 = mu.T  # Model 1
-    X2 = np.tile(tes, (1, n_voxels)) * mu.T / t2s.T  # Model 2
+    # set up design matrices
+    X1 = mu.T  # Model 1: TE-independence model
+    X2 = np.tile(tes, (1, n_voxels)) * mu.T  # Model 2: TE-dependence model
```
handwerkerd: In the current code, X2 = np.tile(tes, (1, n_voxels)) * mu.T / t2s.T
The t2s is now excluded from this calculation and t2S is no longer even passed as a variable to dependence metrics. I'll need to look a bit more carefully at what is happening in the new lines 158-183 (That section could use a few more comments), but it doesn't seem to have a replacement from the removable of t2s here. Am I missing something?

tedana/metrics/kundu_fit.py
```diff
@@ -126,16 +123,17 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
     # compute Betas and means over TEs for TE-dependence analysis
     betas = get_coeffs(utils.unmask(catd, mask),
                        mmix,
-                       np.repeat(mask[:, np.newaxis], len(tes), axis=1))
+                       np.repeat(mask[:, np.newaxis], len(tes), axis=1),
```
handwerkerd: Since len(tes) is fixed here and the adaptive mask isn't used, it seems like betas is being calculated on echoes that might be outside the adaptive mask.

tedana/workflows/tedana.py
```diff
@@ -463,7 +463,7 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
         io.filewrite(s0_full, op.join(out_dir, 's0vG.nii'), ref_img)
 
     # optimally combine data
-    data_oc = combine.make_optcom(catd, tes, mask, t2s=t2s_full, combmode=combmode)
+    data_oc = combine.make_optcom(catd, tes, masksum, t2s=t2s_full, combmode=combmode)
```
handwerkerd: `mask`, `masksum`, and `adaptive_mask` are used inconsistently throughout the code. Here is one example where masksum in the input and make_optcom calls it adaptive_mask.
My suggestion is that `mask` should always refer to a boolean mask. `adaptive_mask` should refer to what is sometimes called `masksum`. `masksum` is a confusing variable name and should be removed.

tedana/metrics/kundu_fit.py
tsalo: That is correct. The thing is, `t2s` has no impact on any calculated values, so it doesn't need to be passed into this function.

tedana/workflows/tedana.py
tsalo: Totally agreed, but I think that renaming the variable throughout the package is outside the scope of this PR. In this case, I'm only editing `tedana.py` as much as I have to, but we do have #448, which I think should have its own PR that addresses this.

tedana/combine.py
tsalo: Great point. I will expand the docstring.

tedana/metrics/kundu_fit.py
tsalo: `betas` are calculated for all echoes at all voxels, and then only good voxels and good echoes are used later on.

tedana/metrics/kundu_fit.py
handwerkerd: I'm either missing something here or need to dig into what happens next to see if t2s just cancels itself out. In the current code `t2s` is used to calculate `X2` This is removed from this pull request.
`X2` is used to calculated `coeffs_R2` and `pred_R2` which ends up in a bunch of other variables including `F_R2_maps` On the face, that seems fairly significant and removing the division by `t2s` in the `X2` calculation is essentially removing all T2* information from the fit to the R2 model.

tedana/workflows/tedana.py
handwerkerd: Sounds good. While we're editing `make_adaptive_mask` I could see making the name change just within that function, but that function-specific change can also be done as part of #491 where getting consistent names right will be more important within a few sub-functions. I'll resolve this critique for now.

tedana/metrics/kundu_fit.py
tsalo: `t2s` is an (S) array, while `X2` is an (S x E) array. Essentially, `t2s` acts as a scalar for the design matrix of E values for each voxel.  Scaling the design matrix will impact `coeffs_R2` and `SSE_R2` (though not `pred_R2`), but those cancel each other out. Since it's just a scalar for each voxel's equation and F-statistics are calculated independently across voxels, it also shouldn't impact goodness of fit for the model.

If you run it both ways (with vs. without `t2s`), the Kappa values end up the same.

handwerkerd: OK. Took a while to warp my head around this & then run on a data set to confirm, but this makes sense now. I'd say we should comment this code more, but it's going to be replaced by the clearer and better commented #475 anyway.

handwerkerd:
Besides the one minor requested doctoring addition and ci merge_coverage failure, this looks good to me. That failure looks like it has more to do with the testing code rather than a specific output. Since I don't understand the testing code as well, perhaps @emdupre can review & focus there. Otherwise, I'm fine approving this merge.

dowdlelt:
Is this failing because the 4 extra comps here:
https://github.com/ME-ICA/tedana/blob/5ce829675bec9c3c4c7010f1ffabd1af67d7249b/tedana/tests/data/cornell_three_echo_outputs.txt#L75-L78
are not being generated/ decomposed/etc now?

tsalo:
Thanks @dowdlelt. Updated the output list and everything's passing now.

handwerkerd:
I've come across something in this PR that may be worth discussion (and also raises a separate issue #534). I ran this PR locally to look at the results more closely. In processing the 5 echo data. It is very clear that more voxels are completely excluded from the mask when the adaptive mask is used. I think this makes sense. You're using a different thresholding test and some voxels that survived the previous threshold have a 3rd echo that doesn't survive the new threshold test. The lost voxels are all in orbito-frontal regions where dropout is common.
It's possible (probable?) that it's better to exclude voxels with significant dropout in the 3rd echo and the chance is appropriate, but I think this is a breaking change that is worth acknowledgement & discussion.
Should we keep an option for existing functionality? If yes, then I think this PR should be expanded to including the attempts in #491 (which focuses on making masking a bit more centralized in the code and thus more cleanly parametrizable) so that everything can be merged cleanly into master together.
As an aside, I've been running tedana on these data locally instead of viewing the outputs created by pytest, which seem to be deleted after pytest completes. Is there a way save intermediate files when pytest is used to run a test?

handwerkerd:
I'm approving. Related to the comment above, consider expanding the docstring for adaptive_mask.
Since this merge will likely reduce the number of voxels within the mask for many data sets, the effect of this should probably be noted somewhere... not sure where.
I'm not sure if the second reviewer will be @emdupre or @jbteves but the one part of the code I still understand the least is testing so, if you're looking this over, focus there.

handwerkerd:
Just following up on this. Someone else have time to be the second review of this PR, with a bit of a focus on the testing code? @emdupre @jbteves @smoia @rmarkello

tedana/combine.py
```diff
@@ -158,23 +158,32 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
...
             msg = 'Optimally combining data with voxel-wise T2* estimates'
         else:
             msg = ('Optimally combining data with voxel- and volume-wise T2* '
                    'estimates')
-        t2s = t2s[mask, ..., np.newaxis]  # mask out empty voxels/samples
-
         LGR.info(msg)
-        combined = _combine_t2s(data, tes, t2s)
+
+    mask = adaptive_mask >= 3
```
eurunuela, line 172: For what I understand from this, there must be a minimum of 3 good echoes in order to keep a voxel. What happens with datasets that only have two echos?

tedana/metrics/kundu_fit.py
```diff
@@ -69,26 +70,23 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
     mmix_corrected : :obj:`numpy.ndarray`
         Mixing matrix after sign correction and resorting (if reindex is True).
     """
-    # Use t2s as mask
-    mask = t2s != 0
-    if not (catd.shape[0] == t2s.shape[0] == mask.shape[0] == tsoc.shape[0]):
+    # Use adaptive_mask as mask
+    mask = adaptive_mask >= 3
```
eurunuela, line 74: Same comment as above, what if the number of echos is 2?

tedana/combine.py
emdupre: We generally don't support those, as the method of usage is quite different in e.g. the Bright and Murphy method (where the first echo is generally treated as a regressor). Maybe we should open a separate issue to confirm if that's mentioned in the documentation, and to highlight it as a condition of tedana ?

tedana/combine.py
```diff
@@ -158,23 +158,32 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
...
+    data = data[mask, :, :]  # mask out empty voxels/samples
```
emdupre, line 173: I know we've had some confusion on this in the past, so I'm wondering if we can head it off with the code comment :sweat_smile: 

By `empty` you mean not having signal across at least three echos, right ? Maybe we could call them `unstable` instead of `empty` ? Not a huge point, just want to make sure this is clear. Maybe we could actually update this in the docstring for `make_optcom` itself ?

tedana/decomposition/pca.py
```diff
@@ -65,8 +65,8 @@ def tedpca(data_cat, data_oc, combmode, mask, t2s, t2sG,
         Poser 2006
     mask : (S,) array_like
         Boolean mask array
-    t2s : (S,) array_like
-        Map of voxel-wise T2* estimates.
+    adaptive_mask : (S,) array_like
+        Adaptive mask
```
emdupre, line 69:
```suggestion
        Adaptive mask of the data indicating those voxels for which at least three echos are available. If
        this is less than the total number of collected echos, we assume that later (echos do not provide
        meaningful signal
```

jbteves:
PS I'm happy to also review this if it's ready, but not sure if more reviews would be more useful yet.

tsalo:
@emdupre I think I've addressed your comments.

@eurunuela Handling of two-echo data is, I believe, mentioned in passing at some point deep in the documentation, but it's not really well laid out. We should definitely document that better.

eurunuela:
As long as we better document the two echo situation, I’m happy with the changes. Thanks @tsalo!

tedana/combine.py
```diff
@@ -107,8 +107,11 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
         Concatenated BOLD data.
     tes : (E,) :obj:`numpy.ndarray`
         Array of TEs, in seconds.
-    mask : (S,) :obj:`numpy.ndarray`
-        Brain mask in 3D array.
+    adaptive_mask : (S,) :obj:`numpy.ndarray`
+        Adaptive mask of the data indicating those voxels for which at least
```
emdupre, line 111: Since the thresholding for >= 3 voxels occurs internally, I'd pare down this docstring:

```suggestion
                Adaptive mask of the data indicating the number of echos with signal at each voxel
```

And move the admonition re: later echos into the function description !

tedana/combine.py
```diff
@@ -97,7 +97,7 @@ def _combine_paid(data, tes):
     return combined
 
 
-def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
+def make_optcom(data, tes, adaptive_mask, t2s=None, combmode='t2s', verbose=True):
     """
     Optimally combine BOLD data across TEs.
```
emdupre, line 102:
```suggestion
    Optimally combine BOLD data across TEs, using only those echos with reliable signal
    across at least three echos. If the number of echos providing reliable signal is greater
    than three but less than the total number of collected echos, we assume that later
    echos do not provided meaningful signal.
```

tedana/combine.py
```diff
@@ -107,8 +107,11 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
         Concatenated BOLD data.
     tes : (E,) :obj:`numpy.ndarray`
         Array of TEs, in seconds.
-    mask : (S,) :obj:`numpy.ndarray`
-        Brain mask in 3D array.
+    adaptive_mask : (S,) :obj:`numpy.ndarray`
+                Adaptive mask of the data indicating the number of echos with signal at each voxel
+        three echos are available. If this is less than the total number of
+        collected echos, we assume that later echos do not provided
+        meaningful signal.
```
emdupre, line 117:
```suggestion
        Adaptive mask of the data indicating the number of echos with signal at each voxel
```

PR #360, opened 2019-07-11
[NO-MRG, ENH] Replace signal-noise_t with signal-noise_z

### LINKED ISSUE #274: Revisit signal > noise t-value metric
### Summary
The [signal-noise_t metric](https://github.com/ME-ICA/tedana/blob/1e148376634c426faf8c4223e5e1ddae71229b90/tedana/selection/select_comps.py#L150) is used to build the decision table score metric (i.e., the sum of several ranked metrics), which involves ranking t-values across components. The t-tests used to get these values are component-wise comparisons of the _unique_ F-statistics from "signal" (i.e., cluster) and "noise" (i.e., non-cluster) voxels, after log-transformation. There are a couple of elements to this metric that I think need to be rethought, including:
- [ ] T-values will be dependent on degrees of freedom (i.e., numbers of significant voxels for each component), so we should z-transform them before we use them as a metric.
- [ ] I don't think there's any reason to only use unique F-statistics in the t-test. We can just drop the `np.unique` call at that stage.
- [ ] I can't find any basis online for log-transforming F-statistics to make them normally distributed. It seems to generally work, but I think that the [Fisher's z-distribution](https://en.wikipedia.org/wiki/Fisher%27s_z-distribution) (i.e., `0.5 * np.log(f)` instead of `np.log10(f)`) makes more theoretical sense. The results are pretty similar, but at least there's a Wikipedia page for the z-distribution.

### Next Steps
1. [ ] Discuss my above suggestions and agree on which ones (if any) should be implemented.
2. [ ] Implement the chosen changes. This should be fairly easy, since all are done in just one or two lines.

tsalo:
Closes #274.

- Add `t_to_z` (new function) to stats submodule
- Use `t_to_z` to convert t-statistics into z-statistics, to account for variable degrees of freedom across components
    - Due to the large number of voxels per t-test, this will not have a large effect on most components' scores
- Use all F-statistics in t-test, instead of just unique ones
- Use [Fisher’s z-distribution](https://en.wikipedia.org/wiki/Fisher%27s_z-distribution) (`0.5 * log(F)`) instead of `log10(F)` for conversion of F-statistics to normal distributions
- Add `countsignal` metric. It isn't currently used for anything, but it's very easy to calculate and it may be helpful to compare to `countnoise` (a metric that already exists and is used)


emdupre:
Can we add the writing to `--debug` here, too ? Or do you think that's a separate PR ?

_Or_, should this wait until we have a clearer system to validate these sorts of changes ?

tsalo:
I've added the debug info.

tsalo:
@emdupre is this okay to merge?

tedana/metrics/kundu_fit.py
```diff
@@ -362,36 +363,45 @@ def kundu_metrics(comptable, metric_maps):
...
-        # NOTE: Why only compare distributions of *unique* F-statistics?
-        noise_FR2_Z = np.log10(np.unique(F_R2_maps[comp_noise_sel, i_comp]))
-        signal_FR2_Z = np.log10(np.unique(
-            F_R2_maps[Z_clmaps[:, i_comp] == 1, i_comp]))
-        (comptable.loc[i_comp, 'signal-noise_t'],
-         comptable.loc[i_comp, 'signal-noise_p']) = stats.ttest_ind(
+        comptable.loc[i_comp, 'countnoise'] = int(np.sum(comp_noise_idx))
+        noise_FR2_Z = 0.5 * np.log(F_R2_maps[comp_noise_idx, i_comp])
+        signal_FR2_Z = 0.5 * np.log(F_R2_maps[Z_clmaps[:, i_comp], i_comp])
+        n_noise_dupes = noise_FR2_Z.size - np.unique(noise_FR2_Z).size
```
emdupre: `dupes` makes me think of being `duped`. Can we change to something like `dupl` ?

tedana/metrics/kundu_fit.py
```diff
@@ -362,36 +363,45 @@ def kundu_metrics(comptable, metric_maps):
...
-        comptable.loc[i_comp, 'countnoise'] = np.array(
-            comp_noise_sel, dtype=np.int).sum()
-        # NOTE: Why only compare distributions of *unique* F-statistics?
-        noise_FR2_Z = np.log10(np.unique(F_R2_maps[comp_noise_sel, i_comp]))
-        signal_FR2_Z = np.log10(np.unique(
-            F_R2_maps[Z_clmaps[:, i_comp] == 1, i_comp]))
-        (comptable.loc[i_comp, 'signal-noise_t'],
-         comptable.loc[i_comp, 'signal-noise_p']) = stats.ttest_ind(
+        comptable.loc[i_comp, 'countnoise'] = int(np.sum(comp_noise_idx))
+        noise_FR2_Z = 0.5 * np.log(F_R2_maps[comp_noise_idx, i_comp])
```
emdupre: Why are we changing this from `log10` to `log` ?

tedana/metrics/kundu_fit.py
```diff
@@ -362,36 +363,45 @@ def kundu_metrics(comptable, metric_maps):
...
+        signal_FR2_Z = 0.5 * np.log(F_R2_maps[Z_clmaps[:, i_comp], i_comp])
+        n_noise_dupes = noise_FR2_Z.size - np.unique(noise_FR2_Z).size
+        if n_noise_dupes:
+            LGR.debug('For component {}, {} duplicate noise F-values '
+                      'detected.'.format(i_comp, n_noise_dupes))
+
+        n_signal_dupes = signal_FR2_Z.size - np.unique(signal_FR2_Z).size
+        if n_signal_dupes:
```
emdupre: Can we combine these two logger statements ? Might be a little cleaner to read.

tedana/stats.py
```diff
@@ -150,3 +151,76 @@ def get_coeffs(data, X, mask=None, add_const=False):
...
+           description='Introduces T-to-Z transform.')
+@due.dcite(Doi('10.5281/zenodo.32508'),
+           description='Python implementation of T-to-Z transform.')
+def t_to_z(t_values, dof):
+    """
+    Convert t-values to z-values.
+
+    Parameters
+    ----------
+    t_values
```
emdupre: can we fill out these parameter descriptions, types ?

tedana/stats.py
```diff
@@ -150,3 +151,76 @@ def get_coeffs(data, X, mask=None, add_const=False):
...
+    dof
+
+    Returns
+    -------
+    out
+
+    Notes
+    -----
+    From Vanessa Sochat's TtoZ package.
```
emdupre: Can we make this an actual ref ?

tedana/metrics/kundu_fit.py
tsalo: This is [Fisher's z-distribution](https://en.wikipedia.org/wiki/Fisher%27s_z-distribution), which is the only F-to-normal transform I've seen. I don't know the rationale for the original transform.

emdupre: Hmm. We can ask Jen or someone if we track them down, but I'm fine with this !

tsalo:
Will add this as a separate metric as part of modularization.

tsalo:
This is implemented as an alternative metric in #436, so I'm happy to close this PR.

PR #298, opened 2019-05-25
[FIX] Add early escape from TEDICA decision tree

tsalo:
Closes #297.

Changes proposed in this pull request:
- Add early stopping procedure to TEDICA decision for when no potential BOLD components are identified in Step 1.

PR #365, opened 2019-07-17
[REF] Mean-center design matrix within getcoeffs

### LINKED ISSUE #179: get_coeffs run on mean-centered data for no reason

tsalo:
Closes #179. This is a minor refactor with a couple of diverse changes. It should not impact behavior _except_ for minuscule changes to PCA metrics. I believe that the changes actually reflect a bug fix, but they're so small that I don't think it's worth it to refer to this as a "fix" PR.

- Mean-center IV matrix within `getcoeffs`.
- Remove `add_const` argument from `getcoeffs`. Add constant automatically instead. This actually is not necessary, because mean-centering the IV matrix results in the same parameter estimates for some reason. 
- Remove unnecessary mean-centering of DV matrices prior to `getcoeffs` calls.
- Use a pandas DataFrame for filling in comptable within `dependence_metrics` from the beginning. We used to create 1D numpy arrays, fill those in, stack them, and then make the DataFrame from the resulting 2D array. This simplifies things.
- Drop an unreachable I006 classification from the TEDICA decision tree. This was a mistake on my part in #298. The mistake doesn't impact behavior at all, but makes the code less readable.
- Allow `getelbow` to handle empty input arrays. If an input array is empty, return NaN with a warning. For `getelbow` calls where results are averaged or `min`-ed, use `nanmean` or `nanmin` instead.


tsalo:
This PR sadly suffers from bloat, so I'm taking the core changes and putting them in a new PR. Closing this one now.

Issue #335, opened 2019-06-11
Add smoke tests for files that handle the first few steps of tedana pipeline


monicayao:
### Summary
As advised by Taylor, I should add smoke tests to the library. I will start by adding smoke tests to files that correlate to the beginning of the pipeline.

### Additional Detail
By smoke tests, there will be a test function that simply asserts that the function, when called on randomly generated inputs in the correct format, will output a non null value. It will be similar to the [kundu test](https://github.com/monicayao/tedana/blob/pr335/utils/tedana/tests/test_model_kundu_metrics.py).

I will add the smoke tests to the files `decay.py`, and `utils.py`.

These are the first few steps of the tedana pipeline. I had originally wanted to also add smoke tests to combine.py, but it seems unnecessary as the functions in `test_combine.py` are already smoke tests 👍 

### Next Steps

- [ ] Add smoke test to utils.py first
- [x] Add smoke test to decay.py

jbteves:
@monicayao you can see an example of `fit_decay_ts` [here](https://github.com/ME-ICA/tedana/blob/ba29b45a962ce77af5b4e8e95015301088b79edd/tedana/workflows/t2smap.py#L193). 

Re: searching. There is probably a better way to do this with VSCode or some other IDE, but on command line you can do something like the following to quickly find references to a function or related documentation:
`grep -R -A a -B b -n search_term directory_to_search`
with `search_term` whatever your search term is, like a function name, `directory_name` whatever the directory you want to search in is (in this case, you want `/path/to/repo/tedana`), `a` the number of lines after a search hit, `b` the number of lines after a search hit. The flag `-R` means to recursively search all subdirectories and files, the flag `-n` means to show line numbers. For more info on grepping, you can check `man grep`. You will be surprised at how quickly it will find them all; my own check was basically instant. 
Again, VSCode and the fancy non-command-line IDES presumably have something easier than listening to a command line fanboy like myself; I just don't know how they work.

jbteves:
Oh, and there's already a proposal in #250 to move masking into its own module. Not quite what you're proposing but I think quite similar.

PR #367, opened 2019-07-19
[TST] New smoke tests for functions in decay.py 

monicayao:
Addresses part of #335

- Added smoke tests for `fit_decay` and `fit_decay_ts` 
- Both tests are fashioned in the style of `test_smoke_kundu_metrics` from [test_model_kundu_metrics.py](https://github.com/ME-ICA/tedana/blob/master/tedana/tests/test_model_kundu_metrics.py)
- Question for the new tests - should I check the size of the outputs, or is that outside the scope of a smoke test?


monicayao:
> Generally very good. Small technical point that I’d like corrected: the numbers generated are ints, not bools. In Python they can be casted to bool implicitly but the type of the matrix is still int.

@jbteves ah yes! I was aware of this but since 0 and 1s have the same effect as false and trues I thought it would be fine to do this and add a comment to explain it. 

Since it is called from the line `data = data[mask]` in `decay.py`, the effects are the same. Should I still change this?

jbteves:
Oh, I just want the comment to not say bool, you don’t have to change the test. The most common descriptor is probably “binary mask.”

tsalo:
@monicayao Is there anything else you wanted to do here or is this ready to merge?

tsalo:
Can you pull from master to run the new testing setup?

monicayao:
@tsalo Hi! I think this is ready to merge. I do want to make the random inputs a pytest fixture eventually, but I think that can come after writing the more important test functions first. I updated my branch from master, and everything ran as before. 🎉

PR #368, opened 2019-07-19
[TST] Streamline CircleCI workflow

tsalo:
Closes None.

- Change Docker images to miniconda3 one.
- Cache environment based on checksum of requirements file (i.e., don't remake environment unless requirements change). This does not include the source code for tedana, which is installed fresh each time.
- Move TravisCI unit tests into CircleCI.
- Run coverage on unit tests _and_ integration tests, then merge those in a final job that should push to CodeCov!


tsalo:
I'm still stuck on how to push the coverage files from different jobs to a shared space before merging them and pushing them to CodeCov. If anyone has any ideas on how to do this, that would be extremely helpful.

tsalo:
@emdupre I think everything is ready. We just need to opt in to using orbs in our CircleCI settings. I don't have admin access there, but you should.

Here's the build error message:

```
#!/bin/sh -eo pipefail
# Orb codecov/codecov@1.0.5 not loaded. To use this orb, an organization admin must opt-in to using third party orbs in Organization Security settings.
# 
# -------
# Warning: This configuration was auto-generated to show you the message above.
# Don't rerun this job. Rerunning will have no effect.
false
Exited with code 1
```

tsalo:
The linting step is fixed now. I just had to fix how flake8 was installed and trigger the environment-building step by slightly changing requirements.txt.

The PR still expects Travis, even though I removed the travis file from the repo, and I'm not sure how to fix that. @emdupre is that a setting you can change?

Also, the CodeCov reports are uploading but not rendering, although I think that will resolve itself once we merge.

tsalo:
I've updated the list of required status checks, which will cause problems for other PRs until these changes are merged into them.

tsalo:
Merging now.

PR #371, opened 2019-07-23
[FIX, TST] Fix CodeCov report upload

tsalo:
Closes None. I thought that the CodeCov report generation problem from #368 would be resolved by merging the PR, but it wasn't. Sorry about that!


tsalo:
It seems to fixed now that I'm converting to xml format before uploading.

Issue #376, opened 2019-08-12
Add smoke tests for helper files


monicayao:
### Summary
Continuing from issue #335, this issue addresses the smoke tests for files that mainly "help" other files, and less directly involved in the workflow.

### Additional Detail
By smoke tests, there will be a test function that simply asserts that the function, when called on randomly generated inputs in the correct format, will output a non null value. It will be similar to the kundu test.

The files this issue addresses include `stats.py`, `io.py`, and `viz.py` for the functions that have an output. Functions which only generate a file or image are not tested. 

### Next Steps

- [x]  Add smoke test to `stats.py`
- [x]  Add smoke test to `io.py`
- [x] Add smoke test to `viz.py`

emdupre:
Amazing, thanks @monicayao !! So excited to see #377 #378 and #380 opened ! :sparkles: :rocket:

PR #378, opened 2019-08-14
[TST] Additional smoke tests for stats.py 

monicayao:
Addresses #376  .

- Additional smoke tests for the stats functions in three different files 
- A bug was discovered, in which get_coeff cannot take in a 3D data when the docstring for the function had claimed that it could


tsalo:
@emdupre @jbteves Does anyone have any idea how the coverage is _decreasing_ here?

jbteves:
@monicayao great catch on that bug!
@tsalo I'm pretty baffled too. If I'm reading the report correctly, it thinks that `test_decay` is no longer covering anything, and coverage is decreasing by 5 lines. But I don't see anything in the diffs for this PR that would cause that.

jbteves:
Okay I looked through all of the commits individually and the only thing I noticed is that at some point we added coverage for Python 3.6, but that coverage addition should be in both `master` and the PR if I understand how the bot works correctly. The only other thing I can think of is that the bot ran last time without the 3.6 coverage somehow and this time it added the 3.6 coverage for the PR but not for the last time it ran with master, but that still doesn't make a whole lot of sense...

emdupre:
Closing as a duplicate of #386. If that's not correct, @monicayao , please re-open !

PR #380, opened 2019-08-14
[TST] Add smoke tests to io.py and viz.py

### LINKED ISSUE #376: Add smoke tests for helper files

monicayao:
Closes #376 .

- Adds a smoke test for almost every function in io.py 
     - with the exception of write_results and write_results_echos, due to the nature of them not having an output and the files should be generated with no issue
     - of course, these could be written!

- Creates a new file, viz.py, which is only a smoke test for the trim_edge_zeros function as of now
    - This file does not have tests to test the different images generated yet (it could be written and similar to some of the tests in io.py, where we could check that the file is generated and delete it right after)


monicayao:
I'm so weirded out that this is failing and the data isn't loading in. Since my other code have a similar line, I've been checking out this issue

jbteves:
@monicayao can you paste in the "similar" code to a comment for us to compare against your PR, or link to the lines? If you didn't know, you can select a few lines and permalink them.

emdupre:
It looks like the failures come from this one line:

```python
ref_img = 'data/mask.nii.gz'
```

Could you try changing it to be a relative path to the actual file ? So it would be:

```python
ref_img = os.path.join('..', 'data', 'mask.nii.gz')
```

Issue #361, opened 2019-07-15
Tip in the doc to write the logs in a file


benoitberanger:
### Summary

To save the logs (outputs in the terminal), I use the same strategy described in `afni_proc.py` : 

`tedana [...] 2>&1 | tee $outdir/tedana.log` (if you use bash)

What do you think if this tip could be written in the *readthedocs* & the *README.md* ?

### Additional Detail

I'm trying `tedana` with different datasets based on different sequences, populations of subjects, activation tasks & resting-state. 

The complete log is written into this `tedana.log` file, and I can parse it latter to check, depending on the *tedpca* strategy chosen, the number of components selected (even if I can have it with `comp_table_pca.txt`), and more importantly if the ICA failed to converged after reaching the maximum iterations attempts.
Parsing `comp_table_ica.txt` gives me the number of accepted/ingored/rejected components.

On our computing cluster which uses *slurm* so it's possible to use an option to redirect stdout and stderr into a files for diagnostic. However I found it's more convenient to have a log of `tedana` in the $outdir.

tsalo:
You should be able to use `--debug` to get the log in a text file.

benoitberanger:
Oh I see ! And the option also write the timestamp, that's wonderful.
I didn't see the option, its not on the docs, the readme.md, and does not appear when you just check the help printed when you write `tedana` <enter>.

Is there a reason this option is a bit "hidden" ?

tsalo:
You're right. For some reason we're suppressing the help message for that option at the moment. I'm not sure why, but that's something we will definitely fix. @emdupre @jbteves do you have a preference for how we should change this? Should be just make `--debug` visible or should we separate log file generation from the logging level (DEBUG) set by `-debug`?

jbteves:
We should probably have a log file made by default and a verbose flag instead. Sorry I never followed up on this.

emdupre:
Commented in #385 -- and now I see where the discussion (that I said I missed) happened ! Too many GitHub notifications :sweat_smile:

jbteves:
Okay, I have a proposal that I think will address everybody's concerns surrounding the logfile:
1. We make a logfile appear on every run
1. We make a logfile, perhaps `tedana_YYYY-MM-DDTHH:MM:SS.log` (the ISO datetime format)
1. We IGNORE the logfile in circleci harvesting BUT
1. We create a testing module which verifies that running `tedana` generates the appropriate log information.
[...]

PR #385, opened 2019-08-21
[ENH] --debug flag appear now in the help & documentation

benoitberanger:
Hello,

This PR is a solution for #361, as discussed with @tsalo.
The changes are :
- make `--debug` visible in the help
- added some *help* documentation about this `--debug` option
- this `--debug` option is is now the default behavior

For now, I see 2 drawbacks that can easily be fixed :
1) changing the behavior so the option can be controlled : `--debug yes/no`
**and / or**
2) create a new option separate such as `--report` or `--log` and separate if from the `--debug` option

What do you think ?

Best,
Benoît

### DIFF
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -178,9 +178,11 @@ def _get_parser():
                           default=False)
     optional.add_argument('--debug',
                           dest='debug',
-                          help=argparse.SUPPRESS,
                           action='store_true',
-                          default=False)
+                          help=('Logs in the terminal will have increased '
+                                'verbosity, and also will also be written  '
+                                'into a .txt file in the output directory.'),
+                          default=True)
     optional.add_argument('--quiet',
                           dest='quiet',
                           help=argparse.SUPPRESS,
@@ -194,7 +196,7 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
                     tedort=False, gscontrol=None, tedpca='mle',
                     source_tes=-1, combmode='t2s', verbose=False, stabilize=False,
                     out_dir='.', fixed_seed=42, maxit=500, maxrestart=10,
-                    debug=False, quiet=False, png=False, png_cmap='coolwarm',
+                    debug=True, quiet=False, png=False, png_cmap='coolwarm',
                     low_mem=False):
     """
     Run the "canonical" TE-Dependent ANAlysis workflow.
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -180,9 +180,9 @@ def _get_parser():
                           dest='debug',
                           action='store_true',
                           help=('Logs in the terminal will have increased '
-                                'verbosity, and also will also be written  '
-                                'into a .txt file in the output directory.'),
-                          default=True)
+                                'verbosity, and will also be written into '
+                                'a .txt file in the output directory.'),
+                          default=False)
     optional.add_argument('--quiet',
                           dest='quiet',
                           help=argparse.SUPPRESS,
@@ -196,7 +196,7 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
                     tedort=False, gscontrol=None, tedpca='mle',
                     source_tes=-1, combmode='t2s', verbose=False, stabilize=False,
                     out_dir='.', fixed_seed=42, maxit=500, maxrestart=10,
-                    debug=True, quiet=False, png=False, png_cmap='coolwarm',
+                    debug=False, quiet=False, png=False, png_cmap='coolwarm',
                     low_mem=False):
     """
     Run the "canonical" TE-Dependent ANAlysis workflow.


emdupre:
Thanks so much, @benoitberanger ! 

I missed this conversation, so just to check in -- why are we changing this to be the default behavior ?

benoitberanger:
**Proposition**
Changing the default behavior of the log file is a proposition that I'd like to discuss with everyone.
Writing the log file takes only a few kB, and opening it is a simple way to check if the ICA converged of stalled after the maximum iterations is reached.

**Why ?**
Because when the workflow is done, I don't see how you can check, without the log file, if the ICA converged or not.

**Side notes**
In my usage of `tedana` I'm testing this workflow on several cohorts (controls, patients, different sequences, different pre-processing steps, ...), the procedure is performed on a computing cluster. I could save the logs of the jobs executed on the cluster, but that's messy. Saving the report.txt on the disk is a very comfortable way to save the info. Afterwards, I can parse the `comp_table_ica.txt` and `report.txt` to summarize and compare the number of volumes, number of accepted components after the decision tree, how many iterations for ICA, did it converge, how many accepted, ignored, rejected final components, etc.

emdupre:
Thanks for summarizing, @benoitberanger ! I think I understand the situation better, now.

The test failure is just because it's checking the outputs of the files, and it's not expecting the report.txt to be an output.

That being said, I'm not yet sure that this is something we should have on by default. For your specific situation: usually with e.g. SLURM schedulers, you should be able to request that the logs are written to a specific file. Or, could you just add the `--debug` flag to your calls ?

For a more general case: I think it's useful to have different levels of verbosity, depending on your interest. Maybe this is too optimistic, but I'd like to think (hope, aspire to) that whether the ICA converged won't be something that everyone wants access to, as it's assumed to be true. If it's something we definitely want folks to review, maybe we should think about putting it into the new `report` that @tsalo created or as a disclaimer on the QC images.

What do you think ?

benoitberanger:
@emdupre SLURM is writing log files, but only in one directory, not next to the nifti files being processed, making the parsing very difficult. The `--debug` flag is the best solution, as @tsalo suggested in #361.

I do agree different levels of verbosity is more ergonomic, and an option to write down the log file by default is debatable.
My point is the `--debug` flag does **not** appear in the help with `tedana --help`, or the documentation https://tedana.readthedocs.io/en/latest/usage.html

What do you think ?

tsalo:
I agree that `debug` needs to be included in the documentation. However, I think it would be better to make writing out the report and log files the default (i.e., divorce them from any arguments) and to keep `debug` off by default. The files are small enough that space is not an issue.

Sorry to be so late in joining this discussion.

jbteves:
We should move discussion from this PR to the issue IMO.

emdupre:
> We should move discussion from this PR to the issue IMO.

Since work has started, I'd prefer to keep it linked to the PR. That way we can more easily comment on the changes :smile_cat: 

> My point is the --debug flag does not appear in the help with tedana --help, or the documentation https://tedana.readthedocs.io/en/latest/usage.html

Absolutely ! Yes, this part of the PR I love. My concern is here:

https://github.com/ME-ICA/tedana/pull/385/files#diff-10962f3ba21a7497257aefd948c6c750R199

I think we're all converging on:

> We should probably have a log file made by default and a verbose flag instead.

to quote @jbteves. So, that would be a bit more of an extensive change -- we could keep the `--debug` argument (off by default), and when it was called the log files would have _more_ information. But the log files should be written by default. This would also mean stepping through our logging right now and making sure everything is at the appropriate level we want (e.g., DEBUG).

That might be out of scope for this PR, in which case I think this one could just document that `--debug` exists. WDYT ?

benoitberanger:
OK I try to summarize :
- `report.txt` is written by default
- `tedana_run.txt` is written by default
- `--debug` flag is disabled by default, and when activated increase the verbosity of `tedana_run.txt`, using `logging.basicConfig(level=logging.DEBUG`

Does everyone agree ?

Side question : is everyone ok with "tedana_run.txt" ? I would like to change the name in something a bit more meaningful, such as tedana.log / tedana_log.txt / log.txt

jbteves:
I have no strong feelings about `tedana_run.txt` as the name; I chose the it fairly whimsically and there are no alternatives suggested in the original PR, so I doubt opinions are strong. You'll have to modify the name in the CircleCI config though.

tedana/workflows/tedana.py
```diff
@@ -178,9 +178,11 @@ def _get_parser():
                           default=False)
     optional.add_argument('--debug',
                           dest='debug',
-                          help=argparse.SUPPRESS,
                           action='store_true',
-                          default=False)
+                          help=('Logs in the terminal will have increased '
+                                'verbosity, and also will also be written  '
+                                'into a .txt file in the output directory.'),
+                          default=True)
```
jbteves: @emdupre @tsalo I thought that `store_true` args could only store `True`; in other words, a default of `True` makes the argument always true. Is that incorrect? In some other code I'm working on that was my experience, but maybe I was handling it incorrectly.

tsalo: Good catch. Yeah, debug's default should be set to False here and in the function.

jbteves: Ah, but then in this setup it won't get written because it requires `--debug` to execute. Okay, this PR will unfortunately need changes.

emdupre:
> Does everyone agree ?

Yes, that sounds exactly right to me ! Thanks @benoitberanger 

> Side question : is everyone ok with "tedana_run.txt" ? I would like to change the name in something a bit more meaningful, such as tedana.log / tedana_log.txt / log.txt

tedana.log sounds great. We could optionally timestamp it, too, so it'd look more like tedana-HHMMSS.log No strong preferences, just thinking out loud.

tsalo:
> tedana.log sounds great. We could optionally timestamp it, too, so it'd look more like tedana-HHMMSS.log No strong preferences, just thinking out loud.

I agree that tedana.log is good, but we originally had the timestamps and those made checking the outputs in the CI difficult (impossible maybe?), which is why we switched to a single name with the backup (`tedana_run_old.txt`) system currently used.

jbteves:
@emdupre I strongly oppose timestamps! It makes testing much more difficult since the filenames aren't static. You can put the timestamp at the top of the file instead.

emdupre:
> I agree that tedana.log is good, but we originally had the timestamps and those made checking the outputs in the CI difficult (impossible maybe?), which is why we switched to a single name with the backup (tedana_run_old.txt) system currently used.

> @emdupre I strongly oppose timestamps! It makes testing much more difficult since the filenames aren't static. You can put the timestamp at the top of the file instead.

Yes, I remember this now (thanks both for the reminder), but I wonder if the simpler solution isn't just to ignore them in testing. I think it's a much more natural pattern for users, and if they're written by default I think we should cater to users rather than developers.

jbteves:
@emdupre in what way do you think it's more natural? Timestamps only feel natural to me if the timing has some inherent meaning, for example if it marks an event that is time-sensitive, not a post-hoc data run. I would guess that non-power-users would find no timestamps more natural, and that power-users would be suited well enough by the `old` suffix, but maybe that's not correct. Also, sorry, I hadn't seen @tsalo's comment that made mine redundant.

emdupre:
> Timestamps only feel natural to me if the timing has some inherent meaning, for example if it marks an event that is time-sensitive, not a post-hoc data run. I would guess that non-power-users would find no timestamps more natural, and that power-users would be suited well enough by the old suffix, but maybe that's not correct. 

I think this is a really common pattern in software and I'm reticent to introduce a new one to users, especially since on our end it's straightforward to implement. Why I think it's a common pattern is if, for example, I'm submitting a job on a SLURM cluster I get output logs by default. These are each stamped with the unique job ID, so I can easily trace back to each subject as appropriate.

In the tedana context, a "new" and "old" qualifier won't work if, for example, I'm trying out all three of the PCA flags. it'd be more natural to give the time, so I can continue to experiment with parameters without looking at an "_old_old" file.

tedana/workflows/tedana.py
emdupre: As above, I'm ok if this PR is only to make the debug option visible in the documentation, and if we make the additional changes in a separate PR ! It's up to @benoitberanger how much to tackle, here.
But definitely agree that regardless, this and the other boolean will need to change back to False :+1:

jbteves:
That makes sense, I hadn't thought of that. I do have some programs that use the `old` appending instead, but I agree that on Linux environments it's quite uncommon. My only objection now is an emotional objection to reducing our test cases; I would feel reasonably cozy if we could still test for file existence. @tsalo can we check for something like `tedana_run_*`?

benoitberanger:
@jbteves @emdupre @tsalo thank you all for your comments.
I will correct the PR content about argument parsing behaving correctly (https://github.com/ME-ICA/tedana/pull/385#pullrequestreview-286150821), and rename the PR title to something more related to it's final objective : make `--debug` flag appear in the help & documentation.

- [x] `--debug` flag appear now in the help & documentation : main objective of this PR
- [x] `report.txt` is written by default : already done before, not covered in the PR
- [ ] Log file written by default is accepted, but it's name is still in debate

The log file written by default will be assessed in another PR, to keep development segmented.

jbteves:
That sounds awesome @benoitberanger, thank you!

tedana/workflows/tedana.py
benoitberanger: Corrected in https://github.com/ME-ICA/tedana/pull/385/commits/e9a5fb1e82d2efe71814f739e3bbcdaa42078ad7

emdupre:
This looks great, so I'm going to go ahead and merge. Thanks again, @benoitberanger !

emdupre:
@all-contributors please add @benoitberanger for code contributions !

PR #391, opened 2019-09-10
[FIX][ENH][TST] Adds datetime logfile and removes it from outputs, fixes stream handling

### LINKED ISSUE #361: Tip in the doc to write the logs in a file

jbteves:
Closes #361 per my [proposal](https://github.com/ME-ICA/tedana/issues/361#issuecomment-530128947) .

- Makes logfile name `tedana_<ISO>.txt`
- Removes logfile name from checked inputs
- WIP will add unit testing to `/tests/` to verify that logfiles are generated.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,6 +188,8 @@ jobs:
       - run:
           name: Checking outputs
           command: |
+            # sanity check
+            echo "Is there anybody out there?"
             # find file
             find /tmp/data/five-echo/TED.five-echo/* \
                 -exec basename {} \; > /tmp/data/five-echo/TED.five-echo/outputs.out
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,7 +160,7 @@ jobs:
             fi
             # verify non-log outputs match exactly
             find /tmp/data/three-echo/TED.three-echo/* \
-                -exec basename {} \; | sort > /tmp/data/three-echo/TED.three-echo/outputs.out
+                -exec basename {} \; | grep -v outputs.out | sort > /tmp/data/three-echo/TED.three-echo/outputs.out
             f3=/tmp/data/three-echo/TED.three-echo/outputs_nolog.txt
             cat $f2 | grep -v -E -e '^tedana_[12][0-9]{3}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}.txt$' > $f3
             diff $f1 $f3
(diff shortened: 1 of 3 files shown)


jbteves:
Update: this PR will now check outputs to make sure that there is a logfile with the appropriate ISO stamp (thanks @KirstieJane !)

jbteves:
Additional update: for some reason we at some point deleted the file handlers so that info streams would be written to the file, this is corrected in 9c4ec51

jbteves:
@tsalo can you tell why I'm getting this discrepancy [as shown in CI](https://circleci.com/gh/ME-ICA/tedana/4804?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)? I didn't change the verbosity of the test as far as I'm aware.

tsalo:
`f1` is set to `tedana_outputs_verbose.txt` for both. The test that doesn't have the `--verbose` flag shouldn't check against that list of outputs.

jbteves:
Cool, we'll see if that does it.

.circleci/config.yml
```diff
@@ -178,9 +195,27 @@ jobs:
       - run:
           name: Checking outputs
           command: |
+            # find file
+            find /tmp/data/five-echo/TED.five-echo/* \
+                -exec basename {} \; > /tmp/data/five-echo/TED.five-echo/outputs.out
+            # set filenames
+            f1=/tmp/src/tedana/.circleci/tedana_outputs_verbose.txt
+            f2=/tmp/data/five-echo/TED.five-echo/outputs.out
```
emdupre: Since this file is being generated in the CI, we'll need to remove it when we do the "exact" testing since it's not an expected output of tedana ! That's the source of the current error, I believe.

jbteves: Resolved in f1bc063

PR #393, opened 2019-09-11
fix: use first echo data in t2s optcomb for zero voxels

emdupre:
Closes None. 

In response to a bug report from @shotgunosine on NeuroStars, fMRIPrep expects that optimally combined data has no masking applied. Since voxels with no T2* value create a divide by zero error, these were traditionally excluded from the calculation. This fix uses the first echo's T2* value instead of zero as the filler value.

### DIFF
--- a/tedana/combine.py
+++ b/tedana/combine.py
@@ -12,7 +12,7 @@ LGR = logging.getLogger(__name__)
 @due.dcite(Doi('10.1002/(SICI)1522-2594(199907)42:1<87::AID-MRM13>3.0.CO;2-O'),
            description='T2* method of combining data across echoes using '
                        'monoexponential equation.')
-def _combine_t2s(data, tes, ft2s):
+def _combine_t2s(data, mask, tes, ft2s):
     """
     Combine data across echoes using weighted averaging according to voxel-
     (and sometimes volume-) wise estimates of T2*.
@@ -33,6 +33,8 @@ def _combine_t2s(data, tes, ft2s):
     """
     n_vols = data.shape[-1]
     alpha = tes * np.exp(-tes / ft2s)
+    mdata = data[mask, :, :]
+
     if alpha.ndim == 2:
         # Voxel-wise T2 estimates
         alpha = np.tile(alpha[:, :, np.newaxis], (1, 1, n_vols))
@@ -40,12 +42,14 @@ def _combine_t2s(data, tes, ft2s):
         # Voxel- and volume-wise T2 estimates
         # alpha is currently (S, T, E) but should be (S, E, T) like mdata
         alpha = np.swapaxes(alpha, 1, 2)
-
         # If all values across echos are 0, set to 1 to avoid
         # divide-by-zero errors
         ax0_idx, ax2_idx = np.where(np.all(alpha == 0, axis=1))
         alpha[ax0_idx, :, ax2_idx] = 1.
-    combined = np.average(data, axis=1, weights=alpha)
+    
+    combined = np.average(mdata, axis=1, weights=alpha)
+    combined = unmask(combined, mask)
+    combined[~mask] = data[~mask, 0, :]
     return combined
 
 
@@ -138,7 +142,6 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
         LGR.warning("Argument 't2s' is not required if 'combmode' is 'paid'. "
                     "'t2s' array will not be used.")
 
-    data = data[mask, :, :]  # mask out empty voxels/samples
     tes = np.array(tes)[np.newaxis, ...]  # (1 x E) array_like
 
     if combmode == 'paid':
@@ -151,10 +154,9 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
         else:
             msg = ('Optimally combining data with voxel- and volume-wise T2 '
                    'estimates')
-        t2s = t2s[mask, ..., np.newaxis]  # mask out empty voxels/samples
+        t2s = t2s[mask, ..., np.newaxis]  # add new axis
 
         LGR.info(msg)
-        combined = _combine_t2s(data, tes, t2s)
+        combined = _combine_t2s(data, mask, tes, t2s)
 
-    combined = unmask(combined, mask)
     return combined
--- a/tedana/combine.py
+++ b/tedana/combine.py
@@ -32,8 +32,10 @@ def _combine_t2s(data, mask, tes, ft2s):
         Data combined across echoes according to T2* estimates.
     """
     n_vols = data.shape[-1]
-    alpha = tes * np.exp(-tes / ft2s)
-    mdata = data[mask, :, :]
+    # mask out empty voxels for T2* weighting
+    mt2s = ft2s[mask, ...]
+    mdata = data[mask, ...]
+    alpha = tes * np.exp(-tes / mt2s)
 
     if alpha.ndim == 2:
         # Voxel-wise T2 estimates
@@ -46,7 +48,7 @@ def _combine_t2s(data, mask, tes, ft2s):
         # divide-by-zero errors
         ax0_idx, ax2_idx = np.where(np.all(alpha == 0, axis=1))
         alpha[ax0_idx, :, ax2_idx] = 1.
-    
+
     combined = np.average(mdata, axis=1, weights=alpha)
     combined = unmask(combined, mask)
     combined[~mask] = data[~mask, 0, :]
@@ -154,7 +156,7 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
         else:
             msg = ('Optimally combining data with voxel- and volume-wise T2 '
                    'estimates')
-        t2s = t2s[mask, ..., np.newaxis]  # add new axis
+        t2s = t2s[..., np.newaxis]  # add new axis
 
         LGR.info(msg)
         combined = _combine_t2s(data, mask, tes, t2s)
--- a/tedana/workflows/t2smap.py
+++ b/tedana/workflows/t2smap.py
@@ -181,7 +181,7 @@ def t2smap_workflow(data, tes, mask=None, fitmode='all', combmode='t2s',
         LGR.info('Computing adaptive mask')
     else:
         LGR.info('Using user-defined mask')
-    mask, masksum = utils.make_adaptive_mask(catd, getsum=True)
+    mask, masksum = utils.make_adaptive_mask(catd, mask=mask, getsum=True)
 
     LGR.info('Computing adaptive T2* map')
     if fitmode == 'all':


tedana/combine.py
```diff
@@ -151,10 +154,9 @@ def make_optcom(data, tes, mask, t2s=None, combmode='t2s', verbose=True):
         else:
             msg = ('Optimally combining data with voxel- and volume-wise T2 '
                    'estimates')
-        t2s = t2s[mask, ..., np.newaxis]  # mask out empty voxels/samples
+        t2s = t2s[mask, ..., np.newaxis]  # add new axis
```
tsalo: Should drop mask here I think.

tedana/combine.py
```diff
@@ -33,19 +33,23 @@ def _combine_t2s(data, tes, ft2s):
     """
     n_vols = data.shape[-1]
     alpha = tes * np.exp(-tes / ft2s)
```
tsalo: Should mask ft2s right before this I think.

emdupre:
OK, these changes don't seem to affect component selection, but they're creating odd boundaries in the optimally combined `ts_OC` files:

Screenshot from 2019-09-11 14-06-49
*Right is with this PR, left is without.*

I'd suggest this because the first echo (which we're using to fill empty voxels, here) will have values in the greatest number of voxels, but it will also have the highest contrast. I'm really not sure how nicely fMRIPrep will play with these kinds of images -- do you want to try pinning to this PR, @shotgunosine, and see how if it affects performance for the registration ?

jbteves:
@emdupre it looks like our unit tests will also need to be revised if we go this route.

dowdlelt:
Why the preference to use the first echo value? Could it be preferred to use some some interpolation of the T2* map - that way image contrast/borders are maintained?  Or would that get even worse in some areas? I

jbteves:
@dowdlelt if you did it that way wouldn't you get dropout in frontal areas? Or am I misunderstanding how this mask was generated?

tsalo:
> Why the preference to use the first echo value? Could it be preferred to use some some interpolation of the T2* map - that way image contrast/borders are maintained? Or would that get even worse in some areas?

@dowdlelt Could you clarify what you mean by "interpolation of the T2* map"? I don't think that the T2* map will have valid values for those voxels.

dowdlelt:
Probably I am misunderstanding something:
This is a case where the data have been optimally combined - but now that optimally combined data isn't "whole" - it has been masked because there are bad T2* values. Is that right? 

What is strange to me is that for the image provided the edges are sharp - what made those superior and anterior sections so cleanly have bad T2star values? Or is that just a fabricated example?

@tsalo I'm saying spatially interpolate the bad T2star values - if there is a hole (negative/zero value), then use the closest X neighbors (>0) to estimate it. I suppose this would need to take place within some brain mask (based on e1?), because we don't want to fill in background. 

Of course, this opens a whole new can of worms for what happens during PCA/ICA, so perhaps a terrible solution. Though I still get the feeling that I am just missing something here.

Shotgunosine:
I tried running that fix inside the 1.4.1 and 1.5.0 fmriprep docker containers and either I didn't successfully get the new code to run or the change didn't impact anything. Here's what I ran for 1.4.1:
```
docker run -ti --rm -v /EDB/MBDU/nielsond/multi_echo_testing/multiecho_masking_test/:/data:ro -v /EDB/MBDU/nielsond/multi_echo_testing/test_out:/out -v  /EDB/MBDU/nielsond/multi_echo_testing/fs_license:/license --entrypoint "/bin/bash" poldracklab/fmriprep:1.4.1 

cd /src; git clone https://github.com/ME-ICA/tedana; cd tedana/; git fetch origin pull/393/head:mask_fix; git checkout mask_fix; cd ..; pip install -e ./tedana/ ; cd /out; /usr/local/miniconda/bin/fmriprep /data /out/out participant --fs-license-file /license/license.txt --output-spaces MNI152NLin2009cAsym -w /out/wrk
```

and got basically the same output as last time:
Screen Shot 2019-09-24 at 8 03 16 AM

And for 1.5.0:
```
docker run -ti --rm -v /EDB/MBDU/nielsond/multi_echo_testing/multiecho_masking_test/:/data:ro -v /EDB/MBDU/nielsond/multi_echo_testing/test_out:/out -v  /EDB/MBDU/nielsond/multi_echo_testing/fs_license:/license --entrypoint "/bin/bash" poldracklab/fmriprep:1.5.0

cd /src; git clone https://github.com/ME-ICA/tedana; cd tedana/; git fetch origin pull/393/head:mask_fix; git checkout mask_fix; cd ..; pip install -e ./tedana/ ; cd /out; /usr/local/miniconda/bin/fmriprep /data /out/out_15 participant --fs-license-file /license/license.txt --output-spaces MNI152NLin2009cAsym -w /out/wrk_15
```
Screen Shot 2019-09-24 at 8 03 48 AM

Still looks like it's masked when it's not supposed to be and cutting off bits of brain. @emdupre @effigies I'm not sure 100% that what I did will substitute in the updated Tedana code, is there another way I should run this?

emdupre:
Could you pin this commit in the `requirements.txt` and then [re-build the docker image](https://fmriprep.readthedocs.io/en/stable/contributors.html#rebuilding-docker-image) ?

Shotgunosine:
It looks like fmriprep doesn't have a requirements.txt anymore.

Shotgunosine:
So I added:
`tedana @ git+https://github.com/emdupre/tedana.git@fix/t2smap-wf` to `setup.cfg`
 but now I'm getting a conflict when I'm trying to build the docker container:
```
Collecting tedana@ git+https://github.com/emdupre/tedana.git@fix/t2smap-wf from git+https://github.com/emdupre/tedana.git@fix/t2smap-wf (from fmriprep==0.0.0)
  Cloning https://github.com/emdupre/tedana.git (to revision fix/t2smap-wf) to /tmp/pip-install-5f3m943s/tedana
  Running command git clone -q https://github.com/emdupre/tedana.git /tmp/pip-install-5f3m943s/tedana
  Running command git checkout -b fix/t2smap-wf --track origin/fix/t2smap-wf
  error: Your local changes to the following files would be overwritten by checkout:
        docs/make.bat
  Please, commit your changes or stash them before you can switch branches.
  Aborting
ERROR: Command "git checkout -b fix/t2smap-wf --track origin/fix/t2smap-wf" failed with error code 1 in /tmp/pip-install-5f3m943s/tedana
```

jbteves:
@Shotgunosine you don't need the `-b` flag when using `--track`, so try getting rid of `-b fix/t2smap-wf`

emdupre:
> So I added:
tedana @ git+https://github.com/emdupre/tedana.git@fix/t2smap-wf to setup.cfg
but now I'm getting a conflict when I'm trying to build the docker container:

I assume you haven't made any changes to that file ? I've never seen a git error like that when building a docker container.

jbteves:
@emdupre I don't have much experience with `docker`, but is there a way he could insert `git status` for us to debug further?

Shotgunosine:
@jbteves The git command is being generated by the pip install processing the the setup.cfg file. I'm not adding that -b at all.
When I just run git clone -q https://github.com/emdupre/tedana.git.
It lists docs/make.bat as a modified file. Is there a stashed chance or something like that?

Shotgunosine:
My guess is it's related to this: https://confluence.atlassian.com/bitbucketserverkb/git-status-shows-as-files-modified-directly-after-clones-943969037.html

but I'm unsure how to resolve that inside the docker container.

emdupre:
Ooof. I'm also unsure. Are you on a Mac or Windows, @Shotgunosine ? 

Is this still the right dataset to test on, if so ? https://drive.google.com/file/d/1-FkTf06ln7JWElUdS4Nco8t1C4HaMSr6/view

Shotgunosine:
Both myself and the container are on linux. @emdupre can you commit to that branch from a linux box?

Shotgunosine:
Yep, that's still what I'm trying to test on.

emdupre:
> Both myself and the container are on linux. @emdupre can you commit to that branch from a linux box? 

I'm also on a linux box. Trying to replicate on just one run of that data....

emdupre:
OK, got the error. The fix is just to install from the tarball in `setup.cfg`:

`tedana @ https://github.com/emdupre/tedana/tarball/fix/t2smap-wf`

The build (`docker build -t fmriprep-t2s .`) completes successfully. I've then tried running it with 

```
docker run -it -v /home/emdupre/Desktop/dylan_data:/data:ro \
-v /home/emdupre/Desktop/tmp:/out \
-v /home/emdupre/Downloads:/license:ro \
fmriprep-t2s /data /out participant --t2s-coreg \
--fs-license-file /license/license.txt
```

and I'm getting a versioning error:

```
  File "/usr/local/miniconda/bin/fmriprep", line 10, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/cli/run.py", line 311, in main
    opts = get_parser().parse_args()
  File "/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/cli/run.py", line 49, in get_parser
    currentv = Version(__version__)
  File "/usr/local/miniconda/lib/python3.7/site-packages/packaging/version.py", line 221, in __init__
    raise InvalidVersion("Invalid version: '{0}'".format(version))
packaging.version.InvalidVersion: Invalid version: ''
```

This is still listed as a [recommended development process](https://fmriprep.readthedocs.io/en/stable/contributors.html#rebuilding-docker-image) on the fMRIPrep docs, so I'm not sure why it would be failing ? Would be curious if you see the same error, @Shotgunosine !

Shotgunosine:
I was able to get to at least begin running by passing in a `VERSION` variable in the docker build command:
`docker build --build-arg VERSION=1.5.1rc1 -t fmriprep_local .`

I'll let you know how it turns out.

Shotgunosine:
Running off of master with that change to the setup.py cfg failed at some point, but it ran long enough to produce the boldref and brain mask and, as you can see, the brain is till masked in this step and the brain mask is messed up. @emdupre did you get the same result?
Screen Shot 2019-09-26 at 11 53 07 AM

These are the two commands I ran:
```
docker build --build-arg VERSION=1.5.1rc1 -t fmriprep_local .
docker run -ti --rm -v /EDB/MBDU/nielsond/multi_echo_testing/multiecho_masking_test/:/data:ro -v /EDB/MBDU/nielsond/multi_echo_testing/test_out:/out -v  /EDB/MBDU/nielsond/multi_echo_testing/fs_license:/license  fmriprep_local /data /out participant --t2s-coreg --fs-license-file /license/license.txt
```

And here's my setup.cfg file for fmriprep:
```
install_requires =
    niworkflows ~= 0.10.3
    smriprep ~= 0.3.2
    templateflow ~= 0.4.1
    nibabel >=2.2.1
    indexed_gzip >=0.8.8
    nilearn !=0.5.0, !=0.5.1
    nipype >=1.2.2
    psutil >=5.4
    pybids ~= 0.9.3
    tedana @ https://github.com/emdupre/tedana/tarball/fix/t2smap-wf
    nitime
    numpy
    pandas
    pyyaml
    scikit-image
    sdcflows ~= 0.1.0
    statsmodels
test_requires =
    coverage
    codecov
    pytest
```

emdupre:
It's unfortunate this didn't work.... I'll go ahead and close this, since it seems it won't be the fix we need, but I'll follow-up on https://github.com/poldracklab/fmriprep/issues/1786

PR #408, opened 2019-10-15
[ENH] Add t2smap argument for optional recalculated T2* map

### LINKED ISSUE #407: Add ability to provide T2* map as input
### Summary
We should be able to provide T2* maps directly to the workflow, either as a prior or as a complete replacement for the map calculated within the workflow.

### Additional Detail
The ability to provide T2* maps should be beneficial in two cases (as far as I can see): (1) when the user has a quantitative T2* map from a separate sequence and (2) for applying a two-stage denoising model that incorporates echo-specific denoising. In the latter case, the workflow would be: (1) run tedana on the minimally preprocessed data, (2) apply echo-specific denoising (e.g., phase regression, aCompCor, tCompCor) to the minimally preprocessed data, and (3) re-run tedana workflow with estimated T2* map, ICA mixing matrix, and component table from Step 1.

### Next Steps
1. Add tedana workflow argument for T2* map.
2. Add CLI parameter for T2* map.
3. Check T2* map shape and affine against input data shapes and affines.
4. If the T2* map is provided, use that one for optimal combination instead of running `fit_decay` within the workflow.

tsalo:
Closes #407.

- Add `t2smap` to CLI and tedana workflow. Corresponds to an image file with voxel-wise T2* estimates.
- Allow rerunning *without* `manacc` (i.e., if `manacc` is not supplied, but `ctab` is, then use the classifications from `ctab`.
- Reorganize workflow arguments to group related params together.
- Move rerunning-related arguments into their own subgroup
- Incorporate manual component selection test into the five-echo integration test.


tedana/workflows/tedana.py
```diff
@@ -190,10 +171,38 @@ def _get_parser():
                           action='store_true',
                           default=False)
     parser._action_groups.append(optional)
+
+    rerungrp = parser.add_argument_group('arguments for rerunning the workflow')
+    rerungrp.add_argument('--t2smap',
+                          dest='t2smap',
+                          metavar='FILE',
+                          type=lambda x: is_valid_file(parser, x),
+                          help=('Precalculated T2* map'),
```
dowdlelt: Perhaps this should specify that the T2* map has the same voxel dimensions as the inputted multiecho data?

tsalo: Should be resolved now.

tsalo:
The issue has been resolved and everything appears to be running fine now.

tedana/workflows/tedana.py
```diff
@@ -387,56 +405,75 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
-        LGR.warning('Argument "ctab" requires argument "manacc".')
-        ctab = None
     elif manacc is not None and not mixm:
         LGR.warning('Argument "manacc" requires argument "mixm".')
         manacc = None
 
+    if t2smap is not None and op.isfile(t2smap):
+        t2smap = op.abspath(t2smap)
+        # Allow users to re-run on same folder
+        if t2smap != op.join(out_dir, 't2sv.nii'):
```
emdupre: We might want to update this to gzipped by default given the recent changes in file writing.

tedana/workflows/tedana.py
```diff
@@ -387,56 +405,75 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
+            shutil.copyfile(t2smap, op.join(out_dir, op.basename(t2smap)))
+    elif t2smap is not None:
+        raise IOError('Argument "t2smap" must be an existing file.')
+
     bp_str = ("TE-dependence analysis was performed on input data.")
-    if mask is None:
+    if mask and not t2smap:
+        # TODO: add affine check
+        LGR.info('Using user-defined mask')
+        bp_str += (" A user-defined mask was applied to the data.")
```
emdupre: This will need to be updated with #414

tedana/workflows/tedana.py
```diff
@@ -387,56 +405,75 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
+    elif t2smap and not mask:
+        LGR.info('Using user-defined T2* map to generate mask')
+        t2s = utils.load_image(t2smap)
+        t2sG = t2s.copy()
+        mask = (t2s != 0).astype(int)
+    elif t2smap and mask:
+        LGR.info('Using user-defined mask and T2* map to generate mask')
```
emdupre: Can we rephrase this ? It seems contradictory at first read-through.

tsalo: What about "*Combining* user-defined mask and T2* map to generate mask"?

tedana/workflows/tedana.py
tsalo: Should be done with 623d0452089f91a52d3d4b7ea7dfcf657463b59b.

dowdlelt:
I feel like I'm missing something here - is the `fittype` arg taken out in this pull request?

tsalo:
@dowdlelt I just did a bad job at resolving a merge conflict. Thank you for catching that.

dowdlelt:
Does that mean we are not testing curvefit? I thought it was in the tests, for sure, but I'm still wrapping my head around testing I guess.

tsalo:
The mistake I made was in the CLI, while the tests run the workflow function directly. We don't have any tests for the argument parser at the moment (we did for a while but then we switched to the Python function to make it easier to consolidate everything in pytest/coverage).

dowdlelt:
Thanks for clarifying things, that all makes sense. Back to the topic at hand, I'll try and give this an updated review here shortly.

tedana/workflows/tedana.py
```diff
@@ -410,47 +424,65 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
-    if mask is None:
+    if mask and not t2smap:
+        # TODO: add affine check
+        LGR.info('Using user-defined mask')
+        RepLGR.info("A user-defined mask was applied to the data.")
+    elif t2smap and not mask:
+        LGR.info('Using user-defined T2* map to generate mask')
+        t2s_limited = utils.load_image(t2smap)
+        t2s_full = t2s_limited.copy()
+        mask = (t2s_limited != 0).astype(int)
```
dowdlelt: Are we not concerned about some ugly mask being generated if there are >0 values outside of the brain? Just trying to get a better idea of the use case. 

Perhaps it should ask users to inspect the output for sanity -  I can just imagine that...say someone puts a better t2star map into their EPI space, and interpolation or somesuch creates nonzero ringing - may be invisible when they view it but could generate very ugly mask. 

Maybe not a concern, thoughts? Perhaps line 432/463 deals with this.

jbteves: I agree it's problematic, but I feel like checking your own mask is implicitly assumed. Making something and not checking it is kind of on the user IMO.

jbteves:
Weird, RTD is failing...

jbteves:
I wonder if this is a bug on the CI end, this feature is just in beta.

jbteves:
I think this is a CI issue, see here: https://github.com/pypa/pip/issues/ 7217 (added space to avoid cross-linking on their issue).

PR #409, opened 2019-10-15
[ENH] Adding monoexponential curve fit

### LINKED ISSUE #315: Concerns about log-linear vs weighted log-linear vs direct monoexponential decay model fit
### Summary
From my understanding of the code, which may very well be wrong, we are performing a simple log-linear fit. While this is a very easy computation, I am concerned that it is somewhat incorrect, particularly in areas that drop out quickly. It may be that we are underestimating T2* and weighting the earlier echoes too heavily.  

### Additional Detail
For example, if you look at the thread here:
https://stackoverflow.com/questions/3433486/how-to-do-exponential-and-logarithmic-curve-fitting-in-python-i-found-only-poly
You note that fitting the exponential curve and using weights with log linear perform nearly identical, while log-linear fit with no weights comes up with a curve that drops very quickly. 

I have done a bit of testing (have to get back to computer - still at scanner) that suggest this is true for our data, even in an eight echo test. While I'm working on getting those figures together, I wanted to start the discussion. 

Note that this is seperate from the bi-exponential question in #212 , as the question in this issue is the accuracy of the (approximation) of the mono-exponential fit. 

### Next Steps
- [ ] Discuss - anyone else seen/thought about this?
- [ ] Add in a flag to perform slightly more computationally demanding curve/weighted fit

dowdlelt:
Adding in monoexponential fits to tedana as an alternative method rather than using a log linear estimate to calculate T2* and S0. This does not change requirements (ie, 3 echoes are still sufficient) but may offer more accurate estimates, particularly as the number of echoes increases. This method is slower, certainly, but isn't that much in the grand scheme of running tedana. Results are likely to be different, but limited testing shows that overall the differences are small. 

Closes #315  .

- Add a new monoexponential fit method to calculate T2* and S0
- Uses 'loglin' estimate as starting point
- Method is added for both T2smap and tedana


tedana/decay.py
```diff
@@ -71,14 +89,18 @@ def fit_decay(data, tes, mask, masksum):
...
         n_samp, n_echos = data.shape
         n_vols = 1
 
     data = data[mask]
+    fit_data = fit_data[mask]
     t2ss = np.zeros([n_samp, n_echos - 1])
     s0vs = np.zeros([n_samp, n_echos - 1])
 
+    
+
```
jbteves: This is nitpicky but could you delete the extra space?

tedana/decay.py
```diff
@@ -97,6 +119,30 @@ def fit_decay(data, tes, mask, masksum):
...
+        s0cf = np.zeros([t2s.size,])
+
+        for voxel in range(0, t2s.size):
+            try:
+                popt, cov = scipy.optimize.curve_fit(mono_exp, tes, fit_data[voxel, :],
+                                                     p0=(s0[voxel],
+                                                     t2s[voxel]))
+                s0cf[voxel] = popt[0]
+                t2scf[voxel] = popt[1]
+            except RuntimeError:
```
jbteves: Could you add a comment explaining what might trigger the RuntimeError?

dowdlelt:
Looks like the bulk of these commits are from keeping up with the master branch - lemme know if that is an issue. 

One improvement I have already thought of is that `bounds` could be set for scipy.optimize.curve_fit - right now everything is taking longer because it is unbounded. It currently uses the loglin estimate as a starting point, `p0`, but perhaps setting the bounds to plus/minus an order of magnitude of `p0` would be substantially better than -Inf Int, which is what I assume it currently is.

tedana/decay.py
```diff
@@ -2,10 +2,26 @@
...
+    s0 : :obj:`float`
+        Initial signal parameter
+    t2star : :oj:`float`
+        T2* parameter  
+
+    """
+    return s0 * np.exp(-tes/t2star)
 
-def fit_decay(data, tes, mask, masksum):
+def fit_decay(data, tes, mask, masksum, fittype):
```
jbteves: I think to fix the unit test failures you'll need to change it as follows:
```suggestion
def fit_decay(data, tes, mask, masksum, fittype='loglin'):
```

tedana/decay.py
dowdlelt: begone whitespace!

tedana/decay.py
dowdlelt: Done - sometimes this fails to converge - and that will crash out - added a comment

tedana/decay.py
dowdlelt: Should I do that, or change the tests themselves? I'm still quite vague/confused on the subject of testing so trying to get a feel for best practices...

jbteves: IMO we should use default parameters for this addition, since adding a new parameter is a breaking change for `fit_decay` and we're going to recommend log-linear by default. @emdupre and @tsalo may have differing opinions on this though.

dowdlelt: I just added in the defaults to the tests - and that worked nicely. After I fixed the commas (twice...).

dowdlelt: added the default to testing - this seems  to have resolved the issues.

dowdlelt:
I'm going to wait for @tsalo, since he did just request this pull this morning - so should get his input and make sure its gonna behave as he thought. @emdupre might want take a look as well - but thats not a formal request. 

Thanks for the help to get it ship shape @jbteves

tedana/decay.py
```diff
@@ -22,6 +40,8 @@ def fit_decay(data, tes, mask, masksum):
     masksum : (S,) array_like
         Valued array indicating number of echos that have sufficient signal in
         given sample
+    fittype : :obj: `str`
```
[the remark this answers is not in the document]
dowdlelt: Gotcha, changed.

tedana/decay.py
```diff
@@ -97,6 +119,29 @@ def fit_decay(data, tes, mask, masksum):
         t2ss[..., echo - 1] = np.squeeze(utils.unmask(t2s, mask))
         s0vs[..., echo - 1] = np.squeeze(utils.unmask(s0, mask))
 
+    if fittype == 'curvefit':
+        # perform a monoexponential fit of echo times against MR signal
+        # using the mean of the signal, using loglin estimates
+        # as initial starting points for fit
+
+        t2scf = np.zeros([t2s.size, ])
+        s0cf = np.zeros([t2s.size, ])
```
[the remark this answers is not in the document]
dowdlelt: done!

tedana/decay.py
```diff
@@ -97,6 +119,29 @@ def fit_decay(data, tes, mask, masksum):
...
+
+        for voxel in range(0, t2s.size):
```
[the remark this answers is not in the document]
dowdlelt: makes sense, got rid of it

tedana/tests/test_decay.py
```diff
@@ -70,10 +74,12 @@ def test_smoke_fit_decay():
     tes = np.random.random((n_echos)).tolist()
     mask = np.random.randint(2, size=n_samples) # generate binary mask of random 0s and 1s
     masksum = np.random.random((n_samples))
+    fittype='loglin'
```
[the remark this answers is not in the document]
dowdlelt: Added a smoke test for curvefit method for both smoke tests - guess we'll find out how long the time series approach takes to run. Going to commit and see how it goes

tedana/decay.py
```diff
@@ -97,6 +119,29 @@ def fit_decay(data, tes, mask, masksum):
...
+                                                t2s[voxel]))
+                s0cf[voxel] = popt[0]
+                t2scf[voxel] = popt[1]
+            except RuntimeError:
+                # If curve_fit fails to converge, fall back to loglinear estimate
+                s0cf[voxel] = s0[voxel]
+                t2scf[voxel] = t2s[voxel]
+
+        s0cf_unmask = np.squeeze(utils.unmask(s0cf, mask))
+        t2scf_unmask = np.squeeze(utils.unmask(t2scf, mask))
```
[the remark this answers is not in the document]
dowdlelt: I'll have to think about this a bit - may need an assist here, as I haven't thought about this enough

tedana/tests/test_decay.py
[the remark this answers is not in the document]
dowdlelt: looks like it worked - my first test!

tedana/decay.py
```diff
@@ -97,6 +119,29 @@ def fit_decay(data, tes, mask, masksum):
...
+        # perform a monoexponential fit of echo times against MR signal
+        # using the mean of the signal, using loglin estimates
+        # as initial starting points for fit
+
+        t2scf = np.zeros([t2s.size, ])
+        s0cf = np.zeros([t2s.size, ])
+
+        for voxel in range(0, t2s.size):
+            try:
+                popt = scipy.optimize.curve_fit(mono_exp, tes, fit_data[voxel, :],
```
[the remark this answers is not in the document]
dowdlelt: honest answer - a) no idea, b) not sure if curve fit will let me. I will find out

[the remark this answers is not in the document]
dowdlelt: I'm not seeing how to do it in looking through google and help docs - but there may be a compromise in using the `sigma` parameter https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html

I'll test a commit with the sigma parameter and see if it plays nicely. 

Though there may be a way to unravel the entire time series, and modify the monoexponential fit to work with that.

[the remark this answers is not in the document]
dowdlelt: Feel free to aggressively edit as you see fit. This was very much a hacked in, just fiddling with code solution that I tried to make nice enough to submit.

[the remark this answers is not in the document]
tsalo: I opened a PR (dowdlelt/tedana#8) that fits with what I was thinking. Please take or leave the suggested changes.

BTW it takes a _while_ to run on your eight-echo data, but I think it's reasonable in the long run.

tedana/decay.py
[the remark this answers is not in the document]
tsalo: Should be addressed in dowdlelt/tedana#8.

tedana/decay.py
```diff
@@ -97,6 +121,30 @@ def fit_decay(data, tes, mask, masksum):
...
+
+        for voxel in range(t2s.size):
+            try:
+                popt, cov = scipy.optimize.curve_fit(mono_exp, tes, fit_data[voxel, :],
+                                                     sigma=fit_sigma[voxel, :],
+                                                     p0=(s0[voxel],
+                                                     t2s[voxel]))
+                s0cf[voxel] = popt[0]
+                t2scf[voxel] = popt[1]
+            except RuntimeError:
```
tsalo: It would be nice if we had some kind of summary of how many voxels the curve-fitting method fail for.

dowdlelt: Agree - a counter would be easy enough, and maybe just logging it. If someone is using curvefit, showing the number of failing voxels(and percentage perhaps) for each loop shouldn't cause too much trouble.

dowdlelt: Added logging that reports the number and percentage of failed voxels per # of used echoes. Maybe this is overkill though, since it prints a line per loop...It is interesting, the number is much lower than I expected
`
INFO:tedana.decay:With 3 echoes, monoexponential fit failed on 32 (0.02%) voxel(s), used log linear estimate instead`
`INFO:tedana.decay:With 4 echoes, monoexponential fit failed on 29 (0.02%) voxel(s), used log linear estimate instead`

tsalo: Perhaps it's better suited as a debug-level message. It's definitely interesting (and informative) to see, but most users don't need that info. It's great for developers to have though!

dowdlelt: seems like it that is working nicely

dowdlelt:
Lemme know what you think @tsalo , logging should only occur with debugging, and yes, it does take a while, but it isn't as bad as I thought it could be. Thanks for fixing it up!

tedana/decay.py
```diff
@@ -71,31 +94,62 @@ def fit_decay(data, tes, mask, masksum):
...
+                try:
+                    popt, cov = scipy.optimize.curve_fit(
+                        mono_exp, echo_times_1d, data_2d[:, voxel],
+                        p0=(s0[voxel], t2s[voxel]))
+                    s0[voxel] = popt[0]
+                    t2s[voxel] = popt[1]
+                except RuntimeError:
+                    # If curve_fit fails to converge, fall back to loglinear estimate
+                    fail_count += 1
+                    pass
```
tsalo: You can drop this now.

tedana/decay.py
```diff
@@ -71,31 +94,62 @@ def fit_decay(data, tes, mask, masksum):
...
+            if fail_count:
+                fail_percent = (fail_count / t2s.size * 100)
```
tsalo: `fail_percent = 100 * fail_count / t2s.size` is a bit easier to read IMO.

tedana/decay.py
```diff
@@ -71,31 +94,62 @@ def fit_decay(data, tes, mask, masksum):
...
+                LGR.debug('With {0} echoes, monoexponential fit failed on {1} ({2:.2f}%) voxel(s),'
+                          ' used log linear estimate instead'.format(echo_num + 1, fail_count,
```
tsalo: Just `echo_num`, not `echo_num + 1`.

dowdlelt: I thought +1, or is it actually fitting the monoexp with just 2 TEs? that seems problematic

tedana/decay.py
dowdlelt: agreed

tedana/decay.py
dowdlelt: Makes sense after a little more googling, thanks

tedana/decay.py
tsalo: It does fit with just 2 echoes for the first one, and voxels with only two good echoes are filled in the "full" T2* map (used only for optimal combination), but are left masked out in the "limited" one (used for denoising stuff).

dowdlelt: Gotcha - well thats all corrected now. Think the monoexponential option is ready for the prime time. Another thing to add to the documentation!

tedana/workflows/tedana.py
```diff
@@ -177,6 +177,17 @@ def _get_parser():
                                 'use of IncrementalPCA. May increase workflow '
                                 'duration.'),
                           default=False)
+    optional.add_argument('--fittype',
+                          dest='fittype',
+                          action='store',
+                          choices=['loglin', 'curvefit'],
+                          help='Desired Fitting Method'
```
tsalo: The strings in this help message need trailing spaces in order to render properly.

tsalo:
Okay, other than the help message, all I think we need now is for the title to be in sentence case and to have the appropriate prefix ([ENH]).

tedana/workflows/tedana.py
dowdlelt: spaces added across the help text.

dowdlelt:
Title changed, I'm ready to merge once you approve

PR #410, opened 2019-10-15
add option for external regressors

dowdlelt:
Closes # None!.

This creates an option to write out a text file that just contains only the timecourses of the rejected components (in addition to the other files).

Intended use is for ease of using the rejected components as additional nuisance components in a GLM - and making it a text file makes it the most flexible - it could be plopped into AFNI/SPM/FSL etc without any trouble. 

- Add an option to write out the rejected components as text file `--extreg`


emdupre:
Can we sync up with @afni-rickr to see how afni_proc expects this ? Just to make sure we're standardizing across the ecosystem :)

PR #411, opened 2019-10-19
[DOC] Reorganize RTD sidebar and add section to release drafter

tsalo:
Closes None, but addresses requests made in our most recent [developers call](https://docs.google.com/document/d/1BxS2MapLr3ZLnfVf3HWcJdsFNoHzAA1FaDDLs33jEBk/edit?pli=1#).

- Move FAQ and Support sections into their own pages and up a level in the sidebar.
- Add "Breaking Changes" section to release drafter.


tsalo:
The one other thing we wanted to do in the documentation is render the CHANGELOG, but I realized that we don't actually have a file with that information to render. @emdupre is it possible to automatically export the release info/drafter to a markdown or RST file stored in the repository? From the call, the most important thing is having the summary we draft in the RTD site, for less GitHub-savvy users.

emdupre:
> The one other thing we wanted to do in the documentation is render the CHANGELOG, but I realized that we don't actually have a file with that information to render.

We have the script [update_changes.sh](https://github.com/ME-ICA/tedana/blob/master/update_changes.sh) -- we could add running it to the [release guide](https://tedana.readthedocs.io/en/latest/contributing.html#release-checklist) and then render that in the RTD ?

.github/release-drafter.yml
```diff
@@ -1,7 +1,10 @@
 template: |
   ## Release Notes
 
-  
+
+  ## Breaking Changes
```
emdupre: Is an admin planning to sort through all of the changes and make this determination ? I think that will be harder to decide on than something like "fixes" vs "new features". Do we have recommended criteria, here ?

docs/index.rst
```diff
@@ -148,6 +148,8 @@ tedana is licensed under GNU Lesser General Public License version 2.1.
    usage
    approach
    outputs
+   support
```
emdupre: total nitpick, but could we flip the order, here ? I'd like folks to read the FAQ before posting on Neurostars :smile_cat:

.github/release-drafter.yml
jbteves: @emdupre we did not come up with a set of criteria, but the thought was that things like renaming the log file would go here, since it would break an exiting workflow.

docs/index.rst
tsalo: :+1:

PR #418, opened 2019-11-06
[ENH][TST] Overhauls CI completely

### LINKED ISSUE #412: Add Test to Check .BRIK/.HEAD compatibility
Add a test to check BRIK/HEAD compatibility.

jbteves:
Closes #412 by switching the three-echo test to use BRIK/HEAD pairs. 

- Adds integration test skipping via pytest command-line args
-


jbteves:
@rmarkello  I like most of the changes you made in 494bc63 but is there some reason you don't like using variable names to hold the paths? I figure that made things slightly more obvious to future developers.

rmarkello:
I figured since it was only getting used in one place the variable wasn't totally necessary, but you're right&mdash;it is more obvious for future users! I've modified it back with 796a0d2.

One thing I'm realizing that might be a bit of a pain with the current set-up: using the Docker image for testing means that if someone wants to add an integration test with new data, they would first have to add the data to the Dockerfile, merge that so the Dockerfile build gets triggered, and then open a new PR to update the actual integration tests to use that data that is now in the Docker image... Maybe we can discuss tomorrow some ways around that...

rmarkello:
Ugh, the failure is this stupid `cprint` command that's supposed to be nice and print things prettily. Since it's causing issues that's aren't immediately obvious to ameliorate I kind of vote we just kill it... It was a bit silly and isn't worth this much headache 😅

I'm signing off for the night but we can regroup tomorrow! :+1:

jbteves:
Fortunately it looks like you just need to give it `-T screen` to make it work, it just passed the style check, so I think we're okay. 

I think integration tests should be fairly rare, but we can put something more complete together in RTD to guide someone through that. I don't think it's a huge problem.

rmarkello:
Uh oh, we might have to get @emdupre to update the expected status checks&mdash;"py36_unit_tests_and_coverage" doesn't exist anymore... The "merge_coverage" build now relies on EVERYTHING preceding it, so perhaps that's the only one we should check to consider the CI as passing?

jbteves:
Another small thing to fix: we run the three-echo set twice because of our changes to the defaults. We really only need what's currently labeled as "five_echo_task"

rmarkello:
Hmm, but for timing purposes it might be nice to split the integration tests up to run in parallel, no? Maybe we can look into how exactly we might be able to do that here...

emdupre:
Removed the `py36_unit_tests_and_coverage` build and now only require that `merge_coverage` passes !

jbteves:
I agree. However, the real lesson here is that our integration tests are so long that it's a real bottleneck for testing. I've proposed using synthesized data before; I think we might consider a small set of fake images to do the integration tests. The main concern AFAIK is that the unit tests might pass but a workflow modification breaks the pipeline as a whole. I don't see how fake data wouldn't achieve the same thing as real data for this purpose.

jbteves:
Thanks @emdupre !

emdupre:
>  I've proposed using synthesized data before; I think we might consider a small set of fake images to do the integration tests. The main concern AFAIK is that the unit tests might pass but a workflow modification breaks the pipeline as a whole. I don't see how fake data wouldn't achieve the same thing as real data for this purpose.

What if we just sub-selected our existing data ? Maybe cut down the time series to e.g. the first 50 volumes ?

jbteves:
@emdupre that would also be something worth considering. I'm going to spend a little bit of time working on it this morning.

rmarkello:
@jbteves: I'm so sorry for the MASSIVE, *massive* overhaul I just pulled with the CircleCI build, Dockerfile, and generally everything that we've been working on. To be fair, I _do_ think it's gonna end up being better in the long run, but it's quite a dramatic change from what we'd discussed.

My thought process: using the `tedana-dev` Docker image in the CircleCI build was really time-consuming when, as @emdupre pointed out, this is a pure-Python package that should be able to run things relatively quickly. Granted, the integration tests take a few minutes (though significantly less overall thanks to your cutting down the data!), but still not enough to warrant using a Docker image over just building the environments + caching them with Circle as @tsalo had previously been doing.

That said, I do think the `tedana-dev` docker container could be helpful for people who want to get set up really quickly, which is why I've kept it. And, happily, Docker Hub will continue to automatically build it for us! (I *do* think it might be worthwhile thinking about how to cache most of the integration test data inside the Docker container as we had been doing before; we can talk specifics tomorrow because I have some thoughts on sprucing up the integration test framework, generally, to make it easier to add new tests.)

Anyway, if you have any questions about these changes let me know and we can walk through them because this has turned into an unintentionally gigantic PR 😞

PR #424, opened 2019-11-07
[REF, DOC] Use logging to generate reports

### LINKED ISSUE #414: Incorporate report generation into logging
### Summary
We should be able to [add custom levels to our logging configuration](https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility), which should also make it possible to compile our citable report and associated references without manually compiling the strings within the workflow. 

### Additional Detail
We need to add two new levels to our logging configuration: REPORT and REFERENCES. Each level should have its own formatter, as the report is a paragraph, while the references are a list. We can then load both files, reduce/sort the references, combine the two strings, and write out the full file at the very end of the tedana workflow.

### Next Steps
1. Add new levels to logging configuration.
    - INFO number is 20. Make REPORT=21, REFS=22, & INFO will be what is currently INFO (i.e. the full log)
2. Design level-specific formatters.
    - Report should be stored as a paragraph.
    - References should be something that can be loaded as a list.
3. Move report/references strings from `tedana.workflows.tedana.py` into the appropriate functions, with the methods `LGR.report()` and `LGR.references()`.

tsalo:
Closes #414.

- Add loggers to collect report and references into separate files.
- Compile and clean up these files at the end of the workflow, to generate a publishable workflow description with citations.


tedana/workflows/parser_utils.py
```diff
@@ -12,3 +13,17 @@ def is_valid_file(parser, arg):
...
 
     return arg
+
+
+class ContextFilter(logging.Filter):
+    """
+    This is a filter which injects contextual information into the log.
+
+    Rather than use actual contextual information, we just use random
+    data in this demo.
```
jbteves: What's this demo included for? And is there any reason it's a class instead of a function?

PR #426, opened 2019-11-07
[ENH] Post/melodic mode

smoia:
Closes # .

- Created post_tedana workflow to create a folder that complies with fsleyes melodic mode
-


emdupre:
This will need updating, given our changes to the comptable and the mixing matrix in #456. It's also missing tests.

PR #427, opened 2019-11-08
[ENH] Make png default

jbteves:
Closes None

- Makes png generation default
- Adds `--nopng`
- Adds check to override `--nopng` if `--png-cmap` is supplied


tedana/workflows/tedana.py
```diff
@@ -261,11 +262,11 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
         which is slightly slower but may be more accurate.
     verbose : :obj:`bool`, optional
         Generate intermediate and additional files. Default is False.
-    png : obj:'bool', optional
-        Generate simple plots and figures. Default is false.
+    nopng : obj:'bool', optional
```
emdupre: I'm not sure it's worth exposing this -- it'd make sense to me to generate figures by default. WDYT ?

jbteves:
@emdupre Can you give this a look?

tedana/workflows/tedana.py
jbteves: I think this is fixed in f6cba2e

emdupre:
Total nit-pick: could we change `nopng` to `no_png` ? :smile:

PR #431, opened 2019-11-08
[FIX] Prevents shell expansion in API

jbteves:
Closes None

- Prevents shell-expansion in Python, which is allowed bi `nilearn` but concatenates the data in time, which is incorrect. 
-


jbteves:
Would also fix the five-echo integration test, which currently has t-catted data by accident.

PR #432, opened 2019-11-08
[ENH] Adds a scree plot showing metric values and variance sorted by Kappa. 

dowdlelt:
Closes # None.

Purpose of this pull request is to generate a figure that shows the kappa and rho metric values (organized by kappa) as well as the variance explained by each component. This isn't a major figure, but it was included in some original evaluations. It also may highlight interesting things about variance and kappa values. 

As an example, consider that component # 9 or so has very high variance (its pretty much a linear drift term) - but note that it also has kappa > rho.  

- Add a new figure showing the 'kappa elbow'

image


emdupre:
@dowdlelt , @javiergcas is working on a little report that includes very similar scatterplots -- can you add a screenshot here, Javier ? :smile_cat:

dowdlelt:
I saw some chatter on mattermost about that - excited to see it! Would be far more useful as an interactive figure, after all....

javiergcas:
Here is a screenshot of what we are building. It includes 4 plots (from left to right):

* Kappa vs. Rho Scatter
* Ranked Kappa
* Ranked Rho
* Ranked Variance explained.

All these plots are linked, so that when you click on a component in one plot, the same gets selected in the other three. There is also hovering capabilities, and soon when you click you will also get the corresponing timeseries, fft, and viewer for the component.
Example

dowdlelt:
its beautiful 😍, the images of my dreams even with the color coding!

eurunuela:
I believe this PR does not make sense anymore, now that we are working on the dynamic reports. Should we close this PR?

tsalo:
Are scree plots included in the current version of the dynamic reports? I agree that this PR should be closed, but we should make sure that the figures still end up being created.

dowdlelt:
I believe they are - and while I'm excited to have dynamic reports - I also thought the intention was to have these as fallback/optional figures. For that reason I think it would be good for them to be at parity with the dynamic reports.

jbteves:
@emdupre @dowdlelt @tsalo 
I believe we should include these as part of the static figures. Unfortunately there is now a merge conflict for the outputs; could you resolve it, @dowdlelt ? Happy to help if needed.

dowdlelt:
I have my notes, so maybe I can actually pull this off. I'll give it a shot

jbteves:
Awesome, thanks! If you have trouble just remember to reply here instead of Gitter since we're trying to push everything onto issues and PRs where possible.

dowdlelt:
I've adapted to the minimization  of the 'kitchen conversation' scheme better than I had hoped. And with any luck, I've also resolved the merge conflict - fingers crossed for testing.

jbteves:
I think you did it correctly : - )
That wasn't accusatory, BTW, it was just a reminder. To me just as much as you, and to everyone reading this.
It is a lot easier in that it's easy to find replies.

jbteves:
@dowdlelt since figures are now written by default you actually need to add the scree for all three integration tests, not just three-echo. Sorry, should have caught that when I looked over the PR.

tedana/viz.py
```diff
@@ -245,6 +245,41 @@ def write_kappa_scatter(comptable, out_dir):
...
+        each metric. Requires at least four columns: "classification",
+        "kappa", "rho", and "variance explained".
+    out_dir : :obj:`str`
+        Figures folder within output directory
+
+    """
+
+    fig, ax1 = plt.subplots(figsize=(10, 9))
+
+    ax1.plot(comptable.index.to_numpy(), comptable['variance explained'].to_numpy(),
```
tsalo: Is `to_numpy()` necessary here? I feel like matplotlib should be able to handle pandas.Series.

tedana/viz.py
```diff
@@ -245,6 +245,41 @@ def write_kappa_scatter(comptable, out_dir):
...
+    ax2.plot(comptable.index.to_numpy(), comptable.kappa.to_numpy(),
+             'b-', linewidth=2, label='Kappa')
+    ax2.plot(comptable.index.to_numpy(), comptable.rho.to_numpy(),
+             'r-', linewidth=2, label='Rho')
+    ax2.set_title('Kappa/Rho Metrics', fontsize=28)
+    ax1.set_xlabel('Component Number', fontsize=15)
+    ax2.set_ylabel('Metric Value', fontsize=15)
+    fig.legend(loc='upper right', bbox_to_anchor=(0.82, 0.78))
+    screefig_title = os.path.join(out_dir, 'Kappa_Rho_Scree_plot.png')
+    plt.savefig(screefig_title)
```
tsalo: Could you use `fit.savefig` instead? Mostly a personal preference, but it makes it more explicit.

tedana/viz.py
dowdlelt: lemme see if I can dig up the examples i used for testing. I feel that I did this for a reason, but I am not certain

dowdlelt: It may have been a lingering effect from the testing notebook, because I was loading comp_table directly. Let see what happens!

tedana/viz.py
dowdlelt: seems reasonable to me

dowdlelt:
Thanks folks, glad to close this out.

Issue #317, opened 2019-05-30
Calculation of variance explained appears wrong


tsalo:
### Summary
Variance explained and normalized variance explained are calculated within [dependence_metrics](https://github.com/me-ica/tedana/blob/65f89e1/tedana/model/fit.py#L21) for PCA and ICA components. However, variance explained is also directly returned by the PCA-fitting method, and the varex values from the original method are different from the ones calculated by our function.

### Additional Detail
Here is where we calculate variance explained and normalized variance explained in `dependence_metrics`:
https://github.com/ME-ICA/tedana/blob/65f89e1629bfddcf7f01a91ccb052440a6c22ea4/tedana/model/fit.py#L152-L154

The values are not very similar (for example, for PCA the "real" varex might be 37.8%, while the estimated varex is 57.6% and the estimated normalized varex is 53.8%), but they _are_ highly correlated across components.

jbteves:
Thanks for bringing this up. When you say that the variance explained is highly correlated across components, do you mean that the variance explained appears to at least grow with the amount of variance that a component is purported to explain?

tsalo:
I guess you could say that. The values differ by calculation method, but are correlated across methods.  I have high confidence in one method, but it only works for PCA, so I think we need to either come up with a more accurate method to implement in `dependence_metrics` or to figure out how to calculate variance explained for ICA specifically (i.e., in `tedica`) so we can use the "official" version for both decompositions.

tsalo:
We can calculate voxel-wise variance explained and then average it across voxels to get an estimate of component-wise variance explained. These values are very similar to the PCA-based variance explained values, and can be calculated for both PCA and ICA.

Here's what the code could look like in `dependence_metrics`:
```python
tsoc_dm = tsoc - np.mean(tsoc, axis=-1, keepdims=True)
totvar = np.var(tsoc_dm, axis=1)
...
LGR.info('Fitting TE- and S0-dependent models to components')
for i_comp in range(n_components):
    comp_pred_data = np.dot(mmix[:, i_comp:i_comp+1], tsoc_B[:, i_comp:i_comp+1].T).T
    compvar = np.var(comp_pred_data, axis=1)
    comptable.loc[i_comp, 'variance explained'] = np.mean(compvar / totvar)
```

[...]

PR #436, opened 2019-11-08

### LINKED ISSUE #320: Move component sign determination into a separate function
### Summary
Determination of the optimal signs for components and sorting components by Kappa are two steps performed within `dependence_metrics`. These two steps seem extraneous to the actual calculation of dependence metrics, and therefore I'd argue that they should be moved into their own functions.

### Update(s)
Component sorting can't be separated into its own function with the way things are currently set up, but sign determination is very easy.

### Additional Detail
Sign determination should be performed before any metrics are calculated, because, as mentioned in #318, signs of components matter for metric calculation. Moving this out of `dependence_metrics` would allow us to use the correctly signed components in any number of alternative metric functions.

On the other hand, component sorting can happen after metrics are calculated. I would argue that any metrics that rely on the order of the components should include their own internal sorting (e.g., as implemented in #295), so the order shouldn't matter. Instead, the order of the components should primarily matter for users, when they manually investigate the components (either in the component table or in the component figures). Component sorting can still happen before component selection if needed, but really it can happen any time before component maps, time series, or tables are written out.

### Next Steps
1. Create new function for component sign determination that takes in the mixing matrix and either beta maps or the optimally combined data, and returns the updated mixing matrix.
2. Create new function for component sorting that takes in the mixing matrix and component table and returns the sorted versions.
    - This may not even require its own function. It could just go into the workflow function.
3. Integrate functions into workflow.

tsalo:
Closes #320.

- Move individual metric calculations from one large-scale function into their own smaller functions.
- Add wrapper function that flexibly allow users (or more likely decision trees) to request any set of metrics and to get back a component table with those metrics.


tedana/metrics/collect.py
```diff
@@ -0,0 +1,140 @@
...
+    elif data_cat.shape[1] != len(tes):
+        raise ValueError('Second dimension of data_cat ({0}) does not match '
+                         'number of echoes provided (tes; '
+                         '{1})'.format(data_cat.shape[1], len(tes)))
+    elif not (data_cat.shape[2] == data_optcom.shape[1] == mixing.shape[0]):
+        raise ValueError('Number of volumes in data_cat ({0}), '
+                         'data_optcom ({1}), and mixing ({2}) do not '
+                         'match.'.format(data_cat.shape[2], data_optcom.shape[1], mixing.shape[0]))
+
+    mixing = mixing.copy()
```
rmarkello: I think this copy is unnecessary, here, since the next time it's used is providing `mixing` to `flip_components ()` which will trigger a copy.

tedana/metrics/collect.py
```diff
@@ -0,0 +1,140 @@
...
+from ._utils import determine_signs, flip_components, sort_df, apply_sort
+
+
+LGR = logging.getLogger(__name__)
+RepLGR = logging.getLogger('REPORT')
+RefLGR = logging.getLogger('REFERENCES')
+
+
+def generate_metrics(comptable, data_cat, data_optcom, mixing, mask, tes, ref_img, mixing_z=None,
+                     metrics=['kappa', 'rho'], sort_by='kappa', ascending=False):
```
rmarkello: I think it's generally best to not have a mutable object as a default parameter. Can we set `metrics` to `None` and then do a check in the function?

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+        Parameter estimates for optimally combined data against the mixing
+        matrix.
+    """
+    # compute un-normalized weight dataset (features)
+    weights = computefeats2(data_optcom, mixing_z, normalize=False)
+    return weights
+
+
+def calculate_betas(data_optcom, mixing):
+    assert data_optcom.shape[1] == mixing.shape[0]
```
rmarkello: Can we make this a `ValueError` with a brief message instead of an assertion error?

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+    # demean optimal combination
+    data_optcom_dm = data_optcom - data_optcom.mean(axis=-1, keepdims=True)
+    # compute PSC dataset - shouldn't have to refit data
+    betas = get_coeffs(data_optcom_dm, mixing)
+    return betas
+
+
+def calculate_psc(data_optcom, optcom_betas):
+    assert data_optcom.shape[1] == optcom_betas.shape[0]
```
rmarkello: See comment above re: `ValueError` instead of an assert.

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+def compute_countsigFS0(F_S0_clmaps):
+    countsigFS0 = F_S0_clmaps.sum(axis=0)
+    return countsigFS0
+
+
+def compute_countsignal(Z_clmaps):
+    countsignal = Z_clmaps.sum(axis=0)
+    return countsignal
```
rmarkello: These three functions here are all doing the same thing, but just operating on different arrays. Can we make one function (e.g., `compute_countsig()`) instead, so that we can reduce having to test/maintain all three?

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+def compute_countnoise(Z_maps, Z_cl_maps, z_thresh=1.95):
+    n_components = Z_maps.shape[1]
+    countnoise = np.zeros(n_components)
+    for i_comp in range(n_components):
+        # index voxels significantly loading on component but not from clusters
+        comp_noise_sel = ((np.abs(Z_maps[:, i_comp]) > z_thresh) &
+                          (Z_clmaps[:, i_comp] == 0))
+        countnoise[i_comp] = np.array(comp_noise_sel, dtype=np.int).sum()
```
rmarkello: I don't think this needs to be a for loop. From what I can tell this would broadcast without any issue:

```python
comp_noise_sel = (np.abs(Z_maps) > z_thresh) & (Z_clmaps == 0)
countnoise = np.sum(comp_noise_sel).sum(axis=0)
```

Summing the boolean should automatically cast it to int, too. Let me know if I'm misunderstanding the dimensions of `Z_maps` and `Z_clmaps` though!

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+            LGR.debug('For component {}, {} duplicate signal F-values '
+                      'detected.'.format(i_comp, n_signal_dupes))
+        dof = np.sum(comp_noise_idx) + countsignal - 2
+
+        t_value, signal_minus_noise_p[i_comp] = stats.ttest_ind(
+            signal_FT2_Z, noise_FT2_Z, equal_var=False)
+        signal_minus_noise_z[i_comp] = t_to_z(t_value, dof)
+
+    signal_minus_noise_z[np.isnan(signal_minus_noise_t)] = 0
+    signal_minus_noise_p[np.isnan(signal_minus_noise_p)] = 0
```
rmarkello: `np.nan_to_num` will do the equivalent here and might be more concise?

tedana/metrics/dependence.py
rmarkello: These three functions here are all doing the same thing, but just operating on different arrays. Can we make one function (e.g., `compute_countsig()`) instead, so that we can reduce having to test/maintain all three?

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+    d_table_score = d_table_rank.mean(axis=1)
+    return d_table_score
+
+
+def calculate_z_maps(weights, z_max=8):
+    n_components = weights.shape[1]
+    Z_maps = np.zeros(weights.shape)
+    for i_comp in range(n_components):
+        # compute weights as Z-values
+        weights_z = (weights[:, i_comp] - weights[:, i_comp].mean()) / weights[:, i_comp].std()
```
rmarkello: Why not `scipy.stats.zscore` instead? Also, should we be using `ddof=1` for the standard deviation calculation?

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+def calculate_varex(optcom_betas):
+    n_components = optcom_betas.shape[1]
+    totvar = (optcom_betas**2).sum()
+
+    varex = np.zeros(n_components)
+    for i_comp in range(n_components):
+        varex[i_comp] = 100 * (optcom_betas[:, i_comp]**2).sum() / totvar
+    return varex
```
rmarkello:
```python
compvar = (optcom_betas ** 2).sum(axis=0)
varex = 100 * (compvar / compvar.sum())
```

*should* return the same thing and avoid the for loop.

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+def calculate_varex_norm(weights):
+    n_components = weights.shape[1]
+    totvar_norm = (weights**2).sum()
+
+    varex_norm = np.zeros(n_components)
+    for i_comp in range(n_components):
+        varex_norm[i_comp] = (weights[:, i_comp]**2).sum() / totvar_norm
+    return varex_norm
```
rmarkello: This is the same as `calculate_varex()` but just doesn't multiply by 100? We should consolidate these functions.

tedana/metrics/collect.py
```diff
@@ -0,0 +1,140 @@
+"""
+Collect metrics.
+"""
+import logging
+
+import numpy as np
+import pandas as pd
+
+from .dependence import *
```
rmarkello:
```suggestion
from . import dependence
```

And then update the function calls accordingly. Not a huge fan of * imports within internal module, though that is a stylistic choice.

tedana/metrics/dependence.py
tsalo: Nice catch! This was originally part of a much larger for loop, so I just copied the loop for all of the smaller functions. This makes things much cleaner.

tedana/metrics/collect.py
```diff
@@ -0,0 +1,140 @@
...
+    Z_maps = calculate_z_maps(weights)
+    F_T2_maps, F_S0_maps = calculate_f_maps(mixing, data_cat, tes, Z_maps)
+
+    (Z_clmaps, F_T2_clmaps, F_S0_clmaps,
+     Br_T2_clmaps, Br_S0_clmaps) = spatial_cluster(
+        F_T2_maps, F_S0_maps, Z_maps, optcom_betas, mask, n_echos)
+
+    # Dependence metrics
+    if any([v in metrics for v in ['kappa', 'rho']]):
+        comptable = calculate_dependence_metrics(comptable, F_T2_maps, F_S0_maps, Z_maps)
```
rmarkello: Based on our discussions in-person, it sounds like we're thinking maybe we should aim for these functions to return arrays (_not_ `comptable`), so this function should be updated as with the ones below.

tedana/metrics/dependence.py
rmarkello: Also, the `comp_noise_sel` and `countnoise` calculation appear in this function and the following two functions. If we consolidate this (removing the for loop, as above), what do you think of explicitly calling this in `compute_signal_minus_noise_z()` and `compute_signal_minus_noise_t()`?

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,354 @@
...
+    signal_minus_noise_t[np.isnan(signal_minus_noise_t)] = 0
+    signal_minus_noise_p[np.isnan(signal_minus_noise_p)] = 0
+    return signal_minus_noise_t, signal_minus_noise_p
+
+
+def compute_dice(Br_clmaps, F_clmaps):
+    n_components = Br_clmaps.shape[1]
+    dice_values = np.zeros(n_components)
+    for i_comp in range(n_components):
+        dice_values[i_comp] = utils.dice(Br_clmaps[:, i_comp], F_clmaps[:, i_comp])
```
rmarkello: Any way we could make `utils.dice` broadcastable? I haven't looked at the function in a while but if it's possible it would be really nice!

tedana/metrics/dependence.py
tsalo: I'll clean it up like `calculate_varex()`, but I'd rather wait on merging until I believe in the variance explained calculations a bit more (per #317).

tedana/metrics/collect.py
rmarkello: But we *should* add a check for `if mixing_z is None` and assign `mixing` to `mixing_z`.

tedana/metrics/collect.py
tsalo: Done.

tedana/metrics/dependence.py
tsalo: Should be good now.

tedana/utils.py
```diff
@@ -181,12 +184,20 @@ def dice(arr1, arr2):
...
-    arr_sum = arr1.sum() + arr2.sum()
-    if arr_sum == 0:
-        dsi = 0
+    if axis is not None and axis > (arr1.ndim - 1):
+        raise ValueError('Axis provided {} not supported by the input arrays.'.format(axis))
+
+    if axis is None:
+        arr1 = arr1.flatten()
+        arr2 = arr2.flatten()
+        axis = 0
```
rmarkello: This check isn't necessary; if `axis` is `None` then any calls to `.sum(axis=None)` will implicitly flatten the array.

tedana/metrics/dependence.py
tsalo: I need the noise index in those two more than the counts, so I'll be duplicating most of the work there anyway.

tedana/metrics/dependence.py
tsalo: I don't think this step should even be done. `computefeats2` should return z-values (Fisher's R-to-Z performed on parameter estimates like they were Pearson r-values), but not z-statistics. To get from the Fisher's R-to-Z values to test statistics, we would then divide by sqrt(n-3), rather than z-scoring the resulting values. Although, again, that requires r-values going in and not parameter estimates from a _multiple_ regression.

tedana/metrics/dependence.py
tsalo: I'll add more informative checks (and tests) later on. I just wanted to throw a couple of checks in here as an example.

tedana/utils.py
tsalo: Thanks for catching.

tsalo:
Closing this WIP and reopening as a draft. @rmarkello I'm sorry to dismiss your review by doing this, but I think I've addressed almost everything you've brought up so far.

PR #443, opened 2019-11-10
[REF] Mean-center design matrix within getcoeffs

### LINKED ISSUE #179: get_coeffs run on mean-centered data for no reason

tsalo:
Closes #179. 

- Mean-center IV matrix within getcoeffs.
- Remove add_const argument from getcoeffs. Add constant automatically instead. This actually is not necessary, because mean-centering the IV matrix results in the same parameter estimates for some reason.
- Fix import in viz.
- Simplify existing get_coeffs calls.
- Update get_coeffs tests to match change to function.


tedana/stats.py
```diff
@@ -138,15 +138,19 @@ def get_coeffs(data, X, mask=None, add_const=False):
...
+    if X.shape[0] == 1:
         X = X.T
 
-    if add_const:  # add intercept, if specified
-        X = np.column_stack([X, np.ones((len(X), 1))])
+    # mean-center design matrix
+    X = X - np.mean(X, axis=0, keepdims=True)
+
+    # add intercept
+    X = np.column_stack([X, np.ones((len(X), 1))])
```
rmarkello: Remove.

tedana/stats.py
```diff
@@ -138,15 +138,19 @@ def get_coeffs(data, X, mask=None, add_const=False):
...
 
     betas = np.linalg.lstsq(X, mdata, rcond=None)[0].T
-    if add_const:  # drop beta for intercept, if specified
-        betas = betas[:, :-1]
+
+    # drop beta for intercept
+    betas = betas[:, :-1]
```
rmarkello: Remove.

tsalo:
@rmarkello I do have to say that it’s really weird to me that mean centering the IV array has the same effect as adding a constant, so that also might not be clear to folks reading the code. I added a comment about it to the docstring, but let me know if you think we need more.

PR #447, opened 2019-11-10

### LINKED ISSUE #274: Revisit signal > noise t-value metric

### LINKED ISSUE #320: Move component sign determination into a separate function

tsalo:
Closes #320.

- Move individual metric calculations from one large-scale function into their own smaller functions.
- Add wrapper function that flexibly allow users (or more likely decision trees) to request any set of metrics and to get back a component table with those metrics.
- Add signal-noise_z metric _in addition to_ signal-noise_t (closes #274).


tsalo:
The current failing point is managing masking within the metric calculations, but the general structure is looking good to me.

handwerkerd:
I'll try to look this over during the next few days. In the meantime, the one thing that jumps out to me is that there is a non-trivial decrease in code coverage that seems to be primarily driven by the addition of the t_to_z function in stats.py. Could @tsalo, @jbteves, or someone else add testing code for that function?

tsalo:
I can add the tests, as long as we're happy with the general structure.

tedana/stats.py
```diff
@@ -152,3 +153,76 @@ def get_coeffs(data, X, mask=None, add_const=False):
...
+    t_values
+    dof
+
+    Returns
+    -------
+    out
+
+    Notes
+    -----
+    From Vanessa Sochat's TtoZ package.
```
handwerkerd: Add the package link in this note and possibly also in the reference or report logger: https://github.com/vsoch/TtoZ

handwerkerd:
Going through this PR, it looks like you've added all the modularization functions, but they aren't actually called in workflows/tedana.py That means any testing & CI would confirm that your functions run, but they aren't directly comparing the outputs for the metrics in the new functions with what is currently in the code base. Am I reading this correctly? Have you done comparisons of the comptable outputs? On a practical level, I'd prefer to merge all this code so that everything is in master and then separately try to integrate this into a workflow (with the understanding that there might be modest breaking changes), but I want to confirm that this is your plan as well.

tsalo:
I temporarily ran the new metric calculations in parallel with the old to compare, and it turns out there were a couple of small (but very important) bugs in the new functions! With the new commit, those bugs should all be squashed. That said, there are minor differences in some of the spatial metrics for the following reasons:
- F-statistic maps are cropped at 500 when they're generated in the new version. This means that all functions using those maps are getting the cropped versions. In the current version, those maps are only cropped when kappa and rho are calculated, so other metrics using them don't use the cropped versions. This mostly impacts signal-noise T and P values.
- threshold_to_match isn't implemented in the current version, which means that the Dice values reflect comparisons between thresholded maps with different numbers of significant voxels. In the new version, there's a while loop to get those maps as similar in terms of number of voxels as possible, without taking too long.

I'm quite happy to merge first and investigate after, especially now that I've actually done some investigating now.

tedana/metrics/collect.py
```diff
@@ -0,0 +1,225 @@
...
+        'map beta T2 clusterized': ['map FT2 clusterized', 'map optcom betas',
+                                    'countsigFT2', 'mask', 'ref_img', 'tes'],
+        'map beta S0 clusterized': ['map FS0 clusterized', 'map optcom betas',
+                                    'countsigFS0', 'mask', 'ref_img', 'tes'],
+    }
+    # Apply masks before anything else
+    data_cat = data_cat[mask, ...]
+    data_optcom = data_optcom[mask, :]
+
+    required_metrics = dependency_resolver(METRIC_DEPENDENCIES, metrics, INPUTS)
```
handwerkerd: Just noting that required_metrics is unordered, but by having an ordered list of if statements for each metric calculation, they'll always be calculated after dependencies. This approach might get bloated if the metric list keeps growing, but this should be fine for now.

emdupre:
> On a practical level, I'd prefer to merge all this code so that everything is in master and then separately try to integrate this into a workflow (with the understanding that there might be modest breaking changes), but I want to confirm that this is your plan as well.

> I'm quite happy to merge first and investigate after, especially now that I've actually done some investigating now.

My concern here is that I know @smoia is quite anxious to cut a release, and I think we can get PCA and reporting in there on a reasonable time scale, but I think these metrics (and the corresponding decision tree) will take a bit longer. I might hold off on merging this for now. It gives more time to review, too, which is not a bad thing !

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,527 @@
...
+    Z_maps = stats.zscore(weights, axis=0)
+    extreme_idx = np.abs(Z_maps) > z_max
+    Z_maps[extreme_idx] = z_max * np.sign(Z_maps[extreme_idx])
+    return Z_maps
+
+
+def calculate_f_maps(data_cat, Z_maps, mixing, mask, tes, f_max=500):
+    """
+    Calculate pseudo-F-statistic maps (per component) for TE-dependence
+    and -indepdence models.
```
handwerkerd: indepENdence

handwerkerd: @emdupre I can go either way, but my inclination is to merge the disconnected metric modularization code. If we merge now, others can then already start to use those functions and it will make it slightly easier to test things (both general tests & CI testing) and do a clean swap in workflow/tedana.py when we're ready. For the decision tree part, it would also be nice to just have to sync with the master instead of both the master & @tsalo's fork. That said, if @tsalo & I need to tweak a bunch of things to get our parts working together, that might be easier in a fork than needing to have each edit subject to a PR & review.
What are @tsalo's thoughts?

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,527 @@
...
+
+
+def calculate_dependence_metrics(F_T2_maps, F_S0_maps, Z_maps):
+    """
+    Calculate Kappa and Rho metrics from F-statistic maps.
+    Just a weighted average over voxels.
+
+    Parameters
+    ----------
+    F_T2_maps, F_S0_maps : (S x C) array_like
```
handwerkerd: This does not appear correctly in the readthedocs API. It's: "F_S0_maps (F_T2_maps,)"

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,527 @@
...
+    maps and cropping extreme values.
+
+    Parameters
+    ----------
+    weights
+    z_max
+
+    Returns
+    -------
+    Z_maps
```
handwerkerd: Z_maps : (S x C) array_like
(This happens several places in the docs. When you don't type identify the returns, the "Returns" field is empty in the docs and "Return type" contains the return parameter name)

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,527 @@
...
+    data_cat
+    Z_maps
+    mixing
+    mask
+    tes
+    f_max
+
+    Returns
+    -------
+    F_S0_maps, F_T2_maps
```
handwerkerd: F_S0_maps, F_T2_maps : (S x C) array_like

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,527 @@
...
+    ----------
+    maps : (S x C)
+    mask
+    ref_img
+    threshold
+    csize
+
+    Returns
+    -------
+    maps_thresh
```
handwerkerd: maps_thresh : (S x C) array_like

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,527 @@
...
+    ----------
+    maps : (S x C)
+    n_sig_voxels
+    mask
+    ref_img
+    csize
+
+    Returns
+    -------
+    clmaps
```
handwerkerd: clmaps : (S x C) array_like

tsalo:
> @emdupre I can go either way, but my inclination is to merge the disconnected metric modularization code. If we merge now, others can then already start to use those functions and it will make it slightly easier to test things (both general tests & CI testing) and do a clean swap in workflow/tedana.py when we're ready. For the decision tree part, it would also be nice to just have to sync with the master instead of both the master & @tsalo's fork. That said, if @tsalo & I need to tweak a bunch of things to get our parts working together, that might be easier in a fork than needing to have each edit subject to a PR & review.
What are @tsalo's thoughts?

I'm fine waiting to merge into master until after `0.0.9`. We're almost ready for that anyway, so this PR can just stay open and the release can happen in the mean time. Once the release is made, we can do as @handwerkerd says and merge into master even before switching the workflow over to the new code.

handwerkerd:
I've visually reviewed the code and no concerns stand out. I'm not going to approve based on the discussion above. In the meantime, I'll work to connect the metric calculations in this PR to my decision tree modularization. Depending on how long that takes, we might be able to merge all that at once.

emdupre:
> @emdupre I can go either way, but my inclination is to merge the disconnected metric modularization code. If we merge now, others can then already start to use those functions and it will make it slightly easier to test things (both general tests & CI testing) and do a clean swap in workflow/tedana.py when we're ready. For the decision tree part, it would also be nice to just have to sync with the master instead of both the master & @tsalo's fork. That said, if @tsalo & I need to tweak a bunch of things to get our parts working together, that might be easier in a fork than needing to have each edit subject to a PR & review.

@handwerkerd, it's also a question of how these pieces fit together. My impression is that the metrics PR is not fully independent of the decision tree PR (which seems confirmed by the fact that you're encouraging to merge this in without connecting into the main workflow). It's not great to have a broken master, and especially if we're cutting a release soon I'm not comfortable releasing with a bunch of bloat in the code base (in terms of unintegrated functionality). 

The other option is that we hold off on releasing at all (including this PR, PCA and reporting) until the decision tree PR is also ready. Do you have a timeline for that ?

jbteves:
@emdupre and @handwerkerd : my 2 cents on the matter is that this has a lot of potential for breaking things without meaningfully impacting the user. We should spend some time reviewing and devising careful tests for this, while releasing the new PCA and the report since it has a direct impact on the user's quality of life. This is basically a fancy dev tool, so it's more useful in the long run and therefore not necessary to merge anytime soon.

handwerkerd:
"basically a fancy dev tool"?! 😭 💔 
Assuming, @tsalo is ok, it sounds like we have a plan. I've merged this PR with https://github.com/handwerkerd/tedana/tree/DecisionTreeModularization and I'll be working there.

jbteves:
That was a huge oversimplification @handwerkerd, it’s just that the change is (for now) invisible to users.

handwerkerd:
@tsalo and @emdupre Based on discussion on the dev call, the recommendation was to turn this PR to a branch of master. You'd be able to merge your metrics modularization work on the branch & I'll then submit a PR with the decision tree modularization to that branch. That way, we'll be able to edit and discuss the modularization work in one centralized place. Taylor, does this sound reasonable to you?

jbteves:
@handwerkerd I'm confused; this is a PR to master.

emdupre:
> @handwerkerd I'm confused; this is a PR to master.

Sorry to clarify: the proposal was that @tsalo would update the base branch for this pull-request, so that it's against a new branch (let's call it `fix/decision-metrics`). We could then review this PR to that branch and merge it in there.

@handwerkerd would open a new PR for his updated decision tree, also against the `fix/decision-metrics` branch (which would go through its own rounds of review). _Then_ there would be a third PR which merges these two lines of work together on the same `fix/decision-metrics` branch. After that is reviewed and merged, we could merge the `fix/decision-metrics` branch on to master.

WDYT ?

handwerkerd:
The key thing that took me a while to grok is that, in the current PR, there is one PR and one review of that PR. Thus it's going to be not fully reviewed until everything is done. As a branch, it will be possible to maintain, review, and merge multiple PRs to the code, thus allowing smaller, discrete reviews.
One other nice part of this approach is it would let us have two distinct metric calculation and decision tree pipelines to compare with each other. We can completely remove the old parts from the new branch & still do head-to-head comparisons of the modularized vs un-modularized code until we decide the modularized code is good enough to completely replace the non-modularized code.

emdupre:
@handwerkerd  -- I don't think we had a chance to go through a full review and iteration here so I'm reverting this merge. It'll be much easier to review three independent PRs in-depth than review one whole branch !

I think it'd be great if you can open your own, unlinked code as a PR against this same base branch. When we're happy with both we can worry about linking :smile_cat:

PR #454, opened 2019-11-11
[TST] Adds verbose CI

jbteves:
Closes None

- Makes tests verbose
-

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,3 +12,7 @@ exclude=*build/
 ignore = E126,E402,W504
 per-file-ignores =
     */__init__.py:F401
+
+[tool:pytest]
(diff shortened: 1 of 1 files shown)


jbteves:
Note: this is pretty verbose. If we set it to have INFO for logging level, it's much less like the terminal is screaming at you. However, this will obfuscate DEBUG level info for a DEBUG tedana run in pytest.

jbteves:
I'll leave the branch up in case we change our minds, but per a discussion with @rmarkello and then this morning with @tsalo I don't think we should do this. Verbose pytest is aggravating and impossible to parse, and you should be able to retrieve a logfile if you want one on a failed test.

PR #455, opened 2019-11-12
[DOC] Move notebooks from docs to examples

tsalo:
Closes None.

It would be nice to have some data on OSF (and to use the data grabber) for the processing pipeline notebook to make it more reproducible. I don't think @dowdlelt's eight-echo run can be shared publicly, but we also don't have any other datasets with nearly so many echoes, which I think is integral for illustrating echo dependence.

- Move our two example notebooks from `docs/_static` to new `examples/` folder.
- Add output directory variable to both, in order to keep writing the figures out to `docs/_static`


emdupre:
Can we convert these into .py scripts using Jupytext ? I'd really like to keep our repo free of .ipynb files ! I can explain this in person, too :)

jbteves:
Why no notebook files? You may need to explain in person. Granted, I am not very familiar with notebooks generally.

emdupre:
They're hard to git diff (without nb-diff, which is an additional developer dependency), and Binder can now directly handle Jupytext .py files -- converting them to .ipynb on the fly ! :rocket:

tsalo:
Scripts would kind of defeat the purpose of the walkthrough elements to the notebooks, unless we started actually running the scripts as Sphinx examples (which wouldn't be that computationally intensive, since the tedana derivatives should already be generated beforehand). That said, I have no strong opinions on this.

PR #456, opened 2019-11-12
[ENH] Make BIDS Derivatives-compatible decomposition files

### LINKED ISSUE #180: Output component tables as BIDS Derivatives-compatible jsons

tsalo:
Closes #180. See [here](https://docs.google.com/document/d/17ebopupQxuRwp7U7TFvS6BH03ALJOgGHufxK8ToAvyI/edit#heading=h.dbfk7w8jo2wp) for details on the specification for decomposition derivatives.

- Output component tables as jsons instead of tsvs.
- Output mixing matrices as tsvs instead of txts.
- Change names of several output files.


tedana/io.py
```diff
@@ -404,3 +406,91 @@ def load_data(data, n_echos=None):
     ref_img.header.set_sform(ref_img.header.get_sform(), code=1)
 
     return fdata, ref_img
+
+
+def add_decomp_prefix(name, prefix, max_value):
```
emdupre: Is this one supposed to be private, too ?

emdupre: If not, can we fill out the docstring ?

tsalo: It should be public. thanks for catching this.

emdupre:
@afni-rickr -- is AFNI ok if we pass back the decompositions as a JSON file ?

emdupre:
We're not able to parse the resulting JSON in most browsers since it contains NaN values. One solution might be to set `allow_nans=False` in `json.dump`.

@effigies , does BIDS functional derivatives yet have an opinion on what to do with NaN's in a JSON ?

afni-rickr:
@emdupre Sure, going with JSON seems quite reasonable.  (In AFNI) we will probably extract them back to 1D, but having all of the information in one place seems like a good idea.  Thanks!

effigies:
Could you describe more what you're trying to put into json? If it can be a 1d file, would it make more sense as a TSV?

tsalo:
The commit I just made should convert all NaNs in the DataFrame to "n/a" strings before dumping into the json, which I think follows BIDS convention for missing values.

@effigies We have three decomposition-related elements here- (1) a component table describing the components and associated metrics/classifications and which will be stored in a json, (2) component time series stored in a TSV, and (3) component maps, which are stored in a 4D nifti. I think that the specification requires that the component table be a json file.

PR #457, opened 2019-11-12
[DOC] multi-echo reports

emdupre:
Supersedes #432, #451.

- Initial commit of multi-echo reports (!), building on work from many folks
- Includes bokeh-powered interactive, linked figures for:
    1. Variance Explained,
    2. Kappa-Rho,
    3. Ranked Kappa, and
    4. Rho plots. 
- Interactively displays individual component level "static" figures
- Run report generation directly in workflow; therefore adds bokeh as a project dependency


emdupre:
I know this isn't a draft PR -- I'm sorry ! :cry:

jbteves:
@emdupre also looks like you have a merge conflict, you may want to merge in the latest changes. It's probably because of my reformatting spree.

emdupre:
@all-contributors please add @javiergcas for code, design :art:

emdupre:
I think in the most recent merge I re-introduced the "no-png" CLI option -- since reporting will depend on those, I'd say we should remove that option ! Unless folks want to have an option to toggle off reporting altogether.

emdupre:
@eurunuela , moving the conversation here ! I see two main things left on this PR:

1. Clean up the code (i.e., merge `html_report` and `generate_report`, update the docstring, and update the call in `tedana_workflow`)
2. Update the report layout. It might make sense to have all of the interactive components as one column with two rows (on the left side of the page ?) and have the selected static figure pop up on the right side of the page. I'm also wondering if we should have text that says something along the lines of "please select an individual component to view it in more detail" if no components have been selected.

Those should be quick, though, I hope ! And I think the tests might pass this time :crossed_fingers:

eurunuela:
Those should be quick, yes. I can take care of them. 

The tests did not pass though 😅

emdupre:
> The tests did not pass though :sweat_smile:

I know :joy: :sob: Will patch !

emdupre:
OK, I think that did it ! I also checked off the first item on the to-do list, but I forgot to add one as well :disappointed: 

- [ ] Fix the navbar by removing the unnecessary "Cute Dog" link and correctly linking "About" to the report.txt output. I imagine this would make sense beneath all of the figures, rather than in its own page !

jbteves:
I cannot believe removing a cute dog link is up for discussion on this PR. Surely we can move the cute dog link without erasing this 18/10 goodboi.

emdupre:
I think this is ready for review ! :tada: :sparkles:

eurunuela:
May I suggest that @javiergcas gives it a look too?

emdupre:
I think I requested reviews from @eurunuela @smoia @javiergcas and @dowdlelt ! But if that ping didn't go through, hopefully this one will 😸

eurunuela:
Oh sorry. I didn’t see that! I’m still getting used to the GitHub app on my phone.

Thanks @emdupre!

requirements.txt
```diff
@@ -7,3 +7,4 @@ pandas
 scikit-learn>=0.22
 scipy>=1.3.3
 threadpoolctl
+bokeh
```
dowdlelt, line 10: I hit an error on my first run, as my version of bokeh was apparently outdated, should this be set to specific version num? Updating from 1.3.4 (?!) to 2.0.1 fixed it. 
For completion sake, error:
```
  File "/Users/logan/opt/anaconda3/bin/tedana", line 11, in <module>
  File "/Users/logan/local_code/tedana/tedana/workflows/tedana.py", line 671, in _main
    tedana_workflow(**kwargs)
  File "/Users/logan/local_code/tedana/tedana/workflows/tedana.py", line 616, in tedana_workflow
    reporting.generate_report(out_dir=out_dir, tr=img_t_r)
  File "/Users/logan/local_code/tedana/tedana/reporting/html_report.py", line 82, in generate_report
    kappa_rho_plot = df._create_kr_plt(comptable_cds)
  File "/Users/logan/local_code/tedana/tedana/reporting/dynamic_figures.py", line 151, in _create_kr_plt
    legend_group='classif')
  File "fakesource", line 5, in circle
  File "/Users/logan/opt/anaconda3/lib/python3.7/site-packages/bokeh/plotting/helpers.py", line 857, in func
    glyph = _make_glyph(glyphclass, kwargs, glyph_ca)
  File "/Users/logan/opt/anaconda3/lib/python3.7/site-packages/bokeh/plotting/helpers.py", line 398, in _make_glyph
    return glyphclass(**kws)
  File "/Users/logan/opt/anaconda3/lib/python3.7/site-packages/bokeh/model.py", line 307, in __init__
    super(Model, self).__init__(**kwargs)
  File "/Users/logan/opt/anaconda3/lib/python3.7/site-packages/bokeh/core/has_props.py", line 253, in __init__
    setattr(self, name, value)
  File "/Users/logan/opt/anaconda3/lib/python3.7/site-packages/bokeh/core/has_props.py", line 288, in __setattr__
    (name, self.__class__.__name__, text, nice_join(matches)))
AttributeError: unexpected attribute 'legend_group' to Circle, possible attributes are angle, angle_units, fill_alpha, fill_color, js_event_callbacks, js_property_callbacks, line_alpha, line_cap, line_color, line_dash, line_dash_offset, line_join, line_width, name, radius, radius_dimension, radius_units, size, subscribed_events, tags, x or y
```

eurunuela: I think version 1.4 is the minimum required but I'm not completely sure right now. Let me check.

jbteves: Strange, on OS X I got a working version. Would be good to set a minimum version in the `.txt` file though.

jbteves:
This looks good but unfortunately I can't review the HTML. However, I wonder if we should rename `nopng` to `noreport`. We didn't have `nopng` in 0.0.8, so I don't see a reason to keep the option name.

requirements.txt
dowdlelt: I was using an anaconda install, from not so long ago (~months). As you can see, the error message is super useless - so I just had to assume it was a version issue.

emdupre: Interestingly, if we require Bokeh 2.0.0 then we have to drop python 3.5 support (since it requires python 3.6+). How would you feel about that @tsalo  ?

eurunuela:
Is it okay if we chat today and you show me? It will be easier to go through it if I can see what you have.

javiergcas:
Just checked the previous error I mentioned working with Eneko and it looks like a solution would be to add the following two lines to setup.py (right after line 32) to make sure that both data and data/html subdirectories (within reporting) get copied to the right location during install
            'reporting/data/*',
            'reporting/data/html/*',

emdupre:
Thanks, both ! Merged in https://github.com/emdupre/tedana/pull/24 :smile_cat: 

@javiergcas, do you know the minimum bokeh version to use the features included here ? I'm assuming it's 2.0.0, but would love to know if you're working with another version !

javiergcas:
> Thanks, both ! Merged in [emdupre#24](https://github.com/emdupre/tedana/pull/24) 😸
> 
> @javiergcas, do you know the minimum bokeh version to use the features included here ? I'm assuming it's 2.0.0, but would love to know if you're working with another version !

Right now, I am doing the testing with 2.0.1. That does not mean that the software may work ok with an earlier version. Do I need to check earlier versions?

eurunuela:
> Right now, I am doing the testing with 2.0.1. That does not mean that the software may work ok with an earlier version. Do I need to check earlier versions?

@dowdlelt had issues with Bokeh v1.3.4 and had to update to v2.0.1. I believe @emdupre was asking just in case you happen to know. 

I don't think it's necessary to check earlier versions. IMO we should require Bokeh v2.0.0 as the minimum.

emdupre:
> IMO we should require Bokeh 2.0.0 as the minimum

This will require us to drop Python 3.5 support and have several downstream effects. It's theoretically possible, but I just want to confirm it's something we really need for this functionality ! So yes, if you happen to know the history of Bokeh @javiergcas to know when the features you're using were introduced that would be great ! Otherwise we can do a quick test with the last Bokeh before 2.0.0. Just to make sure we know what we need, here

javiergcas:
> > IMO we should require Bokeh 2.0.0 as the minimum
> 
> This will require us to drop Python 3.5 support and have several downstream effects. It's theoretically possible, but I just want to confirm it's something we really need for this functionality ! So yes, if you happen to know the history of Bokeh @javiergcas to know when the features you're using were introduced that would be great ! Otherwise we can do a quick test with the last Bokeh before 2.0.0. Just to make sure we know what we need, here

I will try a few of earlier versions to see what happens and report, as for the python version, my current environment is Python 3.7 with bokeh 2.0.1 and I did not see incompatibility between these two. Is the need to go to Python 3.5 for something else?

emdupre:
> my current environment is Python 3.7 with bokeh 2.0.1 and I did not see incompatibility between these two. Is the need to go to Python 3.5 for something else?

Yes, exactly. The concern is that, since we currently officially support Python 3.5, if we require a Bokeh version that doesn't also support 3.5 then the whole package will lose 3.5 support. Which is ok on its own, but will impact our eg interaction with fMRIPrep which does support 3.5.

@dowdlelt identified an issue with Bokeh 1.3.4 (see: https://github.com/ME-ICA/tedana/pull/457#discussion_r411452607), which it looks like might be the last supported 3.5 version. I just wasn't sure if you were familiar with these older Bokeh versions -- enough to know if there's some "backwards compatible" Bokeh code we could use !

If not, we'll just have to keep in mind this impact.

jbteves:
I would push to move to the newer Bokeh in the near future and deprecate 3.5 support, given that it's at end of life in [September](https://devguide.python.org/#status-of-python-branches). I'm not sure how fMRIprep plans to handle that.

handwerkerd:
I'm seeing something that may be on my end, but may also be a bug. I ran tedana with this review code on a linux cluster. I can open up tedana_report.html on Firefox in a VNC window and the page renders correctly. I can also map those drives onto my Mac and open the report locally. In that case, the component maps on the right don't render. Screenshot attached. I've tried it using Safari, Chrome, and Firefox on my Mac. My guess is there is some relative call to a file name that isn't rendering correctly, but it's not obvious to me. I can keep looking into this on my end, but I figured I'd bring this up in case there is an issue with the code.

image

handwerkerd:
I also have one suggestion that either could be added to this PR or I can post as a new issue. ica_decomposition.json is a great improvement over the old comp_table.txt, but it's not easily human readable as a table of values. In various other versions of reports, there was a way to view that information as a table (ideally sortable by different keys). I think this is an important functionality. If it's relatively easy to render the json has an html table in this PR, I'd be very thankful if that could be added. I really don't want this PR held up so, I'm also fine if this PR was merged and I'll re-post this feature suggestion as a new issue to be addressed in a future PR.

jbteves:
Hm, did you get any javascript errors when you pulled it up?

handwerkerd:
I may have found the issue. When I click on a component, the console in Firefox on my Mac shows the output at the bottom of this comment. `/gpfs/gsfs12` is an internal representation of the base of my directory tree that's slightly specific to how I connect to the storage. (The universal formulation is `/data/`). What I think this means is that javascript is using the full path for the component images rather than a relative path based on the location of the opened tedana_report.html file. If I'm correct, I assume it's better to use a relative file path in this situation so that it would be possible to move a directory without breaking the dynamic report?

`[bokeh] document idle at 187 ms` 
`bokeh-2.0.1.min.js:157:1351`
`Linea: <span>Component Map<span>`

`bokeh-2.0.1.min.js line 346 > Function:28:17`

`Linea: <span>Component Map<span>`

`bokeh-2.0.1.min.js line 346 > Function:28:17>`

jbteves:
Yup, it should absolutely be a relative path IMO. There are a lot of users like you who will need to use clusters or servers to run `tedana`, and then view reports locally.

eurunuela:
Good catch @handwerkerd ! I never even thought of trying to move the folders around. 

I've got it fixed on a PR to @emdupre 's branch. Should be on this PR soon.

handwerkerd:
Any chance we can get two reviews of this so that it can be merged before OHBM?
@smoia @jbteves @javiergcas  @eurunuela @dowdlelt

eurunuela:
Not sure I should be a reviewer since I worked on it 😅

jbteves:
I can't review code but I can confirm if it works. WDYT @handwerkerd and @emdupre ? Should that be enough for a review?

eurunuela:
> Could put the legend in the "northeast" corner, where there's unlikely to be any data points? If you put the three colors into 3 rows it should fit with plenty of room between the legend and most data points.

The northeast corner would not work on the data I used. I guess we can ask Bokeh to find the "optimal" position for the legend. I know this automatically done by matplotlib.

The data I used was the 5 echo dataset we use for testing. That's probably why I get few components.

If the issue I see is only on my side, I'll approve the PR. We can work on tweaking the UI in another PR.

jbteves:
Hm, from your notebook on the OHBM poster thread your data diverges from what Dan and I experience, as well as the Kundu paper. I'm a little bit curious why your data seems to have interrelated Kappa-Rho increase. In any case, we could place the legend outside of the graph for a simple solution; the color scheme is universal across the report.

There does seem to be a small bug where data points disappear when you click on graph points on my Ubuntu machine with Firefox. I am remoted in to that machine, though, and sometimes VNC doesn't render properly. Could someone with access to an in-person Ubuntu machine see if they get the same behavior? I'm all for merging this anyway, just want to document it for future reference.

eurunuela:
> There does seem to be a small bug where data points disappear when you click on graph points on my Ubuntu machine with Firefox. I am remoted in to that machine, though, and sometimes VNC doesn't render properly. Could someone with access to an in-person Ubuntu machine see if they get the same behavior? I'm all for merging this anyway, just want to document it for future reference.

I believe @smoia has an Ubuntu machine.

handwerkerd:
If tedana is working correctly and brain fMRI data are the input, we shouldn't see dots in the top right corner, since that would mean specific components have both the highest kappa and rho values. As I mentioned before @eurunuela's data seem to have relatively low kappa and rho values in general. Are your data from a very short time series or a phantom? If so, I'd consider this a reasonable. If these are otherwise reasonable looking brain fMRI times series, then something weird is happening elsewhere in the program.

eurunuela:
@handwerkerd it is the 5 echo dataset tedana uses in the integration tests (see [here](https://github.com/ME-ICA/tedana/blob/d338acc57f9fde09c5f8dcc251937f00236d3f52/tedana/tests/test_integration.py#L87), which is a very small dataset, yes. I detrended the data and converted it to signal percentage change.

My guess is, this weird behavior comes from the fact that the dataset itself is very small. I'm in a meeting right now. I can check how long it is when I finish the meeting.

jbteves:
Oh, it's the truncated set? Yes, I recall that one had some strange statistics. It's not representative of what happens when you use the full set.

emdupre:
The last change that I know needs to happen here (outside of UX improvements), is the python 3.5 version check. I think this is as simple as:

```
import sys

if sys.version < 3.6.0:
    # create static figures only
```

I'll push it quickly

emdupre:
Also, sorry @jbteves, I had missed your question ! I'm on an ubuntu 18 machine running Firefox (quantum) and the points don't seem to disappear for me, though they do increase in opacity. But not so much that they render as transparent !

jbteves:
This sounds like it's probably because I'm remoted in, then. We can ignore it unless someone else reports the issue.

emdupre:
I'll leave it to @jbteves @handwerkerd or @dowdlelt (or anyone else who's interested !) to give a final review. Once we have 2 approving, in it goes :smile_cat:

emdupre:
Last ping on this -- if you can / want to review, please let me know ! Otherwise I'll merge this in later this week, and we'll iterate from there :smile_cat:

PR #458, opened 2019-11-13
[DOC] Improve documentation of computefeats2

CesarCaballeroGaudes:
Closes # .

- Adds some more comments explaining some steps to computfeats2
-

### DIFF
--- a/tedana/metrics/kundu_fit.py
+++ b/tedana/metrics/kundu_fit.py
@@ -107,7 +107,7 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
     WTS = computefeats2(tsoc, mmixN, mask=None, normalize=False)
 
     # compute PSC dataset - shouldn't have to refit data
-    tsoc_B = get_coeffs(tsoc_dm, mmix, mask=None)
+    tsoc_B = get_coeffs(tsoc_dm, mmix, mask=None, add_const=False)
     del tsoc_dm
     tsoc_Babs = np.abs(tsoc_B)
     PSC = tsoc_B / tsoc.mean(axis=-1, keepdims=True) * 100
--- a/tedana/stats.py
+++ b/tedana/stats.py
@@ -76,7 +76,7 @@ def computefeats2(data, mmix, mask=None, normalize=True):
         data = data[mask, ...]
     # normalize data (minus mean and divide by std) in the last dimension
     # so that least-squares estimates represent "approximate" correlation values (data_R)
-    # assuming mmix are also normalized
+    # assuming mixing matrix (mmix) values are also normalized
     data_vn = stats.zscore(data, axis=-1)
 
     # get betas of `data`~`mmix` and limit to range [-0.999, 0.999]
--- a/tedana/stats.py
+++ b/tedana/stats.py
@@ -74,7 +74,7 @@ def computefeats2(data, mmix, mask=None, normalize=True):
     # demean masked data
     if mask is not None:
         data = data[mask, ...]
-    # normalize data (minus mean and divide by std) in the last dimension
+    # normalize data (subtract mean and divide by standard deviation) in the last dimension
     # so that least-squares estimates represent "approximate" correlation values (data_R)
     # assuming mixing matrix (mmix) values are also normalized
     data_vn = stats.zscore(data, axis=-1)
--- a/tedana/stats.py
+++ b/tedana/stats.py
@@ -74,10 +74,15 @@ def computefeats2(data, mmix, mask=None, normalize=True):
     # demean masked data
     if mask is not None:
         data = data[mask, ...]
+    # normalize data (minus mean and divide by std) in the last dimension
+    # so that least-squares estimates represent correlation values (data_R)
+    # assuming mmix are also normalized
     data_vn = stats.zscore(data, axis=-1)
 
     # get betas of `data`~`mmix` and limit to range [-0.999, 0.999]
     data_R = get_coeffs(data_vn, mmix, mask=None)
+    # Avoid abs(data_R) = 1 or -1. 
+    # otherwise Fisher's transform will return Inf or -Inf  
     data_R[data_R < -0.999] = -0.999
     data_R[data_R > 0.999] = 0.999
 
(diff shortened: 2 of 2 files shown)


tedana/stats.py
```diff
@@ -74,10 +74,15 @@ def computefeats2(data, mmix, mask=None, normalize=True):
...
     if mask is not None:
         data = data[mask, ...]
+    # normalize data (minus mean and divide by std) in the last dimension
+    # so that least-squares estimates represent correlation values (data_R)
+    # assuming mmix are also normalized
     data_vn = stats.zscore(data, axis=-1)
 
     # get betas of `data`~`mmix` and limit to range [-0.999, 0.999]
     data_R = get_coeffs(data_vn, mmix, mask=None)
+    # Avoid abs(data_R) = 1 or -1.
```
tsalo: Because the standardized parameter estimates from a multiple regression are not the same as correlation coefficients and often go above or below what is possible with correlation coefficients, this step also avoids values > 1 or < -1.

tedana/stats.py
```diff
@@ -74,10 +74,15 @@ def computefeats2(data, mmix, mask=None, normalize=True):
     # demean masked data
     if mask is not None:
         data = data[mask, ...]
+    # normalize data (minus mean and divide by std) in the last dimension
+    # so that least-squares estimates represent correlation values (data_R)
```
tsalo: I'd say they "approximate" correlation values, because this conversion isn't technically valid as far as I can tell.

emdupre:
@javiergcas : in addition to considering @tsalo 's comments above, could you make sure we pass the linter ?

Thanks !

emdupre:
@all-contributors add @CesarCaballeroGaudes for documentation, code !

tedana/stats.py
```diff
@@ -74,10 +74,15 @@ def computefeats2(data, mmix, mask=None, normalize=True):
...
+    # assuming mmix are also normalized
```
tsalo: This is a bug, in my opinion. Would you mind adding something to normalize the mixing matrix here? I think that the following should work appropriately:

```python
mmix = stats.zscore(mmix, axis=0)
```

tsalo:
Also worth noting when it comes to computefeats2: the results seems to be treated as Fisher's R-to-Z-transformed z-values, but not at z-statistics. In order to "convert" them to z-statistics, we do the following:
https://github.com/ME-ICA/tedana/blob/be175c711749b86251803a8977eac4ffeddc39d5/tedana/metrics/kundu_fit.py#L178-L182

That is, we essentially z-score each component's z-value map separately, which I don't think is appropriate. The z-values (really just normally-distributed correlation coefficients) shouldn't necessarily have a mean of zero, so perhaps doing the variance normalization (minus mean-centering), as is done when `normalize=True` in `computefeats2` would be better there.

Moreover, since all of the component z-value maps have the same degrees of freedom, we should be able to normalize all of the maps at once using the same scaling factor.

CesarCaballeroGaudes:
I changed the comments following your suggestions and passed linter. Just doing it to close this PR. Some of these comments will disappear in the new version of stats.py that I am working on where the lines of the Fisher's transformation disappear. Therefore, the only relevant change is in kundu_fit.py

tsalo:
@CesarCaballeroGaudes Do you mind if I merge this?

CesarCaballeroGaudes:
Yes, it is.

tedana/metrics/kundu_fit.py
```diff
@@ -107,7 +107,7 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
     WTS = computefeats2(tsoc, mmixN, mask=None, normalize=False)
 
     # compute PSC dataset - shouldn't have to refit data
-    tsoc_B = get_coeffs(tsoc_dm, mmix, mask=None)
+    tsoc_B = get_coeffs(tsoc_dm, mmix, mask=None, add_const=False)
```
emdupre: Are we adding the last two arguments just for clarity ? We shouldn't need either of them, actually, since they're the defaults !

tedana/stats.py
```diff
@@ -74,10 +74,14 @@ def computefeats2(data, mmix, mask=None, normalize=True):
     # demean masked data
     if mask is not None:
         data = data[mask, ...]
+    # normalize data (minus mean and divide by std) in the last dimension
```
emdupre:
```suggestion
    # normalize data (subtract mean and divide by standard deviation) in the last dimension
```

tedana/stats.py
```diff
@@ -74,10 +74,14 @@ def computefeats2(data, mmix, mask=None, normalize=True):
...
+    # so that least-squares estimates represent "approximate" correlation values (data_R)
+    # assuming mmix are also normalized
```
emdupre:
```suggestion
    # assuming mixing matrix (mmix) values are also normalized
```

tedana/stats.py
```diff
@@ -86,9 +90,11 @@ def computefeats2(data, mmix, mask=None, normalize=True):
     if data_Z.ndim == 1:
         data_Z = np.atleast_2d(data_Z).T
 
-    # normalize data
+    # normalize data (only division by std)
```
emdupre:
```suggestion
    # normalize data (only dividing by standard deviation)
```

tedana/stats.py
```diff
@@ -86,9 +90,11 @@ def computefeats2(data, mmix, mask=None, normalize=True):
...
     if normalize:
+        # minus mean and divided by std
```
emdupre:
```suggestion
        # subtract mean and dividing by standard deviation
```

tedana/stats.py
CesarCaballeroGaudes: Agree

CesarCaballeroGaudes: Agree

KirstieJane:
Hi @CesarCaballeroGaudes! @emdupre made a few suggestions above (https://github.com/ME-ICA/tedana/pull/458#pullrequestreview-331469813) and you marked them resolved ***without*** including them (which means you've dismissed those suggestions).

Can you clarify if that's what you meant to do? My guess is that you clicked "resolve" thinking that you wanted to "commit the suggestion" and accept? If that's the case you need to go back up to those suggestions, click "unresolve suggestion" and then click the "commit this suggestion" button.

Does that make sense? Happy to chat in the [gitter channel](https://gitter.im/ME-ICA/tedana) if not!

tedana/metrics/kundu_fit.py
CesarCaballeroGaudes: Sorry for the delay. Personally, I prefer to have the default option indicated in the call because it is easier to understand/read if you don't remember it.

CesarCaballeroGaudes:
@KirstieJane sorry for the delay in responding. Yes, I was meant to accept it, but I'm not very familiar with the procedure. Please let me know if I did it correctly now.

CesarCaballeroGaudes:
I cannot merge this PR because it seems I don't have write access

tsalo:
@CesarCaballeroGaudes I'll merge it now, and maybe @emdupre can handle the permissions issue?

PR #466, opened 2019-11-15
[FIX] Fix io.load_comptable

tsalo:
Closes None. Fixes a bug reported by @eurunuela. Perhaps due to changes in pandas `from_json()` behavior, `io.load_comptable` no longer works as expected. Instead of using pandas to load the json file, we now read in the json file as a dictionary, clean it up a bit, and then convert to a DataFrame after.

- Fix bug in `io.load_comptable` that appears to be due to changes in pandas' behavior.


emdupre:
_That_ explains it ! I apparently should update my pandas :sweat_smile: 

Thanks @eurunuela and @tsalo  !

eurunuela:
You def should! Haha

tsalo:
We should probably set a minimum pandas version for tedana that works with this change. I don't know when it changed though.

eurunuela:
0.25.1 is the version I had, so somewhere in between that and @emdupre 's pandas version.

emdupre:
Yeah, that was my follow-up question :smile_cat: I currently have 0.23.1 ...But can we confirm that this is backwards compatible ? Logically it should be.

Issue #448, opened 2019-11-10
Rename common variables and functions


tsalo:
### Summary
There are a number of variables and functions that should be renamed across the package, in my mind, because they are confusing or uninformative.

### A partial list
- catd --> data_cat
- tsoc --> data_optcom
- masksum --> adaptive_mask
- t2s --> t2s_limited
- s0 --> s0_limited
- t2ss --> t2s_ascending
- s0s --> s0_ascending
- t2sG --> t2s_full
- s0G --> s0_full
- dd --> data_dimred
- mmix --> mixing (sklearn convention)
- betas --> betas_optcom
- tedana.decomposition._utils.eimask() --> outlier_mask()
- tedana.stats.computefeats2() --> z_comp_maps()
- tedana.io.writeresults() --> tedana.io.write_results()
- tedana.io.writefeats() --> tedana.io.write_feats()
- tedana.io.load_data() --> tedana.io.reshape_data()  (since it just reshaped input 4D data to 2D/3D)
- tedana.gscontrol.gscontrol_mmix() --> tedana.gscontrol.minimum_image_regression()
- tedana.gscontrol.gscontrol_raw() --> tedana.gscontrol.gscontrol()
- Finally, how do we feel generally about `betas` vs. parameter estimates vs. regression parameter, etc? I personally had enough classes with SPSS that I automatically interpret "beta" as "standardized parameter estimate", which is not how we use it throughout the package.

### Next Steps
1. Discussion.
2. Large-scale renaming.

dowdlelt:
I'm fine with betas as the label and it is nice and short.  do we think it should include that these are ICA betas? As in, betas_ica_optcom? If the optcom was not used to generate the betas, would it be betas_sepechoes? or something like that?

Overall I am very much pro renaming to names that kind of mean something. Just reading this has helped me make sense of it.

emdupre:
What do you think of `ctab` to `comp_table` ?

rmarkello:
@emdupre: I think @handwerkerd had updated `ctab` to just `comptable` in his branch of the code, which would also work!

tsalo:
`ctab` refers to the comptable file provided by the user though, while `comptable` is the DataFrame. I think there are times when we need separate variables. Same for the mixing matrix and T2* map (pending #408).

tsalo:
I'd like this to be the final thing we address for 0.0.9.

tsalo:
Not stale.

tsalo:
[...]

PR #468, opened 2019-11-17
[REF] Reorganize fit_decay

### LINKED ISSUE #439: Drop ascending T2*/S0 estimates
### Summary
We currently estimate T2* and S0 for all ascending subsets of echoes, starting at two. We then retain only those estimates associated with the value in the adaptive mask (except for voxels in the mask with a value of 1, where we use the T2*/S0 estimates from the first two echoes). This leads to a lot of calculations that are not used, as well as function outputs that no one wants. Also, while there is minimal time cost with the default log linear method, with the curve method calculating all of these unnecessary T2*/S0 estimates can take a while.

### Additional Detail
We can limit T2*/S0 calculation to voxels and sets of echoes that will actually be used, as well as drop the ascending estimates from our outputs.

### Next Steps
1. Remove ascending T2*/S0 estimates from the outputs of fit_decay.
2. Don't write out the ascending estimates when verbose is True.
3. Remove the corresponding files from our list of outputs in the documentation, as well as the list of test outputs.
4. Limit calculations (at least for curvefit) to only used voxels/sets of echoes.

### LINKED ISSUE #440: Incorporate bounds into curvefit T2*/S0 estimation
### Summary
We can limit T2*/S0 estimates with the curvefit method to realistic values, which hopefully would improve the estimates or at least make fit failures more explicit.

### Additional Detail
T2* should always be greater than zero, and S0 should always be higher than the first echo's value. However, in the latter case, we might have noise in the data, so maybe a more reasonable limit would be the minimum value across echoes? This will be higher than zero, which is better than nothing. I don't know about maximum values, but it would be nice if we had some idea about the maximum T2* for CSF that we could use.

### Next Steps
1. Add `bounds` argument to `scipy.optimize.curve_fit` call.
2. Allow both `RuntimeError` (current) and `ValueError` (when bounds don't work with data) in except statement.

tsalo:
Closes #440 and closes #439.

- Split the actual decay-fitting parts of `decay.fit_decay` into `fit_loglinear` and `fit_monoexponential`.
- Only fit curve to voxels that will be used in the limited or full S0/T2* maps. This should speed up curve-fitting with large number of echoes.
- Remove ascending T2*/S0 estimates from `fit_decay` outputs.
- Write out adaptive_mask.nii.gz by default.

### DIFF
--- a/tedana/decay.py
+++ b/tedana/decay.py
@@ -11,7 +11,7 @@ RepLGR = logging.getLogger('REPORT')
 RefLGR = logging.getLogger('REFERENCES')
 
 
-def mono_exp(tes, s0, t2star):
+def monoexponential(tes, s0, t2star):
     """
     Specifies a monoexponential model for use with scipy curve fitting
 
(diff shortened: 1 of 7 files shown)


tedana/decay.py
```diff
@@ -28,7 +28,153 @@ def mono_exp(tes, s0, t2star):
...
+
+    Returns
+    -------
+    t2s_limited, s0_limited, t2s_full, s0_full
+    """
+    RepLGR.info("A monoexponential model was fit to the data at each voxel "
+                "using nonlinear model fitting in order to estimate T2* and S0 "
+                "maps. For each voxel, the value from the adaptive mask was "
+                "used to determine which echoes would be used to estimate T2* "
+                "and S0.")
```
dowdlelt: This is a bit tricky - the automated methods will have this, and then the paragraph from loglin fit, since it is called within this function. Maybe a flag in log lin to avoid appending the "loglin methods" if they were called from the fit_monoexponential function. Or perhaps even better, the it instead states something like "Log linear Estimates were used to initialize the non-linear fit. "

tedana/tests/test_decay.py
```diff
@@ -32,15 +32,13 @@ def test_fit_decay(testdata1):
...
     """
-    t2sv, s0v, t2ss, s0vs, t2svG, s0vG = me.fit_decay(testdata1['data'],
-                                                      testdata1['tes'],
-                                                      testdata1['mask'],
-                                                      testdata1['mask_sum'],
-                                                      testdata1['fittype'])
+    t2sv, s0v, t2svG, s0vG = me.fit_decay(testdata1['data'],
+                                          testdata1['tes'],
+                                          testdata1['mask'],
+                                          testdata1['mask_sum'],
```
dowdlelt: Should all mentions of mask_sum become adaptive mask?

tedana/tests/test_decay.py
```diff
@@ -75,14 +73,10 @@ def test_smoke_fit_decay():
     mask = np.random.randint(2, size=n_samples)  # generate binary mask of random 0s and 1s
     masksum = np.random.random((n_samples))
     fittype = 'loglin'
-    t2s_limited, s0_limited, t2ss, s0vs, t2s_full, s0_full = me.fit_decay(
-                                                            data, tes,
-                                                            mask, masksum,
-                                                            fittype)
+    t2s_limited, s0_limited, t2s_full, s0_full = me.fit_decay(
+        data, tes, mask, masksum, fittype)
```
dowdlelt: Same thing here - perhaps I'm missing something, but shouldn't masksum here be adaptive mask, just for consistency?

tedana/tests/test_decay.py
```diff
@@ -101,15 +95,10 @@ def test_smoke_fit_decay_curvefit():
     mask = np.random.randint(2, size=n_samples)  # generate binary mask of random 0s and 1s
     masksum = np.random.random((n_samples))
     fittype = 'curvefit'
-    t2s_limited, s0_limited, t2ss, s0vs, t2s_full, s0_full = me.fit_decay(
-                                                                data, tes,
-                                                                mask,
-                                                                masksum,
-                                                                fittype)
+    t2s_limited, s0_limited, t2s_full, s0_full = me.fit_decay(
+        data, tes, mask, masksum, fittype)
```
dowdlelt: another masksum

tedana/tests/test_decay.py
```diff
@@ -128,10 +117,7 @@ def test_smoke_fit_decay_ts():
     masksum = np.random.random((n_samples))
     fittype = 'loglin'
     t2s_limited_ts, s0_limited_ts, t2s_full_ts, s0_full_ts = me.fit_decay_ts(
-                                                                data, tes,
-                                                                mask,
-                                                                masksum,
-                                                                fittype)
+        data, tes, mask, masksum, fittype)
```
dowdlelt: another masksum

tedana/tests/test_decay.py
```diff
@@ -153,10 +139,7 @@ def test_smoke_fit_decay_curvefit_ts():
     masksum = np.random.random((n_samples))
     fittype = 'curvefit'
     t2s_limited_ts, s0_limited_ts, t2s_full_ts, s0_full_ts = me.fit_decay_ts(
-                                                                data, tes,
-                                                                mask,
-                                                                masksum,
-                                                                fittype)
+        data, tes, mask, masksum, fittype)
```
dowdlelt: The last masksum I saw

tedana/tests/test_decay.py
tsalo: As part of #448, absolutely, although I'd prefer to just do that in one fell swoop.

tedana/decay.py
tsalo: Nice catch. I totally missed that.

Since the `fit_loglinear` function is now, in effect, pretty much private, I don't have a problem adding a flag. I'll also update the monoexponential report.

tedana/tests/test_decay.py
tsalo: Actually never mind. Might as well do it now.

PR #469, opened 2019-11-19
[FIX] Fix array preallocation in decay-fitting functions

### LINKED ISSUE #470: If all voxels are good or bad, an index mismatch is triggered
### Summary
When a dataset has voxels that are all good or all bad echoes, it triggers the following error:
```
IndexError: boolean index did not match indexed array along dimension 1; dimension is 3 but corresponding boolean dimension is 1
```

### Additional Detail
Can be recreated with the FIU truncated 4-echo dataset here: https://osf.io/q4ybc/download
and the command
```
tedana -d sub-PILOT_ses-01_task-localizerDetection_run-01_echo-[1,2,3,4]_space-sbref_desc-preproc_bold+orig.HEAD -e 11.8 28.04 44.28 60.52 --tedpca=kundu
```
According to git biset and @tsalo both, the error was introduced in ce72cf13eb87723c7c5d12150662f5f6e26d70c8

### Next Steps
1. Correct the bug

tsalo:
Closes #470. Fixes a bug introduced in #468.

- Use the set of good echoes to preallocate arrays instead of all echoes.

### DIFF
--- a/tedana/tests/test_decay.py
+++ b/tedana/tests/test_decay.py
@@ -70,8 +70,9 @@ def test_smoke_fit_decay():
     n_times = 20
     data = np.random.random((n_samples, n_echos, n_times))
     tes = np.random.random((n_echos)).tolist()
-    mask = np.random.randint(2, size=n_samples)  # generate binary mask of random 0s and 1s
-    adaptive_mask = np.random.random((n_samples))
+    mask = np.ones(n_samples, dtype=int)
+    mask[n_samples//2:] = 0
+    adaptive_mask = np.random.randint(2, n_echos, size=(n_samples)) * mask
     fittype = 'loglin'
     t2s_limited, s0_limited, t2s_full, s0_full = me.fit_decay(
         data, tes, mask, adaptive_mask, fittype)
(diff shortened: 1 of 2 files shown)


tedana/tests/test_decay.py
```diff
@@ -70,8 +70,9 @@ def test_smoke_fit_decay():
     n_times = 20
     data = np.random.random((n_samples, n_echos, n_times))
     tes = np.random.random((n_echos)).tolist()
-    mask = np.random.randint(2, size=n_samples)  # generate binary mask of random 0s and 1s
-    adaptive_mask = np.random.random((n_samples))
+    mask = np.ones(n_samples, dtype=int)
+    mask[n_samples // 2:] = 0
```
jbteves: Why does the test need the first half of the mask to be 0?

tsalo: It doesn't, but the test mask and adaptive mask can't both be random because then test failures can occur randomly.

jbteves: Maybe I haven't had enough coffee, but then wouldn't you want there to be a set number of "good" voxels rather than "bad"?

tsalo: It'll be half good and half bad. The zeros are unnecessary, but they make it a real mask.

jbteves: Cool. Add a quick comment explaining that above the first test and then this looks good to me.

Issue #452, opened 2019-11-11
Add regression tests for T2*/S0 estimation and optimal combination


tsalo:
### Summary
Per discussion with @afni-rickr and @emdupre, we may want regression tests for our more stable steps, like T2*/S0 estimation with log-linear and optimal combination with the t2s method.

### Additional Detail
We will need to tune our tolerances for the regression tests.

### Next Steps
1. Write tests
2. Choose tolerance levels

jbteves:
Boo stale-bot, we're doing our best.

Issue #425, opened 2019-11-07
Only operate on masked data in array format and keep masks as images


tsalo:
### Summary
Per discussion at the hackathon, it is confusing to have our data variously in XxYxZxT-, SxExT-, and MxExT-shaped arrays. Since passing around strings (to load/write files in each function) or images (per #251) are both too slow, we have committed to operating on arrays, but also want to limit our functions to MxExT-shaped arrays. Masks should then be images, and we will need custom functions to mask and unmask multi-echo data (as nilearn automatically concatenates lists of files/imgs in the time dimension).

### Next Steps
1. In all cases where we pass around SxExT data, we should change it to just use the MxExT (i.e., masked) data.
2. Add new functions to apply masks and unmask data that works elegantly with lists of files.

jbteves:
We're doing the work, bot.

PR #475, opened 2019-11-22
[REF] Modularize metrics

emdupre:
This re-implements #447 for review.

edit by @tsalo 2020/07/19:

To do:
- [x] Implement adaptive masking changes from #358
- [x] Check that changes to eimasking haven't been overwritten (#555)
- [x] Unit tests for new metric calculation functions
- [x] Regression test against outputs from current workflow(s)
- [x] Fix breaking tests (probably due to inescapable while loop)


emdupre:
@handwerkerd I know you had started reviewing this previously -- if you have a chance to look over this again we can make sure your suggestions are incorporated !

handwerkerd:
@tsalo added ./docs/module.rst to parse documentation for his code. I've also started to use this module.rst for the decision tree modularization documentation parsing in #476. I don't know enough about how these files should be set up to review it. Particularly, if this is something multiple people should use, it should get reviewed by someone else. @jbteves or @emdupre?

I am seeing a bunch of warnings. The most noteworthy one is, when I go into ./docs/ and type `make html` I get the same warning for a bunch of different functions: 
`/Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/generated/tedana.metrics.collect.rst:15: WARNING: duplicate object description of tedana.metrics.collect.generate_metrics, other instance in generated/tedana.metrics.collect.generate_metrics, use :noindex: for one of them`

This only seems to only be happening for the new functions so I think it's specific to this PR. 

The other warning I see is:
`checking consistency... /Users/handwerkerd/code/tedana_community/handwerkerd/tedana/docs/dependence_metrics.rst: WARNING: document isn't included in any toctree`
I think that might be an issue in master, not this PR, but since it's focused on dependence metrics, I figured I'd mention here first.

tsalo:
@handwerkerd That template is just a slightly adapted version of [this Sphinx template](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/templates/autosummary/module.rst). If we revert to the basic version I linked to, the duplicate entry warnings go away. Do you want to incorporate that into a separate PR (maybe one that fixes #480)?

The file `dependence_metrics.rst` is only linked to in `approach.rst` without being included in any tables of contents. It's part of master.

handwerkerd:
If line 11 `:toctree:` is removed from module.rst and the contents of the ./doc/generated directory are removed before running `make html` the warning goes away. That can probably fixed to resolve this issue in this PR, but someone who better understands how these rst templates are being used to confirm that this makes sense and this template is doing what we want it to do.

FWIW, in the GitHub website GUI for code review, I only see files where code was edited, not newly added files. Thus, I couldn't add this as an in-line review edit. Am I missing something obvious?

I'm going to try to finish up a bit more testing of this code today through tomorrow or finish my review of this PR.

tsalo:
New files are included in the code review GUI, but the `module.rst` template isn't new. It was added ~2 years ago.

handwerkerd:
ok. Not sure why I was missing it, but I hadn't noticed it in master. Now I see it there. I'll move the discussion of module.rst to #480 & try to finish up reviewing other aspects of this today.

handwerkerd:
@tsalo I'm trying to run generate_metrics on a different dataset and came across a problem where the solution might involve some discussion. The testing data ( https://github.com/tsalo/tedana/blob/41d94465b39ac210a938b1972f36bc14c701ef7b/tedana/tests/test_metrics.py#L16 ) uses an artificial optimally combined volume, which is the average of the unmasked echoes. (This might also relate to #452) I tried to run this using actually optimally combined data that was masked & set to 0 outside of the brain. In that case, a call to stats.zscore in computefeats2 ( https://github.com/tsalo/tedana/blob/41d94465b39ac210a938b1972f36bc14c701ef7b/tedana/stats.py#L78 ) includes a bunch of dividing by zeros, which creates NaNs that later crash a least squares fit in a call to stats.get_coeffs.

I can think of two overlapping ways to solve this issue.
1. stats.zscore should test for dividing by 0 and set those values to zero in all cases (with or without a warning). 
   - In practice, this is probably adW functionality, but it assumes there is no zscoring situation where we'd want a divide by 0 to create NaNs and potentially crash the program. 
   - That said, we could add a flag to zscore to make this a non-default option. For example, If a function is calling zscore on the time dimension of an array, (particularly voxels x time) That is a scenario where a time series with all 0s was almost definitely masked and ignoring those voxels would be benign so stats.zscore could be called with a `maskzeros` flag
2. generate_metrics takes a mask as input, but it doesn't use that mask for many of the early processing steps and not all of @tsalo's new functions take a mask as input or use the mask when given. I think making it possible to pass the mask to everything is a good explicit approach, but it presents several challenges.
   - Someone might want to pass an explicit mask, but they don't want everything masked (i.e. the discussion surrounding masked data & integration with fmriprep, which doesn't seem to be in an issue on GitHub). In this case we'd need to not only pass a mask, but also pass some flags saying which steps should or should not include masking
   - Some functions use the variable `mask` as an image and some use it as a binary array. It seems like the goal is to modularize this more towards the image ( #425 ), but that would become a higher priority if we need to pass the mask variable to more functions.

Thoughts?

As I'm running functions, I'm getting other errors, which I'll try to document as I better specify them.

tsalo:
> stats.zscore should test for dividing by 0 and set those values to zero in all cases (with or without a warning).

`stats.zscore` is a scipy function, so I don't know how likely it is that we'll successfully make any changes to it. That said, the input to `computefeats2` should probably always be masked (which would prevent divide-by-zero errors), which is definitely a part of #425. I think that's just an oversight in our current API.

> generate_metrics takes a mask as input, but it doesn't use that mask for many of the early processing steps and not all of @tsalo's new functions take a mask as input or use the mask when given. I think making it possible to pass the mask to everything is a good explicit approach, but it presents several challenges.

We can make some arguments that are only used for some metrics (like `mask`) into optional keyword arguments, but then we'll need to have some kind of check at the beginning of `generate_metrics` in order to compare the requested metrics with the arguments provided, so that it doesn't run too long before crashing when you don't provide a necessary argument.

> Someone might want to pass an explicit mask, but they don't want everything masked (i.e. the discussion surrounding masked data & integration with fmriprep, which doesn't seem to be in an issue on GitHub). In this case we'd need to not only pass a mask, but also pass some flags saying which steps should or should not include masking

I think that metric calculation should be done on masked data only. For fMRIPrep compatibility, we could denoise the unmasked data (after figuring out how to do optimal combination on unmasked data as well), but I don't think we should include non-brain voxels in the metric calculations.

handwerkerd:
All of the above sounds good. I'll work through the code to make sure the mask is passed appropriately. I'm not sure I'll be able to completely deal with #425 now, but I'll try to get something that works. 
I was a bit confused by a zscore function in our code base, but the fun of VSCode is I asked it to open the function definition and I didn't realize I had tedana.stats.py open right next to scipy.stats.stats.py. We can still add a test before or after the call to zscore, but we're obviously not adding this as an edit to scipy

handwerkerd:
It's frustrating that I didn't have time to fully resolve this today, but I think masking might have been ok, but the issue is that, if there are zero voxels outside of the mask, functions are unhappy. This might be an edge case, but it's officially an edge case that caused one user a headache. I'm adding code to check for this situation, log a warning, and expand the mask to exclude voxels without data in metric calculations. It's not done, but the work is in https://github.com/tsalo/tedana/pull/8

tsalo:
I will need to go through changes that have been made to tedana since work on this PR began and come up with a checklist of things to reimplement in the new structure.

tsalo:
@handwerkerd I think that the mask should be reduced to only good voxels _before_ the metrics are calculated, so it might not be worth it to include something to correct the mask in this PR (which is why your work in #491 is great). However, I can add in a check like the one you're proposed in tsalo/tedana#8.

Here's a draft of a function that we can call on both the ME and OC data at the beginning of the collection function. Please let me know what you think.

```python
def check_mask(data, mask):
    """
    Check that no zero-variance voxels remain in masked data.

    Parameters
    ----------
    data : (S [x E] x T) array_like
        Data to be masked and evaluated.
    mask : (S) array_like
        Boolean mask.

    Raises
    ------
    ValueError
    """
    assert data.ndim <= 3
    assert mask.shape[0] == data.shape[0]
    masked_data = data[mask, ...]
    dims_to_check = list(range(1, data.ndim))
    for dim in dims_to_check:
        # ignore singleton dimensions
        if masked_data.shape[dim] == 1:
            continue

        masked_data_std = masked_data.std(axis=dim)
        zero_idx = np.where(masked_data_std == 0)
        n_bad_voxels = len(zero_idx[0])
        if n_bad_voxels > 0:
            raise ValueError('{0} voxels in masked data have zero variance. '
                             'Mask is too liberal.'.format(n_bad_voxels))
```

tsalo:
Just to note for myself from a Jitsi conversation during OHBM poster presentations- the biggest change that I have to update to is the changes to masking.

tsalo:
I will try to review changes from the past... several months 🤦 to make sure we aren't overwriting anything by next Friday. After that, the plan (per our monthly call) is to merge this sooner rather than later and then start opening issues about the branch under the Modularization project (https://github.com/ME-ICA/tedana/projects/2).

tsalo:
Okay, all that's left is a regression test against current outputs. @emdupre and @handwerkerd, is there anything else you'll want done before we can get this merged?

emdupre:
This is just to merge into that branch, right, rather than merge into master ?

Hard to give it a full-review, but I agree on the checklist, generally ! So I'm +1 that regressions tests are what's left.

tsalo:
Okay cool. Should I just merge once the regression test is handled? Maybe open a more reviewable PR from the branch to master?

jbteves:
I'd go ahead and merge this and then open a PR to master to make this reviewable. Currently there's a ton of noise from integrating changes into your branch from master.

handwerkerd:
I'll defer to others on when to merge. This doesn't have to be a perfectly functional branch so I'm fine merging so that additional changes can be discrete and more documented PRs onto this branch (Until we start dealing with connecting the decision tree modularization part)

tsalo:
From some initial comparisons using the full five-echo test data (not the cropped five-echo data), it looks like there are very small differences in metrics, with some components switched in order starting around component 31. It's always one component switched with the one right after it, so it must be due to the small differences in kappa values.

I feel pretty comfortable merging, unless someone wants something more rigorous, in which case I'll probably need a little help.

PR #481, opened 2019-12-03
[MNT] Drop versioneer from requirements

effigies:
The versioneer package is not a requirement to use versioneer, only to update it. `install_requires` should only list packages needed to use the package, not to maintain it. If you like, I could add a `dev` extra that includes versioneer.

Context: fMRIPrep indirectly requires versioneer because it requires tedana. I'd rather keep the dependencies to a minimum.

- Removes the versioneer package from `install_requires`.

### DIFF
--- a/tedana/info.py
+++ b/tedana/info.py
@@ -33,7 +33,6 @@ REQUIRES = [
     'nilearn',
     'nibabel>=2.1.0',
     'scipy',
-    'versioneer',
     'pandas',
     'matplotlib'
 ]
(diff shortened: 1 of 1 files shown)


emdupre:
Thanks, @effigies !

It'd be great to add the DEV extra in this PR too, if you don't mind :smile_cat:

effigies:
Done. Let me know if anything else would be useful to do while I'm here.

effigies:
Realized that y'all use `[TAG]` syntax. Sorry about that. I didn't see any more-applicable tag in your [list](https://github.com/ME-ICA/tedana/blob/master/CONTRIBUTING.md#pull-requests). 🤷‍♀

PR #491, opened 2019-12-11
moving eimask to execute in make_adaptive_mask

### LINKED ISSUE #490: Secondary masking in decomposition.tedpca crashes program
### Summary
decomposition.tedpca takes a mask as input. The following lines of code call _utils.eimask to apply an additional mask that seems to remove voxels with outlier magnitudes.
https://github.com/ME-ICA/tedana/blob/9912c092dd12891aeb4609a2112ce0a8f349d2d2/tedana/decomposition/pca.py#L258-L259
The additional voxels masked in this step are not recorded anywhere and the program later crashes because the masked data volume now has fewer voxels than the mask defined in `mask`

### Additional Detail

I came across this bug while processing a new data set. I suspect that most data haven't actually caused _utils.eimask to remove voxels so this bug never appeared. In my case, the original mask included 60455 voxels and eimask removed a single voxel so that the new data included 60454 voxels. The program continued to run without problem until the following write call where `dd` contained one fewer voxel than `mask`
https://github.com/ME-ICA/tedana/blob/9912c092dd12891aeb4609a2112ce0a8f349d2d2/tedana/workflows/tedana.py#L486-L487

When I change the above code so that `eim` is True for all values (i.e. no additional masking) the full program runs without problems.

This function has existed for a while so I don't know when this problem arose. I think @tsalo worked on in this area of the code 8 months ago and @eurunuela had the most recent PR. Either no one has tested data that has caused additional masking or the recent merges have altered how the additional masking was handled.

### Next Steps

- Decide if this additional outlier masking is necessary.
- If the masking isn't necessary, then remove this function and the few uses of `eim` from the code
- If the masking is necessary, then
    - Option 1: Outlier masking should be moved to the section of the code where the mask is originally calculated. Outlier masking isn't specific to the PCA step and this masking will remove voxels from every resulting image. Thus this should be documented in the mask file.
    - Option 2: Think more about how user provided masks should be handled. If a user provides a mask that includes outliers, there are and should be checks to exclude those voxels, but that might necessitate logging and creating a more restricted mask than was provided by the user.
    - Option 1 makes sense for this specific scenario, but https://github.com/tsalo/tedana/pull/8 points another situation where the user provided mask might need to be additionally restricted in the code. Thinking about how to do that in a consistent way is probably useful.

handwerkerd:
Closes #490

This addresses an issue with a secondary mask is applied to the data during PCA decomposition and the mask volume isn't updated so the program later crashes.

- Moves eimask function from decomposition._utils to utils.py and executes that function within make_adaptive_mask

One general thing to note is that mask_adaptive_mask is currently a few fairly arbitrary thresholds. This will add another arbitrary threshold. Both these arbitrary thresholds are already in the code and this properly places them in the same place so it's a net positive in code clarity.


handwerkerd:
I think I've made the necessary edits to the code, but I haven't had a chance to test this on my data yet.

tsalo:
I also tinkered with a solution on my fork so I was able to identify a couple of issues that need to be resolved.

`eimask` returns an SxE boolean array, where bad voxel-echo combinations are False and good ones are True. This is different to `make_adaptive_mask`, which identifies the number of echoes with good data for each voxel. So for one voxel and five echoes, we might have the following:

- eimask: [True, False, False, True, True]
- adaptive_mask: 3
    - equivalent to [True, True, True, False, False]

Any code throughout the package that uses the adaptive mask just grabs the first X values from the data based on the number in the mask. If we use eimask as-is, we'll need to adapt our mask arrays and our related code to support selecting non-consecutive subsets of echoes.

I see two options there:
- Adapt eimask to do identify the number of consecutive non-outlier echoes starting with the first one, instead of identifying each non-outlier echo independently.
    - Also make sure it returns an (S,) array of integers instead of an SxE array of bools.
- Adapt make_adaptive_mask and all other code to support voxel-specific combinations of "good" echoes. This would mostly impact the T2*/S0 estimation, but if we merge #358 (which we said we'd do at some point), it would impact metric calculation as well.

handwerkerd:
In the current version of the code eimask is actually dysfunctional because it is supposed to have multi-echo data as an input, but we were inputting single echo data. My sense is that a conservative improvement is just to mask out any voxel that has bad data in any echo. The threshold in eimask is really conservative anyway (>5 times or less than 1/1000 of the 98th percentile value) so it's not like it's masking specific voxel/echoes for a subtle model-misfit. This is the approach I'm taking & will try to share tomorrow.

On a bigger picture, going down this rabbit hole made be realize how arbitrary all these masking thresholds are. I want something functional in the short-term, but I'd also need to think more about how this will interact with #358

handwerkerd:
#513 notes another code section where data-less voxels within a mask causes problems. It would be good for this masking issue to be addressed in this PR on a more systematic level.

handwerkerd:
As part of this PR, change `masksum` to `adaptive_mask` within masking-specific functions. See #448 and https://github.com/ME-ICA/tedana/pull/358#discussion_r362030830

tsalo:
@handwerkerd Would you be willing to change the permissions on the files in your repo from 100755 to 100644 to make it easier to compare changes?

tedana/utils.py
```diff
@@ -41,6 +41,45 @@ def load_image(data):
...
+                                         interpolation_method='lower')
+        lthr, hthr = 0.001 * perc98, 5 * perc98
+        LGR.debug('Eimask threshold boundaries: '
+                  '{:.03f} {:.03f}'.format(lthr, hthr))
+        m = dd[:, ee, :].mean(axis=1)
+        imask[np.logical_and(m > lthr, m < hthr), ee] = True
+
+    # if any echo has an outlier set the mask for that voxel to False
+
+    mask = imask.all(axis=1).astype(int)
```
tsalo: The code says "all" but the comment says "any". I noticed in the conversation that you were leaning toward any. Is that still accurate?

handwerkerd: Permissions should now be changed. FWIW, this is still a work in progress. My hope is to get #358 merged and then get this working on that code.

handwerkerd:
@tsalo I'm fairly sure this has been resolved with #555. If you agree, you can close this.

PR #495, opened 2019-12-12
[REF] Removes mle option

### LINKED ISSUE #486: Remove MLE tedpca option
### Summary
Per https://github.com/ME-ICA/tedana/issues/356#issuecomment-560134982, our current plan is to remove MLE as an option, since it just doesn't work for fMRI data.

### Additional Detail
We may wish to replace MLE with a basic variance explained-based threshold, which may be worth discussing in this issue.

### Next Steps
1. Remove MLE option from CLI, workflow function, PCA functions, tests, and documentation.
2. ~~Possibly add simple threshold-based PCA option, though retaining maPCA as the default.~~ See #492.

jbteves:
Closes #486  .

- Removes MLE PCA
-

### DIFF
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -13,7 +13,6 @@ from tedana import metrics, utils, io
 from tedana.decomposition import (ma_pca, _utils)
 from tedana.stats import computefeats2
 from tedana.selection import kundu_tedpca
-from tedana.due import due, BibTeX
 
 LGR = logging.getLogger(__name__)
 RepLGR = logging.getLogger('REPORT')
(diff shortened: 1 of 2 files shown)


tedana/decomposition/pca.py
```diff
@@ -31,34 +30,6 @@
     """),
```
tsalo: The duecredit decorators associated with the MLE function need to removed as well.

tsalo:
Could you update the PR title to include a prefix?

tedana/decomposition/pca.py
jbteves: Yeah, I would have noticed if I'd run flake8 while following *my own checklist* 🤦‍♂. That's in an unpushed commit, waiting to make sure the integration tests to pass on Circle because my conda env is a little screwy (my fault).

PR #496, opened 2019-12-12
[TST] Remove testing directory from coverage tests

### LINKED ISSUE #489: Remove tests subfolder from coverage reporting
### Summary
Per https://github.com/ME-ICA/tedana/pull/482#issuecomment-563423531, we should not count the tests folder toward our code coverage.

### Next Steps
1. Update our testing configuration to ignore the tests folder when generating coverage reports.

jbteves:
Closes #489  .

- Removes tests from codecov checks
-

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,4 @@
 ignore:
-  - "setup.py"
   - "tedana/tests/"
   - "tedana/due.py"
   - "tedana/_version.py"
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,5 @@
 ignore:
+  - "setup.py"
   - "tedana/tests/"
+  - "tedana/due.py"
+  - "tedana/_version.py"
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+  - "tedana/tests/"
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,3 +2,4 @@ ignore:
   - "tedana/tests/"
   - "tedana/due.py"
   - "tedana/_version.py"
+  - "tedana/info.py"


tsalo:
Can you also ignore `tedana/due.py` and `tedana/_version.py`?

jbteves:
Good catch, let me add those.

emdupre:
Wait, sorry, one more time for me -- why are we removing coverage on the entire tests directory ? Don't we want coverage to include the unit tests ?

tsalo:
Tests shouldn't count toward coverage, since they are what covers non-test code. It gives us an artificially inflated estimate of the actual code that is being covered.

emdupre:
> Tests shouldn't count toward coverage, since they are what covers non-test code.

I'm not sure I understand this statement. What do you mean "non-test code" ?

jbteves:
Basically, we're asking it not to count `/tests` for hits and misses because otherwise we're adding nonsense coverage. We don't need tests to cover our tests, or else we'll end up in an infinite loop. So what this will do is have the tests run, but test code lines aren't counted for hits/misses themselves. Does that help?

jbteves:
@emdupre RE: your edit: that coverage report was out of date, when we ask it to ignore a few more things coverage improves. You can see that on the latest codecov edit. The `tests` folder had mostly the same coverage as the rest of the project though, so it didn't change the actual coverage much. It's more the principle of the thing, plus reducing clutter in the report.

emdupre:
I see what you mean now. Thanks @jbteves and @tsalo for clarifying !

jbteves:
Hm, I don't think they're included based on my file browsing. But that means that `setup.py` also likely isn't included. Let's see if removing `setup.py` from the ignored files changes coverage; if not, then I think it's safe to not explicitly ignore docs, since they're top-level.

tsalo:
`tedana/info.py` could also be ignored, I think.

jbteves:
It seems that the top level is ignored, so no need to explicitly leave out docs or the setup file.

PR #498, opened 2019-12-12
[REF] Remove confusing variable copy

### LINKED ISSUE #484: reason for mmix = mmix.copy() in kundu_fit.py
@smoia @eurunuela and myself are wondering the reason of this copy of the mixing matrix mmix in kundu_fit.py (https://github.com/ME-ICA/tedana/blob/9912c092dd12891aeb4609a2112ce0a8f349d2d2/tedana/metrics/kundu_fit.py#L119)

tsalo:
Closes #484.

- Eliminate confusing variable copy by assigning to a new variable and using that.

### DIFF
--- a/tedana/metrics/kundu_fit.py
+++ b/tedana/metrics/kundu_fit.py
@@ -66,7 +66,8 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
         component selection. If `algorithm` is None, then seldict will be None as
         well.
     betas : :obj:`numpy.ndarray`
-    mmix_new : :obj:`numpy.ndarray`
+    mmix_corrected : :obj:`numpy.ndarray`
+        Mixing matrix after sign correction and resorting (if reindex is True).
     """
     # Use t2s as mask
     mask = t2s != 0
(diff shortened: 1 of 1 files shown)


tedana/metrics/kundu_fit.py
```diff
@@ -116,16 +116,15 @@ def dependence_metrics(catd, tsoc, mmix, t2s, tes, ref_img,
     # correct mmix & WTS signs based on spatial distribution tails
     signs = stats.skew(WTS, axis=0)
     signs /= np.abs(signs)
-    mmix = mmix.copy()
-    mmix *= signs
+    mmix_new = mmix * signs
```
eurunuela: I'd call it like this `mmix_signed = mmix * signs`. _New_ is not very explanatory imo.

tsalo: `mmix_new` is used later on for the sorted version, so this is just pushing that variable definition up a bit.

I'm happy to change it to something other than `mmix_new`, but I'd prefer that it be something that covers both the sign changes and the sorting, so that we don't end up with extraneous variables.

eurunuela: Good point. What about `mmix_corrected`? It sounds reasonable to me given that the comment just above states we're correcting `mmix`.

tsalo: Sounds good. I'll push that now.

PR #499, opened 2019-12-12
[REF] Remove t2smap workflow and add no-denoise argument to tedana workflow

### LINKED ISSUE #488: Convert t2smap_workflow into a function that can be called within tedana_workflow
### Summary
[`tedana_workflow`](https://github.com/ME-ICA/tedana/blob/9912c092dd12891aeb4609a2112ce0a8f349d2d2/tedana/workflows/tedana.py#L217) is the focus of most of our efforts, even though [`t2smap_workflow`](https://github.com/ME-ICA/tedana/blob/9912c092dd12891aeb4609a2112ce0a8f349d2d2/tedana/workflows/t2smap.py#L106) is also probably frequently used (via fMRIPrep). As a result, `t2smap_workflow` keeps becoming more and more out-of-date with respect to our current goals. For example, it still uses the `label` argument instead of an `out_dir` one. While we need to maintain compatibility with fMRIPrep, we should still work to keep `t2smap_workflow` current.

Given that the contents of `t2smap_workflow` overlap completely with a portion of `tedana_workflow`, we should convert `t2smap_workflow` into a function that can be called both by the `t2smap` CLI and by `tedana_workflow`.

### Next Steps
1. Update `tedana.workflows.t2smap_workflow` with respect to changes that have been made to `tedana.workflows.tedana_workflow`.
2. Move the CLI portion of `tedana.workflows.t2smap_workflow` into a separate file. Have that CLI argparser call the updated function in `tedana.workflows.t2smap_workflow`.
    - The CLI can still have different restrictions than the CLI for the tedana workflow (e.g., FIT is supported for t2smap but not for tedana).
3. Replace the T2*/S0 estimation and optimal combination sections of `tedana.workflows.t2smap_workflow` with a call to `tedana.workflows.t2smap_workflow`.

### LINKED ISSUE #494: Remove chdir in tedana_workflow
### Summary
Even though any functions that write out files should also include the output directory as an argument, it's possible that not all do. We currently have a line in `tedana_workflow` (see below) that changes the working directory to the output directory so files can be written without paths. We should eliminate this step, as it makes the code more difficult to interpret and it's just not a good idea to change directories inside the code.

https://github.com/ME-ICA/tedana/blob/feddfcb7a2435940a50a295d612df92f07cb7b80/tedana/workflows/tedana.py#L436

### Next Steps
1. Ensure that all functions that write out files include the `out_dir` argument and actually use it.
2. Remove the `os.chdir` line from `tedana_workflow`.

tsalo:
A duplicate of #497, but properly opened as a draft PR. Closes #488.

- Remove the t2smap CLI and workflow function. All of the functionality is already in the tedana workflow.
- Add --no-denoise argument to tedana CLI and denoise parameter to tedana workflow function. When denoise is False, the workflow exits after T2* estimation and optimal combination. Otherwise, it continues on to TEDPCA, TEDICA, etc. as always.
- Update t2smap tests to use tedana workflow with denoise as False.


tsalo:
This is currently breaking due to `chdir` related things that are fixed in #500. If we decide to move forward with this approach, I will fix the bugs, but I'm planning to leave it as-is for now. The core changes are solid and can be reviewed independently.

dowdlelt:
Would it be worth revisiting this now that #500 is merged in? I like this idea personally, and never liked having to run a different call to do combining when it was part of the tedana workflow. If there is some dramatic new thing we do with with the echoes, that makes sense as a separate workflow, but not optcom. Just an opinion though!

PR #502, opened 2019-12-13
[TST] Expand parameters tested in integration tests

tsalo:
Closes None.

- Test gscontrol, png_cmap, and re-running.


tsalo:
Between this and #496 we should finally be within spitting distance of 90%.

tsalo:
Going to merge this now. I'm also hoping that this will either fix the current testing bug or at least clarify what's causing it.

PR #508, opened 2019-12-16
[ENH, REF] Allow re-running without manacc and reorder args

### LINKED ISSUE #514: -ctab option
Hi - In previous versions of tedana, the input for --ctab was a tab delimitated text file (formatted just like the then output file, "comp_table_ica.txt"). If a user wanted to manually accept/reject components, they could edit the classification fields, and then rerun tedana, specifying the edited .txt file as the input for  "--ctab"

In this latest version of tedana, however, the tab delimitated file output I am used to seeing has been replaced by a .json file called "ica_decomposition.json". 

I assumed that I could edit the classification fields in this .json file, just as I had done with the .txt files, and rerun tedana using the edited .json file as the input for the --ctab option. This was not working. 

I looked at the code briefly and it looked like the same bit of code that was used to load the classification table in earlier version of tedana, is still being used now ... 

line 513 in tedana.py 
comptable = pd.read_csv(ctab, sep='\t', index_col='component')

Which prompted me to try using a tab delimitated file instead, and this seems to have done the trick. 

My question is whether this is expected behavior? It is not clear to me if the code just had not yet been updated to handle the new .json based component classification files, or if the user is expected to understand that a tab delimitated text file in a very specific format is expected. If the latter, then I think the documentation should be updated a bit to provide more information for users, because this wasn't obvious to me at first. Right now it just says, 

--ctab = "File containing a component table from which to extract pre-computed classifications."

Thanks!
Chuck

tsalo:
Closes #514. Originally a part of #408 but that PR seems to be a ways away from being merged. This PR prevents sprawl anyway.

- Reorder tedana workflow arguments.
    - Broadly reorder them from early to late in the workflow, but also grouping related arguments together. Add a group for re-running the workflow.
- Remove unnecessary check in workflow when loading the data.
    - Data is a list from the CLI and io.load_data will convert a string to list anyway.
    - Plus io.load_data will raise an error if the file doesn’t exist, so there's no need to raise an error in the workflow.
- Allow re-running workflow without manacc.
    - I.e., if manacc is not supplied, but ctab is, then use the
classifications from ctab.


tedana/workflows/tedana.py
```diff
@@ -497,20 +497,23 @@ def tedana_workflow(data, tes, mask=None, mixm=None, ctab=None, manacc=None,
...
-            comptable = pd.read_csv(ctab, sep='\t', index_col='component')
+            mmix = mmix_orig.copy()
+            comptable = io.load_comptable(ctab)
+            if manacc is not None:
+                comptable = selection.manual_selection(comptable, acc=manacc)
+            betas_oc = utils.unmask(computefeats2(data_oc, mmix, mask), mask)
+            io.filewrite(betas_oc,
+                         op.join(out_dir, 'ica_components.nii.gz'),
+                         ref_img)
             comptable = selection.manual_selection(comptable, acc=manacc)
```
dowdlelt: Should this line still be here?

tsalo: You're totally right. I'll remove this line.

```suggestion
```

tedana/workflows/tedana.py
```diff
@@ -204,16 +191,38 @@ def _get_parser():
...
+                          metavar='FILE',
+                          type=lambda x: is_valid_file(parser, x),
+                          help=('File containing mixing matrix. If not '
+                                'provided, ME-PCA & ME-ICA is done.'),
+                          default=None)
+    rerungrp.add_argument('--ctab',
+                          dest='ctab',
+                          metavar='FILE',
+                          type=lambda x: is_valid_file(parser, x),
+                          help=('File containing a component table from which '
```
jbteves: Not going to disapprove this pull request over it, but I don't think you need the parentheses. Or is that style preferred?

tsalo: I don't know which is more Pythonic, but I just find the parentheses make multiline strings more readable, even when they're not necessary.

jbteves: Okie dokie, I prefer the other way but that's fine, they're both stylistically valid.

PR #518, opened 2020-01-07
[FIX, TST] Fix the failing tests

### LINKED ISSUE #517: Current Master Failing CI
### Summary
For some reason in 4a56876 we're failing CI, which is bizarre since it only changed the license type. According to the failed report [view here](https://circleci.com/gh/ME-ICA/tedana/8600?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification) the 4-echo test generated mismatched outputs, where we expect `s0v.nii` and two other files to be generated; however, they are not.

### Next Steps
@tsalo @emdupre @rmarkello any thoughts on what's happening?

tsalo:
Closes #517.

- Not sure yet.


emdupre:
Failures were introduced at least 6 days ago. The last working test was 24 days ago. Looking over our dependencies in that time:

- Scipy [released a new version](https://github.com/scipy/scipy/releases)
- Scikit-learn [released a new version](https://scikit-learn.org/stable/whats_new/v0.22.html)
- Numpy [released a new version](https://github.com/numpy/numpy/releases)
- Nibabel [released a new version](https://github.com/nipy/nibabel/releases)

Maybe we can try pinning all of these ? My other idea is to change the four echo options to mirror those for e.g. the five echo. If we see the problem resolves then, we can narrow down what part of the code is seeing an update (probably from an environment update of some kind...).

jbteves:
@tsalo is it okay for me to open PRs to your fork to try and mess with options and consolidate work, or should I make a separate PR?

tsalo:
@jbteves Absolutely. Go for it.

tsalo:
I finally started working locally in Docker and I was able to reproduce the error both within the local Docker container _and_ a miniconda environment with the same installation procedure/requirements. Setting the five relevant packages (numpy, scipy, scikit-learn, nilearn, and nibabel) to versions from before the failing tests didn't solve the problem.

One random thing I tried out- I grabbed the first 75 volumes of the four-echo dataset on OSF (in nifti format) and ran the workflow with that instead of the BRIK/HEAD data. The results were actually different (including having actual accepted components). @jbteves did you grab the **first** 75 volumes when you made the test four-echo dataset?

Also, running on the full four-echo dataset produces good-looking results (accepted components obviously corresponding to the finger-tapping and checkerboard blocks and many rejected components that are obviously noise), so ultimately it seems like this effect might be a fluke and not indicative of a larger problem. Which is not to say that we shouldn't find out _why_ it's happening- just that it might not be something that has to be fixed once we figure it out.

tsalo:
I know it's a bit of a band-aid, but I've fixed the outputs and moved the test parameters around a little bit. Now everything's passing, although we still don't know what dependency update caused the four-echo test to fail in the first place.

jbteves:
I'm really confused, originally only the four-echo was failing, now *all* are failing?

jbteves:
I also think I did grab the first 75, so that's a bit strange...

tsalo:
@jbteves None are failing.

emdupre:
> One random thing I tried out- I grabbed the first 75 volumes of the four-echo dataset on OSF (in nifti format) and ran the workflow with that instead of the BRIK/HEAD data. 

Do you think it's an update in BRIK/HEAD handling, then ? @rmarkello are there atomic tests for that somewhere ?

emdupre:
> Setting the five relevant packages (numpy, scipy, scikit-learn, nilearn, and nibabel) to versions from before the failing tests didn't solve the problem.

Sorry, just confirming, that was done locally, right ? In this PR it looks like you're just setting a minimum, not pinning. Or am I misinterpreting ?

tsalo:
>Sorry, just confirming, that was done locally, right ? In this PR it looks like you're just setting a minimum, not pinning. Or am I misinterpreting ?

Yes, I did that locally, and I did pin to the same versions that I've set as the minimums in the PR.

BTW numpy does need at least 1.15 with the new PCA methods due to `np.flip`, which required `axis` as a positional argument up to 1.14 and now has a default.

rmarkello:
> Do you think it's an update in BRIK/HEAD handling, then ? @rmarkello are there atomic tests for that somewhere ?

Looks like the BRIK/HEAD files are volumes 1 through 75 (not 0 through 74) of the full four-echo dataset. From the BRIK/HEAD history:

```bash
3dTcat -prefix sub-PILOT_ses-01_task-localizerDetection_run-01_echo-1_space-sbref_desc-preproc_bold 'sub-PILOT_ses-01_task-localizerDetection_run-01_echo-1_space-sbref_desc-preproc_bold.nii.gz[1:75]'
```

AFNI uses (inclusive) 0-indexing so that should read [0:74] if we wanted the first 75 timepoints.

jbteves:
D'oh, you're right, @rmarkello, I forgot that while I was making the test. 
Also sorry for my lack of reading comprehension @tsalo. I think I'm confused because only 4-echo was failing, but all three tests are modified.

tsalo:
>I think I'm confused because only 4-echo was failing, but all three tests are modified.

Ah. I just reorganized them so that we'd be using more well-known methods with the 4-echo dataset (esp. kundu-stabilize), since that one's being finicky.

jbteves:
I see. So what we've really done is we've shifted the parameters so that the more stable sets use more unstable options, and less stable sets use standard options? This seems like a pretty good philosophy to me.

emdupre:
What's the thought, here ? Should we merge the fix or do we want to spend a bit more time tracking down the source of the issue for the 4-echo dataset ?

jbteves:
I'm kind of torn here because the way I see it, we have two competing priorities:
1. Being able to test revisions
1. Being able to understand why users might experience a weird behavior, and why
However, there's no indication in the projects' changelogs that we should be seeing this kind of change. So it could be quite hard to track down (2).

tsalo:
I'm leaning toward merging now. Otherwise, everything else is put on hold until after this problem is fixed. I spent a few hours on it over the weekend with no success and I just don't have time to keep digging into it right now.

jbteves:
Hm, I'd spent around a few hours just reading through bug reports and changelogs as well. If neither of us saw anything at that effort level, then I agree that this could be extremely hard to track down. Let's go ahead and merge this. Thanks @tsalo

PR #530, opened 2020-02-17
[FIX, DOC] Reorganize and improve documentation

### LINKED ISSUE #464: Add information about quantitative T2* mapping to documentation
### Summary
This is related to #407. We should provide some information on quantitative T2* sequences, and a couple of protocol PDFs if possible.

### Additional Detail
High quality T2* estimates would be helpful for decomposition and component classification, so the ability to provide them as input is useful (i.e., #407). In addition, folks need to be able to actually acquire these data. I, for one, have no clue what sequence I would even use for this, even if I know that it is possible.

### Next Steps
1. Add information about quantitative T2* mapping to the [sequences section on RTD](https://tedana.readthedocs.io/en/latest/considerations.html#available-multi-echo-fmri-sequences-for-multiple-vendors).
2. Discuss the benefits of higher quality T2* estimates in the [tedana pipeline section on RTD](https://tedana.readthedocs.io/en/latest/approach.html).
3. Add a couple of example protocol PDFs (or whatever the preferred format is for each manufacturer) to the RTD. We can include a caveat about trusting these examples implicitly if necessary.
4. Support external T2* maps for the denoising workflow (#407).

### LINKED ISSUE #480: Minor issue with links in approach.rst
### Summary

I noticed a couple of links with no target in https://github.com/tsalo/tedana/blob/modularize-metrics/docs/approach.rst#id4
Around line 188 is:
> combined` data using `Principal Components Analysis (PCA)`_ and then an `Independent Components Analysis (ICA)`_.

Neither the PCA or ICA links have targets. I'm not sure if the plan was to link to general PCA & ICA guides or internally into to another point in the documentation. 

### Next Steps

In the short-term, it probably makes sense to either remove the links or link to a general guide.

tsalo:
Closes #480 and closes #464.

- Link PCA/ICA to respective wikipedia pages (#480).
- Add information about acquisition and processing for quantitative T2* mapping (#464).
- Consolidate acquisition-related sections into one page (**Acquiring multi-echo data**).
- Move considerations from **Considerations** over to **What is multi-echo fMRI?** page and move resources from **Considerations** into new **Resources** page.
- Move **ME-fMRI Parameters & Publications** into new **Acquiring multi-echo data** page.
- Remove "comptable" from **Outputs of tedana** table (we already have the decomposition json in there).
- Fix miscellaneous typos.
- Convert some paragraphs to lists.
- Link to OSF project for protocol PDFs. The actual project is still empty but we should fix that.


docs/acquisition.rst
```diff
@@ -0,0 +1,160 @@
...
+Please reach out to the GE Research Operation team or each pulse sequence’s
+author to begin the process of obtaining this software.
+More information can be found on the `GE Collaboration Portal`_
+
+Once logged in, go to Groups > GE Works-in-Progress you can find the description
+of the current ATSM (i.e. prototypes).
+
+.. _GE Collaboration Portal: https://collaborate.mr.gehealthcare.com
+
+Other available multi-echo MRI sequences
```
handwerkerd: This section is a good addition, but missing a few key points.
1. "While tedana allows the input of a T2* map from any source, and a more accurate T2* map should lead to better results, this hasn't been systematically evaluated yet."
2. There are many ways to calculate quantitative T2* maps. You're listing one way and making some claims that aren't obvious to a reader (i.e. get phase data). This doesn't really tell people where to look for more info. I suggest find your favorite reference for this type of T2* map and reference it here.

handwerkerd:
I'm not sure I like "Move ME-fMRI Parameters & Publications into new Acquiring multi-echo data page" The parameters and publications are their own thing and don't really flow well with the rest of the acquiring page. If you don't want that as a stand-alone page, perhaps put them in resources.rst at the bottom of the publications section?

docs/acquisition.rst
```diff
@@ -0,0 +1,160 @@
...
+More complete recommendations
+and guidelines are discussed in the `appendix`_ of Dipasquale et al, 2017.
+
+.. _appendix: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0173289
+
+.. note::
+    In order to increase the number of contrasts ("echoes") you may need to first increase the TR, shorten the
+    first TE and/or enable in-plane acceleration.
+    For typically used parameters see the `parameters and publications page`_
+.. _parameters and publications page: https://tedana.readthedocs.io/en/latest/publications.html
```
handwerkerd: publications.html is now a dead link. Once there's an agreement where this should be, update the link.

docs/approach.rst
```diff
@@ -235,23 +236,23 @@ in a dimensionally reduced version of the dataset which is then used in the `TED
...
 betas for each of the components.
-The beta values from the linear regression 
-can be used to determine how the fluctutations (in each component timeseries) change 
-across the echo times. 
+The beta values from the linear regression
+can be used to determine how the fluctuations (in each component timeseries) change
+across the echo times.
 
 TE-dependence (:math:`R_2` or :math:`1/T_{2}^*`) and TE-independence (:math:`S_0`) models can then
 be fit to these betas.
```
handwerkerd: :ref:`dependence models` does not seem to be working in the compiled file. I know we had a similar issue before & it worked on the actual RTD site, but the other :ref: links seem to be working and this one does not. Not sure if something changed in this revision or an earlier revision, or if it will compile correctly on the RTD site.

handwerkerd:
Besides fixing broken links, consider all my comments here suggestions. Once the links are fixed, I'll approve the review.

docs/acquisition.rst
```diff
@@ -0,0 +1,167 @@
+Acquiring multi-echo data
+=========================
+
+Available multi-echo fMRI sequences
+-----------------------------------
+
+We have attempted to compile some basic multi-echo fMRI protocols in an `OSF project`_.
+These protocols should not be considered "canonical".
+If you would like to use one of them, please customize it for your own purposes
+and make sure to pilot it before using it in a study.
```
tsalo: @handwerkerd In fixing the OSF link you noticed, I also realized that we have a ME-fMRI sequences project on the OSF as well (https://osf.io/ebkrp/). I want to link to it (even though we don't actually have anything in there yet), but also want to make sure that the language reflects your past concerns that folks would treat any protocols we provide as "best practice" without critically evaluating and customizing them themselves.

What do you think of the above language?

docs/acquisition.rst
tsalo: I'm adding in your first point verbatim, but am a little unsure about what to say for the second. In this section I am basically parroting responses to #464, since I don't have actual experience with quantitative T2* mapping. I can include the paper linked in #464 but I'm not really familiar with any others. My own attempts at searching for T2*-mapping papers have mostly come across methods for scanning knees.

tsalo:
>Perhaps keep the parameter histograms in "acquisition" and move the google doc list of manuscripts into "resources"

That's a great recommendation and the resulting docs make more sense, I think.

docs/acquisition.rst
handwerkerd: How about: "The parameter choices in these protocols run and seem reasonable, but the have not been optimized for a specific situation. They are a good starting point for someone designing a study, but should not be considered canonical. If you would like to use one of them, please customize it for your own purposes and make sure to run pilot scans to test your choices."

docs/approach.rst
tsalo: I can fix this with a hidden toctree and a new parameter in the html configuration in conf.py.

docs/acquisition.rst
tsalo: Added. Thanks!

docs/acquisition.rst
handwerkerd: In that case, here's my suggested re-write of that paragraph. As a general message, I think we should focus on the fact that multi-echo is used in other ways and that some of these ways can calculate potentially useful T2* maps. The original text got specific in ways that left gaps of omission. I've tried to get the main point across without over-specification. (If you like my text, make sure the links are appropriate & work... I'm just drafting here)

"In addition to ME-fMRI, other MR sequences benefit from acquiring multiple
 echoes, including T1-weighted imaging (MEMPRAGE) and susceptibility weighted imaging.
 While most of these kinds of sequences fall outside the purview of this documentation, quantitative T2* mapping is relevant since a baseline T2* map is used in several processing steps including :ref:`optimal combination`.  While the T2* map estimated directly from fMRI time series is noisy, no current study quantifies the benefit to optimal combination or tedana denoising if a higher quality T2* map is used. Some benefit is likely, so, if a T2* map is independently calculated, it can be used as an input to many functions in the tedana workflow.

There are many ways to calculate T2* maps, with some using multi-echo acquisitions. We are not presenting an expansive review of this literature here, but https://doi.org/10.1016/j.ejro.2018.12.006 and https://doi.org/10.1016/j.neuroimage.2012.01.053 are good places to start learning more about this topic."

tsalo: Looks great. Thank you!

docs/approach.rst
handwerkerd: At least in my local running of <make html> the link to dependence models is still not working.

docs/acquisition.rst
handwerkerd: One really minor suggestion to take or leave. When you reference "Cohen-Adad et al. (2012) and Ruuth et al. (2019)" add links to the DOIs for those papers.

docs/approach.rst
tsalo: It's probably a Sphinx version issue. The `:hidden:` directive and `:includehidden:` parameter were both added in like 1.4 I think. Tedana's current requirements use a later version than that and the RTD CI build didn't have any warnings (just checked) so if you update your version that might help.

docs/acquisition.rst
tsalo: There are supposed to be links. I must have messed up the hyperlinking, but I'll fix it now.

tsalo: Ah, I added the actual references below but forgot to actually format the links in the text.

tsalo:
I think I've addressed all comments, and the RTD build is passing even though the CI doesn't show that. I don't know how to retrigger a build in the hopes that it will send the correct signal back to the CI, so I figured I'd just post a comment with that info.

docs/approach.rst
handwerkerd: Not sure what the issue is, but the RTD page that's outputted from the CI  documentation creation that's still pending looks fine so this is resolved.

tsalo:
Thank you both for your reviews (especially @handwerkerd)! Merging now.

PR #537, opened 2020-02-21
[ENH] Control number of threads with threadpoolctl

### LINKED ISSUE #401: More elegant multithreading handling
### Summary
Since we last managed environment variables, `numpy` has begun [recommending that people use the threadpoolctl package](https://github.com/numpy/numpy/issues/11826#issuecomment-522342271). We should add this package as a dependency and use it to manage the multithreading environment.

### Next Steps
- [ ] add threadpoolctl to this project

### LINKED ISSUE #473: importing tedana in python seems to lock my multiprocessing to one core
### Summary
I am writing a preprocessing script in python which mostly involves using nipype to wrap afni terminal commands. Normally afni will either automatically use all available cpus or take a manually given amount to process certain commands, however when I import tendana, nipype will only use one. Could this have to do with #188 ?

### Additional Detail
Here is a sample code for a wrapper in my script:
``` 
import nipype
from multiprocessing import cpu_count
from nipype.interfaces import afni as afni
from tedana.workflows import t2smap_workflow

def warp(self,fileobj1,fileobj2,out_file,transformation=None,args=None,saved_mat_file=None):

        ThreeDWarp = afni.Warp(in_file=fileobj1,out_file=out_file)
        if args is not None:
            ThreeDWarp.inputs.args=args
        if transformation == 'card2oblique':
            ThreeDWarp.inputs.oblique_parent = fileobj2
        elif transformation == 'deoblique':
            ThreeDWarp.inputs.deoblique = True
        elif transformation == 'mni2tta':
            ThreeDWarp.inputs.mni2tta = True
        elif transformation == 'tta2mni':
            ThreeDWarp.inputs.tta2mni = True
        elif transformation == 'matrix':
            ThreeDWarp.inputs.matparent = fileobj2
        elif transformation == None:
            print("Warning: no transformation input given")
        else:
            print("Warning: none of the transformation options given match the possible arguments. Matching arguments are card2oblique,"+
             " deoblique, mni2tta, tta2mni, and matrix")
        ThreeDWarp.inputs.num_threads = cpu_count()

        if saved_mat_file: #this is for if the pipline requires saving the 1D matrix tranformation information
            print('saving matrix')
            ThreeDWarp.inputs.verbose = True
            ThreeDWarp.inputs.save_warp = True
            
        ThreeDWarp.run()
```

### Next Steps
printing os.environ gives me ```({ 'OMP_NUM_THREADS': '1', 'LESSOPEN': '| /usr/bin/lesspipe %s', '_': '/usr/bin/python3', 'MKL_NUM_THREADS': '1', 'NUMEXPR_NUM_THREADS': '1', 'VECLIB_MAXIMUM_THREADS': '1', 'KMP_DUPLICATE_LIB_OK': 'True', 'KMP_INIT_AT_FORK': 'FALSE'})``` 

I changed OMP_NUM_THREADS and fixed my issue, but wanted to post this so that the devs here would know how this can affect nipype.

tsalo:
Closes #401 and closes #473. This is also related to #188 and #215 as well.

- Add threadpoolctl to requirements.
- Add new argument to t2smap and tedana CLI called `--n-threads`.
- Use threadpoolctl to control the number of threads based on `--n-threads` by wrapping _around_ the respective workflow functions.
    - **IMPORTANT**: This change only controls the number of threads if calling the workflows through the CLI, **not** as Python functions.
    - I just didn't know how else to do it, except by making the workflow functions into dummy functions and having the actual workflows in new functions and then wrapping those new functions with threadpoolctl in the existing functions. But that would be hideous, so I went with what I currently have.


tsalo:
If #538 is merged first then there shouldn't be a coverage decrease.

tsalo:
Given that #473 is likely causing performance issues for fMRIPrep users with multi-echo data, I'd like to attach this PR to our 0.0.9 release.

tedana/workflows/t2smap.py
```diff
@@ -89,6 +90,14 @@ def _get_parser():
                                'demanding monoexponential model is fit'
                                'to the raw data',
                           default='loglin')
+    optional.add_argument('--n-threads',
+                          dest='n_threads',
+                          type=int,
+                          action='store',
+                          help=('Number of threads to use. Used by '
+                                'threadcountctl to set the parameter outside '
```
rmarkello:
```suggestion
                                'threadpoolctl to set the parameter outside '
```

tedana/workflows/tedana.py
```diff
@@ -177,6 +171,14 @@ def _get_parser():
                                 'use of IncrementalPCA. May increase workflow '
                                 'duration.'),
                           default=False)
+    optional.add_argument('--n-threads',
+                          dest='n_threads',
+                          type=int,
+                          action='store',
+                          help=('Number of threads to use. Used by '
+                                'threadcountctl to set the parameter outside '
```
rmarkello:
```suggestion
                                'threadpoolctl to set the parameter outside '
```

tedana/workflows/tedana.py
```diff
@@ -2,23 +2,17 @@
...
-os.environ['VECLIB_MAXIMUM_THREADS'] = '1'
-os.environ['OPENBLAS_NUM_THREADS'] = '1'
-
+import os.path as op
 import shutil
 import logging
-import os.path as op
-from glob import glob
 import datetime
+from glob import glob
```
rmarkello: Any reason for the reordering of imports here?

tsalo: I moved os.path up to right below os to group imports from the same package together. I moved glob down to match our general pattern of increasing length (grouped first by built-in, third-party, then local imports).

tsalo:
@rmarkello Thanks for reviewing. And for catching the threadcountctl typo.

tedana/workflows/tedana.py
rmarkello: Gotcha! I generally use alphabetical so I think that threw me off. I see now that none of the rest are alphabetical either so :man_shrugging: Works for me!

rmarkello:
Beyond the extra reviewer, I just wanted to touch base and see if there was anything else holding this PR up @tsalo ? Are you planning to add some extra tests for the new functionality, or should we just leave it as-is? (Personally I'm not opposed to a -0.49% decrease in coverage, but 🤷‍♂)

tsalo:
In order to test the changes, we need to directly test the CLI. I have #538, which tests the CLI, but doesn't directly test the number of cores. We can merge #538 first so that coverage doesn't decrease, or we can merge this PR as-is. There's nothing holding it up except for the second reviewer.

rmarkello:
Gotcha! Thanks for clarifying 😄 

I think it makes sense to wait for #538 before this, but both seem pretty much ready to go, by my estimation! Anyone else want to chime in on these two PRs?

dowdlelt:
In testing does using more cores actually lead to a speedup? Thinking about that in regards to documentation

tsalo:
@dowdlelt I haven't tested it out recently, but when I was working on limiting it to one core in #188, I noticed a speedup from running on a single core compared to using more than one (see https://github.com/ME-ICA/tedana/pull/215#issue-251706129):

> On the bright side, there is anecdotal evidence that limiting to one core will speed the pipeline up a fair bit, although we'll want to test that to make sure it's reproducible.

dowdlelt:
I thought I had a memory of that. Maybe it is worth mentioning somewhere in docs since my (and maybe others) natural inclination would be to go as high on cores as possible. 

Assuming that finding holds with more testing....

rmarkello:
This is going to vary quite a bit based on the size of the input data. I think for most ~normal(ish) input data the extra threads aren't gonna help, but if you have some pretty extreme data then the extra threads _could_ help. (TBD what "normal" versus "extreme" data are.) 

A (small) example (on my quad-core Macbook):

```python
In [1]: import numpy as np                                                      

In [2]: import threadpoolctl                                                    

In [3]: def test(n_threads): 
   ...:     with threadpoolctl.threadpool_limits(n_threads): 
   ...:         np.sum(a @ a)               

In [4]: a = np.random.rand(10000, 10000)

In [5]: %time test(1)
CPU times: user 39.5 s, sys: 1.02 s, total: 40.5 s
Wall time: 41.1 s

In [6]: %time test(4)
CPU times: user 53.9 s, sys: 645 ms, total: 54.5 s
Wall time: 13.9 s

In [7]: a = np.random.rand(10, 10)

In [8]: %time test(1)
CPU times: user 2.74 ms, sys: 1.04 ms, total: 3.78 ms
Wall time: 2.89 ms

In [9]: %time test(4)
CPU times: user 2.94 ms, sys: 1.38 ms, total: 4.32 ms
Wall time: 3.29 ms
```

dowdlelt:
Had some data handy to test on. 
Threads | Time
---|---
1 | 1m9.834s
2| 1m28.272s
4| 2m10.307

Looks like @tsalo's findings from before hold on this data as well (8 echo, relatively low resolution, relatively short duration). More threads just seem to cause trouble in tedana. 

I did notice this while running - 
`INFO:numexpr.utils:NumExpr defaulting to 8 threads.`
just prior to combining the timeseries - not sure if that is a concern.

jbteves:
Sorry guys I really dropped the ball on reviewing this. @dowdlelt that is an interesting observation. That is a concern, because it means that we're still trying to max out threads at that point. Did you observe this too, @rmarkello ?
Also it would be ideal to set the number of threads to 4, not 8.

dowdlelt:
I think the help text should specific that 1 core is the default, and more cores slows things down. to discourage users from thinking more cores = faster.

jbteves:
Sorry, I should have clarified:
`tedana` should set the number of cores to 1. However, that issue where things "default" to 8 is doubly problematic, because (a) we want it single-threaded and (b) that's the number of logical cores, not physical. Because the operations we're doing are repetitive math, it doesn't help to use more than the number of physical cores, which on your MacBook is 4 rather than 8 from what you said. 
I was asking @rmarkello if he also saw the "default to 8" message, rather than the behavior that more cores = slower tedana, which I think we have all observed independently on several machines. 
So I also second help text specifying 1 as default.

rmarkello:
I didn't run `tedana` in my tests—just demonstrated with some random data arrays that multithreading _can_ speed up computations if your data are big enough. 

I can try and check later today if I also get the `numexpr` warning when I run `tedana`.

jbteves:
No, it's okay, I'll pull it and run it since I have things open now, thanks though. Thanks @rmarkello !

tedana/workflows/t2smap.py
```diff
@@ -248,7 +259,11 @@ def _main(argv=None):
     else:
         logging.basicConfig(level=logging.INFO)
 
-    t2smap_workflow(**vars(options))
+    kwargs = vars(options)
+    n_threads = kwargs.pop('n_threads')
+    n_threads = None if n_threads == -1 else n_threads
+    with threadpool_limits(limits=n_threads, user_api=None):
+        t2smap_workflow(**vars(options))
```
rmarkello:
```suggestion
        t2smap_workflow(**kwargs)
```

I believe you want to provide `kwargs` here since you removed `n_threads` (which is not a parameter of `t2smap_workflow`). But I might be mistaken!

tsalo: Thank you for catching this!

PR #538, opened 2020-02-21
[TST] Add tests for the workflow CLIs

tsalo:
Closes None.

- Swap re-runs in the integration tests from using `tedana.workflows.tedana.tedana_workflow` to `tedana.workflows.tedana._main` (i.e., the CLI).
- Add new t2smap unit test that uses `tedana.workflows.t2smap._main`.

### DIFF
--- a/tedana/tests/test_t2smap.py
+++ b/tedana/tests/test_t2smap.py
@@ -131,7 +131,6 @@ class TestT2smap():
         out_dir = 'TED.echo1.t2smap'
 
         # Check outputs
-        assert op.isfile(op.join(out_dir, 'ts_OC.nii.gz'))
         img = nib.load(op.join(out_dir, 't2sv.nii.gz'))
         assert len(img.shape) == 3
         img = nib.load(op.join(out_dir, 's0v.nii.gz'))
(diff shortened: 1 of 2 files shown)


tedana/tests/test_t2smap.py
```diff
@@ -118,6 +116,33 @@ def test_basic_t2smap4(self):
...
+                op.join(data_dir, 'echo2.nii.gz'),
+                op.join(data_dir, 'echo3.nii.gz')]
+        args = (['-d'] + data +
+                ['-e', '14.5', '38.5', '62.5', '--combmode', 't2s',
+                 '--fitmode', 'all', '--label', 't2smap'])
+        workflows.t2smap._main(args)
+        out_dir = 'TED.echo1.t2smap'
+
+        # Check outputs
+        assert op.isfile(op.join(out_dir, 'ts_OC.nii.gz'))
```
rmarkello: Do we need this line since you're loading the file below (L143)? That check should fail if it doesn't exist.

tedana/tests/test_t2smap.py
```diff
@@ -118,6 +116,33 @@ def test_basic_t2smap4(self):
...
+        img = nib.load(op.join(out_dir, 't2sv.nii.gz'))
```
tsalo:
```suggestion
        img = nib.load(op.join(out_dir, 't2sv.nii.gz'))
```

tedana/tests/test_t2smap.py
tsalo: Good catch. Will change.

PR #548, opened 2020-03-08
[ENH] Calculate and use r-squared

### LINKED ISSUE #543: Supplement or replace adaptive mask with R-squared
### Summary
In [Cohen Adad et al. (2012)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3442114/), they used an R^2 threshold of 0.8 to define masks for their ME-GRE-based T2* maps.

### Additional Detail
I don't know if the threshold translates perfectly to the smaller numbers of echoes we typically have with ME-EPI, but I calculated R^2 for both my own four-echo and @dowdlelt's eight-echo data and the results look reasonable.

### Next Steps
1. Calculate R^2 in `fit_decay`.
    - There are a few ways we could do it. The easiest is probably to compare `S0 * exp(-TE/T2*)` to the data averaged over time. We could also take a time-series-based approach, but that would take a very long time.
1. Use R^2 to derive brain mask.
1. Save R^2 to file.

tsalo:
Closes #543.

Pending possible issues:
- When T2* is provided, r-squared is not calculated, as this would require fitting S0 with T2* held constant, which we don't have set up yet. We could further refactor the model fitting functions to allow them to estimate one parameter if the other is already set.
- `tedana.decay.fit_decay_ts` results in surprisingly bad r-squared values. Honestly, I was expecting _over_-fitting rather than underfitting.
- Sometimes there are individual, almost random voxels that are flagged as "bad" by the new adaptive mask. This is probably true of the regular adaptive mask as well. We may wish to smooth out small subsets of bad voxels.

- Calculate r-squared for T2*/S0 estimates against data averaged over time.
- Use r-squared (with threshold of 0.8) to update adaptive mask (rather than replacing it).
- Calculate T2*/S0/r-squared for first 3+ echoes, rather than 2+.
- For "bad" voxels (adaptive mask value less than 3), the full T2*/S0 maps now use the T2*/S0 values from the first three echoes, rather than the first two.
- Refactor `tedana.decay.fit_monoexponential`, `tedana.decay.fit_loglinear`, and `tedana.decay.fit_decay` to do the following:
    - General model fit wrapper (`fit_decay`) loops through echoes.
    - Individual model fit functions (`fit_loglinear` and `fit_monoexponential`) only fit one set of data and echo times. They fit to all voxels, with no voxel selection based on the adaptive mask. This allows us to use r-squared to choose the most appropriate subset of echoes _after_ they're all calculated.


tsalo:
Fixing up test_decay is going to take a bit of thought given that we'll actually need data with good model fit, so I'm adding WIP to the title.

tsalo:
If anyone wants to revive this in their own PR, please feel free to build off of my changes. I'm going to close this PR though.

PR #549, opened 2020-03-13
[FIX,TST] Adds unit tests for ma_pca functions and fixes a bug in ma_pca

### LINKED ISSUE #522: Add rigorous unit testing to PCA module
### Additional Detail
We need to add unit tests to the PCA module.

### Next Steps
Even though the methods are "private" the unit tests can still call them easily since there is no real "private" status in Python. We can simply write the unit tests as calling those methods.

eurunuela:
Closes #522 .

- Adds unit tests to almost every function in `ma_pca.py`.
- Fixes bug in `ma_pca.py`.

### DIFF
--- a/tedana/tests/test_mapca.py
+++ b/tedana/tests/test_mapca.py
@@ -6,6 +6,50 @@ import numpy as np
 import nibabel as nib
 from tedana import decomposition
 from pytest import raises
+from tedana.decomposition.ma_pca import _autocorr, _check_order, _parzen_win
+from tedana.decomposition.ma_pca import _subsampling, _kurtn, _icatb_svd, _eigensp_adj
+
+
+def test_autocorr():
+    """
+    Uni test on _autocorr function
+    """
+    test_data = np.array([1, 2, 3, 4])
+    test_result = np.array([30, 20, 11, 4])
+    autocorr = _autocorr(test_data)
+    assert np.array_equal(autocorr, test_result)
+
+
+def test_check_order():
+    """
+    Unit test on _check_order function
+    """
+    test_order = -1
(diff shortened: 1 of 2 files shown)


tedana/tests/test_mapca.py
```diff
@@ -6,6 +6,50 @@
...
+    test_order = 0
+    ord_out, w, trivwin = _check_order(test_order)
+    assert ord_out == test_order
+    assert trivwin
+
+    test_order = 1
+    ord_out, w, trivwin = _check_order(test_order)
+    assert ord_out == test_order
+    assert w == 1
+    assert trivwin
```
emdupre: Any way to separately test when `trivwin` is `False` ?

eurunuela: Must have missed that one. Just added this. Thanks for catch @emdupre! 💯

tedana/tests/test_mapca.py
```diff
@@ -6,6 +6,55 @@
 import nibabel as nib
 from tedana import decomposition
 from pytest import raises
+from tedana.decomposition.ma_pca import _autocorr, _check_order, _parzen_win
+from tedana.decomposition.ma_pca import _subsampling, _kurtn, _icatb_svd, _eigensp_adj
+
+
+def test_autocorr():
+    """
+    Uni test on _autocorr function
```
tsalo:
```suggestion
    Unit test on _autocorr function
```

eurunuela:
Thanks for the feedback @tsalo!

eurunuela:
That’s right. I checked everything on a Python session.

PR #555, opened 2020-03-25
[FIX] Drop use of eimask

### LINKED ISSUE #490: Secondary masking in decomposition.tedpca crashes program

tsalo:
Closes #490. This acts as a quick fix for the PCA masking issue, but we will ultimately use the outlier-based mask (i.e., `eimask`) in the adaptive masking procedure at the beginning of the workflow (see #491).

- No longer use eimask at all.

### DIFF
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -10,7 +10,7 @@ from scipy import stats
 from sklearn.decomposition import PCA
 
 from tedana import metrics, utils, io
-from tedana.decomposition import (ma_pca, _utils)
+from tedana.decomposition import ma_pca
 from tedana.stats import computefeats2
 from tedana.selection import kundu_tedpca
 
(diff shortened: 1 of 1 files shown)


tsalo:
Regarding the coverage decrease, I think we can either bite the bullet for the quick fix or I can add an eimask test. I'd rather just take the coverage hit, given that proper development on eimask should happen in #491.

PR #559, opened 2020-04-13
[REF] Include daysUntilStale and daysUntilClose in stalebot message

tsalo:
Closes None. This is an extremely minor change, impacting only the message posted by stale-bot. I noticed that the message posted when the stale label is added doesn't include the time frame before the issue will be closed. This just incorporates that information into the message.

- Include daysUntilStale and daysUntilClose in stale-bot message.


tsalo:
Requesting reviews from @handwerkerd and @jbteves because you both reviewed #535.

PR #560, opened 2020-04-15
[FIX] Use SNR for PAID instead of mean data

### LINKED ISSUE #558: Confirm if PAID echo combination method is implemented correctly
### Summary

I'm working on a variety of echo combination methods and was investigating the PAID method by [Poser et al. (2006)](https://doi.org/10.1002/mrm.20900). I looked into the `tedana` code for this method and noticed that [this line](https://github.com/ME-ICA/tedana/blob/master/tedana/combine.py#L94) (the second line in the code block below) only contains the data mean and not tSNR, i.e. the mean divided by standard deviation. Here are the main lines of code from the PAID method in `tedana`:
```
n_vols = data.shape[-1]
alpha = data.mean(axis=-1) * tes
alpha = np.tile(alpha[:, :, np.newaxis], (1, 1, n_vols))
combined = np.average(data, axis=1, weights=alpha)
```
### Additional Detail
Here's a screenshot of where the method is described in Poser et al. (2006):

Screen Shot 2020-04-03 at 15 37 46

I also noticed that [the description](https://github.com/ME-ICA/tedana/blob/master/tedana/combine.py#L67) for the PAID method in `tedana` mentions *"Combine data across echoes using SNR/signal and TE via the..."*. Does this mean it uses SNR ***or*** signal? This might clarify the absence of standard deviation in the equation, but in that case some more clarification of the reasoning behind that decision could be useful.

### Next Steps
Is the absence of the division by standard deviation an oversight or intentional? If intentional, is there some documentation available as to why?

tsalo:
Closes #558.

- Replace data average over time with SNR (data average over time divided by data standard deviation over time) in PAID combination method.

### DIFF
--- a/tedana/combine.py
+++ b/tedana/combine.py
@@ -91,7 +91,8 @@ def _combine_paid(data, tes):
                 "International Society for Magnetic Resonance in Medicine, "
                 "55(6), 1227-1235.")
     n_vols = data.shape[-1]
-    alpha = data.mean(axis=-1) * tes
+    snr = data.mean(axis=-1) / data.std(axis=-1)
+    alpha = snr * tes
     alpha = np.tile(alpha[:, :, np.newaxis], (1, 1, n_vols))
     combined = np.average(data, axis=1, weights=alpha)
     return combined


tsalo:
Awesome. Thanks @dowdlelt and @jbteves!

PR #557, opened 2020-04-02
[REF] Update t2smap workflow for fMRIPrep compatibility

tsalo:
References #540. This is currently limited to the t2smap workflow because the focus is on fMRIPrep compatibility, but I can also rename the output files in the tedana workflow as well, if folks want.

Changes proposed in this pull request:
- Change output files for t2smap_workflow and corresponding outputs in tedana_workflow to be more BIDS-like:
    - t2sv.nii --> T2StarMap.nii.gz
    - s0v.nii --> S0Map.nii.gz
    - t2svG.nii --> desc-full_T2StarMap.nii.gz
    - s0vG.nii --> desc-full_S0Map.nii.gz
    - ts_OC.nii --> desc-optcom_bold.nii.gz
- Add out-dir argument to t2smap_workflow, defaulting to CWD.

PR #562, opened 2020-04-28
[FIX] Correct t2smap outputs

tsalo:
Closes None. I was reading the BEP001 supported extensions list and realized I made a mistake in #557. This will fix that.

- In the `t2smap` workflow: T2StarMap --> T2starmap
- In the `t2smap` workflow: S0Map --> S0map


tsalo:
Thank you both!

Issue #534, opened 2020-02-21
integration tests not warning about more voxels excluded from a mask


handwerkerd:
### Summary

As part of reviewing #358, I was looking at the resulting volumes. I noticed that a non-trivial number of voxels are masked out in #358 than in master. The discussion of why is in that PR, but the reason for this issue is that a change in results that is fairly dramatic seemed to pass the current tests in CI. I haven't yet looked into how the results are being compared in the integration tests, but it seems like some useful comparisons are missing.

### Additional Detail
I locally ran the following command for both the master & PR358:
```
tedana -d /tmp/data/five-echo/p06.SBJ01_S09_Task11_e1.sm.nii.gz \
          /tmp/data/five-echo/p06.SBJ01_S09_Task11_e2.sm.nii.gz \
          /tmp/data/five-echo/p06.SBJ01_S09_Task11_e3.sm.nii.gz \
          /tmp/data/five-echo/p06.SBJ01_S09_Task11_e4.sm.nii.gz \
          /tmp/data/five-echo/p06.SBJ01_S09_Task11_e5.sm.nii.gz \
 -e 15.4 29.7 44.0 58.3 72.6 \
--out-dir tedana_out/PR358 \
--verbose 
```
Here is the resulting first volume of the optimally combined images with master on the left and PR358 on the right
![image](https://user-images.githubusercontent.com/7406227/75042339-40f25780-548c-11ea-8f7f-7005e6522842.png)

### Next Steps

- Have someone else replicate this comparison
- Do others agree this should have thrown on a failed CI test?
- Add code to test for this type of change
- Check to see if other non-trivial data changes aren't currently being examined.

handwerkerd:
This relates to #59 which currently has a suggestion to consider nib-diff. From the Feb 2020 dev call, it sounds like we don't want CI to fail based on changes within files, but there was agreement that there should be an easier way for code reviewers to identify which files have internal changes and to relatively easily view those changes.

tsalo:
What about outputting the difference images in the CI artifacts, so we could look at those directly and evaluate whether the change is too much more qualitatively?

PR #566, opened 2020-05-05
[FIX] Fix t2smap optimal combination

tsalo:
Closes None. Fixes a bug introduced in #358, where make_optcom now takes `adaptive_mask` instead of `mask`. The tedana workflow was updated, while t2smap wasn't.

- Change `mask` --> `masksum` in `t2smap_workflow` call to `make_optcom`.


tsalo:
We'll need to fix this before 0.0.9, for fMRIPrep compatibility (poldracklab/fmriprep#2117).

emdupre:
This code change LGTM, but I really would like to see another test added to catch this kind of error. Out-of-band you suggested a `t2smap_workflow` integration test, which I think is a good idea. Could we add that here as well ?

tsalo:
@emdupre I've added a t2smap integration test. It doesn't increase coverage because we already have a set of smoke tests for t2smap, but it does let contributors access t2smap-related artifacts, which will make it easier to download and visually check the outputs of the workflow. I think that's the best we can aim for until we have a solution for #534.

PR #574, opened 2020-05-22
[ENH] BIDS Derivatives-compatible outputs

### LINKED ISSUE #146: Update output names to match BIDS derivatives RC2 conventions

### LINKED ISSUE #649: Change decomposition files from json back to tsv
### Summary
We originally had our component tables stored in TSV format, but we switched to JSONs in #456 to better match BIDS Derivatives convention. However, after discussing the issue with some of the BIDS folks recently (please see https://github.com/nipreps/fmriprep/issues/2232#issuecomment-671594369), it sounds like a TSV is more flexible and is BIDS Derivatives-compatible. We can still place metadata about the components/decomposition in a JSON.

### Next Steps
1. Review BEP012 to determine what should go in the json and what should go in the tsv.
2. Restructure outputs so that the core component table is outputted in a tsv.
3. Update documentation.
4. Update visualization code.
5. Update component table ingestion step in the tedana workflow.

tsalo:
Closes #146 and closes #649.

- Rename outputs to be as BIDS Derivatives-compatible as possible.
- Reorganize metric and decomposition files. Metrics are now stored in TSVs, with associated JSONs describing the metrics, and decomposition files should match BEP012 convention.

## TODO

- [x] T1gs.nii.gz (from `gscontrol_raw()`)
- [x] desc-[PCA\|ICA]R2ModelPredictions_X.nii.gz
- [x] desc-[PCA\|ICA]S0ModelPredictions_X.nii.gz
- [x] desc-[PCA\|ICA]AveragingWeights_X.nii.gz
- [x] Outputs of tedana page
- [x] Tedana workflow page
- [x] `desc-ICA_metrics.tsv` vs. `desc-[tedana|AROMA]_metrics.tsv`

EDIT: ~~Whoops, I think `desc-tedana_metrics.tsv` isn't specific enough since we'll have metrics from ICA and PCA.... 🤔~~ Circumvented by just having `desc-PCA` for the TEDPCA metrics and `desc-tedana` for the final TEDICA metrics.


tedana/workflows/tedana.py
```diff
@@ -496,6 +512,12 @@ def tedana_workflow(data, tes, out_dir='.', mask=None,
         catd, data_oc = gsc.gscontrol_raw(catd, data_oc, n_echos, ref_img,
                                           out_dir=out_dir)
 
+    io.filewrite(
+        data_oc,
+        op.join(out_dir, 'desc-optcom_bold.nii.gz'),
+        ref_img
+    )
```
tsalo, line 519: Write OC data earlier than before. Much more transparent this way.

tedana/metrics/kundu_fit.py
```diff
@@ -222,20 +222,33 @@ def dependence_metrics(catd, tsoc, mmix, adaptive_mask, tes, ref_img,
...
 
         # Echo-specific maps of predicted values for R2 and S0 models for each
         # component.
-        io.filewrite(utils.unmask(pred_R2_maps, mask),
-                     op.join(out_dir, '{0}R2_pred.nii'.format(label)), ref_img)
-        io.filewrite(utils.unmask(pred_S0_maps, mask),
-                     op.join(out_dir, '{0}S0_pred.nii'.format(label)), ref_img)
+        io.filewrite(
+            utils.unmask(pred_R2_maps, mask),
+            op.join(out_dir, 'desc-{0}R2ModelPredictions_X.nii.gz'.format(label)),
```
tsalo, line 238: X is not a valid suffix.

tedana/io.py
```diff
@@ -202,41 +214,48 @@ def writeresults(ts, mask, comptable, mmix, n_vols, ref_img, out_dir='.'):
...
+    --------
+    tedana.io.write_split_ts: Writes out time series files
+    tedana.io.writefeats: Writes out component files
     """
     acc = comptable[comptable.classification == 'accepted'].index.values
-
-    fout = filewrite(ts, op.join(out_dir, 'ts_OC'), ref_img)
-    LGR.info('Writing optimally combined time series: {}'.format(op.abspath(fout)))
-
-    write_split_ts(ts, mmix, mask, comptable, ref_img, out_dir=out_dir, suffix='OC')
```
tsalo, line 228: Just moved down. Not removed.

tedana/io.py
```diff
@@ -202,41 +214,48 @@ def writeresults(ts, mask, comptable, mmix, n_vols, ref_img, out_dir='.'):
...
+    ===================================    =====================================
+
+    See Also
+    --------
+    tedana.io.write_split_ts: Writes out time series files
+    tedana.io.writefeats: Writes out component files
     """
     acc = comptable[comptable.classification == 'accepted'].index.values
-
-    fout = filewrite(ts, op.join(out_dir, 'ts_OC'), ref_img)
```
tsalo, line 225: Moved to main workflow.

tedana/gscontrol.py
```diff
@@ -78,23 +79,38 @@ def gscontrol_raw(catd, optcom, n_echos, ref_img, out_dir='.', dtrank=4):
     detr = dat - np.dot(sol.T, Lmix.T)[0]
     sphis = (detr).min(axis=1)
     sphis -= sphis.mean()
-    io.filewrite(utils.unmask(sphis, Gmask), op.join(out_dir, 'T1gs'), ref_img)
+    io.filewrite(
+        utils.unmask(sphis, Gmask),
+        op.join(out_dir, 'T1gs.nii.gz'),
```
tsalo, line 84: Invalid, but I don't know what this actually is, mathematically.

tedana/decomposition/pca.py
```diff
@@ -218,9 +218,8 @@ def tedpca(data_cat, data_oc, combmode, mask, adaptive_mask, t2sG,
     comptable['normalized variance explained'] = varex_norm
 
     # write component maps to 4D image
-    comp_ts_z = stats.zscore(comp_ts, axis=0)
-    comp_maps = utils.unmask(computefeats2(data_oc, comp_ts_z, mask), mask)
-    io.filewrite(comp_maps, op.join(out_dir, 'pca_components.nii.gz'), ref_img)
+    comp_maps = utils.unmask(computefeats2(data_oc, comp_ts, mask), mask)
+    io.filewrite(comp_maps, op.join(out_dir, 'desc-PCA_components.nii.gz'), ref_img)
```
tsalo, line 222: This should probably be `get_coeffs` or `desc-PCAZ` since `computefeats2` normalized the maps.

tsalo:
For transparency, here is what the ICA decomposition json will look like, with two example components:

```
{
    "Method": "Independent components analysis with FastICA algorithm implemented by sklearn. Components are sorted by Kappa in descending order. Component signs are flipped to best match the data.",
    "ica_00": {
        "Description": "ICA fit to dimensionally-reduced optimally combined data.",
        "classification": "rejected",
        "countnoise": 1119,
        "countsigFR2": 6128,
        "countsigFS0": 3778,
        "d_table_score": 18.4,
        "d_table_score_scrub": "n/a",
        "dice_FR2": 0.5568785491848324,
        "dice_FS0": 0.243486809765689,
        "kappa": 35.24615192099909,
        "kappa ratio": 2.408554759865709,
        "normalized variance explained": 0.017924586667012896,
        "original_classification": "rejected",
        "original_rationale": "I005",
        "rationale": "I001",
        "rho": 18.026996948546344,
        "signal-noise_p": 0.28063083660822047,
        "signal-noise_t": -1.0960543047701905,
        "variance explained": 9.670718133027448
    },
    "ica_01": {
        "Description": "ICA fit to dimensionally-reduced optimally combined data.",
        "classification": "accepted",
        "countnoise": 1277,
        "countsigFR2": 4097,
        "countsigFS0": 3496,
        "d_table_score": 14.8,
        "d_table_score_scrub": 11.8,
        "dice_FR2": 0.5252645699806231,
        "dice_FS0": 0.3942153186930905,
        "kappa": 33.957022187827526,
        "kappa ratio": 0.13136512366806724,
        "normalized variance explained": 0.0009146448952444163,
        "original_classification": "accepted",
        "original_rationale": "",
        "rationale": "",
        "rho": 33.23361928836013,
        "signal-noise_p": 0.7822498063437031,
        "signal-noise_t": 0.2769873372743333,
        "variance explained": 0.508159640579781
    },
}
```

tedana/metrics/kundu_fit.py
tsalo: This may not matter as much since I'm dropping these intermediate images in #475.

tedana/gscontrol.py
```diff
@@ -78,23 +79,38 @@ def gscontrol_raw(catd, optcom, n_echos, ref_img, out_dir='.', dtrank=4):
...
 
     # find time course ofc the spatial global signal
     # make basis with the Legendre basis
     glsig = np.linalg.lstsq(np.atleast_2d(sphis).T, dat, rcond=None)[0]
     glsig = stats.zscore(glsig, axis=None)
-    np.savetxt(op.join(out_dir, 'glsig.1D'), glsig)
+
+    glsig_df = pd.DataFrame(data=glsig.T, columns=['global_signal'])
+    glsig_df.to_csv(op.join(out_dir, 'desc-globalSignal_regressors.tsv'),
+                    sep='\t', index=False)
```
tsalo, line 95: I think `regressors` was replaced with `timeseries` in the spec. Is that correct?

tedana/tests/data/cornell_three_echo_outputs.txt
```diff
@@ -1,6 +1,20 @@
-adaptive_mask.nii.gz
-betas_OC.nii.gz
-betas_hik_OC.nii.gz
+S0map.nii.gz
+T2starmap.nii.gz
+desc-ICAAcceptedZ_components.nii.gz
+desc-ICAAccepted_components.nii.gz
+desc-ICAZ_components.nii.gz
```
tsalo, line 5: These could possibly all be replaced with `fitlins`-style outputs, in which statistical maps have the `_statmap` suffix and the `stat` entity can tell us if the map is a z-score, parameter estimate, etc. map. The problem is that BIDS-Models (and fitlins' outputs) are not supported in the actual specification at this point. I don't know how in-flux they are either.

tsalo:
Do we want to output sidecar jsons for these files as well? To specify `Sources` and `Units` at least?

tedana/tests/data/cornell_three_echo_outputs.txt
tsalo: So this one would be `desc-ICA_stat-z_components.nii.gz` in this case, I think?

tedana/gscontrol.py
emdupre: Yes, that's also my understanding ! And it matches the newest fMRIPrep outputs, if you want additional confirmation :smile_cat:

tedana/tests/data/cornell_three_echo_outputs.txt
emdupre: can you mix the `components` suffix with the `stat` key in fitlins ? I'm less up to date on BEP002, but can review it

emdupre:
> Do we want to output sidecar jsons for these files as well? To specify `Sources` and `Units` at least?

Sorry, which files ?

tedana/tests/data/cornell_three_echo_outputs.txt
tsalo: I don't think the file formatting is actually in BEP002. I think it's just about the jsons specifying the models. It was a mistake to mention BIDS-Models, since it just seems to be in `fitlins`.

tsalo:
We'd probably want jsons for all of the files. At least the core niftis.

tedana/tests/data/cornell_three_echo_outputs.txt
tsalo: Oh but I'm also not sure if `stat` can be combined with `components`. I feel like it should be allowed though. It's a logical combination...

tsalo:
I think the only things left are:

- Decide between `desc-ICA` and `desc-tedana` for the metrics files. We want to support multiple classification algorithms on the same decomposition.
- Add standard derivatives files, like `derivatives_description.json`.

handwerkerd:
I was just talking with @jbteves about this and I have one non-trivial concern. Sorry for not bringing this up earlier. This is going to be a major breaking change on outputs. Right now it seems to be close, but not quite matching the BIDS derivatives format and BIDS deviates isn't totally fixed either. That means we'll have another breaking change for output names in the near future (probably within the coming year). 

You've put so much work into this that I don't want it sitting stagnate until we agree on perfect file naming. One middle ground might be to add a --BIDSoutput flag that would use the new names when true & the current names when false. The default would currently be false. Doing that would allow for incremental changes in BIDS derivatives output names within main for the people using those outputs (and this would allow for issue and fixes that come from active use) while not breaking things for people using the current output names. Knowing that file naming happens all over the code, this will be a bit of a headache & this whole topic is worth discussion (probably at tomorrows dev call). I figured I'd comment now instead of just speaking up during the dev call so that others could think about this in advance.

I also have a tangential suggestion. If we are making breaking changes to file names, is it time to finally get ride of 'Optimally Combined' / 'OptCom' / 'OC' in file names? There are several weighted summation methods in the code and @jsheunis has just proposed a new one. Having several methods all being saved as 'optimally combined' creates unnecessary confusion. Josh also noted 'Com' can mean 'components' in other situations, which adds to the confusion. Perhaps 'Weighted Sum' / 'WSum' / 'WC' is another option???

emdupre:
> This is going to be a major breaking change on outputs. Right now it seems to be close, but not quite matching the BIDS derivatives format and BIDS deviates isn't totally fixed either. That means we'll have another breaking change for output names in the near future (probably within the coming year).

I think we're well-timed for a breaking change as we approach our first minor-version release, which usually signals backwards-incompatible API changes :) I'd suggest that we pin ourselves to a set BIDS version for a while, and bump at the next minor-version release. But I agree with your broader point that this is worth discussing !

> If we are making breaking changes to file names, is it time to finally get ride of 'Optimally Combined' / 'OptCom' / 'OC' in file names? There are several weighted summation methods in the code and @jsheunis has just proposed a new one. Having several methods all being saved as 'optimally combined' creates unnecessary confusion. 

I'm fine with that change, though it's a good push for us to think about where we can explicitly denote the combination method. I'd argue that this is something that should be more transparently documented to end-users as they can call more methods.

tsalo:
Per today's call, @handwerkerd and @jbteves will look into a name-swapping option next week.

tsalo:
> Quick thing: I noticed that this actually isn't quite compliant because we don't keep the subject and session for the outputs. Should we maybe also add a --prefix option so that this could be specified?

This won't make `tedana` a full-on BIDS App. Instead, the goal is to make the suffixes and formats BIDS-compliant. That way, a BIDS App that uses tedana can take the subject, session, task, run, etc. entities and add them to the filenames we produce.

eurunuela:
Just so we don't forget it, we discussed that updating the file names can be a burden if BIDS guidelines keep changing. 

A solution could be to have a `json` file with the directory tree of `tedana` outputs. This would allow us to update the json file and forget about where in the code these files are being written.

docs/outputs.rst
```diff
@@ -6,118 +6,136 @@ Outputs of tedana
...
+                                                                single echo's value while the limited map has a NaN.
+                                                                Only used for optimal combination.
+desc-full_S0map.nii.gz                                          Full S0 map/time series. Only used for optimal
+                                                                combination.
+echo-[echo]_desc-[PCA|ICA]_components.nii.gz                    Echo-wise PCA/ICA component weight maps.
+echo-[echo]_desc-[PCA|ICA]R2ModelPredictions_components.nii.gz  Component- and voxel-wise R2-model predictions,
+                                                                separated by echo.
+echo-[echo]_desc-[PCA|ICA]S0ModelPredictions_components.nii.gz  Component- and voxel-wise S0-model predictions,
+                                                                separated by echo.
+desc-[PCA|ICA]AveragingWeights_X.nii.gz                         Component-wise averaging weights for metric
```
jbteves, line 93: Instead of `X` should this end with `components`?

tedana/decomposition/pca.py
```diff
@@ -243,38 +244,74 @@ def tedpca(data_cat, data_oc, combmode, mask, adaptive_mask, t2sG,
     comptable['normalized variance explained'] = varex_norm
 
     # write component maps to 4D image
-    comp_ts_z = stats.zscore(comp_ts, axis=0)
```
jbteves, line 246: Why get rid of z-scoring? Unless it's somewhere else and I missed it?

tsalo: `computefeats2` handles z-scoring internally. Hopefully when #644 is merged it'll even do it correctly.

docs/outputs.rst
tsalo: good catch!
```suggestion
desc-[PCA|ICA]AveragingWeights_components.nii.gz                Component-wise averaging weights for metric
```

jbteves:
Alright, I have a significant draft of what's basically a BIDS derivatives enabler + major modularization of `io.filewrite`. I can either open a PR to this branch, which is merged in, or open a separate one to the repository where we can discuss the changes there. What's your preference @tsalo ?

tsalo:
@jbteves I worry about keeping eyes on a PR to my fork. In my experience PRs like that tend to get ignored.. What if we change the target of this PR to a new branch and then merged? Then you could open _your_ PR to that branch and we could have a draft PR from that branch to `main` simultaneously?

jbteves:
I'm not quite sure I follow. In my branch, everything in this one is merged already.

tsalo:
> I can either open a PR to this branch

This would mean opening a PR to `tsalo/tedana`. My concern is that PRs to individuals' branches are often overlooked.

> or open a separate one to the repository where we can discuss the changes there

This should work fine, but it won't let us evaluate your changes in isolation. They'll be mixed in with my changes.

My alternative was to (1) change the target of this PR to a new branch (e.g., `bids-derivs`), (2) merge this PR as-is into `bids-derivs`, (3) open your PR to `bids-derivs`, so that the diff will focus on your proposed changes, and (4) open a draft PR from `bids-derivs` to `main`, which we would merge once your PR to the new branch is merged.

jbteves:
> My alternative was to (1) change the target of this PR to a new branch (e.g., bids-derivs), (2) merge this PR as-is into bids-derivs, (3) open your PR to bids-derivs, so that the diff will focus on your proposed changes, and (4) open a draft PR from bids-derivs to main, which we would merge once your PR to the new branch is merged.

Oh, okay, I follow now. Sorry, I don't know how I misunderstood that. Yes, let's retarget to a branch on this repo that I can open a PR to it and shows up here for everyone to see. That sounds good, thanks!

tsalo:
Done!

PR #572, opened 2020-05-19
[DOC] Include [all] in developer setup install guidelines

tsalo:
Closes #532. I noticed the issue went stale and it seemed easy enough to address, so I took a crack at it.

Changes proposed in this pull request:

- Recommend using `[all]` option when installing tedana in editable mode for development.
- Rename "Developer Guidelines" page to "Developer guidelines" to match general title-case pattern.
- Link to "Developer guidelines" page in README.

PR #575, opened 2020-05-22
[DOC] Miscellaneous improvements to documentation

### LINKED ISSUE #573: Consolidate codecov files
### Summary
If possible, I'd like to consolidate `codecov.yml` and `.codecov.yml` to reduce the number of files we have to maintain.

### Additional Detail
Both files should refer to the same thing, so hopefully we can move the contents of one into the other.

tsalo:
Closes #573. This should be handled after #572.

- Add more rst references.
- Title-case parser arg groups to match “Named Arguments”
- Use admonition boxes whenever appropriate.
- Add custom `make clean` to remove generated files
- Misc. cleanup in approach, acquisition, and dependence_metrics files.
- Remove numpydoc extension. Use Napoleon instead (which we already were using, so now they just won’t step on each other).
- Sort extensions alphabetically in `docs/conf.py` to make diffs clearer.
- Update requirements in info.py based on requirement files.
- Remove requirement files.
- Add RTD config file (`.readthedocs.yml`). This will install `tedana` using `pip install .[doc]`, so we don't need a specific requirements file for it anymore.
- Consolidate codecov files. Not sure if this will work though.


docs/approach.rst
```diff
@@ -35,6 +33,8 @@ manner.
 
 Adaptive mask generation
 ````````````````````````
+:func:`tedana.utils.make_adaptive_mask`
```
tsalo, line 36: This matches the pattern employed in fMRIPrep's documentation, and is just slightly prettier, IMO.

docs/approach.rst
```diff
@@ -84,12 +82,7 @@ transforms the data by default.
 The BOLD data are transformed as :math:`log(|S|+1)`, where :math:`S` is the BOLD signal.
 The echo times are also multiplied by -1.
 
-`Decay model fit code`_
-
-.. _Decay model fit code: https://tedana.readthedocs.io/en/latest/generated/tedana.decay.fit_decay.html#tedana.decay.fit_decay
-
-
-.. note::
+.. tip::
```
tsalo, line 85: @emdupre, do you know the right CSS code to inject into our theme_override.css file to control the colors of our admonition boxes? I spent a fair amount of time trying to get `note` and `see also` to have different colors, but couldn't find the right entity to edit.

emdupre: The CSS incantation is:

```
div.admonition.note {
  background: #hex-color
}
.p.admonition.title {
  background: #darker-hex-color
}
```

Updating `note` to `warning` or any other kind of admonition. We need to add in a new CSS file in `static` to hold these, like we do in Nilearn. In a few days I can make a PR on your PR if that's helpful :smile_cat:

emdupre: Actually, I think the `theme_override.css` _should_ work. Let's try it there first ?!

tsalo: Thanks!

tsalo: The box background works, but not the title.

.codecov.yml
```diff
@@ -13,3 +13,9 @@ coverage:
       default:
         target: 0
         threshold: null
```
tsalo, line 15: We may need to change our thresholds to actual numbers now that 4.5.0 has been released. It seems like "null" no longer works.

docs/approach.rst
emdupre: Try:

```
.rst-content .note .admonition-title {
  background: #darker-hex-color
}
```

?

emdupre: Also, the current color is just for testing, right ? :sweat_smile:

tedana/combine.py
```diff
@@ -79,6 +88,16 @@ def _combine_paid(data, tes):
     -------
     combined : (M x T) :obj:`numpy.ndarray`
         Data combined across echoes according to SNR/signal.
+
+    References
+    ----------
+    * Poser, B. A., Versluis, M. J., Hoogduin, J. M., & Norris,
```
emdupre, line 94: Should we just number all the references, as in `make_optcom` ?

tsalo: We don't have a Notes section for this function, so the reference number wouldn't be pointing to anything, which raises a warning during the build. We _could_ put the reference number in the first part of the docstring, but in cases where that first description appears in a table like the ones in the API page, _that_ raises a warning. This isn't a real problem for this docstring, since it's a private function and won't be built on the site, but I think it's a good pattern to get into.

docs/approach.rst
tsalo: It was, and somehow during my merge commit I pushed a change that I'd been specifically _not_ committing, but 🤷‍♂️.

Why, you don't like it? I think it's fancy!

image

So fancy.

tedana/combine.py
emdupre: Ah, I see. Sorry though -- which "good pattern to get into" did you mean ?

docs/approach.rst
tsalo: Your recommendation worked!

image

tedana/combine.py
tsalo: Just not numbering references when they'll produce a build warning.

docs/approach.rst
emdupre: Great !! Now let's iterate on the colors a bit :joy: :man_artist:

tsalo: I'm going to drop the Notes change but add one for Caution boxes to make it look like below:

image

It's only used once in our documentation, but I think it'll be a good example as we circle back to this. The issue being that several different boxes are blue (like Notes) and sometimes those boxes will come up near each other and when they're the same color it doesn't look great. Also, if we use Napoleon to use admonitions for some of our docstring sections, at least Notes and References are the same color and right next to each other.

WDYT?

tsalo:
Does anyone have any thoughts or concerns with the proposed changes here?

jbteves (commented):
This looks great! For a lot of the text, it looks like the rule is to wrap at punctuation for <80ish characters, is that accurate? Do we have this written somewhere? If not we should add it to this set of changes IMO.

tsalo:
> This looks great! For a lot of the text, it looks like the rule is to wrap at punctuation for <80ish characters, is that accurate? Do we have this written somewhere? If not we should add it to this set of changes IMO.

This is not a hard rule, but I think it's reasonable to adopt.

jbteves:
Let's merge this as-is but if it's not a hard rule we can make a separate issue and PR later.

jbteves (dismissed):
LGTM, awesome, thanks @tsalo !

docs/dependence_metrics.rst
```diff
@@ -9,10 +9,8 @@ In the below equation, :math:`S(t, TE_k)` is signal at a given time :math:`t`
 and for a given echo time :math:`TE_k`.
 :math:`\bar{S}(TE_k)` is the mean signal across time for the echo time
 :math:`TE_k`.
-:math:`{\Delta}{S_0}(t)` is the difference in :math:`S_0` at time :math:`t`
-from the average :math:`S_0` (:math:`\bar{S}_0`).
-:math:`{\Delta}{R_2^*}(t)` is the difference in :math:`R_2^*` at time :math:`t`
-from the average :math:`R_2^*`.
+:math:`{\Delta}{S_0}(t)` is the difference in :math:`S_0` at time :math:`t` from the average :math:`S_0` (:math:`\bar{S}_0`).
```
handwerkerd, line 12: To match this up with the following line, I think:
from the average :math:`S_0` (:math:`\bar{S}_0`). 
should be
from the average :math:`S_0`.

docs/dependence_metrics.rst
```diff
@@ -63,68 +64,85 @@ For a **TE-dependence model**, if there were no fluctuations in :math:`S_0`:
...
 As an example, let us simulate some data.
 We will simulate signal decay for two time points, as well as the signal decay
 for the hypothetical overall average over time.
-In one time point, only S0 will fluctuate.
-In the other, only R2* will fluctuate.
+In one time point, only :math:`S_0` will fluctuate.
+In the other, only :math:`T_2^*` will fluctuate.
 
-.. note::
+.. caution::
```
handwerkerd, line 76: Each individual choice seems reasonable, but it's semi-arbitrary what's a 'note' vs 'tip' vs 'caution.' This isn't a reason to hold up this PR, but perhaps we'll later want to add a brief comment in the developer guide listing each one we're using in the docs & setting up a guideline for which to use.

docs/multi-echo.rst
```diff
@@ -125,50 +126,53 @@ isn't a non-trivial loss in SNR or an increase of artifacts.
...
+acquire multi-echo data.
+If a data set is expected to be used for future analyses in later years, it is
+likely that more powerful approaches to multi-echo denoising will sufficiently
+mature and add even more value to a data set.
+
+Other multi-echo denoising methods, such as MEICA, the predecessor to tedana,
+have shown the potential for much greater data quality improvements, as well as
+the ability to more accurately separate visually similar signal vs noise, such
+as scanner based drifts vs slow changes in BOLD signal.
+More powerful methods are still being improved, and associated algorithms are
```
handwerkerd, line 153: This was also my language, but it's too wordy. Suggest changing too:
"More powerful methods, and associated algorithms are..." (cutting "are still being improved")

handwerkerd:
My only comment that would be nice to deal with before merging this PR is the one in docs/dependence_metrics.rst which looks like a typo to me. Everything else is a suggestion. I'm going to approve the merge since I know you want to get this one done. If you make the suggested change, I can just approve again.

handwerkerd (dismissed):

docs/dependence_metrics.rst
tsalo, line 12:
```suggestion
:math:`{\Delta}{S_0}(t)` is the difference in :math:`S_0` at time :math:`t` from the average :math:`S_0`.
```

docs/multi-echo.rst
tsalo:
```suggestion
More powerful methods and associated algorithms are
```

tsalo:
Thanks @jbteves and @handwerkerd! I added in your edits and will merge now.

handwerkerd:
@tsalo One think I forgot to mention in my review. When I ran `make` locally for the documentation, I saw the following:
`/Users/handwerkerd/code/tedana_community/master/tedana/docs/conf.py:145: RemovedInSphinx40Warning: The app.add_stylesheet() is deprecated. Please use app.add_css_file() instead.
  app.add_stylesheet('theme_overrides.css')
/Users/handwerkerd/code/tedana_community/master/tedana/docs/conf.py:146: RemovedInSphinx40Warning: The app.add_javascript() is deprecated. Please use app.add_js_file() instead.
  app.add_javascript("https://cdn.rawgit.com/chrisfilo/zenodo.js/v0.1/zenodo.js")
`
This seems like something that's fairly easy to deal with, but I wasn't sure if it might force us to unnecessarily increase our minimum version requirements.

tsalo:
Since it's just a warning, we can ignore it until Sphinx completely drops it, I think. At that point, we'll probably need to change our minimum Sphinx version though.

PR #585, opened 2020-07-15
[FIX] Add non-zero floor to T2* values

### LINKED ISSUE #513: Divide by zero error when using explicit mask
Hi all - I am experiencing the following error, which occurs when combining data with voxelwise T2* estimates, and only if I specify an explicit mask.

..."ZeroDivisionError: Weights sum to zero, can't be normalized"

It seems that this could be related to the issue raised here? (https://neurostars.org/t/fmriprep-to-tedana/2685/12).

EDIT: After a bit of troubleshooting, I have noticed that I don't get a divide by zero error when specifying loglin as the curvefit method.

tsalo:
Closes #513.

Todo:
- [x] Write test

- Add a new function to determine smallest non-zero T2* value that will not resolve to zero in optimal combination step. Replace values in the T2* maps > 0 and < this value with the value.
- Add a new unit test for the function. This test doesn't improve coverage, but is important.

### DIFF
--- a/tedana/decay.py
+++ b/tedana/decay.py
@@ -11,6 +11,31 @@ RepLGR = logging.getLogger('REPORT')
 RefLGR = logging.getLogger('REFERENCES')
 
 
+def _apply_t2s_floor(t2s, echo_times):
+    """
+    Apply a floor to T2* values to prevent zero division errors during
+    optimal combination.
+
+    Parameters
+    ----------
+    t2s : (S,) array
+    echo_times : (E,) array
+
+    Returns
+    -------
+    t2s : (S,) array
+    """
+    echo_times = np.asarray(echo_times)
+    if echo_times.ndim == 1:
+        echo_times = echo_times[:, None]
+
+    eps = np.finfo(dtype=t2s.dtype).eps  # smallest value for datatype
+    temp_arr = np.exp(-echo_times / t2s)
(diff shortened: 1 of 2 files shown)


tedana/decay.py
```diff
@@ -11,6 +11,35 @@
...
+        T2* estimates with very small, positive values replaced with a floor value.
+    """
+    t2s_corrected = t2s.copy()
+    echo_times = np.asarray(echo_times)
+    if echo_times.ndim == 1:
+        echo_times = echo_times[:, None]
+
+    eps = np.finfo(dtype=t2s.dtype).eps  # smallest value for datatype
+    temp_arr = np.exp(-echo_times / t2s)  # (E x V) array
+    bad_voxel_idx = np.any(temp_arr == 0, axis=0) & (t2s != 0)
```
jbteves, line 38: Is the pythonic way considered to be using `and` instead?

jbteves: I've been silly and forgot about bitwise operations, ignore me.

tsalo: `and` doesn't do element wise comparisons like that. Replacing `&` with `and` in this case gets the following error:

```python
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all() 
```

jbteves:
@dowdlelt would you mind writing up an issue for that?

PR #591, opened 2020-08-07

### LINKED ISSUE #501: Look for inefficiencies in the interaction between unmasking and spatial clustering
### Summary

As part of examining #490, I noticed that utils.unmask was called 400 times through the tedana workflow. 352 of those 400 times were following the logged line:
"INFO:tedana.metrics.kundu_fit:Performing spatial clustering of components" My data has 323 volumes.

### Next Steps

- This section of the code should be examined to see if there is any way to call unmask less often or to make unmasking more efficient in this situation.
- unmask should be profiled with different dataset sizes to get a sense of the computational costs of these repeated calls (i.e. is this inefficient enough to make optimization worth the effort). Profiling unmask might be a good first issue for someone.
- This inefficiency relates to my general concern that always moving files around in an img format might end up compounding inefficiencies like this.

tsalo:
Closes #501.

- Modularize metrics.
- Add while loop to cluster-extent thresholding to maximize similarity in number of significant voxels between comparison maps.
- Add `signal-noise_z` metric, which should ultimately replace `signal-noise_t`.


tsalo:
The only failing tests are CodeCov, because of a 0.73% decrease in coverage. So that's good at least!

tedana/tests/test_integration.py
```diff
@@ -94,6 +94,7 @@ def test_integration_five_echo(skip_integration):
         out_dir=out_dir,
         tedpca='aic',
         fittype='curvefit',
+        fixed_seed=49,
```
jbteves, line 97: Why the seed change?

jbteves:
Hey @tsalo, thanks so much for this, this is awesome!
While I was looking through this one of the concerns I have is that we'll have to maintain the order of metric calculation and checks for metrics in the core part of the code. I think this could be difficult to maintain, especially if our plan is to add or delete metrics frequently. I know that we're generally opposed to using object-oriented approaches because they can be conceptually difficult, but this is a case where I think it will actually be easier to follow if we use classes. I came up with something along these lines:
- A metric superclass contains a class-specific table telling you which metrics are calculated and if so, what their values are
- Each individual class inherits from metric superclass (easy template to follow), and has a constructor that checks to see if that metric's value exists in the table; if not, calculate its dependencies and then it
- This allows you to do recursion to figure out dependencies, simplifying the core code to just specifying the top-level metrics you want
- Modify #592 to iterate over the needed metrics and calculate them.
- Would have to wipe out the metric values at the end of each component to avoid memory explosion, but that's easy to implement by simply pointing the metric to `None`, so the GC takes out the big variables
- Maybe add a verbosity function which saves intermediates? Currently we're not saving some of these and it might be good to do so.

I would propose to do the work stated above if we decide it's a good idea. I'm open to other suggestions, but I am a little concerned about having to edit core code in order to add metrics, if that makes sense.

tedana/tests/test_integration.py
tsalo: Because 49 is the first one to convert for the five-echo dataset and we're clocking in at ~30 minutes for that one test. I know it's cheating a bit but I'd rather jump to a seed that converges and get our tests finishing much faster.

jbteves:
@tsalo it looks like we end up losing several metric files. I talked to @handwerkerd and he thinks you chatted about that some time ago. Do you recall if you decided that it was okay to lose those metric files?
What would you think about adding an option to save used metrics for later review?
Lastly, would it be okay to push a few commits to this branch to bring it in line with `master`?

tsalo:
Sorry, what metric files?

Also yes, feel free to push changes.

jbteves:
Sorry for the ambiguity. In this branch, we're no longer writing the following:
image

tsalo:
Ah, I see. I believe all of those were only used for the figures in our documentation, but we can figure out how to output them again.

jbteves:
I don't see a need to do that right away. If we miss them we can bring them back. In the future, though, we should probably create an option to get metric maps.

tsalo:
@jbteves have you had a chance to work on the OOP version of this? I'm kind of leaning toward getting this approach merged first, so that we don't have to keep rebasing, but if you are close then we can hold off.

jbteves:
Not yet, we should definitely merge this. I'm glad you merged the new changes in, since I was going to finish tackling that Monday. Let me submit some cluster jobs to make sure this new version is running correctly on Monday before merging, but otherwise we shouldn't delay.

jbteves:
@tsalo I get an interesting error when I try this branch on the cluster:
```
FileNotFoundError: [Errno 2] No such file or directory: '/data/tevesjb/heunis_derivatives/sub-024/tedana/run2/_references.txt'
```
I'll try to track it down later this week. It did the same for subject 12 in that dataset. The good news is that it wrote the report anyway. Will keep you posted on the comparison between this and `main`.

tsalo:
Thanks for digging into that @jbteves.

I'd like to push forward with #574 before we merge this one, so that I can make sure that the approach we use there for tracking metadata and writing out outputs meshes well with these changes.

tsalo:
Per today's call, metrics that change are dice_FS0, dice_FT2, signal-noise_t, d_table_score, and kappa_ratio (rarely). I implemented both [`signal-noise_t`](https://github.com/ME-ICA/tedana/blob/1e07784b47c3b680fca3e618ad65bbe45b05446b/tedana/metrics/dependence.py#L449) and [`signal-noise_z`](https://github.com/ME-ICA/tedana/blob/1e07784b47c3b680fca3e618ad65bbe45b05446b/tedana/metrics/dependence.py#L392), and it's possible that I switched things up somewhat. However, I'm not sure how `dice_FS0`, `dice_FT2`, and `kappa_ratio` would be affected.

tedana/metrics/dependence.py
```diff
@@ -0,0 +1,604 @@
...
+            ref_img, np.squeeze(utils.unmask(maps[:, i_comp], mask))
+        )
+
+        maps_thresh[:, i_comp] = utils.threshold_map(
+            ccimg, min_cluster_size=csize, threshold=threshold, mask=mask, binarize=True
+        )
+    return maps_thresh
+
+
+def threshold_to_match(maps, n_sig_voxels, mask, ref_img, csize=None):
```
tsalo, line 229: @jbteves- `dice_FS0` and `dice_FT2` rely on cluster-extent thresholded maps made using this function, wherein the F-statistic maps for S0 or T2* models are cluster-extent thresholded to have a total number of significant voxels that matches (as closely as possible) the number in the cluster-extent thresholded Z-statistic map. I think that could cause some of the differences, since `tedana` doesn't currently try to match the number of significant voxels very well.

handwerkerd: The inputs to the DICE measures are used in other measures. The big thing that might make the DICE measures change is what is or isn't masked in the volume. signal-noise_t also relies heavily on the volume mask. Maybe the mask has changed between the two versions?
I wouldn't worry about kappa_ratio. That's a measure that's calculated between components within the decision tree, so, anything that affects whether a component is rejected may then affect kappa_ratio. If we fix the other issues, I suspect kappa_ratio (and d_table) will also be fixed.

tsalo: >The inputs to the DICE measures are used in other measures. The big thing that might make the DICE measures change is what is or isn't masked in the volume.

Agreed. That is definitely causing differences here.

> signal-noise_t also relies heavily on the volume mask. Maybe the mask has changed between the two versions?

It's possible, but I can't see where that would happen. `signal-noise_t` relies on the thresholded z-statistic map, which is thresholded using the normal procedure. The F-T2 and F-S0 thresholded maps are the ones that should be affected by the changes in this PR.

jbteves:
Alright, this is mostly in line with main. There is basically one hack that I had to make where we save the verbose outputs every time we collect the metrics, because the verbose outputs require a lot of information that doesn't live outside of that function, unlike the metrics which don't require echo-specific information after that step. This actually slows down the four-echo test a bit since that means each re-run requires a series of file writes. I can't think of a much better way of doing things without further modularization. However, this branch easily gets out of sync with main and I think it's better to get it in, since it touches so many parts of the code.
It additionally contains some modifications to RTD because for some reason our old configuration just stopped working. I copied the gist linked there into a javascript file to more closely match the setup specification on their documentation.
Additionally, unlike main I generate metric metadata from the component table rather than during metric creation because that ended up being a simpler way to handle the metadata while I was fixing the merge conflicts.

jbteves:
@tsalo and @handwerkerd do you mind taking a look?
Anyone else is welcome, but this is a very dense PR!

eurunuela:
I have just compared the results with `main` and this PR on a motor task and a resting-state dataset. I used `np.allclose()` to check if all metrics were identical, and they were. I ran tedana with all the default options by the way.

handwerkerd (approved):
Looks ready to merge to me! I've looked through the code for clarity and run it on several distinct data sets and the results don't change between this PR & main!

handwerkerd:
@emdupre & @eurunuela, Any chance one of you can give this a look-over so that we can approve this week? It would be really nice to just say this is merged, without any qualifiers, on the OHBM poster.

We could have me and @jbteves formally approve and @tsalo merge, but it would be nice to have a few more eyes. As far as I can tell, the resulting metric calculations are effectively unchanged. I've tested it on a few data sets with different voxel sizes, run lengths, and 3 vs 5 echoes. During my review, I saw some things that might be worth revisiting later, but were functional as is.

Please let us know if we should wait for your reviews or if you're ok with us merging.

eurunuela:
> @emdupre & @eurunuela, Any chance one of you can give this a look-over so that we can approve this week? It would be really nice to just say this is merged, without any qualifiers, on the OHBM poster.

I'll review it before Saturday.

emdupre:
I'll also aim to review this weekend, but if you have the necessary reviews don't feel compelled to wait on me !

notZaki:
I compared this PR vs main on the 88 subjects in the Cambridge collection, using both `tedpca='mdl'` and `tedpca='kundu'`. 
With `mdl`, the denoised data was exactly identical. With `kundu`, there were minor differences but they were all below 1e-7 in all cases.

The `pca_metrics.tsv` file has more detail in the PR version. 👍

tedana/metrics/collect.py
```diff
@@ -0,0 +1,683 @@
...
+        metric_maps["map optcom betas"] = dependence.calculate_betas(
+            data_optcom, mixing
+        )
+        if io_generator.verbose:
+            metric_maps["map echo betas"] = dependence.calculate_betas(
+                data_cat, mixing
+            )
+
+    if "map percent signal change" in required_metrics:
+        LGR.info("Calculating percent signal change maps")
```
eurunuela, line 206: Same thing as before, we should add tests for the if statements.

tedana/metrics/collect.py
```diff
@@ -0,0 +1,683 @@
...
+
+    if "signal-noise_z" in required_metrics:
+        LGR.info("Calculating signal-noise z-statistics")
+        RepLGR.info(
+            "A t-test was performed between the distributions of T2*-model "
+            "F-statistics associated with clusters (i.e., signal) and "
+            "non-cluster voxels (i.e., noise) to generate a z-statistic "
+            "(metric signal-noise_z) and p-value (metric signal-noise_p) "
+            "measuring relative association of the component to signal "
+            "over noise."
```
eurunuela, line 355: I think the wording in this sentence is a little complicated to follow.

tedana/metrics/collect.py
```diff
@@ -0,0 +1,683 @@
...
+            Z_maps=metric_maps["map Z"],
+            Z_clmaps=metric_maps["map Z clusterized"],
+            F_T2_maps=metric_maps["map FT2"],
+        )
+
+    if "countnoise" in required_metrics:
+        LGR.info("Counting significant noise voxels from z-statistic maps")
+        RepLGR.info(
+            "The number of significant voxels not from clusters was "
+            "calculated for each component."
```
eurunuela, line 370: Wording

eurunuela (approved):
I'm happy with the changes. This is some excellent work you've done @jbteves @handwerkerd , thanks!

I only have very minor comments that can be addressed after this PR is merged.

jbteves:
@eurunuela please also credit @tsalo as mostly we were adapting his branch.

@handwerkerd any strong feelings about addressing these changes now vs. later?

eurunuela:
> @eurunuela please also credit @tsalo as mostly we were adapting his branch.

True! I had completely forgotten! Sorry about that @tsalo and thanks for your amazing work!

jbteves:
Since this is text I just want to clarify that I wasn't being short with you, I just want to make sure he gets due credit! It's a significant change that was very intricate.

eurunuela:
> Since this is text I just want to clarify that I wasn't being short with you, I just want to make sure he gets due credit! It's a significant change that was very intricate.

Oh, don't worry! I'm actually grateful that you pointed it out :)

handwerkerd:
These look like good changes, but they won't affect functionality and should be easy to do in a later PR. My preference would be to merge before Monday and open a new issue with these suggested changes.
If @emdupre might look at this over the weekend, I'd wait for her review. Otherwise, @tsalo, you want the honor of merging Sunday night?

tedana/metrics/_utils.py
```diff
@@ -0,0 +1,180 @@
...
+                print("Warning: {} not found".format(k))
+            required_metrics_new += new_metrics
+        if set(required_metrics) == set(required_metrics_new):
+            # There are no more parent metrics to calculate
+            break
+        else:
+            required_metrics = required_metrics_new
+        escape_counter += 1
+        if escape_counter >= 10:
+            LGR.warning("dependency_resolver in infinite loop. Escaping early.")
```
emdupre, line 55: We should alert the user why this would happen with the LGR.

emdupre (approved):
LGTM ! I would like to see a few more tests, but we can merge and then add those next ! :rocket:

handwerkerd:
@tsalo You willing to squash & merge today? We can then open another issue or two to start addressing all the comments here.

tsalo:
I'll do it now then. Thanks.

PR #592, opened 2020-08-10
[REF] Decision tree modularization

### LINKED ISSUE #403: Modularize the component selection decision tree

tsalo:
Closes #403 .

- The decision tree steps in selection are modularized in new functions and classes
- decision_tree_class.py contains the new class for initializing and running a decision tree
- selection/data/ contains json files that define a simpler, conservative decision tree and the full MEICA v2.7 decision tree
-Most functions that define each decision tree step are in selection_nodes.py
-There are some minor changes in decision tree steps that might result in breaking the existing tests

Known issues:

- Some functions in the MEICA decision tree haven't been run yet because they require linking to metric calculation functions. Those functions and all of PR #447 is already merged into this code, to make that possible.
- The function documentation isn't being linked to and generated in the ReadTheDocs API section. The documentation also has some gaps. In particular, the documentation for the class structure is not done and there should be a guide explaining the json files for the decision trees and listing rules for creating a new decision tree (i.e. Even though it is possible, it is not advised to create a decision tree that can re-classify accepted or rejected components once they've received those labels)
- More checks and warnings can be added to various functions. For example, there can be a warning if an accepted or rejected component is reclassified at a later step
- There is repetitive code at the beginning of each decision tree function. That can be changed into an initialization function (or all the functions in selection_nodes.py could become a class, but that might be more hassle than it's worth). Either of these changes probably wouldn't make the code much shorter, but it would make it easier to know what is expected for each function
- Some of the functions for the later steps in the decision tree are ugly (i.e. highvariance_highmeanmetricrank_highkapparatio ). These are all functions that tweak metrics, threshold each of them separately, and use the intersection of those thresholded values to make a decision. I can theoretically make a multiple threshold comparison function that would be more generally useful. The minus is that would either break the current way decision nodes are set up, or push a lot more details and function calls into a decision node specification in the json file. On balance, I think these long functions with explanations of what they do are better than adding complexity to the json files, but this deserves a bit more thought. Also, I'm hoping we'll move away from decision trees with some messy joint measurements so I hope the currently messy functions will become legacy functions to maintain a rough replication of the MEICA method, but won't spawn too many additional messy functions.
- No testing functions have been written for this code yet
- All the other improvements that aren't coming to mind right now.


tsalo:
@handwerkerd Is there anything I can do to help with this?

handwerkerd:
I'll need to think a bit more, but here are some rough thoughts that are based on the notes above based on things that don't require deep collaboration or discussion.
There is no testing code for any of this. If you can try to start writing tests for different decision tree functions, it should be easier to expand them to all the functions.
I have a a code structure for all the decision tree modules, that's a bit repetitive. Look at it and see if you can think  of a way to clean them up a bit more.

The other thing on my list is that I renamed some parameters in the decision tree modularization, which are now different from the modularize metrics code. I tried to document those well, but we'll definitely want to harmonize variable names by either expanding the names I use or going back to what's used elsewhere in the code.

The harder goal will be to start linking this to your part, but that may require both of us taking a closer look at this PR together. I can also find time to talk through this with you (and others?) to try to break it down a bit more.

Also, shouldn't this be targeted to merge with MEICA:temp-modularization, not ME-ICA:master?

FWIW, someone is supposed start to work with me in September who should be able to devote more time to stuff like this.

tsalo:
> I have a a code structure for all the decision tree modules, that's a bit repetitive. Look at it and see if you can think of a way to clean them up a bit more.

I can give it a try!

> The other thing on my list is that I renamed some parameters in the decision tree modularization, which are now different from the modularize metrics code. I tried to document those well, but we'll definitely want to harmonize variable names by either expanding the names I use or going back to what's used elsewhere in the code.

I think we can use the current metric names and then adjust once it's time to merge this with the metric calculations, unless folks strongly prefer to push both together into one branch.

> Also, shouldn't this be targeted to merge with MEICA:temp-modularization, not ME-ICA:master?

We can always switch the target branch, but I'd honestly prefer to get the metric modularization (which is compatible with the current decision tree) merged first, and then merge this on top of that. If you and others strongly prefer to put everything in one branch before merging into master, however, that's fine too.

emdupre:
> I think we can use the current metric names and then adjust once it's time to merge this with the metric calculations, unless folks strongly prefer to push both together into one branch. 

I think that was the original plan, but if you think they can be completely self-contained (or, at least, that the metric calculations one can be), then we can separate them. This will be blocked until metrics is merged, in that case.

The two PRs will target the same reviewers / contributors though, I imagine ?

tsalo:
Yes, I think the same reviewers will take on both PRs. I'll need to dig into the code again, but I don't know if this PR will need #591 to work. As long at the metric names end up matching up (which could be a single commit here after #591 is merged), I think that the two branches are sufficiently modularized(!) that we can stitch them together at the end.

handwerkerd:
The decision tree in this PR isn't linked the rest of the code. The plan was to link it to the version with the modularized metrics. We could, but I didn't see a good reason to link it to the current master without #591. 
As for when to merge #591 into master, my big concern is whether if there are known or unknown breaking changes. At least for this PR, I know there are breaking changes. In addition to getting this code working, I'd like to get some more comparisons & tests of the resulting reports to get a better sense of the effects of the breaking changes. This relates to the discussion on #59 and #534 in that we don't currently have a good sense of how results change with code changes. I don't think we can 100% resolve this before merging this PR, but I do want to move to have some more tests of how the result do change. 
The Q is whether #591 also has enough breaking changes to make tests for data changes more critical or if we can just keep the reviews focused on the code rather than the results before merging #591 into master..

tsalo:
Thanks @handwerkerd. 

We can build off of #591 instead of `master`, but I'd still prefer to merge this into master _after_ #591 is merged. 

> The Q is whether #591 also has enough breaking changes to make tests for data changes more critical or if we can just keep the reviews focused on the code rather than the results before merging #591 into master..

It caused some minor changes. Not sure if I'd consider them breaking though.

----

One thing I wanted to bring up is that I think we should take a very hard look at Pydra. I don't recall if we discussed it at the hackathon (I remember only learning about it at the following Code Convergence), but I think that Pydra is extremely compatible with our desired approach. We're essentially constructing a workflow, defined by a json specification, using Python functions as nodes. I bet we could also convince some of the Pydra folks to help us out with the conversion!

tsalo:
@handwerkerd have you had a chance to look into the functionality we would need Pydra to support?

emdupre:
>  the functionality we would need Pydra to support?

Sorry, I know I missed the last call, but I'm a little unclear why we'll need Pydra at all. So knowing a bit about the functionality we're looking for would be really helpful !

jbteves:
@emdupre we were debating if we wanted to make the workflow pydra-compatible, but I don't recall reaching any conclusions on whether it was worth it, other than unless Pydra folks can help it's not worth it.

tsalo:
> Sorry, I know I missed the last call, but I'm a little unclear why we'll need Pydra at all.

We can make it work without Pydra, but the modularization we want is essentially a standard workflow, and I'd rather use a stable workflow engine than try to cobble together one ourselves. I think using a custom "engine" to build workflows will only make things harder down the line (e.g., with debugging and training new contributors), so we should only use one if (1) an existing general-purpose engine won't work for our needs or (2) existing engines have too many requirements to be worth it (as with nipype).

handwerkerd:
I haven't had a chance to look yet.
As @tsalo notes, it would probably be faster to keep the code as-is, but it might be better for long-term maintenance and adding features, if we use a more generalized pipeline deployment. That said, the big question we left the last meeting with with was: 1. What functionality does tedana need that isn't currently in Pydra; 2. Is it realistic to ask them to add needed functionality to Pydra? If yes to both, then we consider Pydra in the short term. If no, we go forward, as is for now.

tsalo:
Should we close this in favor of #756?

jbteves:
I believe so. Any objections @handwerkerd ?

handwerkerd:
No objections.

PR #609, opened 2020-09-24
[REF] Rename T1c to "minimum image regression"

### LINKED ISSUE #605: Rename T1c to "minimum image regression"
### Summary
In a manuscript that I'm working on, I describe, in detail, the method implemented in [`tedana.gscontrol.gscontrol_mmix()`](https://github.com/ME-ICA/tedana/blob/150b5c0dd4e452ed00861e7e8d4347f1583599c5/tedana/gscontrol.py#L111).
We generally refer to this method as either `gscontrol_mmix` (based on the array it works on), `T1c` (because it references a T1-like effect that it removes as "T1-correction"), or `MIR` (aka minimum image regression).  I don't recall the origin of the term "minimum image regression", beyond that regressing out a minimum image from a time series is very much a step in the function. I think it might have come from the original MEICA code, but please correct me if I'm wrong on that.

In any case, the method, while recommended in MEICA, is not actually documented in a publication AFAIK. This means that, unfortunately, the manuscript I'm working on might end up being the first detailed description of the method. To make things as clear as possible, I'd like to standardize how we talk about it, both in the documentation and in the code itself.

### Next Steps
1. Rename `gscontrol_mmix` to `gscontrol_mir`.
1. Incorporate detailed steps from my manuscript (still in-prep, sorry) into the docstring to make the procedure clearer.
1. Rename outputs to include `MIR` or `mir` instead of `T1c` (except for the map of the T1-like effect, of course).

tsalo:
Closes #605.

- Rename function name, variables, and output files for T1c global signal regression to match new designation as "minimum image regression."
- Change multiline string-based comments to regular comments.
- Clean up some steps (e.g., z-scoring of arrays) in the function.

### DIFF
--- a/tedana/gscontrol.py
+++ b/tedana/gscontrol.py
@@ -9,6 +9,7 @@ from scipy import stats
 from scipy.special import lpmv
 
 from tedana import io, utils
+from tedana.due import due, Doi
 
 LGR = logging.getLogger(__name__)
 RepLGR = logging.getLogger("REPORT")
@@ -107,6 +108,9 @@ def gscontrol_raw(catd, optcom, n_echos, ref_img, out_dir='.', dtrank=4):
     return dm_catd, dm_optcom
 
 
+@due.dcite(Doi("10.1073/pnas.1301725110"),
+           description="Minimum image regression to remove T1-like effects "
+                       "from the denoised data.")
 def minimum_image_regression(optcom_ts, mmix, mask, comptable, ref_img, out_dir="."):
     """
     Perform minimum image regression (MIR) to remove T1-like effects from
--- a/tedana/gscontrol.py
+++ b/tedana/gscontrol.py
@@ -5,15 +5,14 @@ import logging
 import os.path as op
 
 import numpy as np
-from numpy.linalg import lstsq
 from scipy import stats
 from scipy.special import lpmv
 
 from tedana import io, utils
 
 LGR = logging.getLogger(__name__)
-RepLGR = logging.getLogger('REPORT')
-RefLGR = logging.getLogger('REFERENCES')
+RepLGR = logging.getLogger("REPORT")
+RefLGR = logging.getLogger("REFERENCES")
 
 
 def gscontrol_raw(catd, optcom, n_echos, ref_img, out_dir='.', dtrank=4):
@@ -108,7 +107,7 @@ def gscontrol_raw(catd, optcom, n_echos, ref_img, out_dir='.', dtrank=4):
     return dm_catd, dm_optcom
 
 
-def minimum_image_regression(optcom_ts, mmix, mask, comptable, ref_img, out_dir='.'):
+def minimum_image_regression(optcom_ts, mmix, mask, comptable, ref_img, out_dir="."):
     """
     Perform minimum image regression (MIR) to remove T1-like effects from
     BOLD-like components.
@@ -144,14 +143,15 @@ def minimum_image_regression(optcom_ts, mmix, mask, comptable, ref_img, out_dir=
     meica_mix_MIR.1D          T1 global signal-corrected mixing matrix
     ======================    =================================================
     """
-    LGR.info('Performing minimum image regression to remove spatially '
-             'diffuse noise')
-    RepLGR.info("Minimum image regression was then applied to the "
-                "data in order to remove spatially diffuse noise.")
+    LGR.info("Performing minimum image regression to remove spatially-diffuse noise")
+    RepLGR.info(
+        "Minimum image regression was then applied to the "
+        "data in order to remove spatially diffuse noise."
+    )
 
     all_comps = comptable.index.values
-    acc = comptable[comptable.classification == 'accepted'].index.values
-    ign = comptable[comptable.classification == 'ignored'].index.values
+    acc = comptable[comptable.classification == "accepted"].index.values
+    ign = comptable[comptable.classification == "ignored"].index.values
     not_ign = sorted(np.setdiff1d(all_comps, ign))
 
     optcom_masked = optcom_ts[mask, :]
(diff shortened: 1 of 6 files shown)


tedana/gscontrol.py
```diff
@@ -137,77 +137,67 @@ def gscontrol_mmix(optcom_ts, mmix, mask, comptable, ref_img, out_dir='.'):
...
 
     all_comps = comptable.index.values
-    acc = comptable[comptable.classification == 'accepted'].index.values
-    ign = comptable[comptable.classification == 'ignored'].index.values
+    acc = comptable[comptable.classification == "accepted"].index.values
+    ign = comptable[comptable.classification == "ignored"].index.values
     not_ign = sorted(np.setdiff1d(all_comps, ign))
 
     optcom_masked = optcom_ts[mask, :]
     optcom_mu = optcom_masked.mean(axis=-1)[:, np.newaxis]
```
jbteves, line 158: If we're naming everything based on stat names, shouldn'lt we make this `optcom_mean` instead of `optcom_mu`? I figure since we have other style changes in this PR, this would be a good one to change this for.

tedana/gscontrol.py
```diff
@@ -137,77 +137,67 @@ def gscontrol_mmix(optcom_ts, mmix, mask, comptable, ref_img, out_dir='.'):
...
+    comp_pes = np.linalg.lstsq(mmix, optcom_z.T, rcond=None)[0].T
+    resid = optcom_z - np.dot(comp_pes[:, not_ign], mmix[:, not_ign].T)
 
-    """
-    Build BOLD time series without amplitudes, and save T1-like effect
-    """
-    bold_ts = np.dot(cbetas[:, acc], mmix[:, acc].T)
-    t1_map = bold_ts.min(axis=-1)
+    # Build BOLD time series without amplitudes, and save T1-like effect
+    mehk_ts = np.dot(comp_pes[:, acc], mmix[:, acc].T)
```
jbteves, line 167: It's a little counter-intuitive to me to change this from "bold_ts" to "mehk_ts" (mid-elbow, high-kappa?) but leaving the comment the same. I'd recommend either changing the comment or reverting the variable name change.

jbteves (changes requested):
Only one slight change requested, otherwise looks good!

tedana/gscontrol.py
tsalo: You're totally right. I will improve the comment. I switched to `MEHK` (multi-echo high kappa) because "BOLD" is so unclear when we're dealing with multi-echo data.

tedana/gscontrol.py
```diff
@@ -5,15 +5,14 @@
...
-from numpy.linalg import lstsq
 from scipy import stats
 from scipy.special import lpmv
 
 from tedana import io, utils
 
 LGR = logging.getLogger(__name__)
-RepLGR = logging.getLogger('REPORT')
-RefLGR = logging.getLogger('REFERENCES')
+RepLGR = logging.getLogger("REPORT")
```
emdupre, line 14: Was this just for style ?

tedana/gscontrol.py
```diff
@@ -137,77 +137,67 @@ def gscontrol_mmix(optcom_ts, mmix, mask, comptable, ref_img, out_dir='.'):
...
 
-    """
-    Compute temporal regression
-    """
-    data_norm = (optcom_masked - optcom_mu) / optcom_std
-    cbetas = lstsq(mmix, data_norm.T, rcond=None)[0].T
-    resid = data_norm - np.dot(cbetas[:, not_ign], mmix[:, not_ign].T)
+    # Compute temporal regression
+    optcom_z = stats.zscore(optcom_masked, axis=-1)
+    comp_pes = np.linalg.lstsq(mmix, optcom_z.T, rcond=None)[0].T
```
emdupre, line 163: What is `comp_pes` short for ?

tedana/gscontrol.py
emdupre: Thanks, I think that helps !! On the same point : can we write out MEHK here ?

tedana/gscontrol.py
```diff
@@ -137,77 +137,67 @@ def gscontrol_mmix(optcom_ts, mmix, mask, comptable, ref_img, out_dir='.'):
...
+    resid = optcom_z - np.dot(comp_pes[:, not_ign], mmix[:, not_ign].T)
 
-    """
-    Build BOLD time series without amplitudes, and save T1-like effect
-    """
-    bold_ts = np.dot(cbetas[:, acc], mmix[:, acc].T)
-    t1_map = bold_ts.min(axis=-1)
+    # Build time series of just BOLD-like components (i.e., MEHK) and save T1-like effect
+    mehk_ts = np.dot(comp_pes[:, acc], mmix[:, acc].T)
+    t1_map = mehk_ts.min(axis=-1)  # map of T1-like effect
```
emdupre, line 168: I know that this is a bigger question, but if we can maybe mention in the docstring that this is how we derive the T1-like effect ? Or should we change this comment / variable to reference the "minimum image" ?

tedana/gscontrol.py
```diff
@@ -137,77 +137,67 @@ def gscontrol_mmix(optcom_ts, mmix, mask, comptable, ref_img, out_dir='.'):
...
-                               glob_sig, mmixnogs_norm])
-
-    """
-    Write T1-GS corrected components and mixing matrix
-    """
-    cbetas_norm = lstsq(mmixnogs_norm.T, data_norm.T, rcond=None)[0].T
-    io.filewrite(utils.unmask(cbetas_norm[:, 2:], mask),
-                 op.join(out_dir, 'betas_hik_OC_T1c'), ref_img)
-    np.savetxt(op.join(out_dir, 'meica_mix_T1c.1D'), mmixnogs)
+    # Find the global signal based on the T1-like effect
```
emdupre, line 173: I'm not sure why the diff is splotting like this, but just to confirm: None of the functionality is changed here, no ?

emdupre: (From my look-through I'd think not, but wanted to double-check !

emdupre (commented):
Thanks, @tsalo ! I think this looks great.
I had a few small clarifications / questions.

tedana/gscontrol.py
tsalo: Yeah, I ran `black` and `isort` on the file and then only kept changes from the top of the file and the `minimum_image_regression` function.

tsalo: Scratch that- I don't think I ran `isort`. But I _did_ run `black`.

tedana/gscontrol.py
tsalo: component parameter estimates

tedana/gscontrol.py
tsalo: Yeah, that's weird... You're correct, though, that the functionality shouldn't change.

tedana/gscontrol.py
jbteves: I think adding to the docstring would be better, because it'll also show up in the API docs.

tedana/gscontrol.py
jbteves: It's splitting like that because of the change from """ -> # 
They're all so close together that the heuristic is counting it as big text blobs since it's multi-line.

jbteves (dismissed):
This looks good, thanks @tsalo !

tedana/gscontrol.py
tsalo: I've improved the docstring.

tsalo:
@jbteves I hate to discard your review, but I improved the docstring and added a reference to make things clearer. I'd love to know what you and @emdupre think of the new section in Notes.

jbteves:
I think it's a good addition, thanks!

jbteves (approved):

tedana/gscontrol.py
emdupre:
```suggestion
    comp_pes = np.linalg.lstsq(mmix, optcom_z.T, rcond=None)[0].T  # component parameter estimates
```

emdupre (approved):
This looks great !! I had one small suggestion, but it's not worth holding up this PR over. Happy to see this in :+1:

tsalo:
Thanks @emdupre and @jbteves. I'll wait until the CI passes and then I'll merge. No need to re-review.

PR #623, opened 2020-11-20
[REF] Oop metric mod

### LINKED ISSUE #501: Look for inefficiencies in the interaction between unmasking and spatial clustering

jbteves:
Closes #501 .

- Create Metric class to simplify some data I/O
- Based off of #591, and designed to simplify some of the operation of decision tree modularization.

Happy to talk about this on the call tomorrow. You can see my start on it in `metrics.py`
Remaining TODO if we like this approach:
- [ ] further modularize some functions
- [ ] revise tests
- [ ] revise workflow


jbteves:
Alright, unfortunately I think my initial assessment of how to approach this was wrong. I'm going to refocus efforts on instead helping test #591 and making it possible for us to extract some metrics from there. Once we've tested it and had some time to analyze data with the "new" `tedana`, it can inform what parts of the code would be really nice to have even more modularized. Closing, but if we decide to modify metrics extensively we can discuss some of the ideas I had.

PR #627, opened 2020-12-07
[TST] Allow CI for all-contributors

### LINKED ISSUE #628: Allow CI for all-contributors
### Summary
Right now we skip CI for all-contributors because presumably changing the contributor table shouldn't change the code results. However, it doesn't properly allow us to merge the pull requests as usual because we don't have the passing checks. I propose we simply allow CI, which is simpler than conditionally requiring checks or not, even though it's wasteful of cloud time.

jbteves:
Closes #628 

- Deploys CI for all-contributors so we can review it like normal pull requests

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -221,5 +221,5 @@
   "projectOwner": "ME-ICA",
   "repoType": "github",
   "repoHost": "https://github.com",
-  "skipCi": true
+  "skipCi": false
 }


tsalo:
This seems like a reasonable change. I will approve once tests pass. Can we close and re-request #626 once this is merged as a test?

jbteves:
I'm fine with that. The other thought is that we may want to add credit for bringing to our attention this potential difference in results from different python versions.

tsalo:
Okay, good point. So maybe instead we merge #626 and then make a new request for the bug report once we have an issue for the Python version weirdness.

eurunuela (approved):

handwerkerd:
It would be nicer to have a way to avoid testing the code every time we make a documentation change, but, currently, this seems like a reasonable solution.

handwerkerd (approved):

Issue #617, opened 2020-10-20
Should we have divergent approaches to optimal combination based on the goal?


tsalo:
### Summary
In #600, @cjl2007 and @mvdoc expressed concerns with our new approach to optimal combination, which incorporates echo signal quality information into the process. @emdupre, @rmarkello, and I just had a short call about this, and I think we've come up with a good decision point to bring to the rest of the community.

Namely, the use-cases for the optimally combined data appear to be somewhat different for the `t2smap` and `tedana` workflows. In the `t2smap` workflow, the goal seems to be to use multi-echo fMRI to increase coverage, with a _slight_ increase in SNR. In the `tedana` workflow, we want to limit the optimally combined data to just _good_ data going into the denoising process, in which case coverage will suffer in the OC data, but SNR should increase more than if we incorporated "bad" echoes.

### Additional Detail
Here is what we were thinking.

For tedana, we keep the optimal combination approach the way it is.

For t2smap, there are three possibilities:
1. Revert to the old way, where we run optimal combination using _all_ echoes, regardless of whether some echoes are "bad" or not.
2. Use the echoes with good signal from each voxel, but drop the threshold we use. So instead of only combining voxels with >=3 good echoes and masking out the others, we also combine 2 good echoes and keep values from the first echo for the rest of the voxels.
3. Like number 2, but for voxels with only one good echo, we use the first two echoes, like we do with the "full" T2*/S0 maps.

### Next Steps
1. Discuss.
2. Add an argument to the optimal combination function that toggles this behavior.

emdupre:
I'd agree with splitting the method used depending on the end-goal of the application.

Personally, I think that (3) is perhaps the most principled (in line with the reasoning we use for T2*/S0 maps), but the benefit I see to (1) is that it's likely what users who are familiar with ME-ICA are expecting.

jbteves:
I'm against option (1) if we're pretty sure that the bad echoes are bad, though I also see the merit in including it for legacy reasons. If we're not as sure then we should probably discuss that as well.

Curious what @handwerkerd and @CesarCaballeroGaudes in particular think about this.

mvdoc:
[...]

PR #635, opened 2020-12-16
[ENH] Add threshold argument to make_adaptive_mask

tsalo:
References #617.

- Add threshold argument to `make_adaptive_mask`.
- Hardcode threshold at 1 for the t2smap workflow and 3 for the tedana workflow.
- When deriving the binary mask from the adaptive mask, change threshold from 3 to 0 (basically assuming that the adaptive mask is thresholded).
- Improve associated documentation a bit.
- Update the `make_adaptive_mask` test.

### DIFF
--- a/tedana/combine.py
+++ b/tedana/combine.py
@@ -131,7 +131,9 @@ def make_optcom(data, tes, adaptive_mask, t2s=None, combmode='t2s', verbose=True
     tes : (E,) :obj:`numpy.ndarray`
         Array of TEs, in seconds.
     adaptive_mask : (S,) :obj:`numpy.ndarray`
-        Adaptive mask of the data indicating the number of echos with signal at each voxel
+        Adaptive mask of the data indicating the number of echos with signal at each voxel.
+        This mask may be thresholded, typically with values less than 3 set to 0 depending
+        on the thresholding method.
     t2s : (S [x T]) :obj:`numpy.ndarray` or None, optional
         Estimated T2* values. Only required if combmode = 't2s'.
         Default is None.
@@ -211,7 +213,7 @@ def make_optcom(data, tes, adaptive_mask, t2s=None, combmode='t2s', verbose=True
                    'estimates')
         LGR.info(msg)
 
-    mask = adaptive_mask >= 3
+    mask = adaptive_mask > 0
     data = data[mask, :, :]  # mask out unstable voxels/samples
     tes = np.array(tes)[np.newaxis, ...]  # (1 x E) array_like
     combined = np.zeros((data.shape[0], data.shape[2]))
--- a/tedana/combine.py
+++ b/tedana/combine.py
@@ -137,7 +137,7 @@ def make_optcom(data, tes, adaptive_mask, t2s=None, combmode='t2s', verbose=True
         Array of TEs, in seconds.
     adaptive_mask : (S,) :obj:`numpy.ndarray`
         Adaptive mask of the data indicating the number of echos with signal at each voxel.
-        This mask may be thresholded, typically with values less than 3 set to 0 depending
+        This mask may be thresholded; for example, with values less than 3 set to 0 depending
         on the thresholding method.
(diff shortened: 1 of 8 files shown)


jbteves (changes requested):
The code looks good! However, I think the existing description of
> This mask may be thresholded, typically with values less than 3 set to 0 depending
        on the thresholding method.

Is a tad confusing, and it's in many places. I'm not sure the precise best way to address this, but maybe through either a piece of API documentation which each docstring can reference?

tsalo:
I could include a "see also" to the `make_adaptive_mask` function in the docstrings of functions that use the adaptive mask. WDYT?

jbteves:
That's perfect.

jbteves (approved):
LGTM; thanks @tsalo !

tedana/utils.py
```diff
@@ -90,19 +93,20 @@ def make_adaptive_mask(data, mask=None, getsum=False):
...
-        mask = (masksum >= 3).astype(bool)
         # reduce mask based on masksum
         # TODO: Use visual report to make checking the reduced mask easier
-        if np.any(masksum[mask] < 3):
-            n_bad_voxels = np.sum(masksum[mask] < 3)
+        if np.any(masksum[mask] < threshold):
+            n_bad_voxels = np.sum(masksum[mask] < threshold)
             LGR.warning('{0} voxels in user-defined mask do not have good '
                         'signal. Removing voxels from mask.'.format(n_bad_voxels))
+            masksum[masksum < threshold] = 0
```
notZaki, line 109: I am not completely familiar with where the adaptive mask is used, but could adding the following line
```python
masksum[maskum == 1] = 2
```
simplify the downstream steps or would it break other parts of the code?

tsalo: I think it depends on where we want to bury the "treat one good echo as two good echoes" trick. The way it is now, we can dig into where the adaptive mask indicates only one good echo, and it's on the other functions to document that those voxels will be treated as if they have two good echoes (depending on the threshold). On the other hand, if we add that, then the other functions can be simplified a bit, but we won't be able to investigate the adaptive mask as well.

notZaki: That makes sense. Thanks.

tedana/combine.py
```diff
@@ -137,7 +136,9 @@ def make_optcom(data, tes, adaptive_mask, t2s=None, combmode='t2s', verbose=True
     tes : (E,) :obj:`numpy.ndarray`
         Array of TEs, in seconds.
     adaptive_mask : (S,) :obj:`numpy.ndarray`
-        Adaptive mask of the data indicating the number of echos with signal at each voxel
+        Adaptive mask of the data indicating the number of echos with signal at each voxel.
+        This mask may be thresholded, typically with values less than 3 set to 0 depending
```
emdupre, line 140: Docstring suggestion for clarity :point_down: 

```suggestion
        This mask may be thresholded; for example, with values less than 3 set to 0.
        For more information on thresholding, see `make_adaptive_mask`.
```

tedana/decay.py
```diff
@@ -73,7 +73,8 @@ def fit_monoexponential(data_cat, echo_times, adaptive_mask, report=True):
         Echo times in milliseconds.
     adaptive_mask : (S,) :obj:`numpy.ndarray`
         Array where each value indicates the number of echoes with good signal
-        for that voxel.
+        for that voxel. This mask may be thresholded, typically with values
```
emdupre, line 76: Same comment as above re: docstring suggestion. Won't comment on other instances !

emdupre (dismissed):
Thanks, @tsalo ! This approach makes sense to me. I made a light suggestion on the docstring, but not a strong feeling.

I think this will definitely help improve the T2*map use case !!

jbteves (dismissed):

tsalo:
Thanks @emdupre! I'm going to count both approvals and merge once tests pass.

PR #636, opened 2020-12-29
[FIX] Normalize data to zero mean and unit variance before dimension estimation

notZaki:
References #653 

Input data is normalized prior to dimension estimation in the gift/matlab implementation, but not in `ma_pca`. 
There is a normalized array `data_z` but it isn't used as the input to `ma_pca`:
https://github.com/ME-ICA/tedana/blob/45e5ad7ab23f89dad091c3f0b6c7cc61062a0acc/tedana/decomposition/pca.py#L186-L193

Not sure if this was intentional, but this PR normalizes the input in `ma_pca`.


tsalo:
Hi @notZaki. Thank you for noticing this bug. @eurunuela and I are actually working on moving the MA-PCA code out of tedana and into another repository, so that folks can use it without having to install tedana. We're just waiting on licensing info from the GIFT devs before we can make that repository public.

In the meantime, I would recommend changing the data that `tedpca` feeds into `ma_pca` rather than changing `ma_pca` itself.

notZaki:
Sounds good. In that case, I'll hold off on this PR until the new repository becomes public.

The matlab code mentions GPLv2 at the top of the script, but the year is listed as 2003-2009, so it's probably better that you're confirming with the devs.

notZaki:
I re-opened this PR because there's something I needed to check with the CI.

eurunuela:
I'm sorry I haven't responded before.

We wrote this code on November 2019 and it's been some time so I cannot quite remember the thinking on some of the decisions made during the hackathon. I don't know why we added [line 483](https://github.com/ME-ICA/tedana/blob/dae9f20d80c6aa365d443faac55c2db3b7b94374/tedana/decomposition/ma_pca.py#L483). In the original GIFT code, they do z-score the data before estimating the number of components to keep, so my guess is we were testing how the z-scoring affected the decomposition and forgot to remove that line.

Now, since we're working on `maPCA` on its own repo, I'd add an if statement to make sure the input data is z-scored, and apply [line 482](https://github.com/ME-ICA/tedana/blob/dae9f20d80c6aa365d443faac55c2db3b7b94374/tedana/decomposition/ma_pca.py#L482) when it's not.

jbteves:
@eurunuela now that you bring that up, I actually do remember that while I was reviewing the code because I'm the one who made you put in that TODO. So that's why it's there.

eurunuela:
> @eurunuela now that you bring that up, I actually do remember that while I was reviewing the code because I'm the one who made you put in that TODO. So that's why it's there.

I guess jetlag happened and we just forgot we had to do that. Sorry about that.

Next steps before we merge this PR should be:

- Use `data_z` in [pca.py line 190](https://github.com/ME-ICA/tedana/blob/45e5ad7ab23f89dad091c3f0b6c7cc61062a0acc/tedana/decomposition/pca.py#L190)
- Add an if statement that checks the data is already z-scored in [mapca.py lines 481-483](https://github.com/ME-ICA/tedana/blob/dae9f20d80c6aa365d443faac55c2db3b7b94374/tedana/decomposition/ma_pca.py#L481-483)

Would you like to do that @notZaki ? Or would you prefer I push changes to this PR?

jbteves:
@eurunuela I'd open an issue in the mapca repo cross-referencing this one so we don't forget a second time ; - )

eurunuela:
Now that I think about it I'm starting to remember why we used `data` instead of `data_z`. For maPCa we only want to z-score in time, while tedana z-scores both in time and space before the PCA step, hence the normalization step inside maPCA.

CesarCaballeroGaudes:
Yes, I also remember that.

eurunuela:
I've been looking into the different normalization approaches and how they differ. I've checked the one used in GIFT, the sklearn scaler we added during the hackathon, and the z-score applied to the temporal dimension. Here's an overview of what I've found:

I will use this dummy matrix to make the comparisons:

```python
array([[4, 1, 8, 8],
       [7, 3, 7, 5]])
```

## GIFT

This is how GIFT normalizes variance ([see here](https://github.com/trendscenter/gift/blob/1a8012eab1c40d4543b5b3a8199717ac29a60242/GroupICATv4.0c/icatb/icatb_estimate_dimension.m#L209-211)):

```matlab
    for n = 1:size(data, 2)
        data(:, n) = detrend(data(:, n), 0) ./ std(data(:, n));
    end
```
This normalization returns the following matrix:

```python
array([[-0.70710678, -0.70710678,  0.70710678,  0.70710678],
       [ 0.70710678,  0.70710678, -0.70710678, -0.70710678]])
```

## Sklearn scaler

This is how we implemented it during the hackathon for maPCA ([see here](https://github.com/ME-ICA/tedana/blob/dae9f20d80c6aa365d443faac55c2db3b7b94374/tedana/decomposition/ma_pca.py#L480-482)):

```python
scaler = StandardScaler(with_mean=True, with_std=True)
scaler.fit_transform(data)
```

And it returns:

```python
array([[-1., -1.,  1.,  1.],
       [ 1.,  1., -1., -1.]])
```

## Z-score in temporal dimension

This is how tedana does z-score in the temporal dimension ([see here](https://github.com/ME-ICA/tedana/blob/dae9f20d80c6aa365d443faac55c2db3b7b94374/tedana/decomposition/pca.py#L186)):

```python
data_z = ((data.T - data.T.mean(axis=0)) / data.T.std(axis=0)).T
```

And it returns:

```python
array([[-0.42409446, -1.44192118,  0.93300782,  0.93300782],
       [ 0.90453403, -1.50755672,  0.90453403, -0.30151134]])
```

## Conclusion

It is clear that the three approaches yield very different results. 

I've gone through the paper again and having zero mean and unit variance in the temporal dimension is what makes sense. Mathematically speaking, we're interested in calculating the entropy rate of a Gaussian random process (x[n], n=1,2,...,N; where N is the number of samples,  E{x[n]} = 0 and E{x^2[n]} = 1).

To me, z-scoring in the temporal domain makes the most sense and is what I'd do (and that's what has been suggested so far in this PR). However, I wanted to make sure we all see what exaclty the different approaches are doing.

notZaki:
Thanks for sharing the results @eurunuela 

I think the difference between GIFT and StandardScaler will become negligible with a larger matrix. They seem very different on the example data because the first dimension (space) only has two values, so the sample and population standard deviation produce very different values.

If z-scoring in the temporal domain is indeed what makes sense, then perhaps this can be relayed to the GIFT group to confirm whether this was a bug or intential. This would be more relevant for the mapca repo because it would be diverging from GIFT.

jbteves:
@eurunuela would you mind quickly opening an issue explaining briefly what `tedana`'s options and current behavior are? I'm afraid that your comment will be missed in a PR rather than an issue.

eurunuela:
> I think the difference between GIFT and StandardScaler will become negligible with a larger matrix. They seem very different on the example data because the first dimension (space) only has two values, so the sample and population standard deviation produce very different values.

You're right. I didn't think of it as I was just trying to keep the demonstration simple. I've just tested the same commands with a bigger matrix (20000 x 100) and the results between GIFT and sklearn are very similar. The biggest difference is in the order of `4e-5`. 
 
> If z-scoring in the temporal domain is indeed what makes sense, then perhaps this can be relayed to the GIFT group to confirm whether this was a bug or intentional. This would be more relevant for the mapca repo because it would be diverging from GIFT.

There will be no need for that. I've checked that the sklearn scaler z-scores in the temporal domain, so we're safe.

notZaki:
> There will be no need for that. I've checked that the sklearn scaler z-scores in the temporal domain, so we're safe.

I thought that the [earlier results](https://github.com/ME-ICA/tedana/pull/636#issuecomment-763166064) were showing that sklearn scaler is z-scoring in spatial domain instead of the temporal domain.

eurunuela:
> I thought that the [earlier results](https://github.com/ME-ICA/tedana/pull/636#issuecomment-763166064) were showing that sklearn scaler is z-scoring in spatial domain instead of the temporal domain.

I´ve looked into that in more depth, see #655

eurunuela (approved):
Thank you @notZaki ! I think we should have this PR in our next release to ensure both normalization steps are done.

emdupre:
> Thank you @notZaki ! I think we should have this PR in our next release to ensure both normalization steps are done.

Thanks everyone for digging into this ! Just to be clear: is this PR ready for reviews ?

eurunuela:
> Thanks everyone for digging into this ! Just to be clear: is this PR ready for reviews ?

I think it's ready to get merged, so yes.

jbteves:
@notZaki could you please confirm by switching this PR from "Draft" to "Ready for Review"?

notZaki:
On my end, I am not completely sure if the normalization should be done along voxels or time. 
If there is no doubt that normalization should be along time, then this is ready to get merged. Otherwise, I think the discussion in #653 and #655 need to be resolved first.

notZaki:
Revisiting this: If the next release is scheduled for next week, then some kind of change should be made because mdl is the default option and I think everyone agrees that the current version (no normalization) is not ideal.  

Options are:

- (A) Normalize in temporal domain
  + Pro: Consistent with how tedana z-scores elsewhere in `decomposition/pca`
- (B) Normalize along spatial dimension
  + Pro: This is what GIFT does

I am learning towards option B because I think it's safer to follow an established library (GIFT), but I'm curious what everyone else's thoughts are.

eurunuela:
Why don't we go with option B (safest and it will select more components) and study option A on the `maPCA` repo?

eurunuela (approved):
Thank you @notZaki

eurunuela:
Can we please merge this before we cut the release? It makes sure that we're normalizing the data as the first step to maPCA (even if the results don't vary too much).

@emdupre @tsalo @jbteves

emdupre:
> Can we please merge this before we cut the release?

Yes, I think we should ! @notZaki had made a great point re: which to use, and I wanted to drag out my old Linear Algebra text to confirm. But I can commit to doing that by Wednesday, which should still leave us enough time to get it merged before cutting on Friday !

eurunuela:
> > Can we please merge this before we cut the release?
> 
> Yes, I think we should ! @notZaki had made a great point re: which to use, and I wanted to drag out my old Linear Algebra text to confirm. But I can commit to doing that by Wednesday, which should still leave us enough time to get it merged before cutting on Friday !

Sounds great! Right now, the PR mimics GIFT. I will have a look at the papers too just to make sure.

emdupre (approved):
Thanks @notZaki ! :rocket:

eurunuela:
Thanks @notZaki and @emdupre .  I'm merging this PR.

PR #641, opened 2021-01-03
[REF] Replace MAPCA code with mapca library

### LINKED ISSUE #698: Replace internal MAPCA code with mapca library
### Summary
We just released 0.0.1 of `mapca` (:tada:), so we can now excise the MAPCA code from `tedana` and add `mapca` as a dependency.

### Additional Detail
I already have a draft PR open (#641), but it's very out-of-date and I might just open a new one.

tsalo:
Closes #698.

- Remove MA-PCA-related code from `tedana`.
- Use new `mapca` library (once it's released on PyPi).


jbteves (dismissed):
This is GREAT! Thanks so much for everyone who moved it into its own package, great job!

notZaki:
The output might be different because `mapca` performs normalization along temporal-domain, while the implementation in `tedana` is along spatial-domain. At least, I think that's the case.

The license might also be worth discussing: `mapca` is currently licensed under GPLv2, and I *think* that merging it will require `tedana` to also change to GPL, and then any downstream dependencies such as `fmriprep`, would similarly have to change to GPL in order to use the next release of `tedana`.

eurunuela (dismissed):
Changes look good, thanks @tsalo ! Something is wrong with the 3 echo test though.

emdupre:
> The license might also be worth discussing: mapca is currently licensed under GPLv2, and I think that merging it will require tedana to also change to GPL, and then any downstream dependencies such as fmriprep, would similarly have to change to GPL in order to use the next release of tedana.

Changing the license is really a non-starter for both fMRIPrep and tedana, since it requires getting all existing contributors to the packages to re-consent to have their code release under the new license.

tsalo:
>The license might also be worth discussing: mapca is currently licensed under GPLv2, and I think that merging it will require tedana to also change to GPL, and then any downstream dependencies such as fmriprep, would similarly have to change to GPL in order to use the next release of tedana.

I don't believe licensing impacts dependencies (i.e., I can have any licensed code as a dependency, but when including the actual code in my library it does matter). If anything, it was just a problem that we had MAPCA code _within_ tedana _before_.

jbteves:
The good news is that `tedana` is already GPLv2.1, so we don't have to change anything.
Per my understanding of the licensing: fMRIPrep can include `tedana` provided that there is a notice that `tedana` code is licensed under GPL. Since we're packaged on PyPI with a license notice both in the repository and on PyPI, they should be fine. They would run into trouble only if they started distributing our code with a different license (which they don't). If you're curious, they use the BSD license (allows derivative works without assent).

notZaki:
> I don't believe licensing impacts dependencies (i.e., I can have any licensed code as a dependency, but when including the actual code in my library it does matter). If anything, it was just a problem that we had MAPCA code _within_ tedana _before_.

True, the situation is kinda sticky with the current `ma_pca` implementation in `tedana`.

> The good news is that `tedana` is already GPLv2.1, so we don't have to change anything.

My understanding was that `tedana` uses L-GPL, which does not have the same restrictions as GPL [[ref]](https://www.gnu.org/licenses/gpl-faq.html#IfLibraryIsGPL).

Based off [this table/matrix](https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility) it's ok to use a GPL library (`mapca`) inside L-GPL code (`tedana`) but the combination must be relicensed under the GPL.

jbteves (approved):

eurunuela (approved):
LGTM! Thanks @tsalo

PR #644, opened 2021-01-13
[FIX] Fix z statistic computaiton in computefeats2

### LINKED ISSUE #178: Calculating coefficients in computefeats2

smoia:
Closes #178 .

@tsalo, apologies for the delay! The branch is incredibly behind `ME-ICA/tedana/master`, but at first sight I didn't see any conflict related to lack of updates.

Also, let me know if instead of a [FIX] this should be a [REF] or a [ENH].

Also, it's still a draft because I see it's missing some final tweaks.

- **Main change**: removed the possibility of normalising the output of `computecoeffs2` because with the new Z-score implementation it doesn't make much sense anymore.
- **Main change**: added the optional steps of computing z values from beta scores in `get_coeffs` if an argument of the function is flagged as `True`. However, the way the output is returned might not be so pythonic - what do you think?
- **Main change**: Rename `get_coeffs` into `get_ls_coeffs` to make the _least square_ approach more explicit, and update the related API documentation as well as its calls throughout the code. If it's an issue, though, we can revert to the former name.
- Add `t_to_z` function to tedana's `stats.py`, from [Vanessa Sochat's TtoZ](https://github.com/vsoch/TtoZ) package, add its reference for duecredit (but please do check the docstring for copyright). However, this could be avoided by adding the package as dependency and importing it. What do you all think? (Now that I'm not a python newbie, I think I should!)
- Update `test_get_coeffs` to `test_get_ls_coeffs` and the calls to `get_coeffs`. However, it's not testing the new part of `get_ls_coeffs`, only what was previously tested. If you have any suggestion on how to test the new part of the code, let me know.

Things left to do and I might or might not (if not required to merge PR) do [and by extension, question for the reviewer to see if they expect them]:
- [ ] Import TtoZ as dependency rather than merely copying code.
- [ ] Think about the output of `get_ls_coeffs` and whether it should always compute z scores or not
- [ ] Update documentation beside API
- [ ] Test the z score computation (any idea for that?)
- [ ] Add an empty commit to acknowledge @CesarCaballeroGaudes as co-author of this PR since we worked on it together.


tedana/stats.py
```diff
@@ -74,28 +135,20 @@ def computefeats2(data, mmix, mask=None, normalize=True):
...
 
-    # get betas of `data`~`mmix` and limit to range [-0.999, 0.999]
-    data_R = get_coeffs(data_vn, mmix, mask=None)
-    data_R[data_R < -0.999] = -0.999
-    data_R[data_R > 0.999] = 0.999
-
-    # R-to-Z transform
-    data_Z = np.arctanh(data_R)
+    # get betas and z-values of `data`~`mmix`
+    # mmix is normalized internally
```
tsalo, line 142: I think we should do it here anyway.

smoia: What if we use `add_const=True` while calling `get_ls_coeffs` instead, now that there's the option? I think it would make more sense given how the program runs.

In fact, I would set add_const default as True, given that if the data is not demeaned, you need the intercept, and if it's demeaned, the intercept doesn't bother you (it's 0).

@CesarCaballeroGaudes is there any other difference between normalising before computing LS and adding the intercept in the model?

tsalo: I like including the intercept by default, but isn't variance normalization required to get standardized parameter estimates?

tedana/stats.py
```diff
@@ -34,7 +93,7 @@ def getfbounds(n_echos):
     return f05, f025, f01
 
 
-def computefeats2(data, mmix, mask=None, normalize=True):
+def computefeats2(data, mmix, mask=None):
```
tsalo, line 96: I'd love to see this renamed. Perhaps `get_ls_zvalues`? It should also return standardized parameter estimates, but those are relatively unimportant compared to the z-values.

tedana/stats.py
```diff
@@ -34,7 +93,7 @@ def getfbounds(n_echos):
     return f05, f025, f01
 
 
-def computefeats2(data, mmix, mask=None, normalize=True):
+def get_ls_zvalues(data, mmix, mask=None):
     """
     Converts `data` to component space using `mmix`
```
tsalo, line 98:
```suggestion
    """Fits mmix to data with OLS and calculates standardized beta values and z-statistics.
```

tedana/stats.py
```diff
@@ -150,11 +206,45 @@ def get_coeffs(data, X, mask=None, add_const=False):
...
+                      'calculation. Stopping!!')
+        elif df <= min_df:
+            LGR.warning('Number of degrees of freedom in least-square '
+                        'estimation is less than {}'.format(min_df + 1))
+        # compute sigma:
+        # RSS = sum{[mdata - (X * betas)]^2}
+        # sigma = RSS / Degrees_of_Freedom
+        sigma = np.sum(np.power(mdata - np.dot(X, betas.T), 2), axis=0) / df
+        sigma = sigma[:, np.newaxis]
+        # Copmute std of betas:
```
tsalo, line 228:
```suggestion
        # compute std of betas:
```

tedana/stats.py
```diff
@@ -7,11 +7,70 @@
...
+@due.dcite(T2Z_TRANSFORM,
+           description='Introduces T-to-Z transform.')
+@due.dcite(T2Z_IMPLEMENTATION,
+           description='Python implementation of T-to-Z transform.')
+def t_to_z(t_values, dof):
+    """
+    From Vanessa Sochat's TtoZ package.
+    Copyright (c) 2015 Vanessa Sochat
+    MIT Licensed
+    """
```
tsalo, line 41: Just recommending the docstring I use in `NiMARE`.

```suggestion
    """Convert t-statistics to z-statistics.

    An implementation of [1]_ from Vanessa Sochat's TtoZ package [2]_.

    Parameters
    ----------
    t_values : array_like
        T-statistics
    dof : int
        Degrees of freedom

    Returns
    -------
    z_values : array_like
        Z-statistics

    References
    ----------
    .. [1] Hughett, P. (2007). Accurate Computation of the F-to-z and t-to-z
           Transforms for Large Arguments. Journal of Statistical Software,
           23(1), 1-5.
    .. [2] Sochat, V. (2015, October 21). TtoZ Original Release. Zenodo.
           http://doi.org/10.5281/zenodo.32508
    """
```

tsalo: The actual code from NiMARE might also be good to use since I improved the internal variable names.

PR #645, opened 2021-01-14
[FIX] Eliminate duplicate lines in logs

### LINKED ISSUE #638: Duplicated text in reports
### Summary
At minimum, the optimal combination step is duplicated in our report text and visual reports. We should review the logging code to ensure that report text is only logged once, as appropriate.

In the five-echo test's current report (emphasis mine):

> TE-dependence analysis was performed on input data. An adaptive mask was then generated, in which each voxel's value reflects the number of echoes with 'good' data. **Multi-echo data were then optimally combined using the T2\* combination method (Posse et al., 1999). Multi-echo data were then optimally combined using the T2\* combination method (Posse et al., 1999). Multi-echo data were then optimally combined using the T2\* combination method (Posse et al., 1999).** Next, components were manually classified as BOLD (TE-dependent), non-BOLD (TE-independent), or uncertain (low-variance). This workflow used numpy (Van Der Walt, Colbert, & Varoquaux, 2011), scipy (Jones et al., 2001), pandas (McKinney, 2010), scikit-learn (Pedregosa et al., 2011), nilearn, and nibabel (Brett et al., 2019). This workflow also used the Dice similarity index (Dice, 1945; Sørensen, 1948).

### Next Steps
1. Review logging code
2. Eliminate duplicated text by adding a `report` argument to the relevant functions and setting to `False` when called 2+ times. See `fit_loglinear` for an example:
https://github.com/ME-ICA/tedana/blob/45e5ad7ab23f89dad091c3f0b6c7cc61062a0acc/tedana/decay.py#L158

tsalo:
Closes #638.

- Add a `report` argument to relevant functions so that we can toggle whether a given function will add to the report logger. In cases where the same function is called multiple times, we turn the report off.

### DIFF
--- a/tedana/combine.py
+++ b/tedana/combine.py
@@ -14,7 +14,7 @@ RefLGR = logging.getLogger('REFERENCES')
 @due.dcite(Doi('10.1002/(SICI)1522-2594(199907)42:1<87::AID-MRM13>3.0.CO;2-O'),
            description='T2* method of combining data across echoes using '
                        'monoexponential equation.')
-def _combine_t2s(data, tes, ft2s):
+def _combine_t2s(data, tes, ft2s, report=True):
     """
     Combine data across echoes using weighted averaging according to voxel-
     (and sometimes volume-) wise estimates of T2*.
@@ -27,6 +27,8 @@ def _combine_t2s(data, tes, ft2s):
         Echo times in milliseconds.
     ft2s : (M [x T] X 1) array_like
         Either voxel-wise or voxel- and volume-wise estimates of T2*.
+    report : bool, optional
+        Whether to log a description of this step or not. Default is True.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


tedana/combine.py
```diff
@@ -215,17 +221,19 @@ def make_optcom(data, tes, adaptive_mask, t2s=None, combmode='t2s', verbose=True
...
             combined[echo_idx, :] = _combine_paid(data[echo_idx, :echo, :],
-                                                  tes[:echo])
+                                                  tes[:echo], report=report)
         else:
             t2s_ = t2s[mask, ..., np.newaxis]  # mask out empty voxels/samples
 
             combined[echo_idx, :] = _combine_t2s(
-                data[echo_idx, :echo, :], tes[:, :echo], t2s_[echo_idx, ...])
+                data[echo_idx, :echo, :], tes[:, :echo], t2s_[echo_idx, ...], report=report)
+        report = False
```
jbteves, line 236: Should this actually be indented here? This means report is only turned on for the first iteration. Maybe that's the desired behavior, and I'm just confused, but I had thought you would want it on all iterations.

tsalo: We only want to add a line about how the echoes were combined to the report once, but if we leave it on for all iterations then we get duplicates of that line.

jbteves: Ah, yes, okay, I was just being confused. My mistake.

jbteves (approved):
I'm approving because I think this works and I can't think of anything better, but we'll have to be careful about maintaining those blocks of report being True and False. I think it could easily trip us up if we're not careful. But as I said, this still looks good to me!
Thanks @tsalo !

eurunuela (approved):
I'm not sure this is the best solution but I get that right now it's a working solution. We should still think of a more elegant way to deal with the duplicate logs.

PR #658, opened 2021-01-21
[ENH] Implement variance explained threshold-based PCA option

### LINKED ISSUE #492: Add threshold-based PCA option
### Summary
Continuing from #486, we want to add in a simple variance explained threshold-based PCA option.

### Next Steps
1. Decide what threshold we want to use.
2. Add new option to CLI, tedana workflow, PCA function(s), and tests.

tsalo:
Closes #492.

- Add an option to support a cumulative variance explained threshold-based PCA option.
- Update tests.


eurunuela (dismissed):
LGTM! Thank you @tsalo

tedana/workflows/parser_utils.py
```diff
@@ -4,6 +4,26 @@
 import os.path as op
 import logging
 
+import argparse
+
+
+def string_or_float(string):
```
jbteves, line 10: This is really nitpicky but I feel like you're not checking "string or float," you're checking "string or fraction." For example, 1.1 would throw an error here, even though 1.1 is a completely valid float. So out of specificity I think we should consider renaming this (I'm not sure that "fraction" is the best descriptor either).

tedana/workflows/tedana.py
```diff
@@ -387,6 +394,10 @@ def tedana_workflow(data, tes, out_dir='.', mask=None,
     if not isinstance(gscontrol, list):
         gscontrol = [gscontrol]
 
+    # Check value of tedpca *if* it is a float
```
jbteves, line 397: Could we just call `string_or_float` here? From CLI it should get checked automatically, but the Python API won't.

tedana/workflows/parser_utils.py
tsalo: Since it's not a generic function I think naming it `string_or_float` was a mistake. I was originally trying to see if I could feed in the valid options from the parser rather than this function, but since that's not the case and the valid options are hardcoded, what about `check_tedpca_value`?

jbteves:
I'd like to hear your thoughts on my comments @tsalo before approving, but aside from those nitpicks this looks great, thanks!

tedana/workflows/parser_utils.py
jbteves: That sounds much better to me, yes!

jbteves (changes requested):

tedana/workflows/tedana.py
tsalo: Good point!

jbteves (approved):
LGTM! If we end up having a demand for adding a threshold option to other PCA algorithms, we may need to create dependent CLI parameter groups. But I see no reason to cross that bridge until we get there. Thanks again @tsalo !

PR #663, opened 2021-01-26
[ENH] Add ability to re-run ICA when no BOLD components are found

### LINKED ISSUE #662: Add ability to re-run ICA when no/few BOLD components are found
### Summary
When no BOLD components are detected by the decision tree, it is possible that the ICA simply failed to produce components that adequately dissociated BOLD and non-BOLD signal sources by chance. Changing the random seed can help with this. It would be awesome to be able to feed information from the decision tree back into the decomposition (combining with the `maxit` and `maxrestart` parameters) to restart the decomposition with an updated seed.

This would require a substantial refactor, and would probably best work with an object-oriented approach...

### Additional Detail
This stems from #659. In that case, changing the random seed was helpful in some subjects.

### Next Steps

1. Discuss.
2. Come up with an implementation, in particular deciding between our current functional approach and an object-oriented one.
3. Implement.

tsalo:
Closes #662. This is just a first pass, although I think it should work.

- Re-run ICA (within constraints of `maxrestarts` parameter) when no BOLD components are found by the decision tree.

### DIFF
--- a/tedana/decomposition/ica.py
+++ b/tedana/decomposition/ica.py
@@ -38,8 +38,8 @@ def tedica(data, n_components, fixed_seed, maxit=500, maxrestart=10):
     mmix : (T x C) :obj:`numpy.ndarray`
         Z-scored mixing matrix for converting input data to component space,
         where `C` is components and `T` is the same as in `data`
-    n_restarts : :obj:`int`
-        Number of restarts used.
+    fixed_seed : :obj:`int`
+        Random seed from final decomposition.
 
     Notes
     -----
--- a/tedana/decomposition/ica.py
+++ b/tedana/decomposition/ica.py
@@ -66,8 +66,8 @@ def tedica(data, n_components, fixed_seed, maxit=500, maxrestart=10):
 
             w = list(filter(lambda i: issubclass(i.category, UserWarning), w))
             if len(w):
-                LGR.warning('ICA attempt {0} failed to converge after {1} '
-                            'iterations'.format(i_attempt + 1, ica.n_iter_))
+                LGR.warning('ICA with random seed {0} failed to converge after {1} '
+                            'iterations'.format(fixed_seed, ica.n_iter_))
                 if i_attempt < maxrestart - 1:
                     fixed_seed += 1
                     LGR.warning('Random seed updated to {0}'.format(fixed_seed))
--- a/tedana/decomposition/ica.py
+++ b/tedana/decomposition/ica.py
@@ -72,8 +72,8 @@ def tedica(data, n_components, fixed_seed, maxit=500, maxrestart=10):
                     fixed_seed += 1
                     LGR.warning('Random seed updated to {0}'.format(fixed_seed))
             else:
-                LGR.info('ICA attempt {0} converged in {1} '
-                         'iterations'.format(i_attempt + 1, ica.n_iter_))
+                LGR.info('ICA with random seed {0} converged in {1} '
+                         'iterations'.format(fixed_seed, ica.n_iter_))
                 break
 
     mmix = ica.mixing_
--- a/tedana/decomposition/ica.py
+++ b/tedana/decomposition/ica.py
@@ -38,6 +38,8 @@ def tedica(data, n_components, fixed_seed, maxit=500, maxrestart=10):
     mmix : (T x C) :obj:`numpy.ndarray`
         Z-scored mixing matrix for converting input data to component space,
         where `C` is components and `T` is the same as in `data`
+    n_restarts : :obj:`int`
+        Number of restarts used.
 
     Notes
     -----
@@ -76,4 +78,4 @@ def tedica(data, n_components, fixed_seed, maxit=500, maxrestart=10):
 
     mmix = ica.mixing_
     mmix = stats.zscore(mmix, axis=0)
-    return mmix
+    return mmix, fixed_seed
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -524,11 +524,11 @@ def tedana_workflow(data, tes, out_dir='.', mask=None,
         n_restarts = 0
         seed = fixed_seed
         while bad_decomp:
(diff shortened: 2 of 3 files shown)


tsalo:
It looks like it's working! The four-echo test has an ICA that converges nicely each time, but fails to detect any BOLD components for the first three decompositions. On the _fourth_ one, it finds BOLD components! :tada:

One unfortunate thing is that the logs say "ICA attempt 1" for each version. Not sure the cleanest way to fix that though.

tsalo:
I will go through and update the outputs later but I'm going to mark this as ready for review now. If anyone has any ideas on how best to log the number of attempts and/or updated random seeds, please review.

jbteves:
https://github.com/ME-ICA/tedana/blob/dd513ae91c4b907f41d52ead8be01edf05820844/tedana/decomposition/ica.py#L67-L68
Let's just not log the attempt number. Maybe just print seed and indicate success/failure at end?

eurunuela:
I like the changes. I'd like to see Josh's suggestion added and the failing test passing.

eurunuela (dismissed):
LGTM! I think it is a very simple yet effective solution. We can always complicate things more. Thank you @tsalo

tedana/workflows/tedana.py
```diff
@@ -514,21 +514,44 @@ def tedana_workflow(data, tes, out_dir='.', mask=None,
...
+                algorithm='kundu_v2', verbose=verbose
+            )
+            comptable = metrics.kundu_metrics(comptable, metric_maps)
+            comptable = selection.kundu_selection_v2(comptable, n_echos, n_vols)
+
+            n_bold_comps = comptable[comptable.classification == 'accepted'].shape[0]
+            if (n_restarts < maxrestart) and (n_bold_comps == 0):
+                LGR.warning("No BOLD components found. Re-attempting ICA.")
+            elif (n_bold_comps == 0):
+                LGR.warning("No BOLD components found, but maximum number of restarts reached.")
```
notZaki, line 550: Is a `bad_decomp = False` needed here to stop the loop?

tsalo: It is. Good catch. I'll change the variable to `keep_restarting` so it makes sense here, and add it here.

tsalo: Should be good now. WDYT?

tedana/workflows/tedana.py
```diff
@@ -548,8 +548,9 @@ def tedana_workflow(data, tes, out_dir='.', mask=None,
                 LGR.warning("No BOLD components found. Re-attempting ICA.")
             elif (n_bold_comps == 0):
                 LGR.warning("No BOLD components found, but maximum number of restarts reached.")
+                keep_restarting = False
```
jbteves, line 551: Should this actually be `True`? Otherwise you'd stop here, with no BOLD components.

tedana/workflows/tedana.py
notZaki: LGTM!

The last condition in that sequence could be converted to an `else` to be super safe, but I think all the cases are covered as it is.

tedana/workflows/tedana.py
tsalo: Right, but at that point we are out of allotted restarts, so we have to quit.

tedana/workflows/tedana.py
tsalo: Good point. That's done too.

tedana/workflows/tedana.py
jbteves: I'm sorry, for some reason it wasn't rendering the above `if` for me, disregard

eurunuela (approved):

jbteves (approved):
LGTM, thanks @tsalo !

tsalo:
We can revisit the minimum number of BOLD components we want to enforce at some point in the future, but I'm happy with the general approach. Given that we have two approving reviews, I think I'll merge this EOD Thursday if there are no other reviews, to give the other devs a chance to take a look.

tedana/decomposition/ica.py
```diff
@@ -64,16 +66,16 @@ def tedica(data, n_components, fixed_seed, maxit=500, maxrestart=10):
 
             w = list(filter(lambda i: issubclass(i.category, UserWarning), w))
             if len(w):
-                LGR.warning('ICA attempt {0} failed to converge after {1} '
-                            'iterations'.format(i_attempt + 1, ica.n_iter_))
+                LGR.warning('ICA with random seed {0} failed to converge after {1} '
```
emdupre, line 69: Since we're updating this user-facing behavior, we should make sure it's noted somewhere (maybe in the release notes ?).

emdupre (approved):

tedana/decomposition/ica.py
jbteves: I think release notes should be okay for this. It's not a significant change.

emdupre: I know when I used to run tedana mors I would watch the logs, so I worry it could scare those users (is why I do want it mentioned somewhere). I know it would have scared me ! 😅

jbteves: That's completely fair. Any thoughts on where to put it aside from release notes? When we do next release we could package it with the newsletter, too.

tsalo: We could downgrade this from a warning to a debug in a separate PR?

emdupre: In general I was thinking that we don't have a descriptive `whatsnew`. Maybe we could consider starting one ?

For now, I think the release notes are fine ! And I always like the newsletter's summaries :)

jbteves: I actually disagree with downgrading to a debug. I've noticed that failure to converge tends to go along with lower-quality data, so I think it's good to let the user know that the software is struggling a bit.
Would you mind opening an issue for a `whatsnew` @emdupre ? That sounds like a good idea to me, especially if we're planning on increasing the release speed.

PR #665, opened 2021-01-28
[ENH] Allow tedpca argument to be a float inside a string

notZaki:
Closes none.

Got a parser error while trying the variance-based PCA introduced in #658.
I think it's because the inputs to `workflows.tedana.main_` are strings but the parser treats `"0.95"` as an invalid input. 

This PR modifies `workflows.parser_utils.check_tedpca_value` to accept floats-inside-strings. The function got refactored a bit while I was debugging.


tsalo:
This is definitely a much cleaner approach. It looks like there are style issues that need to be address and I might have some changes to the whitespace to request, but overall this looks great!

eurunuela (dismissed):
LGTM!

tedana/workflows/parser_utils.py
```diff
@@ -3,32 +3,27 @@
...
+
+    error = argparse.ArgumentTypeError if is_parser else ValueError
+    try:
+        floatarg = float(string)
+    except ValueError:
+        msg = "Argument must be a float or one of: {}".format(
+            ", ".join(valid_options)
+        )
+        raise error(msg)
+    if not (0 <= floatarg <= 1):
```
tsalo, line 24: I'd love to see space between these two lines.

tsalo (changes requested):
One minor change for readability.

tedana/workflows/parser_utils.py
notZaki: Done!

tsalo (approved):
The changes look great!

eurunuela (approved):

Issue #668, opened 2021-02-02
Review and check packaging before 0.0.9 release


tsalo:
### Summary
This stems from [this NeuroStars comment](https://neurostars.org/t/number-of-components-in-tedana-pca-is-equal-to-the-number-of-epi-volumes/18126/6), in which a user was unable to install `tedana` directly from GitHub. In #595, we started using the README as the source of `long_desc` (long description) in `info.py`, which means that we need to bundle the README in with our package data in order for `tedana` to install correctly. It does not appear to be bundled in just yet, and I'd like to make sure we fix/check this before our release on Friday.

### Additional Detail
I tried to test this by adding the README to `MANIFEST.in` in the [`package-readme`](https://github.com/ME-ICA/tedana/tree/package-readme) branch, but directly installing from GitHub (i.e., `pip install git+https://github.com/ME-ICA/tedana.git@package-readme`) failed for the same reasons described in the NeuroStars post. I don't know if this is unique to installing directly from GitHub or if it will impact our PyPi releases.

### Next Steps
1. Investigate.
2. Fix.
3. Release!

tsalo:
My understanding of the MANIFEST.in file is that it should be used to identify files _outside of the package subfolder_ that you want to bundle with the package in your distribution. On the other hand, `package_data` should be used to identify non-code files _inside_ the package subfolder that you want to bundle with the package.

If anyone knows why adding the README to MANIFEST.in doesn't solve the issue, please respond to this issue.

jbteves:
Interestingly, we shouldn't even need to include it per my read of [this page](https://packaging.python.org/guides/using-manifest-in/), where it states that a README.md should get included by `setuptools`.

tsalo:
Hm, so maybe installing directly from GitHub doesn't use `sdist`?

notZaki:
I think the issue is that the parent folder of the github repository doesn't get copied. Even if it did get copied, the path of the readme file would have to be in the site-packages folder, which is an odd place.

In other words
```
# Folder where tedana gets installed
$SOME_PATH/lib/python3.8/site-packages/tedana’
# Folder where tedana expects the readme to be
$SOME_PATH/lib/python3.8/site-packages/README.md
```

One workaround is to load the readme inside `setup.py` instead of placing that code in the module itself, i.e. don't use `info.py`.
[...]

PR #670, opened 2021-02-03
Move long description logic from info.py to setup.py

notZaki:
References #668 

The `README.md` file is used as the module's long description. The relevant code is defined in `tedana/info.py`. This causes an error when importing the module because it tries to load the readme file, but the readme file does not get copied over once the module is installed

This PR moves the relevant code to `setup.py` so that the readme file is only read when the module is first being installed. This should get rid of the error. Remains to be seen if the long description will correctly show up on pypi.

### DIFF
--- a/tedana/info.py
+++ b/tedana/info.py
@@ -23,7 +23,7 @@ __url__ = 'https://github.com/me-ica/tedana'
 __packagename__ = 'tedana'
 __description__ = ('TE-Dependent Analysis (tedana) of multi-echo functional '
                    'magnetic resonance imaging (fMRI) data.')
-__longdesc__ = readme_path.open().read()
+# __longdesc__ = readme_path.open().read()
 
 DOWNLOAD_URL = (
     'https://github.com/ME-ICA/{name}/archive/{ver}.tar.gz'.format(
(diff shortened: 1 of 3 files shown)


tedana/info.py
```diff
@@ -23,7 +23,6 @@
 __packagename__ = 'tedana'
 __description__ = ('TE-Dependent Analysis (tedana) of multi-echo functional '
                    'magnetic resonance imaging (fMRI) data.')
-__longdesc__ = readme_path.open().read()
```
tsalo, line 26: Since this is being dropped, you can drop the line that defines `readme_path` in this file, along with the `Path` import.

jbteves:
@notZaki looks like CircleCI is having dockerhub issues; you'll probably need to wait until those issues resolve and then push an empty commit in order to correctly trigger tests on Circle.

tedana/info.py
notZaki: Nice catch!

tsalo (approved):
LGTM.

jbteves (approved):
LGTM, thanks @notZaki !

tsalo:
@notZaki do you want to add yourself to the Zenodo file in this PR?

notZaki:
@tsalo I'll open another PR later today with a minor change for getting tests to work on windows, and I can add myself then.

PR #672, opened 2021-02-03
[MAINT] Support Windows' paths; Update zenodo & contributor count

notZaki:
Closes None.

Few small changes:

- Use platform-specific path when comparing output files in integration tests
  + This was causing tests on Windows to fails because it expects `paths\like\this` instead of `/like/that`
- Updated the number of contributors on the readme badge
  + There is a [configuration option](https://github.com/all-contributors/all-contributors/blob/master/docs/bot/configuration.md) which might automatically do this
  + Looks like there is a mismatch in the number of total contributors between all-contributors and [github's contributor list](https://github.com/ME-ICA/tedana/graphs/contributors). I went with number of all-contributors.
- Added myself to `.zenodo.json`
  + Should this follow the order in the planned JOSS paper? That might be better for consistency, but that means getting rid of `The tedana Community` and bit of shuffling
  + There is also [`info/__credits__`](https://github.com/ME-ICA/tedana/blob/14be8520da0ba577c130c56d0835c55321cd0aed/tedana/info.py#L14-L15) which is out of sync and could be removed to simplify maintenance

The full test suite passed on windows, mac, and ubuntu with Python 3.7-3.9 [[log]](https://github.com/notZaki/tedana/actions/runs/535359775)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ multi-echo functional magnetic resonance imaging (fMRI) data.
 [![Percentage of issues still open](http://isitmaintained.com/badge/open/ME-ICA/tedana.svg)](http://isitmaintained.com/project/ME-ICA/tedana "Percentage of issues still open")
 [![Join the chat at https://gitter.im/ME-ICA/tedana](https://badges.gitter.im/ME-ICA/tedana.svg)](https://gitter.im/ME-ICA/tedana?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 [![Join our tinyletter mailing list](https://img.shields.io/badge/receive-our%20newsletter%20❤%EF%B8%8F-blueviolet.svg)](https://tinyletter.com/tedana-devs)
-[![All Contributors](https://img.shields.io/badge/all_contributors-14-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-20-orange.svg?style=flat-square)](#contributors)
 
 
 ## About
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -4,6 +4,11 @@
             "name": "The tedana Community"
         },
         {
+            "name": "Ahmed, Zaki",
+            "affiliation": "Mayo Clinic",
+            "orcid": "0000-0001-5648-0590"
+        },
+        {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -141,4 +141,4 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
 This project follows the [all-contributors](https://github.com/all-contributors/all-contributors) specification. Contributions of any kind welcome!
 To see what contributors feel they've done in their own words, please see our [contribution recognition page][contribution].
 
-[contribution]: <contributions.md>
\ No newline at end of file
+[contribution]: <contributions.md>
```
jbteves, line 144: Huh, any idea what's going on here?

jbteves (approved):
Thanks @notZaki !
It's weird that the patch has that one line both added and subtracted with no apparent change, but I'll go ahead and approve.

jbteves:
The reason for the split is that in "all contributors" we add non-code, non-authored contributions. The GitHub contributors will only list people who have committed or co-authored something added to the main branch. So, "all contributors" is the correct listing for the total number of contributors.

I have no strong feelings about the Zenodo ordering.

README.md
notZaki: I think the version on `main` does not end with a newline character, whereas vim probably added it when I edited the file.

jbteves: I see. In that case I think your PR has the preferred style, though we haven't been picky about the line breaks in the past.

emdupre:
Thanks again, @notZaki ! :smile:

re Windows : Thank you ! :pray:  We should probably add Azure testing....

re configuration : nice find ! I want to add "sort alphabetically," too. I'll see if I can get that changed. Don't worry about it in this PR.

re mismatch : @jbteves 's explanation matches my understanding :)

re: the zenodo : let's keep this as-is. The decision was that the software would always have the tedana community as first-author, but that individual outputs (like papers) could have different authorship orders / lists depending on who took on the writing, experiments, etc.

re `__credits__` : good catch ! I'm in favor of removing it.

emdupre:
I'm going to approve and merge this now so it goes into the release (:tada: ) and we can open another issue to remove `__credits__` and set the all-contributors bot config.

:rocket: :rocket: :rocket:

emdupre (approved):

PR #695, opened 2021-03-09
[DOC] Add references to HTML report

### LINKED ISSUE #686: Add references to HTML reports
### Summary
Currently, our HTML reports have the descriptive report text, with citations, but the reference list is missing. We should add it to the end of the report.

### Additional Detail
The code to do this should be roughly the same as the code to load and add the report text.

tsalo:
Closes #686.

- Generate HTML reports _after_ the text is cleaned up and references are added to `report.txt`. This results in the references being included in HTML reports.


jbteves (approved):
LGTM! Thanks @tsalo !

handwerkerd (approved):
Look good... I also realized I put my one comment on the issue rather than this PR.

PR #702, opened 2021-03-16
[REF] Normalize over time in MAPCA

tsalo:
Closes None. The goal of this PR is to harmonize our approach in `tedana` with the approach in `mapca`. I figure, if the results from these changes are the same as the results in #641, we can merge this to have it in our history and then overwrite the changes with #641 very quickly.

- Normalize data over time in ma_pca to match the changes made in ME-ICA/mapca.

### DIFF
--- a/tedana/decomposition/ma_pca.py
+++ b/tedana/decomposition/ma_pca.py
@@ -434,7 +434,7 @@ def _eigensp_adj(lam, n, p):
     return lam_adj
 
 
-def ma_pca(data_nib, mask_nib, criteria='mdl'):
+def ma_pca(data_nib, mask_nib, criteria='mdl', normalize=True):
     """
     Run Singular Value Decomposition (SVD) on input data,
     automatically select components based on a Moving Average
@@ -450,6 +450,9 @@ def ma_pca(data_nib, mask_nib, criteria='mdl'):
     criteria : string in ['aic', 'kic', mdl']
                Criteria to select the number of components;
                default='mdl'.
+    normalize : bool, optional
+        Whether to normalize data (zero mean and unit standard deviation) or not.
+        Default is True.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


eurunuela:
@tsalo changes look good to me but could you please rerun the tests just to make sure everything's ok?

tsalo:
> @tsalo changes look good to me but could you please rerun the tests just to make sure everything's ok?

@eurunuela the tests are passing though? Do you mean run them a second time to see if restarting causes any problems?

eurunuela:
> @eurunuela the tests are passing though? Do you mean run them a second time to see if restarting causes any problems?

Yes, that's what I meant. I want to make sure we don't get different results like @notZaki did some months ago with his tests when normalizing along the temporal dimension.

tsalo:
Ah, well it did change the three-echo results from 67 components to 64, but I can restart.

eurunuela (approved):
Thanks @tsalo 🙏

tedana/decomposition/ma_pca.py
```diff
@@ -476,9 +479,12 @@ def ma_pca(data_nib, mask_nib, criteria='mdl'):
     [Nx, Ny, Nz, Nt] = data_nib.shape
     data_nib_V = np.reshape(data_nib, (Nx * Ny * Nz, Nt), order='F')
     maskvec = np.reshape(mask_nib, Nx * Ny * Nz, order='F')
-    data_non_normalized = data_nib_V[maskvec == 1, :]
+    data = data_nib_V[maskvec == 1, :]
     scaler = StandardScaler(with_mean=True, with_std=True)
```
jbteves, line 483: Since the scaler is only used if `normalize` is True, could we move it into the body of the `if` statement?

jbteves (changes requested):
Thanks, @tsalo ! However I think that as it stands the code is a bit confusing. The original intent of the scikit-learn scaler appears to have been to train on one dataset and then actually run on a different one, which is why the StandardScaler function is broken into two parts. However, in this case we will never need it in two parts and I think it's mildly confusing to have two lines of code, and then repeat them again later. Could we make a small function, called something like `variance_scale`, which combines these steps? Then we could just pass in the transposed data, and then transpose it on its way back out. I feel that would improve the legibility and clarity a bit here. Furthermore, since this will ultimately be part of the mapca package, it could prove useful in later additions. If you don't have time I could code that up quickly myself.
If folks disagree, though, then I only have minor documentation and placement quibbles.

notZaki:
There is the [`scale`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.scale.html) function which does everything in a single step. However, I believe the two-part nature of `StandardScaler` is [used in `mapca`](https://github.com/ME-ICA/mapca/blob/97c232124190279fca00f1635fc263e94d149b84/mapca/mapca.py#L386).

tsalo:
@jbteves To be honest, I'd rather not put too much work into this. Hopefully these changes will be overwritten by #641 within the week (and the affected file will be deleted), so I just want to make sure it does the same thing as `mapca`.

jbteves (approved):

PR #703, opened 2021-03-16
[ENH] Match BokehJS with BokehPy version

notZaki:
Closes none.

Context: There are two bokeh libraries: one is in python, the other in javascript. Ideally, they should both be the same version. Currently, the javascript version is pinned at 2.0.1 in the html report. This can create a conflict if the user has a different python version of bokeh.

- Dynamically set the javascript version of bokeh to match the python version


tsalo (approved):

jbteves (approved):
LGTM, thanks @notZaki !

PR #705, opened 2021-03-21
[FIX] Enable normalization in mapca call

notZaki:
Closes none.

The results for the three-echo test changed when the `mapca` library was added:

- [before](https://13314-110845855-gh.circle-artifacts.com/0/tmp/data/three-echo/TED.three-echo/tedana_report.html)
- [after](https://13347-110845855-gh.circle-artifacts.com/0/tmp/data/three-echo/TED.three-echo/tedana_report.html)
- EDIT: [this PR](https://13446-110845855-gh.circle-artifacts.com/0/tmp/data/three-echo/TED.three-echo/tedana_report.html)

This might be because the data is no longer being normalized. This PR sets `normalize=True` when calling `ma_pca`, and hopefully that should match the old results.


eurunuela (approved):
Good catch! Thanks @notZaki

jbteves (approved):
Thanks @notZaki !

PR #709, opened 2021-03-22
[DOC] Added MAPCA to list of dependencies

handwerkerd:
I noticed that mapca isn't listed in the current dependencies, so I added it.

- Added mapca to the list of dependencies in README.md
- Not sure if there's anywhere else mapca should be listed as a dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ To set up a local environment, you will need Python >=3.5 and the following pack
 * [scikit-learn](http://scikit-learn.org/stable/)
 * [nilearn](https://nilearn.github.io/)
 * [nibabel>=2.1.0](http://nipy.org/nibabel/)
+* [mapca](https://github.com/ME-ICA/mapca)
 
 You can then install `tedana` with
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -10,6 +10,7 @@ packages will need to be installed:
 - numpy
 - scikit-learn
 - scipy
+- mapca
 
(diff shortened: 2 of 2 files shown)


jbteves (approved):
Ultimately `pip` should correctly fetch the packages, but good catch! Thanks @handwerkerd !

tsalo (approved):

tsalo (changes requested):
Just realized it should be updated in https://github.com/ME-ICA/tedana/blob/main/docs/installation.rst as well.

tsalo (approved):
Thanks!

handwerkerd:
FWIW, the reason I noticed this is I used `python setup.py develop` instead of pip. Maybe I should be using pip, but this did help me notice this omission. :)

tsalo:
We should probably recommend using [`python -m pip install -e .`](https://packaging.python.org/tutorials/installing-packages/#installing-from-a-local-src-tree) instead. That _should_ install the dependencies automatically. Not sure why `python setup.py develop` doesn't, but it might be a bug in our packaging?

handwerkerd:
We do recommend `pip -e` in https://github.com/ME-ICA/tedana/blob/main/CONTRIBUTING.md#3-run-the-developer-setup but I did this a while ago & got used to the other method. Thinking about this a bit more, I suspect the issue is that I had already installed tedana as a developer, which updates the code being executed, but probably doesn't automatically download new dependencies when the code is updated.

jbteves (approved):

PR #711, opened 2021-04-07
Use class for managing outputs

tsalo:
Closes None. Currently just a draft.

-
-


jbteves:
I'm happy to help coordinate or contribute to this PR. One thing to note is that I made a bit of a kludge to make the naming for a derivative that has an echo number associated. In that case, all image entries that ended with "split" required an echo argument and used a formatter instead of a plain name. You could work around it by searching for "split" inside of the name rather than a name which ends with "split." It would also let you split other file types.

tedana/io.py
```diff
@@ -259,8 +257,8 @@ def save_tsv(self, data, name):
         """
         if not isinstance(data, pd.DataFrame):
             raise TypeError(
-                "data must be a pandas dataframe, not type "
-                + str(type(data))
+                "data must be a pandas dataframe, not type " +
+                str(type(data))
```
tsalo: I don't think we need the `+`s in these.

jbteves: Would a format string be preferable?

jbteves: Haha, alright, I can fix that for function calls. Tests passed locally so I think we should be in the clear, if I can confirm we are here then I'll mark it ready for review.

tedana/io.py
```diff
@@ -1,245 +1,299 @@
...
+        a passed variable name, and then we fill in the value.
+        """
+        name = self.config[description]
+        extension = self._determine_extension(description, name)
+
+        name_variables = get_fields(name)
+        for key, value in kwargs.items():
+            if key not in name_variables:
+                raise ValueError(
+                    f'Argument {key} passed but has no match in format'
```
tsalo, line 168:
```suggestion
                    f'Argument {key} passed but has no match in format '
```

tedana/io.py
```diff
@@ -1,245 +1,299 @@
...
+            Full file path for output file.
+        """
+        data_type = type(data)
+        if not isinstance(data, np.ndarray):
+            raise TypeError(
+                f"Data supplied must of type np.ndarray, not {data_type}."
+            )
+        if data.ndim not in (1, 2):
+            raise TypeError(
+                "Data must have number of dimensions in (1, 2), not"
```
tsalo, line 220:
```suggestion
                "Data must have number of dimensions in (1, 2), not "
```

tsalo (commented):
Just some trailing spaces that need to be added.

jbteves (approved):

handwerkerd (changes requested):
My main suggested change is minor, but a bit annoying (sorry). For a variable that is now used many places in the code, `generator` is rather nondescript and may be confusing. Perhaps consider another name: `io_generator`? `output_info`?

tedana/io.py
```diff
@@ -1,245 +1,299 @@
...
-
-Notes
------
-The global variables are set by default in the module to guarantee that the
-functions that use them won't fail if a workflow API is not used.
-However, API calls can override the default settings. Additionally, the
-naming functions beginning with "get" all leverage dictionaries defined in
-the `constants.py` module, as the definitions are large.
+Other functions in the module help simplify writing out data from multiple echoes or write very
+complex outputs.
```
handwerkerd, line 4: This description is a bit confusing, particularly what you mean by 'very complex outputs' Consider clarifying.

handwerkerd (changes requested):

handwerkerd (approved):

PR #712, opened 2021-04-08
[FIX] Adds f_max to right place

### LINKED ISSUE #710: F maps are truncated after they are written but before kappa/rho calculation
### Summary
F-maps are truncated to F_MAX for kappa and rho but not for T2/S0 maps and their derived metrics.

### Additional Detail
In this code block [here](https://github.com/ME-ICA/tedana/blob/0f4d54c375ffd2a4b96ebd6a662269ea3b92d691/tedana/metrics/kundu_fit.py#L185-L200) we see that f > F_MAX is truncated to just F_MAX *after* the f maps are written, but *just before* we calculate kappa and rho. As a consequence, kappa and rho are calculated with one version of the f-map while f-map metrics and their derivatives are calculated with the un-truncated version. I don't believe that we should be doing this since that means that f-map might mean different things at different times, and it seems reasonable to threshold as early as possible. While the difference is typically small, in my metric modularization testing (where this bug is not present) some components will flip, sometimes shifting total variance explained in each category by several percent. It seems that some components have very large f-statistic outliers that can drive classification disproportionately. 

### Next Steps
- [x] Decide if this is actually a bug.
- [ ] Merging metric modularization will fix this bug OR
- [ ] We can issue a hotfix by threshold the f-maps sooner; I can issue a PR if everyone agrees this is a bug OR
- [ ] This is not a bug and we should close the issue.

jbteves:
Closes #710 .

- Threshold F-maps before making derivative metrics, rather than just before kappa/rho calculation
-

### DIFF
--- a/tedana/metrics/kundu_fit.py
+++ b/tedana/metrics/kundu_fit.py
@@ -174,6 +174,7 @@ def dependence_metrics(catd, tsoc, mmix, adaptive_mask, tes, ref_img,
             SSE_S0 = (comp_betas[:j_echo] - pred_S0)**2
             SSE_S0 = SSE_S0.sum(axis=0)  # (S,) prediction error map
             F_S0 = (alpha - SSE_S0) * (j_echo - 1) / (SSE_S0)
+            F_S0[F_S0 > F_MAX] = F_MAX
             F_S0_maps[mask_idx[mask], i_comp] = F_S0[mask_idx[mask]]
 
             # R2 Model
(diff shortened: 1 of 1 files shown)


tsalo (approved):

handwerkerd (approved):
The html reports page is blank, as was fixed in #703. On my computer, this isn't happening when I create the report with `main` but is with this PR. I have no clue why there would be this difference since these two branches are nearly identical. I assume this will get fixed with merging, but, just in case, I wanted to mention anyway.

jbteves:
I ran this branch on a few datasets and in all cases it correctly generated reports for me on my local machine. If there's something amiss we can open an issue to address it.

Issue #699, opened 2021-03-15
Document differences between ME-ICA & tedana


handwerkerd:
### Summary

ME-ICA and tedana have some known differences. At the [Feb 2021 Dev call](https://docs.google.com/document/d/1oPN3GmL0fS5Mb2VrevWCsA6tHrdNhfVJQ_t6LofsQDc/edit?usp=sharing), an action item was to document these differences in one clear location

### Additional Detail

Differences include:
- Different default dimensionality reduction in PCA
- Different masking procedure ( #600 & #680)
- Several differences listed in #622

Note: #622 is focused on analyses to better understand differences. This issue is specifically focused on adding documentation for the differences we are already aware of.

### Next Steps
- Make a cleaner list of known differences in this issue's comments
- Decide where in the documentation this should live & put it there

handwerkerd:
From the March 2021 dev call: Dan volunteers to do this, based on Taylor’s #622, but will not take offense if someone else gets to it first. A short-term option might be to make a link to the long list Taylor has already made in 622. Tentative place to put is the FAQ or in the Usage section, or its own tab. Dan leans toward putting a link to it in the FAQ. Will need to make clear which version of old MEICA we’re referring to.
Location should take the RTD tree suggestions from #677 into account.

tsalo:
Also related to #676, since it would be great if there was interlinking between the summary of differences and version-wise updates to the package in the "What's New" page.

Issue #622, opened 2020-11-19
Examine differences in results between original MEICA & current tedana


handwerkerd:
@handwerkerd processed a block design dataset with MEICA and tedana and saw some differences in the resulting denoised time series. This issue is a place-holder to remember to document this better and share the details with others.

eurunuela:
Could you share a report or some figures of those results? Just to have an idea of big/small of a difference we're talking about.

tsalo:
We should also perform an audit of our pull requests/commit history in order to identify likely sources of differences. I am going to create a label that we can use to identify these PRs and issues.

handwerkerd:
@eurunuela I definitely plan to provide more detail here, but I haven't had a chance to run analyses and make figures that would be useful to others. I wanted to make this issue as a place holder to help me remember to follow through on this.

@tsalo, That's a great list. It should live somewhere more permanent than this issue.

tsalo:
Once the list is finalized, we should absolutely incorporate it into our documentation somewhere.

PR #714, opened 2021-04-08
[FIX] Calculate Kappa and Rho on full F-statistic maps

### LINKED ISSUE #713: Kappa and Rho are only calculated from "perfect" voxels
`F_R2` and `F_S0` should only have non-zero values in voxels where `adaptive_mask == j_echo`, while `F_R2_maps` and `F_S0_maps` should be filled in with values at all voxels where `adaptive_mask >= 3` over the `j_echo` loop:
https://github.com/ME-ICA/tedana/blob/0f4d54c375ffd2a4b96ebd6a662269ea3b92d691/tedana/metrics/kundu_fit.py#L185-L186

However, we _calculate_ kappa and rho from `F_R2`, not `F_R2_maps[:, i_comp]`, which means that kappa and rho are probably only calculated from the F-statistics for voxels with good data in all echoes, along with a bunch of zeros:
https://github.com/ME-ICA/tedana/blob/0f4d54c375ffd2a4b96ebd6a662269ea3b92d691/tedana/metrics/kundu_fit.py#L202

EDIT: The good news is that this bug is also not present in #591, since we pass the 2D `F_T2_maps` and `F_S0_maps` instead of the reduced 1D `F_T2` and `F_S0`.

_Originally posted by @tsalo in https://github.com/ME-ICA/tedana/issues/710#issuecomment-812191727_

tsalo:
Closes #713, which was probably introduced in #358.

- Calculate Rho and Kappa from the F-statistic maps compiled across adaptive mask values, instead of maps where only voxels where adaptive mask == number of echoes are filled in.

### DIFF
--- a/tedana/metrics/kundu_fit.py
+++ b/tedana/metrics/kundu_fit.py
@@ -199,8 +199,8 @@ def dependence_metrics(catd, tsoc, mmix, adaptive_mask, tes, ref_img,
         F_S0[F_S0 > F_MAX] = F_MAX
         F_R2[F_R2 > F_MAX] = F_MAX
         norm_weights = np.abs(wtsZ ** 2.)
-        kappas[i_comp] = np.average(F_R2, weights=norm_weights)
-        rhos[i_comp] = np.average(F_S0, weights=norm_weights)
+        kappas[i_comp] = np.average(F_R2_maps[:, i_comp], weights=norm_weights)
+        rhos[i_comp] = np.average(F_S0_maps[:, i_comp], weights=norm_weights)
     del SSE_S0, SSE_R2, wtsZ, F_S0, F_R2, norm_weights, comp_betas
     if algorithm != 'kundu_v3':
         del WTS, PSC, tsoc_B
--- a/tedana/tests/data/fiu_four_echo_outputs.txt
+++ b/tedana/tests/data/fiu_four_echo_outputs.txt
@@ -70,9 +70,3 @@ figures/comp_017.png
 figures/comp_018.png
 figures/comp_019.png
 figures/comp_020.png
-figures/comp_021.png
-figures/comp_022.png
-figures/comp_023.png
-figures/comp_024.png
-figures/comp_025.png
-figures/comp_026.png
--- a/tedana/tests/data/fiu_four_echo_outputs.txt
+++ b/tedana/tests/data/fiu_four_echo_outputs.txt
@@ -70,3 +70,9 @@ figures/comp_017.png
 figures/comp_018.png
 figures/comp_019.png
 figures/comp_020.png
+figures/comp_021.png
+figures/comp_022.png
+figures/comp_023.png
+figures/comp_024.png
+figures/comp_025.png
+figures/comp_026.png


jbteves (approved):
LGTM, thanks for fixing @tsalo !

handwerkerd:
This generally looks good, but could you line it up with the current `main`? Without #712 merged into this PR you'd be thresholding `F_R2` with `f_max` and then discarding the thresholded map before it influences anything.

handwerkerd:
Also, we've been lax on using it, but since this will change the output, shouldn't the PR title also include the [BRK] tag? There fixes that wouldn't change the output. This probably does change the output.

tsalo:
It seems reasonable to identify any PR that changes outputs, although I think a label would be better than a prefix. I will create a new `breaking` label and apply it here. We can also go through past PRs that have changed the integration test outputs and label those. That should help out with #622 and #699.

jbteves:
To clarify an earlier comment by @handwerkerd: it's because the parent commit is holding the outdated version and there were no new commits since the merge. When we merge, it will be lined up properly with the current `main`.

handwerkerd:
I just locally ran tedana looked at the outputs for the 5 echo test data set. They are non-trivially different. 
For `main` component 0 is kappa=181.73 rho=30.78. For this PR, kappa=317.62 rho=30.9. The kappa rho scatter plots look similar, but the scaling is different, particularly for the kappa values.
Knowing what this change does, the new code should be more accurate, and thus better, but, in the new version, it is rejecting component 0 (i.e. the main visual cortex task response). This component is in the top right corner of the scatter plots below.

Others should run and look at this on their own to confirm that I didn't just mess something up. 

If you're seeing this as well, my suspicion is that this specific change _**might**_ be ok, but the changing scaling may be interacting with decision tree thresholding that was tuned for the older, flawed code. Another possibility is, with the recent adaptive mask changes, the voxel-wise scaling of weights for kappa & rho values wasn't done correctly.

`main`
image

#714
image

tsalo:
@handwerkerd thanks for testing the changes. Given that this fixes a bug introduced in #358, I think this may merit a separate issue about the validity of the current thresholds in the decision tree post-#358.

handwerkerd:
@tsalo, The trouble is if, in fixing a bug, it's creating a seriously bad misclassified component, we can't just merge this. I'm not sure I have time today but I may want to go back to older versions of the code to see if the scaling was this different before adding adaptive masking. If it's this PR specifically that's causing a big scaling change, that makes me think there might be another bug we need to catch before merging.
Put another way, before adaptive masking, we'd only include voxels in data with 5 good echoes. If also including voxels with only 3 good echoes scales the kappa values by nearly 2X, there's something else wrong with the code.

tsalo:
> @tsalo, The trouble is if, in fixing a bug, it's creating a seriously bad misclassified component, we can't just merge this.

I would argue that the introduction of a bad misclassification is an issue that must be resolved before a release, but the scope of this PR is fixing a bug in the code, rather than an overarching problem in the decision tree. We definitely can't release 0.0.10 before the decision tree has been investigated, but I don't think that should block this.

>Put another way, before adaptive masking, we'd only include voxels in data with 5 good echoes. If also including voxels with only 3 good echoes scales the kappa values by nearly 2X, there's something else wrong with the code.

If there are as many 3- and 4-echo voxels (together) as 5-echo voxels, and the F-statistics are roughly the same, then that seems like a reasonable scaling. To reproduce:

```python
import numpy as np

# 3 voxels for each value in adaptive_mask
adaptive_mask = np.array([1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4, 5, 5, 5])

# voxels with >= 3 echoes have actual values in the averaging
arr_3andup = np.array([0, 0, 0, 0, 0, 0, 5, 5, 5, 5, 5, 5, 5, 5, 5])
np.mean(arr_3andup)  # 3.0

# voxels with only 5 echoes have actual values in the averaging
# other voxels are incorrectly set to zero
arr_5only = np.array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 5, 5])
np.mean(arr_5only)  # 1.0
```

handwerkerd:
I'm trying to figure out what's going on. ~23000 voxels have 5 good echoes vs ~900 voxels having only 3 or 4 good echoes. I've been digging into this section of code line-by-line and, as far as I can tell, this 2X scaling difference does not occur in the area where you've changed the code (i.e. through line 203). The kappa values are different, but more like 117.1 vs 117.3.
Either I messed something up when I ran the full script or these changes are altering something else downstream.
I'll try to look into this a bit more tomorrow.

handwerkerd:
I've been looking into this a bit and couldn't figure out where the big change in kappa & rho values came from. I reran on this PR and don't see the same badly scaled result. My best guess is I somehow ran a non-updated up version of this PR last time.
Interestingly, the difference in results is that the kappa values are nearly the same, but the rho values are mostly slightly larger (still <0.5 difference). This makes sense since, the voxels were echoes 4 & 5 were bad are lightly to have more dropout and thus more rho weighting. By no longer accidentally excluding these voxels, they increase rho value for a bunch of components.

handwerkerd (approved):

jbteves (approved):
LGTM, thanks again @tsalo !

PR #720, opened 2021-04-16
[enh] initial commit of datasets module

### LINKED ISSUE #684: Add dataset fetching module
### Summary
Per [today's call](https://docs.google.com/document/d/1oPN3GmL0fS5Mb2VrevWCsA6tHrdNhfVJQ_t6LofsQDc/edit#), we want to add a new `datasets` module (similar to [`nilearn.datasets`](https://nilearn.github.io/modules/reference.html#module-nilearn.datasets)) for fetching open datasets.

### Additional Detail
We'll want to focus on minimally preprocessed open datasets, which we will upload to the OSF.

Some things to keep an eye on:

- Managing the "tedana data" folder. I scavenged `nilearn`'s data to do this for `NiMARE`, and the main issue is that `nilearn`'s code is kind of complicated.
- Ensuring standard interfaces across datasets. This shouldn't be a major problem if we stick to BDIS-ish datasets.
- Maintaining good coverage may be difficult.

### Next Steps
1. Add preprocessed datasets to fetchable location (e.g., OSF).
2. Add functions to new `tedana.datasets` module to grab those datasets.
3. Tests.
4. Documentation.

emdupre:
References #684. (changed from "Closes" by @tsalo since we'll move to a new PR from within the ME-ICA repo).

- Adds a new `datasets` module
- Adds functionality for downloading the preprocessed Cambridge ME dataset from OSF


emdupre:
OK, this isn't quite ready but I wanted to check in -- right now I'm just setting it up to fetch all of the echo `bold.nii.gz` files as well as the regressor time series (one per run).

Is there anything else that we really need to grab ? Also, this data directory will be quite large -- usually for Nilearn we'd aggressively downsample to minimize disk space. Should we downsample to eg 3mm isotropic here ? I suppose it will depend if we intend to use this dataset in examples or in analyses. WDYT @notZaki  ?

jbteves:
I think if we choose to use datasets for examples we should downsample for speed/space, but otherwise I'd rather have us just fetch at the original resolution.

notZaki:
The file size will also affect memory usage and this might also be worth considering if the examples are meant to run on the cloud (e.g. mybinder). There are some steps where memory usage reaches 2.5 GB for the cambridge data, so resampling would be useful for this as well.

However, a potential disadvantage is that downsampling could affect the results. Is that something to be concerned about?

tsalo:
Could we just have both? Upload the downsampled data to the OSF, and have a parameter for the download function like `resolution`? Then we'd use downsampled data for examples/tests and normal resolution data for analyses.

jbteves:
I like that solution @tsalo.

emdupre:
> Could we just have both? Upload the downsampled data to the OSF, and have a parameter for the download function like resolution? Then we'd use downsampled data for examples/tests and normal resolution data for analyses.

I think I'd still need the downsampled data uploaded to OSF ! The alternative is to download the full resolution to disk then resample, but if someone only requests the downsampled data it's a waste of compute and disk space in the call.

tsalo:
> I think I'd still need the downsampled data uploaded to OSF ! 

Agreed. I think that's the best solution. I think one of us (not sure who has the time) could handle that part fairly easily though. I'm not sure if we'd want to put them in the same OSF project as the normal-resolution data though. If so then @notZaki might need to be the one to handle it or add one of us as a non-bibliographic contributor.

eurunuela:
> I think I'd still need the downsampled data uploaded to OSF ! The alternative is to download the full resolution to disk then resample, but if someone only requests the downsampled data it's a waste of compute and disk space in the call.

So a folder called `undersampled` and a folder called `full_res`? We could program the option to choose one of them.

emdupre:
> So a folder called undersampled and a folder called full_res? We could program the option to choose one of them.

From the fetcher's perspective, I just need the files accessible on OSF -- it's actually slightly easier on my end if everything is in a flat hierarchy. I'd then plan to pull it on disk and put it in a folder like `~/tedana_data/cambridge`. Naively I'd think to mix full res and downsampled in that same folder, but have them under different names (so e.g. downsampled could have the additional key of `res-3mm`) !

emdupre:
Regardless, thinking on this a bit more carefully, I think that fetching the downsampled data can be handled in a separate PR, if we're OK that we _also_ fetch the full resolution data. 

So I can finish this up under that assumption, and we can open another PR later to add in fetching the downsampled data as it becomes available !

notZaki:
I can upload downsampled versions in a separate parent folder, with a flattened hierarchy without any subfolders.

Some questions:

- I assume `nilearn.image.resample_img` will do the trick, but not sure what arguments would be appropriate. Anyone have an example code snippet?
- Was the `manifest.json` file in the osf repo useful? If so, I'll update it when the downsampled data is uploaded. Otherwise, I'll still update it, but probably not as promptly.

emdupre:
> I assume nilearn.image.resample_img will do the trick, but not sure what arguments would be appropriate. Anyone have an example code snippet?

Here's one I use !

```python
    aff_orig = nib.load(scan).affine[:, -1]
    target_affine = np.column_stack([np.eye(4, 3) * 3, aff_orig])
    dowsample = image.resample_img(img=scan, target_affine=target_affine)
```

Occasionally, depending on the image acquisition this will yield inconsistent dimensions after resampling (e.g. `(65, 77, 65)` vs `(66, 78, 66)`). In that case, I just pass the additional param `target_shape` to `resample_img` with the desired output shape.

> Was the manifest.json file in the osf repo useful? If so, I'll update it when the downsampled data is uploaded. Otherwise, I'll still update it, but probably not as promptly.

Thank you for making that, @notZaki !! I think I had effectively duplicated your work, unfortunately, in fetching UUIDS for each file. So I can just grab them myself again, and we can update the manifest later.

emdupre:
OK, this is ready for review ! You can try it out by fetching the branch and then running:

```python
from tedana import datasets

cambridge = datasets.fetch_cambridge(n_subjects=2)
```

notZaki:
Thanks for the resampling code @emdupre 

The original data has a voxel dimensions between 3.5 to 4 mm, so a larger voxel size will be needed to reduce the file size further. Any suggestions?

Here's the mean image for voxel sizes of 4 to 6 mm³:
cambridge_downsample

The total space needed for a single subject (i.e. all echoes) is approximately 200 MB @ 4mm³, 100 MB @ 5 mm³, and 60 MB @ 6mm³.

emdupre:
> The total space needed for a single subject (i.e. all echoes) is approximately 200 MB @ 4 mm, 100 MB @ 5 mm, and 60 MB @ 6mm.

Thanks, @notZaki !! I guess with 5mm we could return 5 subjects at approximately half a GB, which should be enough for most examples. Though I suppose it will depend in part on the examples we actually want to create...

notZaki:
The downsampled data is [in this OSF repository](https://osf.io/5n968/). It currently has 5mm resolution data and should take ~45 MB per subject (instead of 100 MB), but other resolutions could be added if needed.

The file size was further reduced in ~half by masking out all voxels outside the brain, i.e. doing `unmask(apply_mask(scan, maskfile), maskfile)` before resampling the data. 
Another thing to note is that the original data has a dtype of int16 whereas the output from `resample_img` had a dtype of float64, so this was changed back to int16 via `header.set_data_dtype()` before writing to disk.

jbteves:
Thanks for doing this @emdupre ! Unfortunately I get a slightly bizarre error:
```

Dataset created in /Users/tevesjb/tedana_data/cambridge


Dataset created in /Users/tevesjb/tedana_data/cambridge/cambridge

Downloading data from https://osf.io/download/6033160f34d30403d0e39abb/ ...
Downloaded 49160192 of 50982496 bytes (96.4%,    0.0s remaining) ...done. (3 seconds, 0 min)
Downloading data from https://osf.io/download/603316164ad52e03b06e395d/ ...
Downloaded 48422912 of 49219000 bytes (98.4%,    0.0s remaining) ...done. (2 seconds, 0 min)
Downloading data from https://osf.io/download/6033161f88ea1a03aeee978a/ ...
 ...done. (4 seconds, 0 min)
Downloading data from https://osf.io/download/6033162b34d30403d1e39c3f/ ...
 ...done. (3 seconds, 0 min)
  File "<stdin>", line 1, in <module>
  File "/Users/tevesjb/repositories/tedana/tedana/datasets/fetchers.py", line 238, in fetch_cambridge
    resume=resume, verbose=verbose)
  File "/Users/tevesjb/repositories/tedana/tedana/datasets/fetchers.py", line 96, in _fetch_cambridge_functional
    funcs.append[tuple(participant_funcs)]
TypeError: 'builtin_function_or_method' object is not subscriptable
```

when running the code snippet above. In my environment I have Python 3.7.9, and all of the packages required for development.

tedana/datasets/fetchers.py
```diff
@@ -0,0 +1,246 @@
...
+            echo_id = entry['echo_id']
+            # Download bold images for each echo
+            func_url = url.format(entry['key_b'])
+            func_file = [(func.format(participant_id, echo_id),
+                          func_url,
+                          {'move': func.format(participant_id, echo_id)})]
+            path_to_func = _fetch_files(data_dir, func_file, resume=resume,
+                                        verbose=verbose)[0]
+            participant_funcs.append(path_to_func)
+        funcs.append[tuple(participant_funcs)]
```
tsalo, line 96:
```suggestion
        funcs.append(tuple(participant_funcs))
```

This should fix the bug @jbteves noted.

tsalo:
After addressing @jbteves' bug, I get the following using Python 3.8.5 and numpy 1.21.1.

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/opt/miniconda3/lib/python3.8/site-packages/numpy/lib/npyio.py in savetxt(fname, X, fmt, delimiter, newline, header, footer, comments, encoding)
   1434                 try:
-> 1435                     v = format % tuple(row) + newline
   1436                 except TypeError as e:

TypeError: must be real number, not numpy.bytes_

The above exception was the direct cause of the following exception:

TypeError                                 Traceback (most recent call last)
<ipython-input-1-df1850d765df> in <module>
      1 from tedana import datasets
      2 
----> 3 cambridge = datasets.fetch_cambridge(n_subjects=2)

~/Documents/tsalo/tedana/tedana/datasets/fetchers.py in fetch_cambridge(n_subjects, reduce_confounds, data_dir, resume, verbose)
    243         resume=resume, verbose=verbose)
    244     if reduce_confounds:
--> 245         regressors = _reduce_confounds(regressors, keep_confounds)
    246 
    247     return Bunch(func=funcs, confounds=regressors, description=fdescr)

~/Documents/tsalo/tedana/tedana/datasets/fetchers.py in _reduce_confounds(regressors, keep_confounds)
     17             selected_confounds = confounds[keep_confounds]
     18             header = '\t'.join(selected_confounds.dtype.names)
---> 19             np.savetxt(out_file, np.array(selected_confounds.tolist()),
     20                        fmt=('%.18e %.18e %.18e %.18e %.18e %.18e %f '
     21                             '%.18e %.18e %.18e %.18e %.18e %.18e %.18e %.18e'),

<__array_function__ internals> in savetxt(*args, **kwargs)

/opt/miniconda3/lib/python3.8/site-packages/numpy/lib/npyio.py in savetxt(fname, X, fmt, delimiter, newline, header, footer, comments, encoding)
   1435                     v = format % tuple(row) + newline
   1436                 except TypeError as e:
-> 1437                     raise TypeError("Mismatch between array dtype ('%s') and "
   1438                                     "format specifier ('%s')"
   1439                                     % (str(X.dtype), format)) from e

TypeError: Mismatch between array dtype ('|S32') and format specifier ('%.18e %.18e %.18e %.18e %.18e %.18e %f %.18e %.18e %.18e %.18e %.18e %.18e %.18e %.18e')
```

tedana/datasets/fetchers.py
```diff
@@ -0,0 +1,246 @@
...
+        out_file = in_file.replace('desc-confounds',
+                                   'desc-reducedConfounds')
+        if not os.path.isfile(out_file):
+            confounds = np.recfromcsv(in_file, delimiter='\t')
+            selected_confounds = confounds[keep_confounds]
+            header = '\t'.join(selected_confounds.dtype.names)
+            np.savetxt(out_file, np.array(selected_confounds.tolist()),
+                       fmt=('%.18e %.18e %.18e %.18e %.18e %.18e %f '
+                            '%.18e %.18e %.18e %.18e %.18e %.18e %.18e %.18e'),
+                       header=header, delimiter='\t', comments='')
```
tsalo, line 22:
```suggestion
            confounds = pd.read_table(in_file)
            selected_confounds = confounds[keep_confounds]
            selected_confounds.to_csv(out_file, sep="\t", line_terminator="\n", index=False)
```

tedana/datasets/fetchers.py
```diff
@@ -0,0 +1,246 @@
+import os
+import numpy as np
+from sklearn.utils import Bunch
```
tsalo, line 3:
```suggestion
import numpy as np
import pandas as pd
from sklearn.utils import Bunch
```

PR #741, opened 2021-07-06
[REF] Eliminate component sorting from metric calculation

### LINKED ISSUE #564: Revisit component sorting within metric calculation
### Summary
In order to support multiple component selection methods performed on a single decomposition (e.g., running tedana and AROMA on the same ICA results) within fMRIPrep, we should not change the mixing matrix at all by resorting or flipping components.

### Next Steps
1. Remove sorting steps from metric calculation functions.
2. If any steps in the metric calculation functions require components sorted by kappa, perform the sorting within the step and update the unsorted component table.
3. Remove `reindex` argument from `dependence_metrics()`.
4. Update figures and reports to maximize their utility without sorting the actual component order.

tsalo:
Closes #564.

**NOTE**: While the proposed changes do affect the order of rows in the component table and the mixing matrix, they should not impact the actual classifications at all.

I have one question- should we also eliminate the component flipping step? I'm thinking specifically that we could (1) keep components with the same signs in our outputs, but (2) add a new "flip" column to the metrics TSV indicating if the component was flipped prior to calculating metrics. Then, the mixing matrix would not be changed at all by `tedana`, but the results would be the same. The only things that would change would be the saved weight maps and the signs of the maps in the report. We could even use the "flip" column in the report figure generation to retain the preferred signs.

To do:

- [x] Resolve sign-flipping question
    - Opened #748.

- Remove the functions `metrics._utils.sort_df()` and `metrics._utils.apply_sort()`.
- Remove `sort_by` and `ascending` parameters from `metrics.collect.generate_metrics()`.
- Update docstrings and metadata to reflect that components are not re-sorted.

### DIFF
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -236,7 +236,7 @@ def tedpca(data_cat, data_oc, combmode, mask, adaptive_mask, t2sG,
     comptable, _ = metrics.collect.generate_metrics(
         data_cat, data_oc, comp_ts, adaptive_mask,
         tes, io_generator, 'PCA',
-        metrics=required_metrics, sort_by=None
+        metrics=required_metrics,
     )
 
     # varex_norm from PCA overrides varex_norm from dependence_metrics,
--- a/tedana/metrics/_utils.py
+++ b/tedana/metrics/_utils.py
@@ -109,44 +109,6 @@ def flip_components(*args, signs):
     return [arg * signs for arg in args]
 
 
-def sort_df(df, by="kappa", ascending=False):
-    """Sort DataFrame and get index.
-
-    Parameters
-    ----------
-    df : :obj:`pandas.DataFrame`
-        DataFrame to sort.
-    by : :obj:`str` or None, optional
-        Column by which to sort the DataFrame. Default is 'kappa'.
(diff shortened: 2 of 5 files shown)


eurunuela (approved):
I understand you went for option (1) keep components with the same signs in our outputs.

LGTM!

tsalo:
>I understand you went for option (1) keep components with the same signs in our outputs.

At the moment, the signs are still being flipped, which means that our mixing matrix won't be the same as the one used by AROMA (assuming we ran both workflows on the same decomposition). I don't think it's a huge deal- it's definitely not as much of a problem as having the components sorted differently- but I would prefer to not have tedana alter the mixing matrix at all if possible.

jbteves (approved):
LGTM, thanks @tsalo !

tsalo:
Thanks @eurunuela and @jbteves! I can hold off on the sign flipping question for another PR. I'll probably open an issue about it, so that we can merge this ASAP. WDYT?

jbteves:
I think that's a good plan. We might have more discussion around that vs. sorting.

tsalo:
I've opened #748, so I think this is ready to merge.

Issue #687, opened 2021-02-22
Include command used and tedana version in HTML reports


tsalo:
### Summary
It would be nice to include the command (or parameters) used for the `tedana` run, along with the version of `tedana`, in the HTML reports.

### Additional Detail
This stems from #450.

eurunuela:
I can take care of these additions to the reports. I know the code quite well.

PR #696, opened 2021-03-09
[ENH] Add carpet plot to outputs

tsalo:
Closes #688.

Changes proposed in this pull request:

- Generate a set of carpet plots in a single figure and output to the figures directory.
- Add the carpet plots to the HTML report.
- Add new function (`tedana.io.denoise_ts`) to handle the denoising part.
- Drop unused output (`varexpl`) from `tedana.io.write_split_ts`.

To do:

- [x] Add info and image of carpet plots to our RTD documentation.
- [ ] Regenerate example report in [our demo](https://me-ica.github.io/tedana-ohbm-2020/). We actually might want to do this _after_ we've handled all of the current report-related PRs.

PR #747, opened 2021-07-10
Add section to reports that show system info, tedana call and version

### LINKED ISSUE #650: Add system info to report outputs
### Summary
From today's dev call, we want to collect and save system/environment info among out report outputs, in order to improve reproducibility.

### Additional Detail
I thought we could take something from `rapidtide` for this, but it turns out they just output the parameters used in the run (which is still useful information that we could output) in a json. I _know_ there are some packages that output system information, but for some reason I can't find any right now. Will need to revisit later.

### Next Steps
1. Find a Python library that outputs system information and adopt their approach.
2. Write out a file with system info at the beginning of each workflow run.

eurunuela:
Closes #650 and #687 .

- [x] Add a function to return system info, then save it into a json file.
- [x] Save tedana call into a variable and add to the dictionary with system info.
- [x] Add tedana version into the dictionary with system info.
- [x] Get Python version and add it to the dictionary with system info.


eurunuela:
This PR is pretty much ready for review but I would like to wait until #696 is merged to make some final changes for the sake of coherence.

jbteves:
@eurunuela this looks good but there's actually one potential downfall: if you call `tedana` with the API instead of on the command line, I'm not sure you would get a good idea of what the API was called with. We may instead want to create a list of parameters that get captured from the API call and display those. We could also add some logic by having a new workflow variable that captures the mode as `cli` at the `__name__ == '__main__'` check and add logic to this PR to format appropriately for CLI/API. Happy to help with that.

tedana/resources/config/outputs.json
```diff
@@ -182,5 +182,9 @@
     "ICA orthogonalized mixing tsv": {
         "orig": "ica_orth_mixing",
         "bidsv1.5.0": "desc-ICAOrth_mixing"
+    },
+    "System info json": {
+        "orig": "system_info",
+        "bidsv1.5.0": "system_info"
```
tsalo, line 188: I think we can put all of this info into the `dataset_description.json` instead. See [here](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#derived-dataset-and-pipeline-description). We could also probably steal some field names from [task events metadata](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html#stimulus-presentation-details). They wouldn't be completely BIDS-compliant, but they're going to be fairly close.

tsalo:
> We may instead want to create a list of parameters that get captured from the API call and display those. We could also add some logic by having a new workflow variable that captures the mode as `cli` at the `__name__ == '__main__'` check and add logic to this PR to format appropriately for CLI/API.

I would be pretty happy with just a list/table/dict of parameters, rather than a reconstructed CLI call. Just my personal preference though.

tedana/resources/config/outputs.json
eurunuela: Done in aded008

eurunuela:
@tsalo @jbteves do you like it now?

tedana/utils.py
```diff
@@ -384,3 +385,54 @@ def get_resource_path():
...
+        "Version": system_info.version,
+        "Machine": system_info.machine,
+        "Processor": system_info.processor,
+    }
+
+    return system_dict
+
+
+def write_tedana_command(arg_vals):
+    """[summary]
```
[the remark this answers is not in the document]
eurunuela: I did! Thanks!

tedana/utils.py
```diff
@@ -384,3 +385,54 @@ def get_resource_path():
...
+
+    command = "tedana"
+
+    for key in arg_vals.keys():
+        # Data and TE are lists
+        if type(arg_vals[key]) is list:
+            values = " ".join(str(v) for v in arg_vals[key])
+        else:
+            values = arg_vals[key]
+        command += f" <b>--{key}</b> {values}"
```
jbteves, line 436: You can iterate over both keys and values in Python, which lets you make this a bit neater:
```suggestion
    for k, v in arg_vals:
        # Data and TE are lists
        if type(v) is list:
            values = " ".join(str(x) for x in v)
        else:
            values = v
        command += f" <b>--{k}</b> {values}"
```

jbteves:
I'm not entirely sure we want to display it CLI-style instead of printing a dict of parameters, but I'm guessing that it's easier to make pretty if we do it CLI-style and it should still be plenty legible. Plus, people could just copy and paste it to replicate.

tedana/utils.py
eurunuela: I didn't know, thanks!

eurunuela:
> I'm not entirely sure we want to display it CLI-style instead of printing a dict of parameters, but I'm guessing that it's easier to make pretty if we do it CLI-style and it should still be plenty legible. Plus, people could just copy and paste it to replicate.

I believe the way I put it it won't be exactly the same as the CLI but it should be pretty close.

tedana/utils.py
```diff
@@ -384,3 +385,54 @@ def get_resource_path():
...
+    -------
+    str
+        Arguments given to tedana and their values formatted in CLI-style.
+    """
+    # Deleta frame from dictionary
+    del arg_vals["frame"]
+
+    command = "tedana"
+
+    for key, val in arg_vals:
```
tsalo, line 430:
```suggestion
    for key, val in arg_vals.items():
```

eurunuela: Thanks!

jbteves: Oops, you're right, I was typing too quickly, thanks @tsalo

tsalo:
Here's what it currently looks like for the four-echo test:

Screen Shot 2021-07-15 at 12 09 44 PM

I really like it. I see that there are some issues with unset arguments (e.g., `--prefix` followed by nothing, `--mixm None`) that make me think that a table might be better, but it's still pretty interpretable. We may need a warning for naive users though. Just something telling them not to just copy and paste the command, since it's not exact.

One minor thing I'd love to see is a section title for the methods description, so that it's clear what the boundary is between command/system info and the report boilerplate.

jbteves:
Hm, would it be straightforward to do a simple table of each parameter getting a line? You could switch the format to 
```
f"{key}: {value}<br>"
```
I think it's probably not great to have something that looks pastable but isn't, unfortunately...

eurunuela:
> Hm, would it be straightforward to do a simple table of each parameter getting a line?

We could create a table like the one below.

Re section title: that's coming on #696

tedana/utils.py
```diff
@@ -384,3 +385,54 @@ def get_resource_path():
...
+        "Release": system_info.release,
+        "Version": system_info.version,
+        "Machine": system_info.machine,
+        "Processor": system_info.processor,
+    }
+
+    return system_dict
+
+
+def write_tedana_command(arg_vals):
```
tsalo, line 412: Following up from https://github.com/eurunuela/pySPFM/pull/28- I think it would be better if we could use the CLI commands directly, and _not_ compile a tedana command when users don't use the CLI. WDYT of swapping out this function with the changes I proposed in pySPFM?

tedana/workflows/tedana.py
```diff
@@ -692,7 +703,17 @@ def tedana_workflow(data, tes, out_dir='.', mask=None,
...
+                "Node": {
+                    "Name": info_dict["Node"],
+                    "System": info_dict["System"],
+                    "Machine": info_dict["Machine"],
+                    "Processor": info_dict["Processor"],
+                    "Release": info_dict["Release"],
+                    "Version": info_dict["Version"]
+                },
+                "Python": info_dict["Python"],
+                "Command": tedana_command
```
tsalo, line 716: When users use the Python function instead of the CLI, perhaps we could have the function call instead? I don't know if that's an easy thing to do though.

eurunuela: That would be cool! I guess we could just write a silly function that turns the function call into a string as part of our utils file.

eurunuela:
I think this is ready. If anything, we may want to create the `call.sh` file with the `io_generator` object.

Any thoughts on that @tsalo @handwerkerd ?

eurunuela:
Ha, I just realized the tests are failing because `ica_reclassify` also uses the reports! I will add the command to it too.

eurunuela:
Hey @tsalo @handwerkerd, you might have missed the notification for this one.

This PR is ready for review now. Let me know if you would make any changes.

tsalo:
Could you copy over the changes to `.circleci/config.yml` from #949 so we can look at the test outputs as well? It'll make it easier to check over the effects of your changes.

eurunuela:
> Could you copy over the changes to `.circleci/config.yml` from #949 so we can look at the test outputs as well? It'll make it easier to check over the effects of your changes.

Done in [11b9701](https://github.com/ME-ICA/tedana/pull/747/commits/11b970167901bba3dbe4b8a8b3d2a252946fa653).

tsalo (approved):
Thanks for this @eurunuela! I like the formatting of the system info, but I do think the command formatting could be improved. That said, we can always tweak the style in a later PR.

One unfortunate thing is that the "command" shows a test script when one is used. See the "command" for the three-echo test:

>tedana --log-cli-level=INFO --cov-append --cov-report term-missing --cov=tedana -k test_integration_three_echo tedana/tests/test_integration.py

I don't think there's a way to get around this easily. It's probably just a limitation we'll have to acknowledge.

Regardless, everything looks really awesome.

eurunuela:
How about this new design for the code block @tsalo?

tsalo (approved):
The new design looks great! Thanks @eurunuela!

handwerkerd (approved):
This looks great.

I'm approving, but I suggested one small change that you can revise or ignore.

One additional thing to consider is to add the version numbers of python modules that are most likely to affect results:  mapca, nilearn, scipy, numpy, pandas, sklearn

eurunuela:
I have moved the tedana command creation closer to the report, references and log creation.

> One additional thing to consider is to add the version numbers of python modules that are most likely to affect results: mapca, nilearn, scipy, numpy, pandas, sklearn

I would leave the addition of these module versions for a future PR.

tsalo (approved):
LGTM

PR #749, opened 2021-07-12
[REF] Do not modify mixing matrix with sign-flipping

### LINKED ISSUE #748: Do not modify mixing matrix with sign-flipping
### Summary
Currently, `tedana` updates the ICA mixing matrix (whether derived internally or provided by the user) based on optimal component signs and Kappa values. While the latter will be resolved by #741, the former is still an issue. I propose that we _not_ update the mixing matrix _at all_ (except for tedort or mir), and we add a new metric called something like "optimal sign", with values of 1 and -1, to the component table.

### Additional Detail
By not updating the mixing matrix at all, we will ensure that any time tedana is run on a pre-run decomposition, the resulting component table will apply to that decomposition, which is crucial if we want users to be able to run multiple classifiers (such as tedana _and_ AROMA) on the same ICA.

This stems from a conversation in #741.

### Next Steps
0. Wait for #741 to be merged.
1. Add metric to component table indicating optimal sign.
2. Do not return an updated mixing matrix from `metrics.collect.generate_metrics`.
3. Write out the mixing matrix _before_ metric calculation.
4. Use original mixing matrix and "optimal sign" metric in figure generation.

tsalo:
Closes #748.

- Add new metric, "optimal sign", to component tables.
- Do not return sign-flipped mixing matrix from `metrics.collect.generate_metrics()`.
- Write out _unmodified_ mixing matrix in main workflow.
- Coerce signs to integers rather than floats. This will make the metric values in the component table easier to interpret.

### DIFF
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -281,15 +281,9 @@ def tedpca(data_cat, data_oc, combmode, mask, adaptive_mask, t2sG,
     io_generator.save_file(mixing_df, "PCA mixing tsv")
 
     # Save component table and associated json
-    temp_comptable = comptable.set_index("Component", inplace=False)
-    temp_comptable.to_csv(
-        io_generator.get_name("PCA metrics tsv"),
-        index=True,
-        index_label="Component",
-        sep='\t',
-    )
+    io_generator.save_file(comptable, "PCA metrics tsv")
 
-    metric_metadata = metrics.collect.get_metadata(temp_comptable)
+    metric_metadata = metrics.collect.get_metadata(comptable)
     io_generator.save_file(metric_metadata, "PCA metrics json")
 
     decomp_metadata = {
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -233,7 +233,7 @@ def tedpca(data_cat, data_oc, combmode, mask, adaptive_mask, t2sG,
         'variance explained', 'normalized variance explained',
         'd_table_score'
     ]
-    comptable, _ = metrics.collect.generate_metrics(
+    comptable = metrics.collect.generate_metrics(
         data_cat, data_oc, comp_ts, adaptive_mask,
         tes, io_generator, 'PCA',
         metrics=required_metrics,
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -291,7 +291,6 @@ def tedpca(data_cat, data_oc, combmode, mask, adaptive_mask, t2sG,
         "Method": (
             "Principal components analysis implemented by sklearn. "
             "Components are sorted by variance explained in descending order. "
-            "Component signs are flipped to best match the data."
         ),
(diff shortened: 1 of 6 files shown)


tedana/decomposition/pca.py
```diff
@@ -282,7 +282,12 @@ def tedpca(data_cat, data_oc, combmode, mask, adaptive_mask, t2sG,
 
     # Save component table and associated json
     temp_comptable = comptable.set_index("Component", inplace=False)
-    io_generator.save_file(temp_comptable, "PCA metrics tsv")
+    temp_comptable.to_csv(
+        io_generator.get_name("PCA metrics tsv"),
+        index=True,
+        index_label="Component",
+        sep='\t',
+    )
```
tsalo, line 290: `io_generator.save_file` doesn't keep the "Component" index.

tedana/metrics/collect.py
```diff
@@ -361,9 +360,11 @@ def generate_metrics(
                     echo=(i_echo + 1)
                 )
 
-    # Build new comptable, re-ordered like previous tedana versions
-    previous_order = (
-        "kappa", "rho", "variance explained",
+    # Reorder component table columns based on previous tedana versions
+    # NOTE: Some new columns will be calculated and columns may be reordered during
+    # component selection
+    preferred_order = (
```
tsalo, line 366: I simplified this section, but the thing I want to note is that the component table will be re-sorted in the selection stage, so I think we should move this part to the selection functions instead. I don't want to do that in this PR though.

tedana/decomposition/pca.py
jbteves: We may want to add that functionality instead.

tedana/metrics/collect.py
jbteves: I think that's fine, this was a hack to make it straightforward to compare "old" and "new" metrics after the PR. We can have a discussion about it in a different issue/PR.

jbteves (changes requested):
This is a great refactor, thanks @tsalo ! Good job cleaning up the logic I had for ordering the table. However, I think it would be good to stick with our `save_file` interface. Either you can add something to add the component index or I can if you'd prefer.

tedana/decomposition/pca.py
tsalo: I think it's fixed if we just don't set the index as Component, so I'll test that out now.

tsalo:
@jbteves thanks for reviewing. I think I was just overcomplicating the TSV saving procedure. If I just keep the "Component" ID column as a column instead of the index, it should be fine.

tedana/decomposition/pca.py
tsalo: Everything looks good now.

jbteves (approved):
LGTM, thanks so much @tsalo !

eurunuela (approved):
Changes look good to me. Thank you @tsalo ! I'll leave it to you to merge this PR.

PR #754, opened 2021-07-15
[FIX] Correct manacc documentation and check for associated inputs

### LINKED ISSUE #753: `manacc` documentation and validation is incorrect
### Summary
`manacc` requires both `ctab` and `mix`, but only `mix` is checked in our input validation. Additionally, there is no information about which arguments are required in the help.

When a user provides `manacc` without `mix`, there is a warning that the manual components will be ignored. However, if the user provides `manacc` and `mix` without `ctab`, there is no warning and it just skips manual selection with no information about why.

### Additional Detail
This stems from https://neurostars.org/t/tedana-issue-with-manacc/19765.

### Next Steps
1. Improve parser help strings.
2. Improve workflow docstring.
3. Improve input validation.

tsalo:
Closes #753.

- Describe mutually dependent parameters in argument parser and workflow docstring.
- Check that both ctab and mmix are provided if manacc is, rather than just checking for mmix.

### DIFF
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -233,15 +233,21 @@ def _get_parser():
                           dest='ctab',
                           metavar='FILE',
                           type=lambda x: is_valid_file(parser, x),
-                          help=('File containing a component table from which '
-                                'to extract pre-computed classifications.'),
+                          help=(
+                              'File containing a component table from which '
+                              'to extract pre-computed classifications. '
+                              "Requires --mix."
+                          ),
                           default=None)
     rerungrp.add_argument('--manacc',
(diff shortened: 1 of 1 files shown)


eurunuela (approved):
LGTM!

jbteves (approved):
LGTM, thanks @tsalo ! I fiddled around with trying to make things required via `argparse`, but you don't get as informative of error messages. The only upside would be that the parser fails much faster than loading the images and then checking some of these args, so I think your implementation is a better trade-off!

tsalo:
Thanks @jbteves and @eurunuela.

@jbteves, yeah that seems like an unfortunate gap in argparse.

PR #756, opened 2021-07-15
[REF] Decision Tree Modularization

### LINKED ISSUE #403: Modularize the component selection decision tree

### LINKED ISSUE #808: Address divide by zero when getelbow called with only one component
### Summary

This is a continuation of issue #752 and PR #760. That discussion focused on removing a really uninformative error when getelbow was called on an empty array. I now have a situation where getelbow is called on only one component and it runs, but has a divide by zero warning.

The basic solution to this is extremely easy: The error goes away if getelbow is only run with a minimum of 2 components. I'm creating an issue rather than a PR because there may be some relevant discussion here. One can calculate an elbow of a curve with only two values, but that's essentially meaningless. If we're setting a threshold for a number of components than than which we shouldn't calculate an elbow, what should it be? 3 is mathematically plausible, but still non-ideal. I'm leaning towards "5" while acknowledging that's arbitrary.

### Additional Detail
The block of code where I get the divide by zero error is:
https://github.com/ME-ICA/tedana/blob/5399ca5368d7fab1a450208d13752c067de8503e/tedana/selection/tedica.py#L261-L269

The error goes away by replacing `if not kappas_nonsig.size` with  `if kappas_nonsig.size<=1` but the question is whether it should be something like `<=5` instead.

The last PR also added a check and warning in getelbow. We can adjust that too to throw a warning if the number of components are under a threshold rather than just empty:
https://github.com/ME-ICA/tedana/blob/5399ca5368d7fab1a450208d13752c067de8503e/tedana/selection/_utils.py#L97-L105

As far as I can tell, getelbow is called in two situations. (1) On most or all of the components. In that situation, this will never be an issue because, if there are only 5 total components, then tedana will have bigger problems. (2) A subset of components where that second elbow threshold is used in conjunction with a threshold on all components. The current approach was to just ignore that second threshold when there are zero components.

Another approach is that the second threshold on non-significant components can only lower the kappa elbow and result in accepting more components and T2* signal. Excluding this option if there are less than X components means that denoising may be slightly more aggressive.

### Next Steps

- Decide on the appropriate threshold for setting an elbow to `nan` rather than using just a few components to calculate an elbow.
Options:
    1. If there's only one component, set that value as the elbow
    2. Set the minimum to 2, which means the code will run as is and the elbow won't actually be an elbow, but will be somewhere around those two value
    3. Set a minimum closer to 5 where an elbow calculation has a chance of being meaningful
    
    I originally was leaning towards iii, but I might be overthinking this, and ii might be simplest.
- Once a decision is made, implement this.

### LINKED ISSUE #889: Additional fields in dataset_description.json identifying source files
### Summary
I think it would be very useful to have information about which source files Tedana was applied to in the dataset_description.json file perhaps for posterity. When conditions/sessions are relabeled it's easy to make mistakes/have uncertainty about what were the underlying files involved in a making the specific files.

### LINKED ISSUE #892: Add "n/a" to accepted components rationale
### Summary

This issue stems from the following neurostars post: https://neurostars.org/t/tedana-v0-0-12-cant-read-metrics-tsv-file/23810

### Next Steps

- Add na_rep='n/a' to to_csv calls.

### LINKED ISSUE #927: Add MRI Together conference presentations to documentation
### Summary

@eurunuela gave a few presentations at the MRI Together Conference in December 2022. We should add these as resources in the documentation

### Additional Detail

Presentations:
- Tedana: Analysis of echo-time dependent fMRI data: https://youtu.be/4wsEodepyI8?t=96
  - Slides: https://eurunuela.github.io/tedana_MRITogether_2022/
- A tour of tedana: https://youtu.be/P4cV-sGeltk?t=10 Slides are also available as a link  - 
  - Slides: https://eurunuela.github.io/tedana_tutorial_MRITogether_2022/

### LINKED ISSUE #931: Switch to CalVer after decision tree modularization
### Summary
Starting with the release after #756 is merged, we want to switch from semantic version (SemVer) to calendar versioning (CalVer).

### Additional Detail
For more information on CalVer, see https://calver.org/.

### Next Steps

1. Release 0.0.13 before #756 is merged.
2. Merge #756.
3. Release 2023.0.1 (assuming I have that right)

### LINKED ISSUE #936: Updating contributor information
### Summary

We haven't updated contributor information in a while. The readthedocs page is updated with version releases so now is a good time to check it over and make updates.

### Additional Detail

Contributor information is in a a few places
- https://github.com/ME-ICA/tedana/blob/main/.all-contributorsrc lists everyone's name and the types of contributions they've made This populates the grid of contributors and the emojis under their names at the bottom of our [frontpage readme](https://github.com/ME-ICA/tedana#contributors).
- https://github.com/ME-ICA/tedana/blob/main/contributions.md is a more narrative-style list of contributions
- https://github.com/ME-ICA/tedana/blob/main/docs/contributing.rst#leadership appears on [our readthedocs page](https://tedana.readthedocs.io/en/latest/contributing.html#leadership). According to that guide, "Once per year, there should be an explicit call to the larger contributor community asking if anyone wants to self nominate for a leadership role." This is that call. Will you answer the call!

### Next Steps

 Use this issue to request updates and then we'll open a PR with the updates. In theory we should vote at a dev call on anyone who is interested in a new leadership role.

For example:
- In .all-contributorsrc, my listed types of contributions are design, doc, example, & review. I'd like to add code, ideas, test, question, bug.
- In contributions.md "once in a while, even contributes to the code." should change to "contributes to the code"

jbteves:
Closes #403, closes #808, closes #809, closes #889, closes #892, closes #936, closes #931, closes #927
Supercedes #592
 
See #592 
This replaces the inflexible decision tree in tedica.py with a modular structure that will allow for multiple default and user-defined decision trees along with a more interpretable and flexible system for tracking and understanding the results. 
 
Noteworthy implemented features / changes:
- Main workflow in `tedana.py` is simplified and a separate workflow `ica_reclassify.py` can be used to manually change component classifications. (Removed 1/5 of the lines of code in `tedana.py` that are just used to handle the manual classification condition)
- Minimal and kundu decision trees are fully functional
- Moved towards using the terminology of “Component Selection” rather than “Decision Tree” to refer to the code that’s part of the selection process. “Decision Tree” is still used to more specifically to refer to the steps to classify components.
- `ComponentSelector` object created to include common elements from the selection process including the component_table and information about what happens along every step of the decision tree. Additional information that will now be tracked and stored includes:
  - `component_table` has columns for `classification_tags` instead of `rationale` so that each decision tree can include multiple descriptive tasks per component. Also an executed tree should only include `accepted` or `rejected` components. `ignored` components are now `accepted` with `classification_tags` like `Low variance` or `Accept borderline`
  - `cross_component_metrics` will save values calculated across components, such as the kappa & rho elbows
  - `component_status_table` contains how component classifications changed along every node of the decision tree
  - `tree` contains the executed tree along with everything calculated or changed during execution
  - `used_metrics` a list of all metrics used when running the tree. (Can potentially be used to calculate a subset of metrics given an inputted tree)
- The new class is defined in `./selection/component_selector.py`, the functions that define each node of a decision tree are in `./section/selection_nodes.py` and some key common functions used by selection_nodes are in `./selection/selection_utils.py`
  - By convention, functions in selection_nodes.py that can change component classifications, begin with `dec_` for decision and functions that calculate cross_component_metrics begin with `calc_`
  - A key function in selection_nodes.py is `dec_left_op_right` which can be used to change classifications based on the intersection of 1-3 boolean statements. This means most of the decision tree is modular functions that calculate cross_component_metrics and then tests of boolean conditional statements.
- `io.py` is now used to output a registry (default is `desc-tedana_registry.json`) of all generated file names that can then be read in to load data rather than requiring follow-up programs, like RICA, to have multiple inputted file names.
- New documentation changes include:
  - `building decision trees.rst` which is designed to explain the whole process in depth
  - `outputs.rst` was bloated and now links to two other files
    - `classification_output_descriptions.rst` is an explanation of the new outputs that's targetted towards a user rather than a potential developer
    - `output_file_descriptions.rst` is an expanded and updated explanation of all the file names that also explains how the filename registry.json is used
- Some terminology changes, such as using `component_table` instead of `comptable` in code
- integration tests now store testing data in `.testing_data_cache` and only download data if the data on OSF was updated more recently than the local data.
- Project package management now uses `pyproject.toml`
- 100% testing coverage on all new component selection functions and and a net increase in total coverage.
- Minimum python version is now 3.8 and minimum pandas version is now 2.0
 
**Remaining work to do before merging**
- [x] Improve documentation
  - [x] Link `building_decision_trees.rst` to the rest of the documentation
  - [x] `building_decision_trees.rst` is both a guide for developers and an explanation of outputs that any user might want. Make sure the user-guide parts are more accessible within the existing documentation
  - [x] `rationale` and descriptions of rationale are in several places in the documentation. Remove all and replace with an explanation of `classification_tags`
  - [x] Make sure documentation and function docstrings render correctly. 
  - [x] Proofread documentation and docstrings
- [x] Triple check the kundu tree in `./resources/decusion_trees/kundu.json` matches the decision tree code in Main
- [x] Testing coverage is already very good. While this is fresh in memory, try to improve testing coverage more.
- [x] tedana reports
  - [x] Add visualization of kappa and rho elbows to tedana reports
  - [x] Add `classification_tags` to hover text
- [x] Take a look at the rho elbow for the minimal tree to see if it’s too aggressive (easier to do after updates to reports)
- [x] Change `provisionalreject` to `unclassified` in the kundu decision tree since those are accepted if not rejected by other criteria
- [x] Get the CLI working for ica_reclassify
- [x] @jbteves is working on mini-tools to directly compare the results in the current main with the results of this PR. This will include:
  - Script(s) and instructions to make it easier to run both main & this PR on the same data
  - Count and identify which components changed classifications between versions of the code and the % variance explained by those changes
  - `ignored` classification in main should be `accepted` classification in this PR with one of the following classification tags: “Low variance” “Accept borderline” “No provisional accept”
 - [x] Add the names of the inputted files for each echo into `desc-tedana_registry.json` so that those file names can be automatically accessed
 - [x] Make sure all written tsv files use "n/a" rather than an empty "" in rationale fields and other places because some programs don't like the empty fields.
 - [x] [Get RICA working with the changes from this PR](https://github.com/ME-ICA/rica/issues/57)
 - [x] Make sure works with AFNI
 - [x] Make sure works with fMRIPrep

**Remaining work where we can really use more help**
- [x] Once the mini-tools are created, have multiple people run on multiple datasets to make sure it runs, gives plausible results, and users with various levels of expertise understand what’s happening.
  - If you want to help with this, leave a comment or contact @handwerkerd
  - Does Main match the output of the kundu decision tree?
  - Is the minimal decision tree reliably more conservative? The minimal tree should accept some component that were rejected by the kundu tree. It’s possible the kundu tree will accept a few low variance components that are rejected by the minimal tree.
- [x] After documentation is cleaned up, we will need both developers and non-developer users to read them for clarity.

There are several improvements that aren’t necessary before merging this PR which are opened as stand alone issues:
- #888 For built-in tedana reports, better use classification tags to allow for dynamic coloring or selection of components This would be better done by someone who is a more experienced bokeh coder
- #888 Write code to automatically create a flow chart or other summary/visualization for a decision tree. This can also be used to make a run-specific visualization that shows how components were reclassified at each node of the decision tree.
- #919 For example consistently using `component_table` vs `comptable` and reducing places where tags with capitalization differences can cause problems
- #920
- #921
- #922
- #929 Decided to merge this PR and then making a separate PR to make a change that will cause a divergence between this PR and the current Main, but will line up better with the older MEICA
- #942

Discussed but not going to open an issue unless others have specific use-cases planned for this:
- `ica_reclassify.py` currently just works for manually changing accepted and rejected classifications. Either that function or another can be used to run a follow-up decision tree. Figure out if there are potential use-cases for this functionality and then update code.


tedana/resources/decision_trees/kundu.json
```diff
@@ -0,0 +1,267 @@
+{
+    "tree_id": "kundu_MEICA27_decision_tree",
+    "info": "Following the full decision tree designed by Prantik Kundu",
+    "report": "This is based on the minimal criteria of the original MEICA decision tree without the more agressive noise removal steps",
+    "refs": "Kundu 2013",
+    "necessary_metrics": [
```
tsalo, line 6: I think just `metrics` would be cleaner here.

tedana/resources/decision_trees/kundu.json
```diff
@@ -0,0 +1,267 @@
+{
```
tsalo, line 1: YAML files are easier to format and allow comments. They might be preferable to JSONs for the trees.

tsalo (commented):
I have a couple of initial comments. Also, it looks like `tedana/selection/DecisionTree.py` and `tedana/selection/decision_tree_class.py` are duplicates.

handwerkerd:
@ME-ICA/tedana-devs Josh and I have been working on this and there's been lots of progress. I updated the initial comment to keep track of what's done and what still needs to get done. The big accomplishment is that the minimal decision tree is fully functional with all the structural and functionality changes that were recently discussed.
Function docstrings are similarly updated (though I'm sure the won't all render prefectly) and I wrote document explaining the whole process at: https://github.com/jbteves/tedana/blob/JT_DTM/docs/building_decision_trees.rst

As you can see, there's still a good bit to do, including making the functions underlying the kundu decision tree functional again. This obviously isn't ready for a full review, but feedback is welcome.

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
+########################################################
+Understanding and building a component selection process
+########################################################
+
+``tedana`` involves transforming data into components via ICA, and then calculating metrics for each component.
+Each metric has one value per component that is stored in a comptable or component_table dataframe. This structure
+is then passed to a "decision tree" which through a series of binary choices categories each component as **accepted** or
```
eurunuela, line 7:
```suggestion
is then passed to a "decision tree" which through a series of binary choices labels each component as **accepted** or
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+**rejected**. The time series for the rejected components are regressed from the data in the final denoising step.
```
eurunuela, line 8: I wonder if we should provide more conservative approaches to denoise the data. See [Moia et al., 2021](https://www.sciencedirect.com/science/article/pii/S1053811921001919).

TL;DR: regressing out the rejected components without orthogonalizing them to the accepted components can be too aggressive.

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+
+There are several decision trees that are included by default in ``tedana`` but users can also build their own.
+This might be useful if one of the default decision trees needs to be slightly altered due to the nature
+of a specific data set, if one has an idea for a new approach to multi-echo denoising, or if one wants to integrate
+non-multi-echo metrics into a single decision tree.
+
+Note: We use two terminologies interchangeably. The whole process is called "component selection"
+and much of the code uses variants of that phrase (i.e. the ComponentSelector class, selection_nodes for the functions used in selection).
+Instructions for how to classify components is called a "decision tree" since each step in the selection
+process branches components into different intermediate or final classifications
```
eurunuela, line 18:
```suggestion
process branches components into different intermediate or final classifications.
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+Expected outputs after component selection
+******************************************
+
+*New rows in the component_table or comptable*
+
+The default file name for the component table is: ``desc-tedana_metrics.tsv``
+
+classification:
+    In the final table, the only values should be 'accepted' or 'rejected'.
+    While the decision table is bring running, there may also be intermediate
```
eurunuela, line 33: What do you mean by "the decision table is bring running"?

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+    classification labels. Note, nothing in the current code requires a tree to
+    assign one of these two labels to every component. There will be a warning
+    if other labels remain.
+
+classification_tags:
+    Human readable tags that explain why a classification was reached. These can
+    be things like 'Likely BOLD', 'Unlikely BOLD', 'low variance' (i.e. accepted
+    because the variance is too low to low a degree of freedom by calling it noise).
```
eurunuela, line 41: Typo? "too low to low"

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+
+  intermediate_classifications:
+      A list of intermediate classifications (i.e. "provisionalaccept", "provisionalreject"). It is very important to prespecify these
+      because the code will make sure only the default classifications ("accepted" "rejected" "unclassified") and intermediate classifications
+      are used in a tree. This prevents someone from accidentially losing a component due to a spelling error or other minor variation in a
+      classification label
+
+  classification_tags:
+      A list of acceptable classification tags (i.e. "Likely BOLD", "Unlikely BOLD", "Low variance"). This will both be used to make sure only
+      these tags are used in the tree and allow programs that interact with the results one place to see all potential tags
```
eurunuela, line 140:
```suggestion
      these tags are used in the tree and allow programs that interact with the results to see all potential tags in one place
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+- "parameters": Specifications of all required parameters for the function in functionname
+- "kwargs": Specification for optional parameters for the function in functionname
+
+The only parameter that is used in all functions is "decidecomps" which is used to identify, based on their classifications,
+the components a function should be applied to. It can be a single classification, or a comma separated string of classifications.
+In addition to the intermediate and default ("accepted" "rejected" "unclassified") component classifications, this can be "all"
+for functions that should be applied to all components regardless of their classifications
+
+Most decision functions also include "ifTrue" and "ifFalse" which specify how to change the classification of each component
+based on whether a the decision criterion is true or also. In addition to the default and intermediate classification options,
```
eurunuela, line 163:
```suggestion
based on whether a the decision criterion is true or false. In addition to the default and intermediate classification options,
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+Key parts of selection functions
+********************************
+
+There are several expectations for selection functions that are necessary for them to properly execute.
+In selection_nodes.py, manual_classify, dec_left_op_right, and calc_kappa_rho_elbows_kundu are good
+examples for how to meet these expectations.
+
+Create a dictionary called "outputs" that includes key fields that should be recorded. 
+The following line should be at the end of each function ``selector.nodes[selector.current_node_idx]["outputs"] = outputs`` 
+Additional fields can be used to log funciton-specific information, but the following fields are common and may be used by other parts of the code:
```
eurunuela, line 185:
```suggestion
Additional fields can be used to log function-specific information, but the following fields are common and may be used by other parts of the code:
```

I like how "funciton" sounds though haha

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+
+- "decision_node_idx" (required): the ordered index for the current function in the decision tree.
+- "node_label" (required): A decriptive label for what happens in the node.
+- "numTrue" & "numFalse" (required for decision functions): For decision functions, the number of components labels true or false within the function call.
+- "used_metrics" (required if a function uses metrics): The list of metrics used in the function. This can be hard coded, defined by input parameters, or empty.
+- "used_cross_component_metrics" (required if a function uses cross component metrics): A list of cross component metrics used in the function. This can be hard coded, defined by input parameters, or empty.
+- "calc_cross_comp_metrics" (required for calculation functions): A list of cross component metrics calculated within the function. The key-value pair for each calculated metric is also included in "outputs"
+
+Before anything data are touched in the function, there should be an ``if only_used_metrics:`` clause that returns ``used_metrics`` for the function call
```
eurunuela, line 194:
```suggestion
Before anything data are touched in the function, there should be an ``if only_used_metrics:`` clause that returns ``used_metrics`` for the function call.
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+
+Existing functions define ``function_name_idx = f"Step {selector.current_node_idx}: [text of function_name]`` This is used in logging and is cleaner to initialize near the top of each function.
+
+
+Each function has code that creates a default node label in ``outputs["node_label"]``. The default node lable
```
eurunuela, line 199:
```suggestion
Each function has code that creates a default node label in ``outputs["node_label"]``. The default node label
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+may be used in decision tree visualization so it should be relatively short. Within this section, if there is
+a user-provided custom_node_label, that should be used instead.
+
+Calculation nodes should check if the value they are calculating was already calculated and output a warning if the function overwrites and existing value
```
eurunuela, line 203:
```suggestion
Calculation nodes should check if the value they are calculating was already calculated and output a warning if the function overwrites an existing value.
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+
+Code that adds the text log_extra_info and log_extra_report into the appropriate logs (if they are provided by the user)
```
eurunuela, line 205:
```suggestion
Code that adds the text log_extra_info and log_extra_report into the appropriate logs (if they are provided by the user).
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+If there are no components with the classifications in ``decide_comps`` this logs that there's nothing for the function to be run on, else continue.
+
+For decision functions the key variable is ``decision_boolean`` which should be a dataframe column which is True or False for the components in ``decide_comps``
+based on the function's criteria. That column is an input to ``change_comptable_classifications`` which will update the component_table classifications,
+update the classification history in component_status_table, and update the component classification_tags. Components not in ``decide_comps`` retain their
+existing classifications and tags.
+``change_comptable_classifications`` also outputs and should assign values to ``outputs["numTrue"]`` and ``outputs["numFalse"]``.
+These log how many components were identified as true or false within each function.
+
+For calculation functions, the calculated values should be added as a value/key pair to both ``selector.cross_component_metrics`` and ``outputs``
```
eurunuela, line 229:
```suggestion
For calculation functions, the calculated values should be added as a value/key pair to both ``selector.cross_component_metrics`` and ``outputs``.
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+Every function should end.
+
+.. code-block:: python
+
+      selector.nodes[selector.current_node_idx]["outputs"] = outputs
+      return selector
+
+  functionname.__doc__ = (functionname.__doc__.format(**decision_docs))
+
+This returns makes sure the outputs from the function are saved in the class structure and the class structure is returned.
```
eurunuela, line 242:
```suggestion
This "return" makes sure the outputs from the function are saved in the class structure and the class structure is returned.
```

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+The following line should include the function's name and is used to make sure repeated variable names are compiled correctly for the API documentation.
+
+If you have made it this far, congratulations. 
+If you follow these steps you'll be to impress your colleagues, friends, and family by designing your very own decision tree functions.
```
eurunuela, line 246: Love it!

eurunuela (commented):
I have just gone through the docs page as requested in our last meeting and have made some comments mostly related to typos.

The text is clear to me but I might biased after you explained the ideas behind the modularization in our last meeting.

I'll have a look again once the code is ready for reviews.

docs/building_decision_trees.rst
```diff
@@ -0,0 +1,246 @@
...
+    For calculation (calc) functions, cross component metrics that were
+    calculated in this function. When this is included, each of those
+    metrics and the calculated values are also distinct keys in 'outputs'.
+    While the cross component metrics table does not include where each component
+    was calculated, that information is stored here.
+
+
+*********************************
+Defining a custom a decision tree
+*********************************
```
tsalo, line 99:
```suggestion
*******************************
Defining a custom decision tree
*******************************
```

tedana/resources/config/outputs.json
```diff
@@ -171,6 +171,18 @@
         "orig": "ica_metrics",
         "bidsv1.5.0": "desc-tedana_metrics"
     },
+    "ICA cross component metrics json": {
+        "orig": "ica_cross_component_metrics",
+        "bidsv1.5.0": "desc-ICA_cross_component_metrics"
```
tsalo, line 176: This isn't BIDS compliant.

jbteves: How would we make it BIDS compliant?

jbteves: `desc-ICA_metrics-xcomp`?

eurunuela:
I was reading the docs again and I think it would be super helpful if every time we talk mention, e.g., `necessary_metrics` or `functionname`, we linked those keywords to where they are in the minimal decision tree. This way we direct users to an example, which will probably help them understand the docs better.

jbteves:
@handwerkerd had to overwrite your changes for simplicity for mmix/black formatting, sorry.

jbteves:
@eurunuela as an FYI I have created a new class, InputHarvester, that reads an OutputGenerator's "registry" of files from a previous run. You can use this to get all of the information you might want from a tedana run. See "tedana_reclassify.py" for an example of how this is done.

tedana/io.py
```diff
@@ -25,6 +25,28 @@
 RefLGR = logging.getLogger("REFERENCES")
 
 
+class CustomEncoder(json.JSONEncoder):
+    """Convert some types because of JSON serialization and numpy
+    incompatibilities
+
+    # noqa: E501
+    See here: https://stackoverflow.com/questions/50916422/python-typeerror-object-of-type-int64-is-not-json-serializable/50916741
```
tsalo, line 33:
```suggestion
    See here: https://stackoverflow.com/q/50916422/2589328
```
I grabbed the share link.

docs/building_decision_trees.rst
tsalo: I'm hesitant to drop the denoised data completely, but I agree that users' general workflow should probably be to get out the components and perform the regression separately. That way, they can include other regressors (e.g., motion regressors) and orthogonalize as needed afterwards. I'm not sure if orthogonalizing w.r.t. the accepted components should happen within tedana via the `tedort` option or afterwards with the other regressors included though.

CesarCaballeroGaudes: Regarding [Moia et al., 2021](https://www.sciencedirect.com/science/article/pii/S1053811921001919), I would emphasize that that's a very special experimental condition, such as cerebrovascular reactivity, where the task produces a global, widespread response and multiple task-related components were classified as non-BOLD by TEDANA. Therefore, we had to do a manual classification, and even within the noisy components we observed substantial correlation with the task because breath-holding respiratory and motion effects are highly correlated with the task.

tedana/resources/decision_trees/kundu.json
jbteves: Thoughts @handwerkerd ?

tedana/resources/decision_trees/kundu.json
jbteves: This would involve a few substantial changes... I'm not convinced that YAML is worth the switch. You can always add a comment by making a field called `_comment`, which would get ignored in the current code.

docs/building_decision_trees.rst
jbteves: This is a pretty big workflow change, I'd like to separate it from the decision tree modularization just because that's already really big. Could we discuss this in an issue?

tsalo: Yes, this seems more like a separate documentation issue than something to be dealt with in this PR. #823 is probably a good place to follow up.

tedana/resources/decision_trees/kundu.json
handwerkerd: There are two terms used in the code `necessary_metrics` and `used_metrics`. `necessary_metrics are declared up-front. With this input, we can take a tree, make a list of all necessary metrics and calculate only those metrics. `used_metrics` is added to has metrics are used by the decision tree. At the end, there is a check to make sure no used metrics were undeclared in `necessary_metrics`. 

The other reason I used this terminology is, if we eventually do set up the code to calculate metrics based on a decision tree, then we can have `necessary_metrics` that are used in the tree and something like `additional_metrics` which should be calculated, but not used.

I'm not wed to this exact terminology, and am open to ideas for better descriptive terms, but I think `metrics` alone is insufficient.

handwerkerd:
I created a useful/ugly bash script to run the same dataset through Main and then this DTM branch with both the kundu and minimal tree & then compare using [dtm_tools](https://github.com/jbteves/dtm_tools). [Here's the script.](https://gist.github.com/handwerkerd/1d42114c00c01e06d819cc53e887a280) It might require adjustments for other systems, but hopefully sharing this will save time for anyone who wants to run comparisons.

I think Kundu in this branch now matches main so I hope several of you can test this out on more data.

jbteves:
For some reason ComponentSelector's docs do not render all of the method docstrings in our readthedocs build. I can't seem to figure out why, though it looks like the only templates are for functions. Does anybody know of a good primer on how to make one for a class and integrate it into our doc structure? Pinging @tsalo and @emdupre in particular.

pyproject.toml
```diff
@@ -1,5 +1,13 @@
+[project]
+name = "tedana"
+version = "0.0.12"
+
+[project.scripts]
+tedana = "tedana.workflows.tedana:_main"
+tedana_reclassify = "tedana.workflows.tedana_reclassify:_main"
```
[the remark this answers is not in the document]
handwerkerd: I generally agree, but there are now a lot of functions that will can be used more generally. I'd slightly lean towards keeping `tedana_` so that people know where this program came from, but I'd also be fine if other's prefer `ica_` Let's discuss at the next dev call.

pyproject.toml
```diff
@@ -1,5 +1,13 @@
+[project]
+name = "tedana"
+version = "0.0.12"
```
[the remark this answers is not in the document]
handwerkerd: No clue why this got hard coded. I suspect adding `[project]` and `[project.scripts]` was needed to get both workflows to compile for the command line, but I'm not sure if the version number was also a necessary addition. @jbteves?

tedana/workflows/tedana.py
```diff
@@ -270,49 +281,34 @@ def _get_parser():
...
-        type=int,
-        nargs="+",
-        help=("List of manually accepted components. Requires --ctab and --mix."),
-        default=None,
+    parser.add_argument(
+        "--force",
+        "-f",
+        dest="force",
+        action="store_true",
+        help="Force overwriting of files. Default False.",
```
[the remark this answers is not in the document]
handwerkerd: I'm fine changing `force` to `overwrite`

tedana/workflows/tedana.py
```diff
@@ -680,67 +657,70 @@ def tedana_workflow(
...
+            "countnoise",
+            "countsigFT2",
+            "countsigFS0",
+            "dice_FT2",
+            "dice_FS0",
+            "signal-noise_t",
+            "variance explained",
+            "normalized variance explained",
+            "d_table_score",
+        ]
```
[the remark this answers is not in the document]
handwerkerd: Not yet. We can rewrite tedana.py to take any tree as an input and only calculate the required metrics listed in that tree, but we decided that should be a separate PR. To do it well, it would be best to make that change as part of a refactored tedana.py #920. Everything is set up to make that change possible, but we messing with enough of the code base already that we wanted to keep tedana.py relatively stable (We removed functionality from tedana.py but we didn't change the remaining functionality)

tedana/selection/component_selector.py
```diff
@@ -0,0 +1,560 @@
...
+            )
+            self.start_idx = 0
+        else:
+            # Since a status table exists, we need to skip nodes up to the
+            # point where the last tree finished
+            self.start_idx = len(tree_config["nodes"])
+            LGR.info(f"Start is {self.start_idx}")
+            self.component_status_table = status_table
+
+    def select(self):
```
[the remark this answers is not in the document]
handwerkerd: Thinking this over. Assuming we want to keep the option open to load a tree and then create a component table from the `required_metrics` in the tree then we wouldn't want those parameters in `init`. We might not want them in `select` because `select` can be called on a tree with existing information and an existing component_table (i.e. by `tedana_reclassify`). To account for multiple use cases, it might make sense to have 3 steps & 4 functions: 
`init`
`load_component_table` OR `calc_component_table`
`selection`
I think that would be a better change to make when we map out those two intermediate functions & how they interact with workflows. Until then, I think it would make sense to better separate `component_table`, `cross_component_metrics`, and `status_table` within `init` OR `select` so that a future change would be easier.

tedana/selection/selection_nodes.py
```diff
@@ -0,0 +1,1810 @@
...
+
+    selector.tree["nodes"][selector.current_node_idx]["outputs"] = outputs
+
+    return selector
+
+
+manual_classify.__doc__ = manual_classify.__doc__.format(**decision_docs)
+
+
+def dec_left_op_right(
```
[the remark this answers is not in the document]
handwerkerd: I tried. Nearly everything that's consistent across all functions is modularized. This is the simplest I could make these while preserving the functionality and flexibility we wanted. A lot of the remaining complexity is documenting provenance info that's unique to each function. I'm happy to work with someone else to simplify these more, but, unless there are obvious ways to simplify, maybe that can happen after this PR is merged.

tedana/resources/decision_trees/minimal.json
```diff
@@ -0,0 +1,231 @@
...
+            "parameters": {
+                "ifTrue": "provisionalaccept",
+                "ifFalse": "provisionalreject",
+                "decide_comps": "unclassified",
+                "op": ">=",
+                "left": "kappa",
+                "right": "kappa_elbow_kundu"
+            },
+            "kwargs": {
+                "log_extra_report": ""
```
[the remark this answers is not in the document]
handwerkerd: What part specifically? The `log_extra_report` fields?

pyproject.toml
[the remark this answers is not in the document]
jbteves: Because versioneer was called in `setup.py`, which no longer exists.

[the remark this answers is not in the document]
emdupre: we can maintain version information using pyproject.toml using the [guidelines outlined here](https://effigies.gitlab.io/posts/python-packaging-2023/). I'm not sure, though, how you want that to interact with this PR ? Like @tsalo , hardcoding in versions generally makes me nervous.

tedana/workflows/tedana_reclassify.py
```diff
@@ -0,0 +1,484 @@
...
+
+    # Write out BIDS-compatible description file
+    derivative_metadata = {
+        "Name": "tedana Outputs",
+        "BIDSVersion": "1.5.0",
+        "DatasetType": "derivative",
+        "GeneratedBy": [
+            {
+                "Name": "tedana_reclassify",
+                "Version": __version__,
```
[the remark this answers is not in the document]
handwerkerd: Not sure what you're asking here. Do you want to store the version of `tedana_reclassify` and also maintain the version for the original `tedana` that was run?

pyproject.toml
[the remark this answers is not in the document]
handwerkerd: At our dev call, everyone was neutral or preferred `ica_reclassify` so we should change the name

pyproject.toml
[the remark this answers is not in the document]
handwerkerd: The previous versioning system ran into problems with this PR, so we do need a fix in this PR.

There was no consensus at the dev call on the best approach, but my opinion was that we really want the git hashes logged for anyone running the development branch. If I accurately interpreted the consensus, it was: If someone has the time & skills to do this right, that's the person who gets to chose the approach.

@jbteves also suggested we could log the git hashes/versions for every module which would make it easier to track.

tedana/workflows/tedana.py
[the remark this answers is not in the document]
handwerkerd: Dev call consensus was change to `overwrite` 
I think @jbteves was joking when he said to keep the short version as `-f` but please don't. It's already non-ideal that the list of "files" is `-d`. `-o` is for the output directory so maybe there should just not be a short option name.

tedana/selection/component_selector.py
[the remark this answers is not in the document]
handwerkerd: Dev call consensus was to table this until after the PR is merged. (If I have the time I might clean things up so it will be easier in the future). Let's keep this thread open until I either make the partial change we we open an issue so that this doesn't slip.

tedana/workflows/tedana_reclassify.py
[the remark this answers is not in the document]
tsalo: I was thinking that the reclassification workflow could compare its own version against the tedana version listed in the incoming dataset's dataset_description, and raise a warning if the versions are different, since major version differences would mean that the incoming data and workflow being run might be incompatible (e.g., if filenames were changed). We are going to try to ensure that doesn't happen, so I think a warning makes more sense than an error.

tedana/resources/decision_trees/minimal.json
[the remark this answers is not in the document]
tsalo: Yeah, I think one key thing tedana has always been missing is a sense of what each step is for and why it should be included. If we could have all of these messages filled in (at least for the minimal tree), it would really make the tree more transparent.

eurunuela:
I see the documentation does not mention the _reclassify_ workflow. The RICA paragraph will need to get updated based on what we write about it.

This comment is a follow-up of https://github.com/ME-ICA/tedana/pull/756/commits/a65a4cfe2c3fa60476b37471d20add68049f1561

handwerkerd (dismissed):

handwerkerd (approved):

handwerkerd:
@ME-ICA/tedana-devs Today at 2:00PM EST ([Your time zone](https://www.timeanddate.com/worldclock/fixedtime.html?msg=tedana+decision+tree+modularization+merge&iso=20230511T02&p1=2214&ah=1)), we are are planning to relase the last version of tedana before the major refactor (v0.0.13) and then merge this PR and release the more modularized version (v23.0.0). 

Since this is way-too-many years in the making, we'll do this over zoom. Come join the "fun"  at  https://nih.zoomgov.com/j/1612837388?pwd%3DK1drdXVkK0xER1hEbkNzbUljQ0ZoUT09&sa=D&source=calendar&usd=2&usg=AOvVaw1cvaBfqiQE-iNPsBQwHmk5
Meeting ID: 161 283 7388
Passcode: 153769

tsalo (approved):
LGTM!

PR #759, opened 2021-07-19
[ENH] Coerce images to 32-bit

### LINKED ISSUE #757: Integer-based output images are unreadable with AFNI
### Summary

@handwerkerd noticed in #736 that the adaptive mask output is unreadable in AFNI and MicroGL. This appears to be due to the fact that the reference image contains floats, so there may be a mismatch between the image header and the data.

This stems from https://github.com/ME-ICA/tedana/pull/736#issuecomment-878605551.

### Additional Detail

@jbteves thinks that `new_img_like` (nilearn's function) might be coercing the datatype incorrectly, so he will look into that element. If there _is_ a bug on nilearn's side and we can get it fixed, then we'll need to update our requirements (and wait on the next nilearn release).

In the meantime, we will coerce the data to float and save that in the next release, then try to address the underlying issue for the release after that.

### Next Steps
1. Coerce img data to float in `new_nii_like` when creating new images as a patch for the next release.
2. Look into saving integer imgs for the release after next.

jbteves:
Closes #757  .

- Downscales 64-bit float, int arrays to 32-bit arrays just before image save
- Adds some notes to the `io.save_img` docstring

### DIFF
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -214,6 +214,10 @@ class OutputGenerator():
             Data to save to a file.
         name : str
             Full file path for output file.
+
+        Notes
+        -----
+        Will coerce 64-bit float and int arrays into 32-bit arrays.
         """
         data_type = type(data)
         if not isinstance(data, np.ndarray):
@@ -225,6 +229,14 @@ class OutputGenerator():
                 "Data must have number of dimensions in (1, 2), not "
                 f"{data.ndim}"
             )
+        # Coerce data to be 32-bit max in the cases of float64, int64
+        # Note that int64 niftis cannot be read by mricroGL, AFNI
+        vox_type = data.dtype
+        if vox_type == np.int64:
+            data = np.int32(data)
+        elif vox_type == np.float64:
+            data = np.float32(data)
+        # Make new img and save
         img = new_nii_like(self.reference_img, data)
         img.to_filename(name)
 


tedana/io.py
```diff
@@ -225,6 +229,14 @@ def save_img(self, data, name):
                 "Data must have number of dimensions in (1, 2), not "
                 f"{data.ndim}"
             )
+        # Coerce data to be 32-bit max in the cases of float64, int64
+        # Note that int64 niftis cannot be read by mricroGL, AFNI
+        vox_type = data.dtype
+        if vox_type == np.int64:
```
tsalo, line 235: What about `np.issubdtype(data.dtype, np.integer)`? That should be more robust to other int types, like `int8`.

tsalo (commented):
The changes you've made look good to me. I have one question/recommendation, but I was also thinking maybe we should have something in the `outputs` page about the change in datatypes?

tedana/io.py
jbteves: I don't want to upscale an int8 to an int32 if we don't have to, as it just wastes space (and time, since the upcast requires allocating a large 32-bit integer array). The main issue is storing something not widely usable (int64) or something that has more precision than needed (float64), so I singled those two types out.

tsalo: That makes sense. Is int64 the only datatype not usable by AFNI and/or mricroGL?

jbteves: I'm guessing that `uint64` is probably not supported if `int64` is not, but there's no reason to suppose that we'll end up with `uint64` arrays unless we explicitly ask for them, which seems unlikely and solvable when we get to it if so.

jbteves:
Hm, I'm not sure what we'd put in the outputs page. This change should be relatively invisible aside from freer disks 🙂. Programs should read everything correctly, and `float32` is the most widely supported data type of them all. I think only developers need to be aware, but maybe there's something I'm not thinking of.

tedana/io.py
tsalo: Thanks for clarifying. The way you have sounds good then.

tsalo (approved):
That's fair. I guess not many people will care. Approving!

handwerkerd (approved):

PR #760, opened 2021-07-19
[FIX] Ignore non-significant kappa elbow when no non-significant kappa values exist

### LINKED ISSUE #752: `getelbow` raises uninformative exception when supplied with an empty array
### Summary
`tedana.selection._utils.getelbow()` and `getelbow_cons()` are functions to identify elbows in one-dimensional arrays (like you would look for in a scree plot). They are used in multiple places in the component selection decision tree. However, in each of those cases, one or more of the arrays being supplied to these functions may be empty, in which case the functions will return something like the following traceback:

```
  File "/opt/anaconda3/bin/tedana", line 33, in <module>
    sys.exit(load_entry_point('tedana', 'console_scripts', 'tedana')())
  File "/Users/handwerkerd/code/tedana_community/master/tedana/tedana/workflows/tedana.py", line 805, in _main
    tedana_workflow(**kwargs)
  File "/Users/handwerkerd/code/tedana_community/master/tedana/tedana/workflows/tedana.py", line 586, in tedana_workflow
    comptable, metric_metadata = selection.kundu_selection_v2(
  File "/Users/handwerkerd/code/tedana_community/master/tedana/tedana/selection/tedica.py", line 250, in kundu_selection_v2
    kappa_elbow = np.min((getelbow(kappas_nonsig, return_val=True),
  File "/Users/handwerkerd/code/tedana_community/master/tedana/tedana/selection/_utils.py", line 85, in getelbow
    p = coords - coords[:, 0].reshape(2, 1)
IndexError: index 0 is out of bounds for axis 1 with size 0
```

I think there are a few places where we can improve how these arrays are handled:

1. Within the `getelbow` functions, we should raise a more informative Exception (or a warning) when an empty array is provided.
2. The functions could return a NaN when an empty array is provided.
3. In cases where multiple calls to the function are combined (see below), we could use the NaN-robust equivalent (e.g., `numpy.nanmean` instead of `numpy.mean`).
    https://github.com/ME-ICA/tedana/blob/d449925378c54c51e7dce55769372f6b2a6242b3/tedana/selection/tedica.py#L250-L251
4. We should review and identify any cases in the decision trees where the functions _might_ fail. That means any time the functions are run on an array that might be empty. There are some places where that's possible and some where it's not.

### Additional Detail
This has come up as a sort of random error in multiple issues: #181, #297, #370, and https://github.com/ME-ICA/tedana/pull/736#issuecomment-878605551, although it does seem to be fairly rare and unpredictable.

### Next Steps
1. Decide whether a more informative exception or returning NaN is preferable.

tsalo:
Closes #752. In the ICA decision tree, there is a step where two elbows are calculated: one from all kappa values and one from non-significant kappa values. When there are no non-significant kappa values, this step fails.

- Ignore the non-significant kappa value-based elbow when there are no non-significant kappa values. Log a warning about this when it comes up.
- For any other cases where `getelbow` or `getelbow_cons` fail, raise a more informative exception and include in that message a request to open an issue here.
- Add new smoke test for `getelbow` and `getelbow_cons`.

### DIFF
--- a/tedana/selection/_utils.py
+++ b/tedana/selection/_utils.py
@@ -45,9 +45,11 @@ def getelbow_cons(arr, return_val=False):
     if not arr.size:
         raise ValueError(
             "Empty array detected during elbow calculation. "
-            "The cause behind this failure is not currently known, so please open an issue at "
+            "This error happens when getelbow_cons is incorrectly called on no components. "
+            "If you see this message, please open an issue at "
             "https://github.com/ME-ICA/tedana/issues with the full traceback and any data "
-            "necessary to reproduce this error."
+            "necessary to reproduce this error, so that we create additional data checks to "
+            "prevent this from happening."
         )
 
     arr = np.sort(arr)[::-1]
@@ -92,9 +94,11 @@ def getelbow(arr, return_val=False):
     if not arr.size:
         raise ValueError(
             "Empty array detected during elbow calculation. "
-            "The cause behind this failure is not currently known, so please open an issue at "
+            "This error happens when getelbow is incorrectly called on no components. "
+            "If you see this message, please open an issue at "
             "https://github.com/ME-ICA/tedana/issues with the full traceback and any data "
-            "necessary to reproduce this error."
+            "necessary to reproduce this error, so that we create additional data checks to "
+            "prevent this from happening."
         )
 
     arr = np.sort(arr)[::-1]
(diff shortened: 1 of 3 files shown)


jbteves (approved):
LGTM, thanks @tsalo ! Thanks in particular for adding tests as well so that this doesn't regress.

handwerkerd (dismissed):
The warning message in the getelbow functions seems a bit more dramatic than necessary, but otherwise fine.
If you feel like changing it, I'd change to "This error happens when a getelbow function is incorrectly called on no components. If you see this message, please open an issue at https://github.com/ME-ICA/tedana/issues with the full traceback and any data necessary to reproduce this error, so that we create additional data checks to prevent this from happening."

handwerkerd (approved):

jbteves (approved):
LGTM

PR #763, opened 2021-07-21
[DOC, MAINT] Updated dependency version numbers

### LINKED ISSUE #761: Updating minimum required python package versions (nilearn)
### Summary

I just found out that nilearn added carpet plots in version 0.7 (I was still on 0.6). We currently don't have any minimum version requirement for nilearn, so we should add 0.7. 

### Additional Detail

Checking into this, I also noticed nilearn is soon phasing out nibabel 2.0 support. I don't know if we should already restrict the minimum version, but we should probably keep track of this. nilearn also lists it's minimum reqs at numpy 1.16, SciPy 1.2, scikit-learn 0.21, nibabel 2.5, & pandas 0.24. Since we are using a bunch of these through nilearn, we might want to consider changing some of those versions to our new minimums.

### Next Steps

- Use this as an excuse to check if there are any other version minimums that are also out-of-date. Decide which version minimums to change now.

Change the versions listed in:
- README.md
- docs/installation.rst
- tedana/info.py
- Anywhere else?

handwerkerd:
Closes #761  .

- Updates minimum version requirements for python dependencies in tedana/info.py to match nilearn>=0.7
- Removes minimum version requirements from docs/installation.rst & README.md so those will only need to be updated in one place

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,11 @@ More information and documentation can be found at https://tedana.readthedocs.io
 You'll need to set up a working development environment to use `tedana`.
 To set up a local environment, you will need Python >=3.6 and the following packages will need to be installed:
 
-* [numpy>=1.14](http://www.numpy.org/)
+* [numpy](http://www.numpy.org/)
 * [scipy](https://www.scipy.org/)
 * [scikit-learn](http://scikit-learn.org/stable/)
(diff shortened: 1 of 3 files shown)


eurunuela (approved):
LGTM!

tsalo (approved):
Looks great. Thanks!

PR #765, opened 2021-07-25
[DOC] Improved walkthrough with GIFs

### LINKED ISSUE #690: Refine multi-echo fMRI walkthrough
### Summary
I'm working on improving our walkthrough of multi-echo fMRI and sources of noise in fMRI in my [`improved-walkthrough` branch](https://github.com/tsalo/tedana/tree/improved-walkthrough). Here I'd like to document the approach I want to take.

### Additional Detail

1. How does single-echo fMRI work?
    1. A sequence diagram.
    1. What is BOLD signal?
        1. Basic physics refresher: excitation, T2 relaxation, signal decay, T2*, blood oxygenation, BOLD signal
    1. Links to references.
    1. Sources of noise: non-neural BOLD signal, non-BOLD noise, any others?
2. What is multi-echo fMRI?
    1. A sequence diagram.
    1. Links to references.
    1. What does the signal looks like? I can recycle some figures from our "processing pipeline details" page here.
    1. Sources of noise. T2* and S0.
    1. How do fluctuations in S0 and T2* impact signal?
        1. Show the simulated decay curve change as S0/T2* fluctuate (see **Figure 1** and **Figure 2** below).
        1. Show the simulated multi-echo signal (points without decay curve).
        1. Show both simulated multi-echo signals from equivalent S0 and T2* fluctuations, so we can see how they _differentially_ impact the signal.
    1. What do these fluctuations look like in single-echo data? Show how it's impossible to dissociate T2* and S0 fluctuations in single-echo data.
1. Denoising multi-echo data
    1. Why don't we just estimate T2*/S0 directly from the data?
        1. Motivate TEDICA and other denoising approaches.
        1. Discuss FIT, maybe.
    1. Can basically recycle the TE-dependence models page for this.
1. What _can't_ multi-echo fMRI deal with?
    1. BOLD-based noise, like sLFOs and physiological signals.
    1. Brain vs. vein (i.e., mis-localization of neural activity due to nearby macrovasculature).

#### Figure 1: The impact of T2* fluctuations on multi-echo signal

t2star_fluctuations

#### Figure 2: The impact of S0 fluctuations on multi-echo signal

s0_fluctuations

tsalo:
Closes #690. Still a WIP.

https://tedana--765.org.readthedocs.build/en/765/multi-echo_walkthrough.html

The more I think about it, the more I wonder if maybe we should move things like the multi-echo walkthrough into a different location, like a Jupyter Book, with https://naturalistic-data.org as inspiration.

GIFs are going to take up a lot of space and bulk up our git repository, but I _do_ think they're really helpful for learning about multi-echo fMRI.

- Add gifs to help readers build a more intuitive understanding of multi-echo denoising.


eurunuela:
I agree that a Jupyter Book would fit the walkthrough much better.

GIFs are definitely useful to understand the content, but I wonder if a dash app or an [Observable](https://observablehq.com) notebook would be nicer. Folks would get to "learn by playing" with interactive plots.

emdupre:
I'd vote for "all of the above" if folks have time and interest !

Specifically on this point:

> GIFs are going to take up a lot of space and bulk up our git repository, but I do think they're really helpful for learning about multi-echo fMRI.

Do we have to host the GIFs here ? Could we host them on e.g. OSF and just render them in the docs ? I've also successfully hosted animated SVGs as Gists in the past, which could be another option ?

tsalo:
Would it be difficult to set up the Jupyter book in another repo and just let folks add to it as they're willing/able?

>Could we host them on e.g. OSF and just render them in the docs ?

That seems feasible. I'll try it out.

docs/multi-echo_walkthrough.rst
```diff
@@ -0,0 +1,314 @@
...
+BOLD signal and BOLD contrast
+*****************************
+
+As mentioned above, energy is released after the excitation pulse, and this is our observed fMRI signal.
+The echo time is the point at which that signal is recorded.
+
+Let's take a look at how fMRI signal varies as a function of echo time.
+
+.. image:: https://osf.io/m7aw3/
+   :alt: physics_signal_decay.png
```
tsalo, line 51: @emdupre this doesn't seem to be working. Any ideas?

emdupre: I just requested access to that OSF link to see what it includes !

emdupre: I think you should modify to the "render" link, e.g. 
https://mfr.osf.io/render?url=https://osf.io/mx4ku/?direct%26mode=render%26action=download%26mode=render

But updating the url as appropriate !

tsalo: Hm... it seems like it should work, but it doesn't seem to help:

image

emdupre: Here's what's happening and (at least) two suggestions.
During the build process, Sphinx is trying to copy the files into the generated `_build` directory. Because these are embedded images on a web page, this process fails and results in what you're showing above. To resolve this, we could (in no particular order) :

- Update those image tags to iframes, to embed the web pages
- Fetch the relevant files using cURL or the OSF API during the documentation build process so they would exist appropriately in the `_build` directory.
- Switch to another provider (or revert to GitHub) where we can access the files themselves.

Do you have a preference, @tsalo  ?

tsalo: Thanks for digging into this.

>Do you have a preference, @tsalo ?

I'm not sure. I'm leaning toward iframes for right now, but I also really want to try out the Jupyter Book idea, in which case the GIFs would be generated by, and rendered in, the notebooks.

tsalo:
I'm just going to close this in favor of the Jupyter Book. If that doesn't work out for some reason, I can reopen this.

PR #769, opened 2021-07-29
[DOC] Generalize preprocessing recommendations

### LINKED ISSUE #755: Generalize "Processing multi-echo fMRI" section of docs
### Summary

With #740, the documentation were reorganized and the "Processing multi-echo fMRI" section was moved from "Using tedana" to "About ME-fMRI". Unfortunately, at the moment, that section is pretty tedana-specific, and includes very little general information about processing ME-fMRI data. We'd like to add some stuff about processing this kind of data in general, as well as change `tedana` to `multi-echo denoising` in the processing descriptions.

This stems from https://github.com/ME-ICA/tedana/pull/740#pullrequestreview-707672224.

tsalo:
Closes #755.

- Replace `tedana` with "multi-echo denoising" in recommendations for preprocessing multi-echo data.
- Add a step about motion correction.

### DIFF
--- a/docs/multi-echo.rst
+++ b/docs/multi-echo.rst
@@ -401,17 +401,33 @@ The following plots reflect the average values for studies conducted at 3 Tesla.
 Processing multi-echo fMRI
 **************************
 
-``tedana`` must be called in the context of a larger ME-EPI preprocessing pipeline.
+Most multi-echo denoising methods, including ``tedana``,
+must be called in the context of a larger ME-EPI preprocessing pipeline.
 Two common pipelines which support ME-EPI processing include `fMRIPrep`_ and `afni_proc.py`_.
 
-Users can also construct their own preprocessing pipeline for ME-EPI data from which to call ``tedana``.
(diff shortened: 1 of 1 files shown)


handwerkerd (approved):
LGTM.
My one critique is that, when I do normalize, I've tended to do one normalization before tedana because it's better to do only one spatial transform. That step shouldn't mess up the relationship between echoes more than motion correction (though normalization with aggressive non-linear warping might). I don't think we clearly know what is better or worse yet so the next could be more nuanced in this area. This is a critique of the current recommendations, not the changes in this PR, so there's no reason to hold this PR up to discuss this issue.

tsalo:
Maybe we could include the caveat that that part, at least, is based on theory rather than empirical evidence?

handwerkerd (approved):

jbteves (approved):
LGTM, thanks @tsalo ! I think this will be a bit easier to point people to since I've heard questions about motion correction with multi-echo several times now.

tsalo:
Thanks @jbteves and @handwerkerd! Merging now.

PR #770, opened 2021-07-31
Replace current Bokeh-based reports with React-based reports

### LINKED ISSUE #583: Manual component classification in reports
### Summary
We've discussed this on calls before, but I'd like to open an issue about it. Now that we have lovely reports, we should be able to update our component classifications _within_ the report.

### Additional Detail
This would take a lot of effort and is more of a long-term goal, but I think it's worth doing.

eurunuela:
Closes #583 and potentially a couple of other issues too (#650 #687 ?).

- Main change is replacing the Bokeh-based (Python) report with a React-based (JavaScript) report. This should allow us to update the ICA classification while exploring the reports because the data will actually be a variable of the JavaScript code.
- Replacing the one-page layout of the reports with a tabbed layout, where different data is shown in different tabs (e.g., ICA components, carpet plots, system info, ...).

TODO:

- [ ] Simplify the Python functions that generate the reports.
- [ ] Add toggle button to change the classification of components.
- [x] Add React code to the `reporting` directory.
- [x] Add a final step to `tedana` to build the reports with `npm`. By building the reports the user will only need to open an HTML file as they do now.
- [x] Make sure all reports-related files are saved in the correct folder structure for the reports to work.
- [ ] Update the docs to reflect the changes made by this PR and describe how the new reports work.


eurunuela:
Following the discussion in today's meeting, I'm closing this PR in favor of the [ICA reports](https://github.com/ME-ICA/ica-reports) repository.

ICA reports will aim to provide a manual classification app for ICA performed with `tedana` and `aroma`.

PR #772, opened 2021-08-16
[FIX] Add resources folder to package data

tsalo:
Closes None. I was just testing a new install from the `main` branch and got the following:

```
  File "/home/data/nbc/misc-projects/Salo_PowerReplication/code/processing/04_run_multi-echo_denoising.py", line 255, in <module>
    run_tedana(project_dir, dset)
  File "/home/data/nbc/misc-projects/Salo_PowerReplication/code/processing/04_run_multi-echo_denoising.py", line 127, in run_tedana
    prefix=prefix,
  File "/home/data/nbc/misc-projects/Salo_PowerReplication/code/conda_env/lib/python3.7/site-packages/tedana/workflows/t2smap.py", line 220, in t2smap_workflow
    make_figures=False,
  File "/home/data/nbc/misc-projects/Salo_PowerReplication/code/conda_env/lib/python3.7/site-packages/tedana/io.py", line 84, in __init__
    config = load_json(config)
  File "/home/data/nbc/misc-projects/Salo_PowerReplication/code/conda_env/lib/python3.7/site-packages/tedana/io.py", line 308, in load_json
    with open(path, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/data/nbc/misc-projects/Salo_PowerReplication/code/conda_env/lib/python3.7/site-packages/tedana/resources/config/outputs.json'
```

It looks like we forgot to add the new json files to our package data so that they'll be bundled with the library in distributions.

- Add `resources/config/*` to package data in `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ def main():
             'tests/data/*',
             'reporting/data/*',
             'reporting/data/html/*',
-            'resources/*',
+            'resources/config/*',
         ]
     }
 
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ def main():
             'tests/data/*',
             'reporting/data/*',
             'reporting/data/html/*',
+            'resources/*',
         ]
     }
 


jbteves:
Intriguingly, I pulled `main` and it still executed the four-echo test just fine, so it's difficult for me to tell if this PR actually fixes the issue since I can't reproduce it. Question, though: why not use `find` instead of manually listing the directories in `setup.py`?

tsalo:
>Intriguingly, I pulled main and it still executed the four-echo test just fine

When you say you pulled main, do you mean you used git? Or did you use pip? If you pulled the repository, then the data files will still be available.

To reproduce:

```
pip install git+https://github.com/ME-ICA/tedana.git
```
If you then run tedana on example data, it should raise the FileNotFoundError.

>Question, though: why not use find instead of manually listing the directories in setup.py?

I was just focused on fixing the bug, rather than improving our packaging. If we want to change our packaging, I would probably use `setup.cfg` ([like I do in NiMARE](https://github.com/neurostuff/NiMARE/blob/3f3bfa75942dfe709bbd1ec1d096874711673d00/setup.cfg#L20-L26)), based on Chris M's packaging recommendations: https://gist.github.com/effigies/9bbb424535d6a1d838d6325191c0a736

jbteves:
Ah, I see. I didn't realize that was the failure mode; now reproduced. Okay, I think this will patch it. We may want to change packaging as part of our next release, but we can discuss tomorrow and there's no reason to hold this up. Thanks for the fix!

jbteves (approved):
LGTM! Thanks @tsalo !

eurunuela (approved):
LGTM!

tsalo:
Thanks @jbteves and @eurunuela! Merging now.

PR #786, opened 2021-08-25
[REF] Suppresses divide by 0 warning

### LINKED ISSUE #762: Warning in true_divide from dice calculation
### Summary
While running `main`, we can currently experience a RuntimeWarning from dividing by zero during the dice coefficient calculation. This is due to some changes to the dice calculation function during the metric modularization.

### Next Steps
- [ ] Patch

jbteves:
Closes #762  . Note that there is some root issue causing the divide by 0 in the first place, but I haven't figured out why it's happening. Since we at least want to suppress the warning for now, this is a small patch to only suppress the warning.

- Simply suppresses warning
-


tedana/utils.py
```diff
@@ -198,7 +199,12 @@ def dice(arr1, arr2, axis=None):
         dsi = np.zeros(arr_sum.shape)
     else:
         intersection = np.logical_and(arr1, arr2)
-        dsi = (2. * intersection.sum(axis=axis)) / arr_sum
+        if np.any(arr_sum == 0):
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
```
tsalo, line 204:
```suggestion
                warnings.simplefilter("ignore", RuntimeWarning)
```

If we're specific enough about the warning we catch, we could just get rid of the entire if statement starting at line 198. Right now we check if _all_ thresholded maps are empty, then here we check if some are empty, and in the new `else` we run when none are empty. We should be able to just filter the warning and calculate across all maps without checking if any are empty, right?

tsalo: Or I guess we could mask things and unmask them? Maybe with a debug message reporting the number of empty maps?

jbteves: I agree with your suggestion, I actually meant to add that filter, whoops.
Yes, we can just not check for emptiness, but I'm inclined to think we actually _should_ warn users if _all_ of the maps are empty. The issue is it's difficult to make a meaningful warning in that context.

tsalo: What about a warning of how many are empty? Then we don't have to check if all are empty, then any are empty. We check if any are empty, raise a warning with how many, and then outside that if statement we apply the warning filter?

jbteves: That would be pretty sensible and straightforward. We also can add a prompt for the users to examine the component table for dice values of 0.

tsalo: I like it!

tedana/utils.py
```diff
@@ -208,7 +209,21 @@ def dice(arr1, arr2, axis=None):
...
+                "zero-elements in the denominator. "
+                "Please check your component table for dice columns with 0-"
+                "values"
+            )
+
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+                dsi = (2.0 * intersection.sum(axis=axis)) / arr_sum
+        else:
+            dsi = (2.0 * intersection.sum(axis=axis)) / arr_sum
```
tsalo, line 226:
```suggestion

        with warnings.catch_warnings():
            warnings.simplefilter("ignore")
            dsi = (2.0 * intersection.sum(axis=axis)) / arr_sum
```

Can we have the filter target the divide-by-zero warning specifically?

tedana/utils.py
```diff
@@ -208,7 +209,21 @@ def dice(arr1, arr2, axis=None):
...
         intersection = np.logical_and(arr1, arr2)
-        dsi = (2.0 * intersection.sum(axis=axis)) / arr_sum
+        # Count number of zero-elements in the denominator and report
+        total_zeros = np.count_nonzero(arr_sum == 0)
+        if total_zeros > 0:
+            LGR.warning(
+                f"Calculating dice coefficient with {total_zeros} "
+                "zero-elements in the denominator. "
+                "Please check your component table for dice columns with 0-"
+                "values"
```
tsalo, line 219: How about something like `f"{total_zeros}/{arr_sum.size} components have empty maps, resulting in Dice values of 0. Please check your..."`

tedana/utils.py
```diff
@@ -208,7 +209,21 @@ def dice(arr1, arr2, axis=None):
         dsi = np.zeros(arr_sum.shape)
     else:
```
tsalo, line 210: I can't tag the whole if statement above, but it can be removed now that you have the warning and filter below.

tsalo (approved):
The changes look good to me. I don't know if you're still running into a styling issue, but the linter job in the CI is passing, so I'm happy to see it merged.

eurunuela (approved):
LGTM! Thank you!

jbteves:
I forgot, did we pause this merge for some reason? I feel like I recall some hesitation due to not being sure about precisely what caused the underlying issue.

eurunuela:
> I forgot, did we pause this merge for some reason? I feel like I recall some hesitation due to not being sure about precisely what caused the underlying issue.

I honestly do not remember we paused it for any reason.

PR #795, opened 2021-09-10
Add kappa, rho and variance ranks to metrics table for React reports.

eurunuela:
Closes none.

- Add kappa, rho and variance explained rank columns to `desc-tedana_metrics.tsv` for React-based reports.

In order to replicate the Bokeh-based reports, we need the ranks of kappa, rho, and variance explained too.

Edit: Most of the changes here are due to Black and isort not being implemented throughout tedana yet.

### DIFF
--- a/tedana/resources/config/metrics.json
+++ b/tedana/resources/config/metrics.json
@@ -13,7 +13,15 @@
             "map FT2",
             "map Z"
         ],
+        "kappa rank": [
+            "map FT2",
+            "map Z"
+        ],
         "rho": [
+            "map FS0",
+            "map Z"
+        ],
+        "rho rank": [
             "map FS0",
             "map Z"
         ],
@@ -41,6 +49,9 @@
             "map FT2"
         ],
         "variance explained": [
+            "map optcom betas"
+        ],
+        "variance explained rank": [
             "map optcom betas"
         ],
         "normalized variance explained": [
--- a/tedana/metrics/collect.py
+++ b/tedana/metrics/collect.py
@@ -4,14 +4,11 @@ import os.path as op
 
 import numpy as np
 import pandas as pd
-
-from . import dependence
-from ._utils import determine_signs, flip_components, dependency_resolver
-
-from tedana import io
-from tedana import utils
+from tedana import io, utils
 from tedana.stats import getfbounds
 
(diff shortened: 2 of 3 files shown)


tedana/resources/config/metrics.json
```diff
@@ -13,10 +13,18 @@
             "map FT2",
             "map Z"
         ],
+        "kappa rank": [
+            "map FT2",
+            "map Z"
+        ],
```
tsalo, line 19: Shouldn't this just be calculated from `kappa`?

eurunuela: It should. But since `kappa rank` is the same data but ranked I thought it should contain the same attributes.

tsalo: I think we should just consider basic "rank" versions of metrics as derivatives of just the metrics they're ranking. The dependency graph will have more layers (i.e., `kappa rank` --> `kappa` --> `map Z` + `map FT2` instead of `kappa rank` --> `map Z` + `map FT2`), but it should be more interpretable that way.

handwerkerd: I agree with this. I like the idea of using the attributes to more clearly show which ones are derivative. Perhaps it's obvious that kappa_rank is just kappa values ranked, setting up this rule now might make future additions of metrics a bit more clear. This is already what we do with d_table_score.

handwerkerd:
This seems to be something that risks bloating up the component table with a bunch of other metrics that can be easily calculated from existing metrics. It would be my preferences to do the ranking calculation to be done in the React based report. If that's not possible, I guess this is conceptually fine.

handwerkerd:
This seems to be something that risks bloating up the component table with a bunch of other metrics that can be easily calculated from existing metrics. It would be my preferences to do the ranking calculation to be done in the React based report. If that's not possible, I guess this is conceptually fine.

eurunuela:
> This seems to be something that risks bloating up the component table with a bunch of other metrics that can be easily calculated from existing metrics. It would be my preferences to do the ranking calculation to be done in the React based report. If that's not possible, I guess this is conceptually fine.

That's a fair point @handwerkerd . I'll try to implement it in React and will reopen this PR if I don't find a way to do it.

PR #805, opened 2021-09-20
[ENH] Create new `datasets` module, with Cambridge dataset included

### LINKED ISSUE #684: Add dataset fetching module

tsalo:
Closes #684. Follows up on #720.

- Add `datasets` module.
- Include Cambridge dataset.
- Add new "Fetching methods" page to the documentation.
- Tests.

Commit log:

* initial commit of datasets module
* Light clean-up, add in Cambridge regressor fetching
* Update docstrings with correct references, details
* Functioning cambridge fetcher
* save in tedana directory rather than nilean directory
* Add missing function
* return list of tuples for iterating
* Apply suggestions from code review


tsalo:
For posterity, here's the code I used to get the low-resolution file hashes. I'm not going to pretend it's well-written (except the parts I took from @emdupre's OSF gist):

```python
import os.path as op
import pandas as pd
import json
import re
import requests


def compile_name(arg):
    sub, echo = arg
    name = f"{sub}_task-rest_{echo}_space-scanner_desc-partialPreproc_res-5mm_bold.nii.gz"
    return name

df = pd.read_csv("cambridge_echos.csv")
df["filename"] = df[["participant_id", "echo_id"]].apply(compile_name, axis=1)

if not op.isfile("TEMP.csv"):
    # If statement for re-running.
    # From https://gist.github.com/emdupre/3cb4d564511d495ea6bf89c6a577da74
    repo = '5n968'  # my example repository, update as appropriate
    query = ''  # my example query, update or leave blank as appropriate

    url = 'https://api.osf.io/v2/nodes/{0}/files/osfstorage/{1}/60864ffba9746601032f9392/'.format(repo, query)
    guids = []
    while True:
        resp = requests.get(url)
        resp.raise_for_status()
        data = json.loads(resp.content)
        for i in data['data']:
            sub = re.search(r'sub-(\S+)_task', i['attributes']['name']).group(1)
            guids.append((sub, i['attributes']['name'], i['id']))

        url = data['links']['next']
        if url is None:
            break
    df_lowres = pd.DataFrame(data=guids, columns=["participant_id", "filename", "hash"])
    df_lowres.to_csv("TEMP.csv", index=False)
else:
    df_lowres = pd.read_csv("TEMP.csv")

df_lowres = df_lowres[["filename", "hash"]]
df2 = pd.merge(df, df_lowres, how="outer", left_on="filename", right_on="filename")
df2 = df2[["participant_id", "echo_id", "key_bold_nativeres", "hash"]]
df2 = df2.rename(columns={"hash": "key_bold_lowres"})

df2.to_csv("cambridge_echos.csv", index=False)
```

tsalo:
I've addressed everything I wanted to, so of course the three-echo integration test is stalling out at the ICA step 😭

tsalo:
Per today's call, I'm working on a JSON file linking filenames to URLs _and_ the filenames within the BIDS folder structure, like this:

```json
{
    "sub-04570_task-rest_desc-confounds_timeseries.tsv": {
        "key": "6033160934d30403d1e39b74",
        "out_file": "sub-04570/func/sub-04570_task-rest_desc-confounds_timeseries.tsv"
    },
    "sub-04570_task-rest_echo-1_space-scanner_desc-partialPreproc_res-5mm_bold.nii.gz": {
        "key": "6086502a7166fa00eae90cee",
        "out_file": "sub-04570/func/sub-04570_task-rest_echo-1_space-scanner_desc-partialPreproc_res-5mm_bold.nii.gz"
    }
}
```

There are two blockers that I can think of:

1. We need a generic filtering function to identify the files we want to download. We can feed in values via keyword arguments (e.g., `filter_files(sub="04570", task="rest")`), but I'm concerned about how to deal with files that don't have matching entities. For example, confounds won't have the resolution (res) entity because they have no associated resolution, so we would either have to have two highly specific searches (so we don't grab both 5mm and native-res BOLD data) or one search that is greedy (which might grab too many files). Does that make sense?
2. How should we structure the output? Presumably we will support many types of data, including masks, transforms, confounds files, etc., so the nilearn approach where there are a couple of reserved fields probably doesn't make sense.

tsalo:
Ohhh, wait a second. What about something like this:

```json
{
    "sub-04570/func/sub-04570_task-rest_desc-confounds_timeseries.tsv": {
        "key": "6033160934d30403d1e39b74",
        "groups": [
            "basic_nativeres",
            "basic_lowres"
        ]
    },
    "sub-04570/func/sub-04570_task-rest_echo-1_space-scanner_desc-partialPreproc_res-5mm_bold.nii.gz": {
        "key": "6086502a7166fa00eae90cee",
        "groups": [
            "basic_lowres"
        ]
    }
}
```

And then `group` could be an optional keyword, and users could grab a specific subset of files that they want _or_ filter files based on BIDS entity-based keyword arguments. It doesn't completely resolve my second blocker (how to structure outputs), but I think it would adequately address the first one.

jbteves:
I'm a fan of this approach, it would alleviate most of my concerns. Great idea @tsalo !

tsalo:
This is so out-of-date that I think it would be better to start over from scratch than try to update this PR.

PR #822, opened 2021-10-22
[ENH] Remove accepted and rejected outputs

### LINKED ISSUE #819: Remove accepted and rejected time series from the standard file outputs
### Summary

### Next Steps

tsalo:
Closes #819.

- Drop accepted and rejected component-related outputs.
- Remove unused `n_vols` parameter from `io.writeresults`.
- Remove "optcom" from certain output names.
- Update `docs/outputs.rst`.
- Update integration test outputs.

## Removed derivatives

- desc-optcomRejected_bold.nii.gz
- desc-optcomAccepted_bold.nii.gz
- desc-ICAAccepted_components.nii.gz
- desc-ICAAccepted_stat-z_components.nii.gz
- echo-[echo]_desc-Accepted_bold.nii.gz
- echo-[echo]_desc-Rejected_bold.nii.gz
- desc-optcomAcceptedMIRDenoised_bold.nii.gz
- desc-ICAAcceptedMIRDenoised_components.nii.gz

## Renamed derivatives

- desc-optcomMIRDenoised_bold.nii.gz —> desc-MIRDenoised_bold.nii.gz
- echo-{echo}_desc-Denoised_bold.nii.gz —> echo-{echo}_desc-denoised_bold.nii.gz
- desc-optcomDenoised_bold.nii.gz —> desc-denoised_bold.nii.gz

To do:
- [x] Document prefix a bit more, so it's clear that the prefix isn't automatically inferred. From https://github.com/ME-ICA/tedana/issues/146#issuecomment-982676687.


tedana/resources/config/outputs.json
```diff
@@ -27,25 +27,9 @@
         "orig": "betas_OC",
         "bidsv1.5.0": "desc-ICA_stat-z_components"
     },
-    "ICA accepted components img": {
-        "orig": "betas_hik_OC",
-        "bidsv1.5.0": "desc-ICAAccepted_components"
```
handwerkerd, line 32: Is this the first place we should add a new file name version code? That is we are changing file names and preventing people from saving files with these older names. We can rename the current ones something like "bidsv1.5.0old". While I don't think anyone would want to use these older names, the big benefit is, if someone has files with these old names, they'd/we'd be able to quickly remember what they now correspond to.

handwerkerd (commented):
This looks good overall.

tedana/resources/config/outputs.json
tsalo: BIDS is on v1.6.0 right now, so I'd lean toward adding that and changing the default over renaming v1.5.0. WDYT?

docs/outputs.rst
```diff
@@ -120,16 +112,14 @@ desc-optcomNoGlobalSignal_bold.nii.gz               Optimally combined time seri
                                                     removed.
 ================================================    =====================================================
 
-If ``gscontrol`` includes 't1c':
+If ``gscontrol`` includes 'mir':
```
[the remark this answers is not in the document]
tsalo: I've added links to the individual functions' docstrings here:

image

WDYT?

tedana/resources/config/outputs.json
tsalo: I now realize that we should have done something like this for #736 as well... and we should probably have our different versions reflected in the outputs tables too!

handwerkerd: Aligning with v1.6.0 with this PR sounds sounds like a great solution.

Did #736 change any file names or just change what was in some of the files? If it changed a file name, I could see back-adding that into the output filename table, but this seems really minor. I'd say don't bother or consider it a really low priority.

tsalo: It changed the T2* and S0 maps. T2starmap.nii.gz --> desc-limited_T2starmap.nii.gz and desc-full_T2starmap.nii.gz --> T2starmap.nii.gz.

tsalo: Beyond adding the v1.6.0 naming convention (and reverting name changes, but not removals, from v1.5.0) in 280771c, here's what I'm thinking could help:

image

Unfortunately, this would involve working through our commit history to pin down when each filename was changed and which tedana versions the changes apply to, which I don't want to tackle in this PR, but I think the overall idea is solid.

handwerkerd: I suggest just opening a new issue for adding in the intermediate file naming convention.

FWIW, when we made this file naming decision, I think we talked about creating new file naming options that were linked to a released version. That is, we'd always be able to say "Save files as they were in tedana v0.0.9" I don't see that in the current code, but this discussion shows why such an option might be useful. Whatever we call each naming convention, there would also be an ordered list saying in which version each convention was the default. This wouldn't be too hard to add, but it would definitely be a new issue.

jbteves: @handwerkerd that is correct, I proposed that if this becomes an issue we could alias a few tedana versions to certain naming conventions (not track them in the JSON file itself).

handwerkerd: The logical place to do this would be in the input parser in the init of `io.OutputGenerator` The one quirk is that would  mean every time we edit outputs.json, we'd also want to add to the parser to peg the current version with the previous file names.

tsalo:
I believe we decided recently that it would be easier to just start over from scratch on this instead of trying to fix up the PR.

PR #824, opened 2021-11-03
[FIX] Fixes broken component figures in report when there are more than 99 components

manfredg:
Closes # .

This is a very minor bug fix that makes the 0-padding in the report Javascript code more robust. Specifically, the component figures didn't display correctly for any components > 99 since they were padded to 0100, 0101 etc instead of just 100. 

-
-

### DIFF
--- a/tedana/reporting/dynamic_figures.py
+++ b/tedana/reporting/dynamic_figures.py
@@ -17,11 +17,8 @@ tap_callback_jscode = """
         // -----------------------------
         var components = data['component']
         var selected = components[selected_idx]
-        var selected_padded = '' + selected;
-        while (selected_padded.length < 2) {
-            selected_padded = '0' + selected_padded;
-        }
-        var selected_padded_forIMG = '0' + selected_padded
+        var selected_padded = String(selected).padStart(3,0);
+        var selected_padded_forIMG = selected_padded
         var selected_padded_C = 'ica_' + selected_padded
 
         // Find color for selected component
--- a/tedana/reporting/dynamic_figures.py
+++ b/tedana/reporting/dynamic_figures.py
@@ -17,7 +17,7 @@ tap_callback_jscode = """
         // -----------------------------
         var components = data['component']
         var selected = components[selected_idx]
-        var selected_padded = String(selected).padStart(3,0);
+        var selected_padded = String(selected).padStart(3,0)
         var selected_padded_forIMG = selected_padded
         var selected_padded_C = 'ica_' + selected_padded
 


eurunuela (approved):
LGTM! Thank you @manfredg !

I have a question though. How long is your acquisition? >100 sounds like too many components. I wonder if the PCA step was not good.

manfredg:
> LGTM! Thank you @manfredg !
> 
> I have a question though. How long is your acquisition? >100 sounds like too many components. I wonder if the PCA step was not good.

That's something I wanted to bring up in a separate issue. My first run on the dataset I am currently processing gave me 0 accepted components and only a handful rejected components. From my previous experience with Prantik's original ME-ICA code I would expect at least 20 accepted components for a dataset to pass QC. Also, since this is a resting state dataset which will be fed into a correlation based connectivity analysis, I need more than just a handful of DoFs in order to get a meaningful correlation matrix. It seems the current code is tuned towards minimizing the number of components, which is nice for the visualisation in the report, but is getting rid of too much information for my purposes.
I managed to increase the number of components (beyond 100 total) by requiring 90-95% variance explained using the --tedpca argument.

eurunuela:
> > LGTM! Thank you @manfredg !
> > I have a question though. How long is your acquisition? >100 sounds like too many components. I wonder if the PCA step was not good.
> 
> That's something I wanted to bring up in a separate issue. My first run on the dataset I am currently processing gave me 0 accepted components and only a handful rejected components. From my previous experience with Prantik's original ME-ICA code I would expect at least 20 accepted components for a dataset to pass QC. Also, since this is a resting state dataset which will be fed into a correlation based connectivity analysis, I need more than just a handful of DoFs in order to get a meaningful correlation matrix. It seems the current code is tuned towards minimizing the number of components, which is nice for the visualisation in the report, but is getting rid of too much information for my purposes. I managed to increase the number of components (beyond 100 total) by requiring 90-95% variance explained using the --tedpca argument.

What `--tedpca` option did you use initially? I assume you used `mdl`, which is the most "aggressive" (and default) option. You could try to use `aic` instead and see if you get more components. However, the fact that you got 0 accepted components could be related to the decision tree that gets run after ICA.

FYI: you could also use [Rica](https://github.com/ME-ICA/rica) to manually classify your components.

emdupre (approved):
This LGTM -- thank you, @manfredg !

eurunuela:
Thanks @emdupre . I'll merge.

We can keep the discussion in an issue if you think it's necessary @manfredg .

tsalo:
@allcontributors add @manfredg for code

Issue #827, opened 2021-11-15
Less aggressive masking for first echo


handwerkerd:
### Summary

There are several anecdotes where CSF and even white matter voxels are being masked. This isn't necessarily a problem for tedana denoising, but may affect how tedana interacts with other pipelines.  (For example: https://github.com/NBCLab/power-replication/issues/14 )
 
These voxels probably shouldn't be masked. Tedana masking is fairly simple and focuses on voxels with relatively lower raw magnitudes. We could see if there is a way to still mask out dropout regions while retaining the ventricles.

### Next Steps

- Agree this is something worth addressing
- Look at other masking methods to see if there's an existing light-weight method

tsalo:
What I ended up doing in the fork I'm using for that analysis is to assume that all voxels in any user-provided mask has good data in at least the first two echoes. It probably means that we end up relying on our T2* value restrictions more in those voxels, which is unfortunate, but given how we perform optimal combination, it shouldn't be a big problem.

handwerkerd:
For your fork, what if you assume the first echo only is good and you'd remove voxels that drop too much from the first to second echo? Right now, we're independently creating a mask for each echo. In the revised version, you'd find the median value for each echo & get median_pchange = (median_echo1-median_echo2)/(median echo1). If pchange in a voxel > X*median_pchange (say for X=2) then assuming you're seeing significant dropout rather than a decay curve and mask that voxel.

This is something I could see playing with in tedana proper, but I'm not sure how to prioritize. I originally set this as medium priority, but I just shifted to low priority. Do you think this would be better to address sooner rather than later or is it more of an edge case?

tsalo:
>For your fork, what if you assume the first echo only is good and you'd remove voxels that drop too much from the first to second echo?

My biggest issue in my analysis was that CSF and WM voxels were being removed, which prevented me from aCompCor. Dropping _any_ voxels from the fMRIPrep brain mask would have been a problem.

>In the revised version, you'd find the median value for each echo & get median_pchange = (median_echo1-median_echo2)/(median echo1). If pchange in a voxel > X*median_pchange (say for X=2) then assuming you're seeing significant dropout rather than a decay curve and mask that voxel.

[...]

PR #837, opened 2022-02-03
[NO-MRG] Changes from Power replication

tsalo:
Closes None. This PR includes changes I made to tedana for a manuscript of mine. The big problem I had to solve was that voxels necessary for CSF/WM extraction (e.g., for aCompCor) were being masked out with the adaptive mask (see #827 and https://github.com/NBCLab/power-replication/issues/14).

- Retain all voxels in the explicit mask.


tsalo:
This was never meant to be merged. My main goal was/is to highlight changes I needed to make to keep CSF and WM voxels for aCompCor calculations _after_ multi-echo denoising. Those concerns still apply, but I don't think we're going to end up adopting these specific changes.

Issue #47, opened 2018-05-14
Finish adding math to function docstrings


tsalo:
We started adding in relevant math and detailed descriptions to the Notes sections of several function docstrings in #43, but it's far from done.

emdupre:
It sounds like @handwerkerd and Jen Evans would be great people to help with this !

jbteves:
I'm interested in helping with this, it would be a good exercise to familiarize myself with the codebase.

jbteves:
Due to @tsalo's large PR #247 and issues #214, #249, #250 I think this issue should be paused.

jbteves:
We have numerous refactors and documentation discussions that have happened in the past month; I think that we should discuss unpausing this and converting it to a larger documentation project on the call, WDYT @handwerkerd @emdupre @tsalo ?

jbteves:
I think this has sat open long enough. As part of the sprint we should look for individual gaps in documentation and fill them as needed; closing.

PR #839, opened 2022-02-08
Print optimal number of maPCA components and plot optimization curves

### LINKED ISSUE #834: Make MA-PCA criteria curves more accessible to users
### Summary

The number of PCA components is defined using one of several methods. [A recent neurostars question](https://neurostars.org/t/tedana-mask-and-components/21117/10) and an issue I've been examining would have been easier to answer if the criteria curves used to calculate the number of components were more accessible. I propose outputting these curves either by default or when the --verbose option is used.

### Next Steps

- @eurunuela has already made these values slightly more accessible in this PR: https://github.com/ME-ICA/mapca/pull/43 
- Have that PR also return either the full PCA object or the three criteria curves so that tedana can access that information
- Add code [after this line](https://github.com/ME-ICA/tedana/blob/64fa9d0b96df949a8ef38d450c37ca528f40f283/tedana/decomposition/pca.py#L236) to save the numbers in the criteria curves for aic, kic, and mdl.

eurunuela:
Closes #834 .

- [x] Gets the optimal number of components and optimization curves for all three criteria in maPCA.
- [x] Plots the optimization curves of all three criteria in maPCA.
- [x] Prints the optimal number of components given by all three criteria to the logs. 
- [x] Print and plot optimal number of components given by 90% and 95% of variance explained when maPCA returns enough components.

Edit: this draft PR will fail until [#47](https://github.com/ME-ICA/mapca/pull/47) and [#48](https://github.com/ME-ICA/mapca/pull/48) are merged in maPCA.


eurunuela:
@handwerkerd  I don't know what you want the `PCA_cross_component_metrics` file to contain exactly, so I am willing to leave that to you. What do you think of the changes I've made so far in this PR?

handwerkerd:
I was hoping to read this carefully today, but it's not going to happen. I'll try to give a closer read soon. When I was discussing, `PCA_cross_component_metrics` it was in reference to `cross_component_metrics` that I've added to the component selection class in #756 Here is one example where it is used:
https://github.com/ME-ICA/tedana/blob/f011ba2c3a51eafd0a4d6429bbbfa4686baebd9d/tedana/selection/selection_nodes.py#L647-L652
The basic idea is to have a dictionary for all values that are calculated based on metrics across components (i.e. the kappa and rho elbows). The complements the component table, where each metric has a value for each component. In the current code, these cross component values are either in the log and nowhere else or not saved at all. Since you're criterion thresholds are the same general concept, I figured you could create a similar dictionary with a similar naming style.

eurunuela:
The three-echo test should pass once we merge [PR #50 in maPCA](https://github.com/ME-ICA/mapca/pull/50).

eurunuela:
The test will pass once [PR #50](https://github.com/ME-ICA/mapca/pull/50) is merged in maPCA.

handwerkerd:
I just tried to locally run the three-echo test with mapca after [#50](https://github.com/ME-ICA/mapca/pull/50) was merged. The test is still failing in io.py with
```
Object of type ndarray is not JSON serializable
  File "[/Users/handwerkerd/code/tedana_community/me-ica/tedana/tedana/io.py]()", line 253, in save_json
    json.dump(data, fo, indent=4, sort_keys=True)
  File "[/Users/handwerkerd/code/tedana_community/me-ica/tedana/tedana/io.py]()", line 201, in save_file
    self.save_json(data, name)
  File "[/Users/handwerkerd/code/tedana_community/me-ica/tedana/tedana/decomposition/pca.py]()", line 322, in tedpca
    io_generator.save_file(mapca_results, "PCA cross component metrics json")
  File "[/Users/handwerkerd/code/tedana_community/me-ica/tedana/tedana/workflows/tedana.py]()", line 638, in tedana_workflow
    dd, n_components = decomposition.tedpca(
  File "[/Users/handwerkerd/code/tedana_community/me-ica/tedana/tedana/tests/test_integration.py]()", line 181, in test_integration_three_echo (Current frame)
    tedana_cli.tedana_workflow(
```
The four-echo test is passing so this isn't a universal issue. I'm not sure I have time to dig into this more right now, but wanted to share.

tsalo:
I think we need to automatically convert numpy arrays to lists in `save_json`.

tedana/decomposition/pca.py
```diff
@@ -234,9 +235,92 @@ def tedpca(
...
-        voxel_comp_weights, varex, varex_norm, comp_ts = ma_pca(
-            data_img, mask_img, algorithm, normalize=True
+        ma_pca = MovingAveragePCA(criterion=algorithm, normalize=True)
+        _ = ma_pca.fit_transform(data_img, mask_img)
+
+        # Extract results from maPCA
+        voxel_comp_weights = ma_pca.u_
+        varex = ma_pca.explained_variance_
+        varex_norm = ma_pca.explained_variance_ratio_
+        comp_ts = ma_pca.components_.T
```
jbteves, line 245: This looks like we're accessing private components of the object. Could we either link to some documentation about the returned object or elaborate on what we're accessing here?

tedana/reporting/static_figures.py
```diff
@@ -288,3 +288,144 @@ def comp_figures(ts, mask, comptable, mmix, io_generator, png_cmap):
         compplot_name = os.path.join(io_generator.out_dir, "figures", plot_name)
         plt.savefig(compplot_name)
         plt.close()
+
+
+def pca_results(criteria, n_components, all_varex, io_generator):
```
jbteves, line 293: This function has a large amount of repetition. Could we break it into smaller functions, parametrized perhaps by something like the input data and the label only, and if matplotlib requires this, the figure itself?

jbteves:
Hi Eneko, I left some comments but didn't formally review. Let me know if you'd like for me to fix the merge conflicts and some of the refactors I mentioned as possibilities for you. However I do think it's worth considering some more documentation about how we're extracting these values from the mapca object.

tedana/decomposition/pca.py
tsalo: The trailing underscore indicates that the attribute is estimated from the data (via `fit`), rather than that it's private. We adopted this convention from scikit-learn.

jbteves:
Hm, I merged the changes but it seems like some expected attribute is missing. It looks like CircleCI is just caching the last environment, so that it's not checking the versions in `setup.cfg` to update them appropriately, or perhaps some other error I'm missing. Any thoughts @eurunuela @tsalo ?

jbteves:
@eurunuela I think you need to cut a new release with the changes so that we can pin that release.

eurunuela:
> @eurunuela I think you need to cut a new release with the changes so that we can pin that release.

Done!

jbteves (dismissed):
LGTM, thanks @eurunuela !

handwerkerd (changes requested):
Mostly LGTM.
The one thing I notices is that a file called `./figures/pca_variance_explained.png` It looks similar to the relevant file `./figures/pca_criteria.png` but it just contains the vertical dashed lines without the criteria curves. My guess is that this was useful while working on the code, but isn't useful as a saved output. If I'm right, just don't save that file?

eurunuela:
> Mostly LGTM. The one thing I notices is that a file called `./figures/pca_variance_explained.png` It looks similar to the relevant file `./figures/pca_criteria.png` but it just contains the vertical dashed lines without the criteria curves. My guess is that this was useful while working on the code, but isn't useful as a saved output. If I'm right, just don't save that file?

I'm sorry @handwerkerd, that was a mistake on my end. Here's what the figure should look like:
pca_variance_explained

I'm committing the fix now.

handwerkerd:
Yay! That does look useful and I'm glad I noticed. In your sample figure above, I don't see a blue dashed line for AIC.

eurunuela:
> Yay! That does look useful and I'm glad I noticed. In your sample figure above, I don't see a blue dashed line for AIC.

That's because it's right behind the KIC line. The figure was generated with the 3 echo integration test by the way.

PR #847, opened 2022-02-20
[DOC] Add information about using tedana with fMRIPrep v21.0.0

tsalo:
Closes None.

- Add information about fMRIPrep's new `--me-output-echos` parameter.
- Add instructions for applying fMRIPrep's transforms to warp native-space data to standard space.
- Fix a couple of small bugs in the documentation.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -161,14 +161,14 @@ htmlhelp_basename = "tedanadoc"
 
 # The following is used by sphinx.ext.linkcode to provide links to github
 linkcode_resolve = make_linkcode_resolve(
-    "tedana", "https://github.com/me-ica/" "tedana/blob/{revision}/" "{package}/{path}#L{lineno}"
+    "tedana", "https://github.com/me-ica/tedana/blob/{revision}/{package}/{path}#L{lineno}"
 )
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3/", None),
     "numpy": ("https://numpy.org/doc/stable/", None),
-    "scipy": ("https://docs.scipy.org/doc/scipy/reference/", None),
+    "scipy": ("https://docs.scipy.org/doc/scipy/", None),
     "matplotlib": ("https://matplotlib.org/", None),
     "nibabel": ("https://nipy.org/nibabel/", None),
     "pandas": ("https://pandas.pydata.org/pandas-docs/stable/", None),
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -57,32 +57,9 @@ The actual filenames of fMRIPrep derivatives depend on the filenames in the BIDS
 but in this example we chose to use the simple example of "sub-01" and "task-rest".
 The standard space template in this example is "MNI152NLin2009cAsym", but will depend on fMRIPrep settings in practice.
 
-.. code-block:: bash
+.. raw:: html
 
-    # First, let's define some paths to the fMRIPrep and tedana outputs
-    fmriprep_dir="/path/to/fmriprep/derivatives/of/subject"
-    tedana_dir="/path/to/tedana/derivatives"
-
-    # Result of denoising the scanner-space multi-echo data with tedana
-    file_to_warp="${tedana_dir}/sub-01_task-rest_desc-optcomDenoised_bold.nii.gz"
-
-    # Name of the standard-space denoised data file to be written out
-    out_file="${tedana_dir}/sub-01_task-rest_space-MNI152NLin2009cAsym_desc-optcomDenoised_bold.nii.gz"
-
-    # An existing standard-space
-    standard_space_file="${fmriprep_dir}/func/sub-01_task-rest_space-MNI152NLin2009cAsym_boldref.nii.gz"
-
-    # Transforms
-    xform_native_to_t1w="${fmriprep_dir}/func/sub-01_task-rest_from-scanner_to-T1w_mode-image_xfm.txt"
-    xform_t1w_to_std="${fmriprep_dir}/anat/sub-01_from-T1w_to-MNI152NLin2009cAsym_mode-image_xfm.h5"
-
-    antsApplyTransforms \
-        -e 3 \
-        -i ${file_to_warp} \
-        -r ${standard_space_file} \
-        -o ${out_file} \
-        -n LanczosWindowedSinc \
-        -t ${xform_native_to_t1w} ${xform_t1w_to_std}
+    <script src="https://gist.github.com/tsalo/f9f38e9aba901e99ddb720465bb5222b.js"></script>
 
 ************************************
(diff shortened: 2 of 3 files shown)


docs/conf.py
```diff
@@ -161,14 +161,14 @@ def setup(app):
...
-    "tedana", "https://github.com/me-ica/" "tedana/blob/{revision}/" "{package}/{path}#L{lineno}"
+    "tedana", "https://github.com/me-ica/tedana/blob/{revision}/{package}/{path}#L{lineno}"
 )
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3/", None),
     "numpy": ("https://numpy.org/doc/stable/", None),
-    "scipy": ("https://docs.scipy.org/doc/scipy/reference/", None),
+    "scipy": ("https://docs.scipy.org/doc/scipy/", None),
```
tsalo, line 171: I was getting the following when I built the docs locally:

```
intersphinx inventory has moved: https://docs.scipy.org/doc/scipy/reference/objects.inv -> https://docs.scipy.org/doc/scipy/objects.inv
```

docs/faq.rst
```diff
@@ -24,18 +35,46 @@ Unfortunately, fMRIPrep's working directory structure is not stable across versi
...
 Nevertheless, we have some code (thanks to Julio Peraza) that works for version 20.2.1.
 
-.. warning::
-    We will try to keep the following gist up-to-date, but there is no guarantee that it will work for a given version.
-    Use it with caution!
-
-    If you do find that the gist isn't working for an fMRIPrep version >= 20.2.1,
-    please comment on `Issue #717 <https://github.com/ME-ICA/tedana/issues/717>`_ (even if it's closed)
-    and we will take a look at the problem.
-
```
tsalo, line 34: Given that fMRIPrep added an option to output individual echoes, we don't really need to keep the gist up to date.

docs/faq.rst
```diff
@@ -24,18 +35,46 @@ Unfortunately, fMRIPrep's working directory structure is not stable across versi
...
+========================================================
+
+Here is a basic approach to normalizing scanner-space tedana-denoised data created from fMRIPrep outputs,
+using ANTS's antsApplyTransforms tool.
+The actual filenames of fMRIPrep derivatives depend on the filenames in the BIDS dataset
+(e.g., the name of the task, run numbers, etc.),
+but in this example we chose to use the simple example of "sub-01" and "task-rest".
+The standard space template in this example is "MNI152NLin2009cAsym", but will depend on fMRIPrep settings in practice.
+
+.. code-block:: bash
```
tsalo, line 52: Could someone take a look at this code? I've used it, but I don't know if it's the optimal set of settings.

handwerkerd (dismissed):
I'm fine with this being merged.
That said, I have two comments:
- I've reviewed the demo code, but I have not run it. I've never run fMRIprep so I'm really not the right person to confirm it works as intended. It would be nice if an fMRIprep user could try it, but having this in the documentation in it's current form is better than not documenting this at all.
- The length of this section stretches what should be in a FAQ. We really should have a clearer page on pipelines (likely related to #823 ). I again think it's better to merge into the FAQ and move to a more logical place later.

docs/faq.rst
jbteves: I don't know enough about either ants or fmriprep to make an informed choice about the settings, but if you ran it and it works as a starting point it's probably better to have something than nothing (especially since you note it will depend on the settings).

jbteves (approved):
LGTM, thanks @tsalo ! I believe that we don't have any other fmripreppers (is that the formal term?) to check this out but it looks good as a resource for people to start from, so I'm approving.

docs/faq.rst
```diff
@@ -10,7 +10,18 @@ FAQ
...
+
+Starting with version 21.0.0, `fMRIPrep`_ added the ``--me-output-echos`` argument,
+which outputs individual echoes after slice timing, motion correction, and distortion correction have been performed.
+These preprocessed echoes can be denoised with tedana,
+after which warps written out by fMRIPrep can be applied to transform the denoised data to standard space.
+
+fMRIPrep versions < 21.0.0
+==========================
+
+Prior to version 21.0.0, `fMRIPrep`_ outputted the preprocessed, optimally-combined fMRI data, rather than echo-wise data.
```
emdupre, line 24: Technically, it could also output preprocessed, individual echos using the `--echo-idx` option. It might be worth bringing this up only in that previously, folks might have thought they could use those echos in tedana (which we tried to discourage).

docs/faq.rst
emdupre: This is silly, but can we put this in a gist and link it as we did above ? It would both standardize the formatting of the code snippets on this page and decouple updates from the docs.

emdupre: Also the code looks right from the last time I ran something like this ! I'll see if I can pull up a directory to actually run it on.

jbteves: I second the idea of a gist

docs/faq.rst
```diff
@@ -24,18 +35,46 @@ Unfortunately, fMRIPrep's working directory structure is not stable across versi
...
+    xform_native_to_t1w="${fmriprep_dir}/func/sub-01_task-rest_from-scanner_to-T1w_mode-image_xfm.txt"
+    xform_t1w_to_std="${fmriprep_dir}/anat/sub-01_from-T1w_to-MNI152NLin2009cAsym_mode-image_xfm.h5"
+
+    antsApplyTransforms \
+        -e 3 \
+        -i ${file_to_warp} \
+        -r ${standard_space_file} \
+        -o ${out_file} \
+        -n LanczosWindowedSinc \
+        -t ${xform_native_to_t1w} ${xform_t1w_to_std}
```
emdupre, line 77: Just kidding, I see now that you say you've already ran it. In which case, only one small comment:

```suggestion
        -t ${xform_native_to_t1w} \
        -t ${xform_t1w_to_std}
```

tsalo:
Just a note for myself. Jeff Dennison wrote a gist for running tedana on fMRIPrepped data that I'd like to link to in the fMRIPrep-related documentation: https://gist.github.com/jbdenniso/73ec8281229d584721563a41aba410cf

docs/faq.rst
tsalo: Done!

handwerkerd (approved):
LGTM

docs/conf.py
emdupre: It seems like either now work, so I'm fine with this as-is !

emdupre (approved):
I think we can merge as-is ! This will likely only continue to evolve for subsequent versions, so it'd be good to get these updates in -- and we can re-visit as necessary.

Thanks, @tsalo  !

tsalo:
So I completely forgot about this PR. Going to merge it now since I have the necessary approvals.

PR #849, opened 2022-02-22
Made AIC the default maPCA option

### LINKED ISSUE #843: Change default maPCA option to AIC
### Summary

After discussion at our last meeting, and considering the number of questions re maPCA on Neurostars, we should probably change the default maPCA criterion to AIC (the most conservative). This way, users will get more components in what should give a more similar solution to the kundu_stabilize option.

### Next Steps

- Change maPCA criterion to AIC.
- Update test results.

eurunuela:
Closes #843 .

- Updated all references to the default maPCA option from `mdl` to `aic`.
- Made `aic` the default maPCA option.
- Updated three-echo test to use `aic`.

### DIFF
--- a/docs/approach.rst
+++ b/docs/approach.rst
@@ -244,8 +244,8 @@ to select the PCA components based on three widely-used model selection criteria
 
 .. note::
     Please, bear in mind that this is a data-driven dimensionality reduction approach. The default
-    option ``aic`` might not yield perfect results on your data. We suggest you explore the ``kic``
-    and ``mdl`` options if running ``tedana`` with ``aic`` returns more components than expected.
+    option ``aic`` might not yield perfect results on your data. Consider ``kic``
+    and ``mdl`` options if running ``tedana`` with ``aic`` returns more components than expected. There is no definitively right number of components, but, for typical fMRI datasets, if the PCA explains more than 98% of the variance or if the number of components is more than half the number of time points, then it may be worth considering more aggressive threshold.
 
 The simplest approach uses a user-supplied threshold applied to the cumulative variance explained
 by the PCA.
--- a/docs/approach.rst
+++ b/docs/approach.rst
@@ -219,7 +219,7 @@ These components are subjected to component selection, the specifics of which
 vary according to algorithm.
 Specifically, ``tedana`` offers three different approaches that perform this step.
 
-The recommended approach (the default ``mdl`` option, along with the ``aic`` and ``kic`` options, for
+The recommended approach (the default ``aic`` option, along with the ``kic`` and ``mdl`` options, for
 ``--tedpca``) is based on a moving average (stationary Gaussian) process
 proposed by `Li et al (2007)`_ and used primarily in the Group ICA of fMRI Toolbox (GIFT).
 A moving average process is the output of a linear system (which, in this case, is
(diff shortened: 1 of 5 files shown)


eurunuela:
I think this PR will fix the failing three-echo test in #839.

tsalo:
Thanks for this. One thing I would like to see added to our documentation is a very short explanation of why we use AIC by default. Something like...

> We have chosen AIC as the default PCA criterion because it tends to result in fewer components than the Kundu methods, which increases the likelihood that the ICA step will successfully converge, but also, in our experience, retains enough components for meaningful interpretation later on.

I doesn't have to be that, but something to that effect would be nice. I also think this probably deserves a `breaking-change` label.

eurunuela:
> Thanks for this. One thing I would like to see added to our documentation is a very short explanation of why we use AIC by default. Something like...
> 
> > We have chosen AIC as the default PCA criterion because it tends to result in fewer components than the Kundu methods, which increases the likelihood that the ICA step will successfully converge, but also, in our experience, retains enough components for meaningful interpretation later on.

Let me know what you think of the changes added in [c610667](https://github.com/ME-ICA/tedana/pull/849/commits/c61066788779da683d6246dc905b326a74d0bd93).

> I doesn't have to be that, but something to that effect would be nice. I also think this probably deserves a `breaking-change` label.

Done!

handwerkerd:
I'll try to look at this more carefully soon, but one thing I've noticed while running tedana is that there's no logging for various pca options. When the default option is used, I'm not sure anywhere in the log will say what the default option is. I think we should add some logging to places [like this](https://github.com/eurunuela/tedana/blob/c61066788779da683d6246dc905b326a74d0bd93/tedana/decomposition/pca.py#L234-L238)

docs/approach.rst
```diff
@@ -233,17 +233,19 @@ For this PCA method in particular, ``--tedpca`` provides three different options
...
+  criterion because it tends to result in fewer components than the Kundu methods, which increases
+  the likelihood that the ICA step will successfully converge, but also, in our experience, retains
+  enough components for meaningful interpretation later on.
 
 .. note::
     Please, bear in mind that this is a data-driven dimensionality reduction approach. The default
-    option ``mdl`` might not yield perfect results on your data. We suggest you explore the ``kic``
-    and ``aic`` options if running ``tedana`` with ``mdl`` returns less components than expected.
+    option ``aic`` might not yield perfect results on your data. We suggest you explore the ``kic``
+    and ``mdl`` options if running ``tedana`` with ``aic`` returns more components than expected.
```
handwerkerd, line 248:
```suggestion
    option ``aic`` might not yield perfect results on your data. Consider ``kic``
    and ``mdl`` options if running ``tedana`` with ``aic`` returns more components than expected. There is no definitively right number of components, but, for typical fMRI datasets, if the PCA explains more than 98% of the variance or if the number of components is more than half the number of time points, then it may be worth considering more aggressive threshold.
```

I just tested AIC on some 5-echo data with 160 volumes and I got 99.28% variance explained and 134 components. Given the recent spate people with too few components, I think it's better to default to giving too many rather than too few, but I wanted to expand this comment a bit.

handwerkerd (changes requested):
I expanded one comment a bit.
I can't figure out how to suggest a change in an area where you didn't make any edits, but I'd replace
https://github.com/eurunuela/tedana/blob/c61066788779da683d6246dc905b326a74d0bd93/tedana/decomposition/pca.py#L228
with 
`LGR.info(f"Computing PCA of optimally combined multi-echo data with selection criteria: {algorithm}")`

That will have slightly ugly parsing if someone enters a number for `algorithm`, but, if someone is entering a number, they'll hopefully understand what they're doing.

Otherwise, this all looks good to me.

docs/approach.rst
eurunuela: Sounds good to me!

eurunuela:
Thank you for your feedback @handwerkerd, [b28a282](https://github.com/ME-ICA/tedana/pull/849/commits/b28a282f0d10f5b665467d38086e47540c7b2a6d) should address your comments.

handwerkerd (commented):
LGTM

handwerkerd (approved):

jbteves (approved):
LGTM, thanks @eurunuela !

PR #852, opened 2022-03-01
[DOC] Improve logging of component table-based manual classification

### LINKED ISSUE #848: Denoising with manual classification is not transparent enough
### Summary

We should make [denoising with the manual classification](https://tedana.readthedocs.io/en/stable/usage.html#Arguments%20for%20Rerunning%20the%20Workflow) more transparent so that users know exactly what tedana is doing. Currently, tedana does not update the component figures (it should probably output a new folder with new images), nor does it say that the manual classification is being used to denoise the data. This can be confusing for users, so we should probably be more explicit on the prints/logs.

### Additional Detail

From a couple of emails with a tedana user:

> the png's in the 'figures' folder created by tedana appear to be unchanged (e.g., a component that was accepted via RICA is still shown as rejected).
The 'tedana.tsv' does state 'Using supplied mixing matrix from ICA', but there is no mention that a provided manual classification is used as well.
However, quickly eyeballing some voxel timecourses of the resulting 'desc-optcomDenoised_bold.nii' and comparing it to the pre-RICA version of the same file does seem to suggest some changes did occur (but it's kind of difficult to gauge).
It would be great to have this step work somewhat more transparent so users can be confident the updated denoising was actually implemented.

### Next Steps

- Make prints/logs more informative.
- Generate new folder of component figures?

tsalo:
Closes #848.

- Add a manual classification step in the four-echo integration test that uses the component table for the classifications.
- Change the classifications in the five-echo integration test's manual classification step.
- Add a logger message when a component table is provided.

### DIFF
--- a/tedana/tests/test_integration.py
+++ b/tedana/tests/test_integration.py
@@ -72,13 +72,19 @@ def download_test_data(osf, outpath):
 
 
 def test_integration_five_echo(skip_integration):
-    """Integration test of the full tedana workflow using five-echo test data"""
+    """Integration test of the full tedana workflow using five-echo test data."""
 
     if skip_integration:
         pytest.skip("Skipping five-echo integration test")
+
     out_dir = "/tmp/data/five-echo/TED.five-echo"
+    out_dir_manual = "/tmp/data/five-echo/TED.five-echo-manual"
+
     if os.path.exists(out_dir):
         shutil.rmtree(out_dir)
+
+    if os.path.exists(out_dir_manual):
+        shutil.rmtree(out_dir_manual)
 
     # download data and run the test
     download_test_data("https://osf.io/9c42e/download", os.path.dirname(out_dir))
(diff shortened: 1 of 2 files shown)


tsalo:
I'm unable to reproduce the following problem from #848:

>the png's in the 'figures' folder created by tedana appear to be unchanged (e.g., a component that was accepted via RICA is still shown as rejected).

## First pass

Note that the selected component is ignored.

image

## Manual classification

All ignored components were reclassified as accepted in this step, via the component table.

image

eurunuela:
I will try to reproduce it sometime this week. I believe we have another user reporting the same issue in [Neurostars](https://neurostars.org/t/question-about-rica-tedana-manual-ica-classification/21654/12?u=e.urunuela).

tsalo:
@eurunuela it looks like https://github.com/ME-ICA/rica/pull/41 solves the component classification issue, correct? If so, then I think this PR can be reviewed purely from the perspective of documentation.

eurunuela:
> @eurunuela it looks like https://github.com/ME-ICA/rica/pull/41 solves the component classification issue, correct? If so, then I think this PR can be reviewed purely as an improvement to documentation/logging.

Indeed. I'm sorry I didn't have time to report back, it's been a busy evening.

As you say, we should improve documentation/logging.

eurunuela (approved):
I think these changes should be enough.

Should we add something to the docs? And if so, would you rather do that on this PR or open a new one?

tsalo:
I think improving the documentation is a good idea, and I'm happy to incorporate that into this PR. Do you have anything specific in mind?

handwerkerd (approved):
LGTM
I don't have any strong opinions on adding documentation here vs a future PR.
My one general concern is how many different things are being tested within some of these integration tests (i.e. manual classification could be a separate test). This is going to create a lot of "fun" when we get to the point of fully reviewing #756 I suspect non-trivial changes to some of these existing tests will be required. It's beyond the scope of this PR, but I'll think about this more, and possibly open up a new issue so that we can start identifying & cleaning up tests that will interface with the modularized decision tree.

PR #42, opened 2018-05-11
[RF] Reorganize package further.

tsalo:
Based on discussion in #30, I have reorganized the package to have the following structure:

- `tedana.model`
    - `tedana.model.monoexponential`: contains `fit_decay` and `fit_decay_ts`
    - `tedana.model.fit`: contains `fitmodels_direct` and associated functions, minus the `getelbow` functions
    - `tedana.model.combine`: contains `make_optcom`
- `tedana.decomposition`
    - `tedana.decomposition.eigendecomp`: contains `tedpca` and `tedica`
    - `tedana.decomposition._utils`: contains `eimask`
- `tedana.selection`
    - `tedana.selection.select_comps`: contains `selcomps`, to be broken into separate functions
        - Also might need a better name, to distinguish it from other potential selection methods (if anyone has any they want to contribute)
    - `tedana.selection._utils`: contains the `getelbow` functions, because I think they belong here, rather than in `tedana.model`

PR #855, opened 2022-03-09
[FIX] The rationale column of comptable gets updated when no manacc is given

### LINKED ISSUE #854: Reclassifying accepted to rejected raises an error when using Rica's ctab but not manacc
### Summary

The main goal of Rica is to make manual classification easier, and that involves making tedana easier to rerun. Currently, it is possible to rerun tedana without the `--manacc` option using Rica's component table with `--ctab`. However, when an accepted component is rejected with Rica, tedana fails to generate new component figures and raises an error. The error comes from the fact that accepted components have empty cells in their `rationale` column, which are used when plotting rejected and ignored components.

### Next Steps

- Extract the indices of the accepted components from `comptable`.
- Update `comptable` with `selection.manual_selection(comptable, acc=manacc)`, the same way it is done when `--manacc` is given.

eurunuela:
Closes #854 .

- Check that the given `ctab` is Rica's by looking at the column names.
- Update the rationale of `comptable` with `selection.manual_selection(comptable, acc=manacc)`.

### DIFF
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -738,7 +738,12 @@ def tedana_workflow(
         else:
             comptable = pd.read_table(ctab)
 
-            if manacc is not None:
+            # If manacc is None and if comptable has a column named original_classification
+            if manacc is None and "original_classification" in comptable.columns:
+                # Get index of rows that have accepted on their classification column in comptable
+                manacc = comptable.index[comptable['classification']=="accepted"].tolist()
+                comptable, metric_metadata = selection.manual_selection(comptable, acc=manacc)
+            elif manacc is not None:
                 comptable, metric_metadata = selection.manual_selection(comptable, acc=manacc)
 
     # Write out ICA files.
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -737,8 +737,8 @@ def tedana_workflow(
             comptable, metric_metadata = selection.kundu_selection_v2(comptable, n_echos, n_vols)
         else:
             comptable = pd.read_table(ctab)
-            # Change rationale value of rows with classification "accepted" to empty strings
-            comptable.loc[comptable.classification == "accepted", "rationale"] = ""
+            # Change rationale value of rows with NaN to empty strings
+            comptable.loc[comptable.rationale.isna(), "rationale"] = ""
 
             if manacc is not None:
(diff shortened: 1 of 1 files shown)


tedana/workflows/tedana.py
```diff
@@ -738,7 +738,12 @@ def tedana_workflow(
         else:
             comptable = pd.read_table(ctab)
 
-            if manacc is not None:
+            # If manacc is None and if comptable has a column named original_classification
+            if manacc is None and "original_classification" in comptable.columns:
+                # Get index of rows that have accepted on their classification column in comptable
+                manacc = comptable.index[comptable["classification"] == "accepted"].tolist()
+                comptable, metric_metadata = selection.manual_selection(comptable, acc=manacc)
```
tsalo, line 745: I don't think we can assume that an updated component table reflects manual classification. For example, if we get AROMA working, we could run the same data through both tedana and AROMA to produce merged classifications derived from both workflows.

eurunuela: Okay, would you be happy if we updated the rationale column instead?

tsalo: Hopefully once we have the decision tree refactor done and have switched from rationales to the messages we plan to use, we can just require classification tools to update that information. In the mean time, I think you empty string solution is great. On RICA's side, could you update the rationale column as well?

eurunuela: I agree things will be different soon, but users need a fix now.

Re Rica: I could make Rica update the rationale to say "manual" or something similar.

tsalo: Yes, if Rica could updated the rationale for any component with an updated classification, that would be optimal. "Manual" or "I001" would both work well.

tsalo (approved):
LGTM!

tedana/workflows/tedana.py
eurunuela: [PR #42 in Rica](https://github.com/ME-ICA/rica/pull/42) adds this functionality.

handwerkerd (approved):
Some of this might get superseded by the changes with the decision tree modularization, but this is a good change for the current code.

eurunuela:
> Some of this might get superseded by the changes with the decision tree modularization, but this is a good change for the current code.

Thank you Dan! Just wanna make sure users using the current version don't run into issues :)

PR #859, opened 2022-03-17
[FIX] Add function to prep data for JSON serialization

### LINKED ISSUE #858: Object of type ndarray is not JSON serializable
### Summary

This issue is the result of a failing test in #839 as reported by @handwerkerd: https://github.com/ME-ICA/tedana/pull/839#issuecomment-1069523729

It looks like our `save_json` function is having trouble saving a numpy array to the json dictionary.

### Next Steps

As mentioned by @tsalo, we might need to automatically convert numpy arrays to lists in `save_json`.

jbteves:
Closes #858  .

- adds prep_data_for_json to inspect a dictionary for ndarrays and change them into lists for saving data to JSON
-

### DIFF
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -707,7 +707,7 @@ def prep_data_for_json(d) -> dict:
     To speed things up since there are a small number of conversions, this
     function does not check for serializability, but does use conversion
     rules for cases encountered where things were not serializable.
-    Add more conversion rules to this function in cases where a 
+    Add more conversion rules to this function in cases where a
     tedana-generated object does not serialize to JSON.
     """
     if not isinstance(d, dict):
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -8,6 +8,7 @@ import json
 import logging
 import os
 import os.path as op
+from copy import deepcopy
 from string import Formatter
 
 import nibabel as nib
@@ -198,7 +199,8 @@ class OutputGenerator:
         if description.endswith("img"):
             self.save_img(data, name)
         elif description.endswith("json"):
-            self.save_json(data, name)
+            prepped = prep_data_for_json(data)
+            self.save_json(prepped, name)
         elif description.endswith("tsv"):
             self.save_tsv(data, name)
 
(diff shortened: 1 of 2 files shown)


eurunuela (dismissed):
LGTM! Thanks @jbteves

eurunuela (approved):

tedana/io.py
```diff
@@ -680,3 +682,43 @@ def split_ts(data, mmix, mask, comptable):
...
+    ValueError if it cannot force the dictionary to be serializable
+    TypeError if you do not supply a dict
+
+    Returns
+    -------
+    An attempt at JSON serializable data
+
+    Notes
+    -----
+    Use this to make something seriazable when writing JSON to disk.
```
handwerkerd, line 706:
```suggestion
    Currently converts numpy arrays to lists and raises an error if there's
    another data structure that can be automatically written to JSON.
    Use this to make something serializable when writing JSON to disk.
```

Really minor suggested change to make clear this is really only for correcting one thing right now (accept or reject this suggestion). Otherwise LGTM.

handwerkerd (changes requested):

tedana/io.py
tsalo: I like this suggestion, but I think it should say "another data structure that _cannot_ be automatically written to JSON."

jbteves: I think it's distinctly possible that more could come, and I'd rather not have to change the docstring and lead people to believe they should expand the docstring rather than just chucking another type into the if-block below.

handwerkerd: Like I said. I'm fine other way. The plus is, if I look at in a year, the note will help me remember why this function was needed. The minus is, as you note, it might over-promise future additions or over-specify. Resolve this review with or without making the change and I'll then approve.

jbteves: Latest commit modifies the language to make it clear that this function should be used to add more conversion cases without specifying them, that feels pretty split down the middle.

handwerkerd (approved):

handwerkerd:
@eurunuela You previously approved this PR, but there were a few changes and it needs a re-approval. This is what's holding up getting the PCA curves accessible in tedana. If you get a chance, can you check this out and re-approve. If you're in transit, maybe @tsalo can look at this? Thank you.

eurunuela:
Will try to have a look before Friday.

eurunuela (approved):
LGTM! Thank you @jbteves !

PR #871, opened 2022-04-14
[MAINT] Adds duecredit as requirement

### LINKED ISSUE #196: DueCredit not imported
When using a miniconda3 environment or pip install, the following import error occurs:
```
Failed to import duecredit due to No module named 'duecredit'
```
Unless we are avoiding acknowledging the devs and their hard work, we should probably fix that :-)

jbteves:
Closes #196 .

- Adds duecredit as a dependency
-

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -73,18 +73,6 @@ the v3.2 code, with the goal of revisiting it when ``tedana`` is more stable.
 Anyone interested in using v3.2 may compile and install an earlier release (<=0.0.4) of ``tedana``.
 
 
-*************************************************
-[tedana] What is the warning about ``duecredit``?
-*************************************************
-
-``duecredit`` is a python package that is used, but not required by ``tedana``.
-These warnings do not affect any of the processing within the ``tedana``.
-To avoid this warning, you can install ``duecredit`` with ``pip install duecredit``.
-For more information about ``duecredit`` and concerns about
-the citation and visibility of software or methods, visit the `duecredit`_ GitHub repository.
-
-.. _duecredit: https://github.com/duecredit/duecredit
-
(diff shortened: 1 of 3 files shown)


tsalo:
If duecredit is a requirement, I believe we can also remove `due.py`, since it is only there to deal with duecredit import issues.

tsalo:
Also, you can remove the `duecredit` requirement group.

jbteves:
The way we have this set up, removing `due.py` actually creates a circular dependency at install... We import from `info.py` to get our requirements, but `info.py` would then require `duecredit`. We could only resolve this by moving the citations elsewhere in the package.

tsalo:
Ah, I see. I think that's a compelling reason for us to switch from `info.py` to `setup.cfg`, but that can be addressed in a separate PR.

tsalo:
In that case, the only other things that need to be done in this PR are:
1. Remove the duecredit dependency group from `info.py`.
2. Potentially add `duecredit` to the list of dependencies in the README. Personally, I don't like duplicating that information outside of the actual dependencies file (i.e., `info.py` or `setup.cfg`), so I won't push for this.
3. Remove the FAQ about the duecredit warning from `docs/faq.rst`.
4. Remove the line about duecredit being an optional dependency from `docs/installation.rst`.

jbteves:
For (1), I'd rather not move that from `info.py` and then lump it in with other changes to the config. Happy to skip adding in (2), and then will push shortly for (3) and (4).

tsalo (approved):

tsalo:
Since we converged on dropping duecredit entirely, do you want to repurpose this PR or just close it?

jbteves:
We can just close this.

PR #874, opened 2022-04-28
[MAINT] Switch to setup.cfg-based configuration

### LINKED ISSUE #872: [MAINT] Change setup configuration
### Summary
The setup for python packages has changed over the last few years. There is now the option of a `pyproject.toml` (which tools like `poetry` use), a `setup.cfg` (which is what pip is designed to use), and there is now a circular dependency from `info.py` if we leave `duecredit` there but wish to remove `due.py`. We should refine the setup configuration.

### Next Steps
- [ ] Pick which items belong in `setup.cfg`
- [ ] Potentially: switch to pyproject.toml

tsalo:
Closes #872.

- Delete `tedana/info.py` and move its contents into `setup.cfg`.


tedana/__init__.py
```diff
@@ -54,3 +45,5 @@
     path="tedana",
     cite_module=True,
 )
+
+del get_versions
```
jbteves, line 49: Why is this needed?

tsalo: It's not needed, but I think it's nice to have a clean set of locals for the package. I can remove it though.

jbteves: I see. We can leave it, just wanted to make sure I understood why it was there.

jbteves (approved):

eurunuela (approved):
LGTM!

PR #877, opened 2022-05-12
[ENH, FIX] PCA variance enhancements and consistency improvements

### LINKED ISSUE #876: understanding variance and normalized variance explained
### Summary 

I went down a bit of a rabbit hole on the calculations for variance explained and normalized variance explained. The core observation is that the calculations underlying these values vary slightly depending which method is used. I’m not sure if anything is wrong, but we can definitely improve clarity in the code and/or docs. One thing I'm fairly sure about is that that variance explained by the PCA does not seem to be consistently logged across options. I wanted to highlight a few things I noticed. 

### Additional Detail

This started because I was playing with setting `--tedpca` using the fixed percent variance explained option. The variance I inputted was not the variance explained that was logged in any of the outputs.

Here's the code were the PCA and variance explained is calculated for a fixed variance explained:
https://github.com/ME-ICA/tedana/blob/50f21abef0b5cc01bf604c132e0be275554846ed/tedana/decomposition/pca.py#L242-L248
What’s interesting is that we’re using varex = ppca.explained_variance_ to set the variance explained and then normalizing that for varex_norm. Thus varex_norm.sum()==1 while varex.sum() is a non-standardized large number. The PERCENT of variance explained by each component is actually in `ppca.explained_variance_ratio` and when I sum those values up I do get my pre-set variance explained level.  The same varex_norm calculation is used when we're not reducing the number of PCA components
https://github.com/ME-ICA/tedana/blob/50f21abef0b5cc01bf604c132e0be275554846ed/tedana/decomposition/pca.py#L258

When MAPCA is used, varex_norm is calculated with explained_variance_ratio:
https://github.com/ME-ICA/mapca/blob/90ca567230dce39681ed5b826b438dd8391dbd8e/mapca/mapca.py#L512

We should probably line things up so that all options use `ppca.explained_variance_ratio` for `varex_norm`

The second part of my confusion is that the logger output lines that include `io.denoise_ts   INFO        Variance explained by decomposition:` show values that are  are much bigger than the variance explained by the PCA. For example, I set variance explained to 0.88 and the logger is showing variance explained values around 96%. I'm not sure if these numbers are just describing different thing or there's an underlying issue.

I'm working on some other sections, but wanted to document this in case other people (@tsalo @eurunuela ) already understand this aspect of the code better than I do.

handwerkerd:
Closes #876.

- We changed AIC to the default PCA selection criterion in #849 for the API. I noticed it was still defaulting to MDL for the command line interface, so I fixed that
- As discussed in #876, normalized variance explained was calculated one way for mdl, kic, & aic, and a different way for the low_mem option or if a specific variance explained as selected. Everything now matches the calculation used for mdl, kic, & aic
- The sum of the normalized variances across PCA components is the % variance of the original data that's explained by the PCA according to the PCA function in sklearn. I am now including this value in the info logger. It doesn't match the percent variance explained values that are given as later outputs. Figuring out how to make sense of the difference is a good future goal.
- The PCA function from sklearn has an option to input a pre-specified number of components as an integer. We were unnecessarily blocking off that option (which I wanted to use for testing some stuff). I updated `check_tedpca_value` to allow integers >=1 and added unit tests for valid and invalid integers.
- doc strings and logging messages were edited in response to the above changes.

Notes: 
- Just to this is cleanly documented, the current version of the code calculated `varex_norm = varex / varex.sum()` for the pre-defined variance-explained option and also for the low-mem option. It uses `varex_norm = ppca.explained_variance_ratio_` for mdl, kic, & aic.
- The testing coverage drops slightly because only integration testing covers `decomposition/pca.py` and `tedana.py` We can add another integration test for when a fixed integer number of components are predefined, but since that would non-trivially increase the duration of the tests, I didn't want to do that without some discussion. Adding direct tests for `pca.py` and modularizing `tedana.py` more would be a better approach, but well beyond the scope of this PR.

### DIFF
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -99,7 +99,7 @@ def tedpca(
         (see Li et al., 2007).
         If a float is provided, then it is assumed to represent percentage of variance
         explained (0-1) to retain from PCA.
-        If an int is provide, then it is assumed to be the number of components
+        If an int is provided, then it is assumed to be the number of components
         to select
         Default is 'aic'.
     kdaw : :obj:`float`, optional
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -34,6 +34,8 @@ def low_mem_pca(data):
         Component weight map for each component.
     s : (C,) array_like
         Variance explained for each component.
+    varex_norm : array-like, shape (n_components,)
+        Explained variance ratio.
     v : (C x T) array_like
         Component timeseries.
     """
@@ -44,7 +46,8 @@ def low_mem_pca(data):
     v = ppca.components_.T
     s = ppca.explained_variance_
     u = np.dot(np.dot(data, v), np.diag(1.0 / s))
-    return u, s, v
+    varex_norm = ppca.explained_variance_ratio_
+    return u, s, varex_norm, v
 
 
 def tedpca(
@@ -96,6 +99,8 @@ def tedpca(
         (see Li et al., 2007).
         If a float is provided, then it is assumed to represent percentage of variance
         explained (0-1) to retain from PCA.
+        If an int is provide, then it is assumed to be the number of components
+        to select
         Default is 'aic'.
     kdaw : :obj:`float`, optional
         Dimensionality augmentation weight for Kappa calculations. Must be a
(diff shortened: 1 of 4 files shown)


eurunuela (approved):
LGTM! Thank you @handwerkerd !

PR #880, opened 2022-07-06
[FIX] Fix beta calculation in OLS functions

tsalo:
Closes # .

-
-


tsalo:
I thought this would be easier to update than it is. I'm just going to close it. 😞

PR #896, opened 2022-11-10
Add recommendations for distortion correction

### LINKED ISSUE #894: Add info on distortion across echoes to documentation
### Summary
This stems from [this NeuroStars post](https://neurostars.org/t/multi-echo-pepolar-fieldmaps-bids-spec-sdcflows-grayzone/23933) by @bpinsard.

Here are some relevant bits from the thread:

@dowdlelt:

>Signal drops will absolutely increase over TE, and in areas like the orbitofrontal cortex, this will, in some visual sense, follow the patter of distortion but it isn’t a manifestation of increased distortion over echo time.
>
>There can be effects over echo time, particularly in ultra high resolution MRI (how topical @\ofgulban), but I think for most use cases in fMRI it isn’t worth the worry.

@handwerkerd:

>Agree with @dowdlelt The two of us have talked about this before and qualitatively looked into it a while bad. Dropout will increase with later echoes, but distortion should be fairly stable. If you want to calculate distortion using AP/PA then do it on the shortest echo with the least drop out. That said, if your first echo is really show and has lousy gray-white contrast, it might not have enough info to guide alignment (probably not an issue for currently typical multi-echo sequences, but might be a problem if people start using really short echoes)
>
>I’m not currently using the SBRef images, but that’s a good idea that I might want to try.
>
>On a more general level you really want to apply the same spatial transforms to all echoes since multi-echo combination methods assume that the same underlying tissue is in each voxel for each echo. I’d say any small benefits from calculating separate transforms by echoes would be out-weighed by problems caused from disconnecting the echoes.
>
>I’m not an an fmriprep user, but AFNI is already configured to process multi-echo with AP/PA distortion correction. Even if you don’t plan to use afni, you can set up a processing script using [afni_proc.py](https://afni.nimh.nih.gov/pub/dist/doc/program_help/afni_proc.py.html) with the -blip_forward_dset -blip_reverse_dset and maybe -volreg_align_to MEDIAN_BLIP to generate a processing script to see how they’ve combined these steps.

tsalo:
Closes #894.

- Add a FAQ on distortion correction.
- Add a subsection to the multi-echo processing section on distortion correction.


docs/faq.rst
```diff
@@ -123,3 +123,25 @@ Yes, these techniques are complementary.
...
+
+What this means is that, in the vast majority of multi-echo datasets,
+standard distortion correction methods will work, and should be applied in the same manner on all echoes.
+For example, if you acquire a blip-up/blip-down set of images for all of your echo times,
+you should use the first echo time's images to generate the undistortion transform,
+as long as that first echo has sufficient gray/white constrast to be useful for alignment
+(in which case, use the earliest echo that does have good contrast).
+
+One edge case that will become more common over time is ultra-high resolution multi-echo fMRI.
+In ultra-high resolution data (voxels <1 mm3), minute differences in distortion can become meaningful.
```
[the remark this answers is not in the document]
dowdlelt: Honestly I'm not even sure it bears mentioning. I think even with ultra-high res fMRI the B0 distortions would be the same, there may just be other effects kind of look like distortion that vary from echo to echo. But maybe I am wrong. I think <1mm is fine. It'll be a while before folks are regularly hitting that without a lot of effort (for multiecho).

dowdlelt (approved):

docs/faq.rst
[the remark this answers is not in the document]
tsalo: Oh, okay. I think I misunderstood what you were saying in the thread. Do you know what these other effects might be?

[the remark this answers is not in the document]
tsalo: I'll probably make the text a little vaguer anyway.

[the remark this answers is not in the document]
dowdlelt: https://www.biorxiv.org/content/10.1101/2021.11.25.470023v3.full#F10 is an example of it - basically, over echoes, there is a blood flow artifact that moves around (as in the blood signal shifts over TE), but usually we wouldn't see this i typical fMRI for various reasons. I just didn't want to go on the record saying that there was no effect when I knew there was at least a possibility of a small effect. I've looked for it in a few different datasets and never seen it, but usually blood isn't super visible anyway, so all these words just to say, mostly that

I don't think people need to worry about it. Maybe even ever. But as my MRI knowledge is incomplete, I thought it worth hedging a bit, in case someone digs up some effect sometime. With that comment in there, at least people could in theory be on the look out for edge cases, which is good.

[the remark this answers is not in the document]
tsalo: Ah, that's good to know. I've removed the mentions of ultra-high resolution data for now. We can add more detailed info in the future.

docs/faq.rst
```diff
@@ -123,3 +123,22 @@ Yes, these techniques are complementary.
...
 fMRI is instead related to collecting multiple unique volumes.
 These techniques can be combined to reduce the TR in a multi-echo sequence.
+
+********************************************************************************
+[ME-fMRI] How do field maps and distortion correction work with multi-echo fMRI?
+********************************************************************************
+
+There are many different approaches to susceptibility distortion correction out there- far too many to discuss here.
+However, the good news is that distortion does not typically change across echoes in multi-echo fMRI.
+In most cases, the ``EffectiveEchoSpacing`` is the same across echoes,
```
handwerkerd, line 133:
```suggestion
In most cases, the readout acquisition type and total readout time are the same for each echo,
```
"Echo spacing" is really confusing in this context because the space between each echo in a multi-echo sequence is also called the echo spacing. The real thing is that the readout is identical for all echoes (not just one parameter in the readout).

docs/faq.rst
```diff
@@ -123,3 +123,22 @@ Yes, these techniques are complementary.
...
+so distortion will remain relatively constant, even as dropout increases with echo time.
+
+What this means is that, in the vast majority of multi-echo datasets,
+standard distortion correction methods will work, and should be applied in the same manner on all echoes.
+For example, if you acquire a blip-up/blip-down set of images for all of your echo times,
+you should use the first echo time's images to generate the undistortion transform,
+as long as that first echo has sufficient gray/white constrast to be useful for alignment
+(in which case, use the earliest echo that does have good contrast).
+
```
handwerkerd, line 142:
```suggestion
One potential issue is that the nonlinear spatial transforms that are part of distortion correction methods have the potential to skew the relationship between echoes. Specific examples of nonlinear alignments causing problems have not yet been observed.
```

docs/faq.rst
```diff
@@ -123,3 +123,22 @@ Yes, these techniques are complementary.
...
+For context, please see
+`this NeuroStars thread <https://neurostars.org/t/multi-echo-pepolar-fieldmaps-bids-spec-sdcflows-grayzone/23933/5>`_.
```
handwerkerd, line 144: I like the idea of starting to link to NeuroStars from our docs so that people know to go there with questions and we don't have to reword entire discussions with examples in the docs.

docs/multi-echo.rst
```diff
@@ -462,6 +462,19 @@ time point.
     research to better understand how these steps can be applied before denoising.
 
 
+4. Apply susceptibility distortion correction consistently across echoes
+========================================================================
+
+One key feature of susceptibility distortion is that it is primarily a factor of echo spacing, rather than echo time.
```
handwerkerd, line 468:
```suggestion
One key feature of susceptibility distortion is that it is primarily a factor of readout pattern and total readout time, rather than echo time.
```

docs/multi-echo.rst
```diff
@@ -462,6 +462,19 @@ time point.
...
+This means that, for most multi-echo sequences, even though dropout will increase with echo time,
+distortion will not (at least not to a noticeable/meaningful extent).
+
+For this reason, if you are applying TOPUP-style (blip-up/blip-down) "field maps",
+we recommend using your first echo time, as this will exhibit the least dropout.
+If your first echo time is very short, and exhibits poor gray/white contrast, then a later echo time may be preferable.
+In any case, you should calculate the spatial transform from just one of your echoes and apply it across all of them.
+
```
handwerkerd, line 476: Lots of repetition in these two files. Not sure if we should add the warning about nonlinear transforms here as well.

handwerkerd (changes requested):

docs/faq.rst
tsalo: Does this apply when the same transform is applied to all echoes? It seems like this is the problem we're circumventing via our recommendation.

handwerkerd: I saw this issue on a dataset a while back that had non-trivial intensity inhomogeneity from the surface to the center of the brain. Even though interpolation is calculated using the same voxels in all echoes, the interpolation affected each echo a bit differently and messed up the relationship between the echoes. I really don't know how systematic or serious this issue is, but I've generally been cautious on using aggressive non-linear interpolations before combining echoes.

jbteves:
@tsalo since we have the 3.10 checks now, I made it mandatory, looks like your PR accidentally got caught. If you merge in main it should run and pass.

tsalo:
Thanks @jbteves!

docs/faq.rst
dowdlelt: I feel like this has to be rare, though that observation does worry me. The main issue I see with stating something like it is preferable to avoid non-linear alignment is that this is in direct contrast with the default settings in AFNI, for example. 

That said, I think the suggestion is good, because it will keep people on their toes. It does make me think it might be an issue that needs to be raised with AFNI in regards to what data goes into tedana (though its possible I missed some of the flexibility there).

docs/multi-echo.rst
dowdlelt: I think this is ok, given that the nonlinear issues are relatively rare (I think and also hope). It also seems that the fix for this would not be on the users to generate special workflows, but instead on AFNI/fmriprep to run or save things at specific stages.

dowdlelt (approved):
I think it is good to go - there is enough detail there, even if it will require some (slight) modification in the future as more edge cases are discovered. I think it would be worth including Dan's suggestion about potential nonlinear issues, so that we have a warning there that we can build on in the future.

handwerkerd (approved):

dowdlelt:
I maintain my approval here - but I will note that section 3 (which was left unchanged, and encourages distortion correction and normilization to be applied after tedana) somewhat disagrees with the new section 4. I don't think it is a major issue worth holding this up, but it just reflects that this isn't super tested.

handwerkerd:
I'm going to merge this and we can tweak this text a bit more as part of #948

handwerkerd:
New text is here: https://github.com/ME-ICA/tedana/pull/948/commits/290a7a853c483b7bcb4afc6ebfa03baab85a8256

I removed the recommendation to do distortion correction & normalization after tedana and I added a bit about normalization to the text on distortion correction. I also switched 3 & 4 & so we talk about what to do before tedana before talking about what goes after. The note was adjusted to justify why we do recommendation normalization & distortion correction first, but is retained as a warning to keep an eye on potential issues.

Issue #860, opened 2022-03-17
Make kundu tedpca selection results more accessible 


dowdlelt:
### Summary
The # of components selected by `--tedpca kundu` is currently invisible to the user if it is not selected. 

### Additional Detail
#834 adds curves for MAPCA, which allows users to see how unselected options perform without running the full tedana pipeline. It would be nice if the number of component selected by the kundu option were shown to the user and even better if these details were shown in relation to the MAPCA curves. 

### Next Steps

- [ ] minimally run the kundu section criteria even when not chosen. 
- [ ] Add details about number of components, perhaps as an additional line on the soon to exist MAPCA curves
- [ ] Potentially include the locations of the components chosen on the variance explained plots (as they could be anywhere)

PR #909, opened 2022-11-25
Exp/3 pca seq

pablosmig:
Experimenting with issue #860.
Non functional changes and breaks many tests.


eurunuela:
Thank you @pablosmig! It's been fun to hack with you at Brainhack Donostia.

I'll finish the PR and share with the core developers team.

eurunuela:
@tsalo @jbteves @handwerkerd where do you think would be a good place to save the PCA variance explained plots that have the Kundu options incorporated?

handwerkerd:
> 

I guess the core issue is that the kundu method isn't finding a local minimum so it's not actually a plot and there isn't even a threshold. The number of components retained would be an easy addition to `desc-PCA_cross_component_metrics.json` The other bit of info from kundu is which components are retained. That could either be a list in the same file or a new column with accepted & rejected in `desc-PCA_metrics.tsv`

Any other information that's looking for a home?

dowdlelt:
Hopefully I'm not completely missing the point here, but this could be added to the existing plots - not as a threshold, but as maybe individual filled circles on the lines (or along the x-axis) indicating which components were selected. Such that the user can see that the three main methods would have picked threshold x, y, z and 'kundu' would have picked most of the components below that threshold and then a few beyond it.

handwerkerd:
The code has non-trivially changed since this PR was opened. This might be easier to add by working from scratch (and just using this as a reference), so I'm closing this PR. Issue #860 is still open to keep track of this suggested addition.

Issue #782, opened 2021-08-19
Numpy Deprecating `int` and `float` aliases


jbteves:
### Summary
As of Numpy 1.20.0, the aliases `np.int` and `np.float` are deprecated. We have an instance of this in the codebase which we should patch. The overwhelming majority are outside of our control as they originate from third-party packages such as `sklearn`, specifically in the `linear_model` sub-package.

### Additional Detail
Read more [here](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations).
Instance occurs [here](https://github.com/ME-ICA/tedana/blob/e8a43ebe1ac896a7c4cd69bc7ba473c28775ecc6/tedana/selection/_utils.py#L59)

### Next Steps
- [ ] Revise to specify precision of cast
- [ ] Consider suppressing third-party warnings

tsalo:
I like the idea of suppressing the warnings. I assume we'll need to bump our minimum versions for several packages once numpy removes these aliases?

jbteves:
I don't believe we will need to. I'm much more concerned about the versioning that's causing #781. For this one we should be able to filter out the specific warning from numpy about `np.int` and `np.float`.

tsalo:
I think we're in good shape at this point. Going to close this.

PR #924, opened 2023-01-08
Sorting varex for decision tree criterion I011

### LINKED ISSUE #923: Decision tree step incorrectly assumes a sorted component table
### Summary

There is a step in the decision tree where the first X remaining components should be used for something where the components are sorted by variance. MEICA sorted components by variance but tedana does not. That means tedana is incorrectly using a semi-random subset of components instead of the highest variance components.

### Additional Detail
These are the key lines
https://github.com/ME-ICA/tedana/blob/f00cb25152142611b8e289ab59c7b5b8ab6eaf08/tedana/selection/tedica.py#L388-L392

This will be fixed in #756 but we might want to also fix in the current code so that pre and post decision-tree- modularization results will perfectly match.

This is a bug, but it will not affect the denoised time series. The `new_varex_lower` threshold is used to decide if components are `accepted` or `ignored` With the current version, it's possible for more components than intended to end up `ignored` but they'll still be retained in the denoised time series.

### Next Steps

- Confirm this is a real bug
- I'm fairly sure MEICA sorted components with the highest variance being component 0 so the highest variance components should be retained (the comment in the code might be wrong). Look back at some MEICA outputs to confirm this correct
- Decide if we want to fix this in the current Main so that it will match the modularized output.

handwerkerd:
Closes #923

- Fixes bug in the decision tree where variance explained should have been sorted before calculating a percentile for a subset of components.
- The relevant decision tree step used code I011.
- With this change, the variance threshold will be higher and a few more components may end up as `accepted` rather than `ignored` This won't affect the denoised time series since either classification is retained in the denoised time series.

- Also changed an `np.int` to `np.integer` in `_utils.py` that was causing tests to fail in CI, but not locally.

### DIFF
--- a/tedana/selection/_utils.py
+++ b/tedana/selection/_utils.py
@@ -58,7 +58,7 @@ def getelbow_cons(arr, return_val=False):
         (arr[nk - 5 - ii - 1] > arr[nk - 5 - ii : nk].mean() + 2 * arr[nk - 5 - ii : nk].std())
         for ii in range(nk - 5)
     ]
-    ds = np.array(temp1[::-1], dtype=np.int)
+    ds = np.array(temp1[::-1], dtype=np.integer)
     dsum = []
     c_ = 0
     for d_ in ds:
--- a/tedana/selection/_utils.py
+++ b/tedana/selection/_utils.py
@@ -58,7 +58,7 @@ def getelbow_cons(arr, return_val=False):
         (arr[nk - 5 - ii - 1] > arr[nk - 5 - ii : nk].mean() + 2 * arr[nk - 5 - ii : nk].std())
         for ii in range(nk - 5)
     ]
-    ds = np.array(temp1[::-1], dtype=np.integer)
+    ds = np.array(temp1[::-1], dtype=np.int32)
     dsum = []
     c_ = 0
     for d_ in ds:
--- a/tedana/selection/tedica.py
+++ b/tedana/selection/tedica.py
@@ -386,10 +386,10 @@ def kundu_selection_v2(comptable, n_echos, n_vols):
         unclf = np.setdiff1d(unclf, midk)
 
         # Find components to ignore
-        # Ignore high variance explained, poor decision tree scored components
-        new_varex_lower = stats.scoreatpercentile(
-            comptable.loc[unclf[:num_acc_guess], "variance explained"], LOW_PERC
-        )
+        # Of the remaining components, ignore ones with higher variance even if their
+        # decision tree scores are poor
+        sorted_varex = np.flip(np.sort((comptable.loc[unclf, "variance explained"]).to_numpy()))
+        new_varex_lower = stats.scoreatpercentile(sorted_varex[:num_acc_guess], LOW_PERC)
         candart = unclf[comptable.loc[unclf, "d_table_score_scrub"] > num_acc_guess]
         ign_add0 = candart[comptable.loc[candart, "variance explained"] > new_varex_lower]
         ign_add0 = np.setdiff1d(ign_add0, midk)
--- a/tedana/selection/tedica.py
+++ b/tedana/selection/tedica.py
@@ -387,7 +387,7 @@ def kundu_selection_v2(comptable, n_echos, n_vols):
 
         # Find components to ignore
         # Of the remaining components, ignore ones with higher variance even if their
-        # decision tree scores are poor
+        # decision tree scores are poor, instead of rejecting them
         sorted_varex = np.flip(np.sort(comptable.loc[unclf, "variance explained"].to_numpy()))
         new_varex_lower = stats.scoreatpercentile(sorted_varex[:num_acc_guess], LOW_PERC)
         candart = unclf[comptable.loc[unclf, "d_table_score_scrub"] > num_acc_guess]
--- a/tedana/selection/tedica.py
+++ b/tedana/selection/tedica.py
@@ -388,7 +388,7 @@ def kundu_selection_v2(comptable, n_echos, n_vols):
         # Find components to ignore
         # Of the remaining components, ignore ones with higher variance even if their
         # decision tree scores are poor
-        sorted_varex = np.flip(np.sort((comptable.loc[unclf, "variance explained"]).to_numpy()))
+        sorted_varex = np.flip(np.sort(comptable.loc[unclf, "variance explained"].to_numpy()))
         new_varex_lower = stats.scoreatpercentile(sorted_varex[:num_acc_guess], LOW_PERC)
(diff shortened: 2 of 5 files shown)


tedana/selection/_utils.py
```diff
@@ -58,7 +58,7 @@ def getelbow_cons(arr, return_val=False):
         (arr[nk - 5 - ii - 1] > arr[nk - 5 - ii : nk].mean() + 2 * arr[nk - 5 - ii : nk].std())
         for ii in range(nk - 5)
     ]
-    ds = np.array(temp1[::-1], dtype=np.int)
+    ds = np.array(temp1[::-1], dtype=np.integer)
```
tsalo, line 61: I think `numpy.int` is equivalent to the built-in `int`, rather than `numpy.integer`.

tedana/selection/tedica.py
```diff
@@ -386,10 +386,10 @@ def kundu_selection_v2(comptable, n_echos, n_vols):
         unclf = np.setdiff1d(unclf, midk)
 
         # Find components to ignore
-        # Ignore high variance explained, poor decision tree scored components
-        new_varex_lower = stats.scoreatpercentile(
-            comptable.loc[unclf[:num_acc_guess], "variance explained"], LOW_PERC
-        )
+        # Of the remaining components, ignore ones with higher variance even if their
+        # decision tree scores are poor
+        sorted_varex = np.flip(np.sort((comptable.loc[unclf, "variance explained"]).to_numpy()))
```
tsalo, line 391:
```suggestion
        sorted_varex = np.flip(np.sort(comptable.loc[unclf, "variance explained"].to_numpy()))
```

This shouldn't impact behavior, but I think it's more readable to get the numpy array, sort it, then flip it, instead of sorting the pandas Series, getting the numpy array from that, sorting it, and then flipping it.

tedana/selection/tedica.py
```diff
@@ -386,10 +386,10 @@ def kundu_selection_v2(comptable, n_echos, n_vols):
         unclf = np.setdiff1d(unclf, midk)
 
         # Find components to ignore
-        # Ignore high variance explained, poor decision tree scored components
-        new_varex_lower = stats.scoreatpercentile(
-            comptable.loc[unclf[:num_acc_guess], "variance explained"], LOW_PERC
-        )
+        # Of the remaining components, ignore ones with higher variance even if their
+        # decision tree scores are poor
```
tsalo, line 390:
```suggestion
        # Of the remaining components, ignore ones with higher variance even if their
        # decision tree scores are poor, instead of rejecting them
```

Thanks for improving the comment. I think making it clear that they would otherwise be rejected will also help. However, I'm also confused. `candart` (the artifact candidate components) are never classified as rejected, so won't all of the un-ignored components from this step actually be accepted?

tsalo (commented):
The changes look good to me, although I'm a little confused about the logic of this classification step.

tedana/selection/_utils.py
handwerkerd: @jbteves and I have been dealing with this in several places so I'll prod him to reply. We had `np.int` in multiple places in the code. At some minor numpy version change, that stopped working so a bunch have been switched to `np.integer` Not sure why this one got triggered here & not earlier. For the use cases we have, I'm not sure if it matters which one we use as long as it doesn't crash with a future version change.

tedana/selection/tedica.py
handwerkerd: Thank you for opening a can of worms. :)
This is a bit confusing, so I decided to go back to the original MEICA to explain better & it turns out there's another divergence in the original port.
To first answer your question, `candart` is used in both versions and is simply a bad variable name. It should be `candign` since these are candidates to ignore.
Now the difference. In the MEICA code, all the steps we now label I009-I012 were done on the same subset of components.
https://github.com/ME-ICA/me-ica/blob/53191a7e8838788acf837fdf7cb3026efadf49ac/meica.libs/select_model.py#L344-L360
`ncl` (now `unclf`) is assigned before this section and isn't changed until the end of the section. In tedana Main, after each classification step, we see:
https://github.com/ME-ICA/tedana/blob/f00cb25152142611b8e289ab59c7b5b8ab6eaf08/tedana/selection/tedica.py#L375

In practice, what this means is that the components rejected with I009 or I010 are NOT candidates to be ignored in I011 and I012 while they can change from rejected (actually midk) to ignored in the original MEICA. This vaguely makes sense since the goal of these final steps is to retain a few things that might have otherwise been rejected. That said, I011 and I012 operate on all remaining unclassified components so they can and do change components from `accepted` to `ignored`. All four of these steps are classifying based on different thresholds for `variance explained` & `d_table_score_scrub` so it would take a bit more thinking to see how they all interact.

That all said, there is the question about what to do now. This is the version of `tedana` people have been running for years. Do we want to:
1. Add a real breaking change into Main and then try to match that version in DTM?
2. Keep Main as is and (possibly) add another decision tree into DTM that will match the original MEICA?

I spoke to @jbteves and I lean towards (2) since breaking Main just before the DTM merge might cause unnecessary headaches. I still think it's worth merging this PR since it's correcting something that clearly wrong - Treating an unsorted array as if it was sorted. The difference between MEICA and tedana Main is a difference between pipelines where neither is inherently right.

tedana/selection/_utils.py
tsalo: I think @jbteves opened #782 about this problem. The numpy documentation says that `numpy.int` is just an alias for `int`, though, so I think it makes more sense to replace `numpy.int` with `int` instead of `numpy.integer`. However, it's not a big deal, since I'm sure `numpy.integer` will work really well too.

tsalo: Actually, even though `np.int` _is_ closer to `int` than `np.integer`, whether that defaults to `int64` or `int32` depends on the system. It might be best for tedana to use `np.int32` across the board (not just in niimgs).

tsalo (approved):
Sorry for the delay. LGTM.

I'd still like to use `np.float32` over `np.integer` in the future, but it's not a big deal for this PR.

handwerkerd:
I just opened #929 to document the additional issue that came up in this discussion. I'll try to get @jbteves to approve this tomorrow. 

I'm also fine going through all the code and replacing all `np.int` and `np.integer` with `np.int32`. I think there are also some non-numpy `integer` datatypes floating around. We might want to see if any/all are also better as `np.int32`
 Since this headache keeps happening, maybe we need to just do it as a stand-alone PR.

handwerkerd:
@tsalo & @jbteves I made some small tweaks at Josh's request & then it was failing style checks. Those now seem to be fixed and CI is passing. If you're both ok with this, then approve.

tsalo (approved):
The changes look good to me. Thanks!

jbteves (approved):

PR #951, opened 2023-06-07
logging MAPCA subsampling

handwerkerd:
Once https://github.com/ME-ICA/mapca/pull/57 is merged into MAPCA, it will be possible for tedana to log the estimated subsampling value and several related parameters.

- Logs several values related to the IID subsampling estiamte that are related to `desc-PCACrossComponent_metrics.json`
- A few PCA-related documentation changes that were noticed when writing this PR

I decided to add a conditional statement to only log these new values if they exist. The alternative is, once the above PR is merged into MAPCA, we release a new version of MAPCA and then update the minimum MAPCA version allowed with tedana. I'd rather just merge for now and then we can delete the conditional in the same PR where the minimum version of MAPCA is increased.

### DIFF
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -103,11 +103,11 @@ def tedpca(
         to select
         Default is 'aic'.
     kdaw : :obj:`float`, optional
-        Dimensionality augmentation weight for Kappa calculations. Must be a
-        non-negative float, or -1 (a special value). Default is 10.
+        Dimensionality augmentation weight for Kappa calculations when `algorithm` is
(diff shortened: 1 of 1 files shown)


eurunuela:
LGTM!

dowdlelt (approved):
Nice step in the right direction, going to try and do more reviewing of the other, related PR soon - but this one LGTM.

handwerkerd:
@eurunuela If you're also ok with this, could you approve so that I can merge?

eurunuela (approved):
I'm happy with the changes. I was just waiting to see if Logan had anything to say.

Issue #888, opened 2022-09-15
Improve summaries and visualizations of results with modularized decision tree


handwerkerd:
Write code to automatically create a flow chart or other summary/visualization for a decision tree. This can also be used to make a run-specific visualization that shows how components were reclassified at each node of the decision tree.

### Summary

The decision tree modularization (#756) is getting close to done. The "Decision Tree" is the series of steps through which each ICA component is classified as accepted or rejected. A nice feature of this work is a lot of useful information about what happened during processing is logged in structured formats. This information can be visualized and summarized in ways that will help users better understand what is happening and better understand their own data. This can be done by someone who enjoys writing code for interactive visualizations and flow charts even if they don't know much about the internal workings of tedana.

#### Flow chart for decision tree

- #355 has an extended discussion on different ways to visualize the decision tree in flow charts. All of these were semi-manually visualized.
- [./tedana/resources/decision_trees](https://github.com/jbteves/tedana/tree/JT_DTM/tedana/resources/decision_trees) contains JSON files for each decision tree that is distributed with tedana. There are a series of elements that detail each step in the process and these can be used to generate a flow chart
- After tedana is run on a dataset, a file called `desc-ICA_decision_tree.json` is saved with the rest of the program's outputts. That file contains the same decision tree that was ran along with an `output` key for each step in the process. Information on what how many component classifications changed and what they changed to is stored there. Other values, such as calculations for various thresholds are also stored. It should be possible to generate a flow charts that highlights exactly which steps affected component classifications.
- Another potentially useful output is `desc-ICA_status_table.tsv` which has a row for each component and a column for each step (node) in the decision process. Each components classification after each node is stored there.

#### Interactive report visualizations

- When tedana is run, the output includes an interactive visualization of the results in `tedana_report.html` This currently includes the components with there kappa values, rho values, variance explained, final classifications, spatial maps, and time series. This information is displayed across several subplots.
- Currently accepted components are green and rejected components are red.
- We will now be saving classification tags for each component and each component could have more than one tag. It would be useful to be able to highlight subsets of components using these classification tags.
- Additional information, such as a mini summary of when components changed their classifications and thresholding metrics that were calculated across components could also be visualized or summarized in this report.

### Next Steps

- If you are interested in working on this, and haven't previously contributed to tedana, let @handwerkerd know so that he can provide additional guidance
[...]

PR #959, opened 2023-08-03
Document decision tree steps in report

tsalo:
Closes None.

- Start adding comments explaining each of the steps in the minimal decision tree. There are still some things I don't understand.
    - I'd like to add more layperson's descriptions in the `_comment` fields.


tedana/resources/decision_trees/minimal.json
```diff
@@ -132,7 +147,7 @@
                 "subset_decide_comps": "unclassified",
                 "rho_elbow_type": "liberal",
                 "log_extra_info": "",
-                "log_extra_report": ""
+                "log_extra_report": "The 'liberal' rho elbow is calculated from all (unclassified?) components, for later steps."
```
tsalo, line 150: I need to look at the rho elbow function to figure out how both unclassified and all components are used.

tedana/resources/decision_trees/minimal.json
```diff
@@ -194,11 +214,12 @@
             "kwargs": {
                 "var_metric": "variance explained",
                 "log_extra_info": "",
-                "log_extra_report": "",
+                "log_extra_report": "Any provisionally rejected components with variance explained less than ? are accepted.",
```
tsalo, line 217: @handwerkerd do you know how the parameters here can be described?

tedana/resources/decision_trees/minimal.json
tsalo: Honestly, the elbow logic doesn't make sense to me, but the end result seems to be that this step calculates the elbow as the maximum of two values: the elbow calculated from all components and elbow calculated from "unclassified" components at this point in the decision tree.

tsalo:
@handwerkerd do you know why the paths to the artifacts in the CircleCI config were wrong? Is it just that we changed the actual output folders in the DTM PR and forgot to update the artifact paths, or did we choose to stop uploading artifacts because of free plan limits or something similar?

tedana/resources/decision_trees/minimal.json
```diff
@@ -64,17 +68,24 @@
...
                 "metric_name": "variance explained",
                 "median_label": "varex"
-            }
+            },
+            "kwargs": {
+                "log_extra_info": "",
+                "log_extra_report": "The median variance explained is calculated across all components, for later steps."
+            },
+            "_comment": ""
         },
```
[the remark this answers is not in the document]
handwerkerd: Looks good. Note that many of the early nodes in minimal are also in kundu.json. They're identical until the elbow thresholding. If you're adding comments to one, it's probably good to add the same comment to the other (when it's the same thing)

tedana/resources/decision_trees/minimal.json
```diff
@@ -108,9 +121,11 @@
...
-                "log_extra_report": "",
+                "log_extra_info": "",
+                "log_extra_report": "Any components with negative t-statistic comparing distribution of T2* model F-statistics from voxels in clusters to those of voxels not in clusters and variance explained greater than median are rejected.",
                 "tag_if_true": "Unlikely BOLD"
-            }
+            },
+            "_comment": "This essentially rejects any high-variance components exhibiting more 'speckled' T2*-associated voxels than 'clustered' ones."
         },
         {
             "functionname": "calc_kappa_elbow",
```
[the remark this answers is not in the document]
handwerkerd: The first four rejection criteria for both trees are run on all components and can be in any order. You can see that a bit more clearly in the flow charts: https://tedana.readthedocs.io/en/stable/included_decision_trees.html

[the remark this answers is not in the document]
handwerkerd: The reason I didn't change this in either tree is that, with provenance tracking, it's good to see if a component is ever rejected by only one of these four. In my experience, most components are rejected by all 4 and we might be able to simplify in the future.

tedana/resources/decision_trees/minimal.json
handwerkerd: This is one place worth consideration. I really don't like the rho elbow calculation in the kundu tree which seems to skew low (i.e. more component are skewed towards reject). For the minimal tree, I used 2 of the 3 same calculations, but skewed it to be as high as possible. Since the minimal tree rejects nearly everything that is greater than the rho elbow threshold, I wanted the tree to be conservative and not reject edge cases.

I'll need to relook to confirm, but I think the elbow is calculated on all components and on just the remaining unclassified components. The used threshold is the greater value of the two.

tedana/resources/decision_trees/minimal.json
handwerkerd: I think this function gives an automatic description when run. The log_extra_report field is only useful if you want to give additional explanation.

The current automatic message will be something like
`accepted if variance explained<0.1. All variance<1.0, else nochange`

tsalo: Thanks for pointing me to the current message. That said, the `log_extra_report` field does show up in the HTML report, which is what I was hoping for. See https://output.circle-artifacts.com/output/job/500e9760-3014-49f4-a17e-a503ce9a1c32/artifacts/0/tmp/src/tedana/.testing_data_cache/outputs/five-echo/tedana_report.html for an example of the updated report text from this PR.

tedana/resources/decision_trees/minimal.json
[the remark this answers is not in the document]
tsalo: My initial thought was that we could put the cross-component metrics that are calculated across all components at the top of the decision tree. I thought that would be easier to read, but you're totally right that it doesn't impact anything. Having the metric calculations closer to the decision nodes where they're actually used is probably more readable than what I was thinking.

tsalo:
I didn't get through all of the Kundu decision tree steps, because some are quite complicated, so the report will be incomplete for that one.

handwerkerd:
> Is it just that we changed the actual output folders in the DTM PR and forgot to update the artifact paths

Looking at your change, I think that's what happened. I changed the location of the outputs for testing, but didn't realize I should have also made this change.

tsalo:
I think I've successfully documented the remaining nodes of the kundu tree.

tedana/resources/decision_trees/kundu.json
```diff
@@ -98,7 +109,9 @@
                 "left2": "variance explained",
                 "op2": ">",
                 "right2": "median_varex",
-                "tag_if_true": "Unlikely BOLD"
+                "tag_if_true": "Unlikely BOLD",
+                "log_extra_info": "",
+                "log_extra_report": "Any components with higher S0 model beta map-F-statistic map Dice similarity index than T2 model beta map-F-statistic map Dice similarity index and greater than median variance explained are rejected."
```
eurunuela, line 114: I don't understand the wording here: `beta map-F-statistic map`.

tedana/resources/decision_trees/kundu.json
```diff
@@ -116,7 +129,9 @@
                 "left2": "variance explained",
                 "op2": ">",
                 "right2": "median_varex",
-                "tag_if_true": "Unlikely BOLD"
+                "tag_if_true": "Unlikely BOLD",
+                "log_extra_info": "",
+                "log_extra_report": "Any components with negative t-statistic comparing distribution of T2* model F-statistics from voxels in clusters to those of voxels not in clusters and variance explained greater than median are rejected."
```
eurunuela, line 134: Also not easy to understand.

tedana/resources/decision_trees/kundu.json
```diff
@@ -325,8 +363,11 @@
                     "unclass_highvar"
                 ]
             },
-            "kwargs": {},
-            "_comment": "Add more here"
+            "kwargs": {
+                "log_extra_info": "",
+                "log_extra_report": "If any components are still labeled as unclassified or unclassified high-variance, then a revised decision table score is calculated from the provisionally accepted, unclassified, and unclassified high-variance components."
+            },
+            "_comment": "Calculate a new d_table_score. Why can't we use the original d_table_score function? Because it's a metric function?"
```
eurunuela, line 370: Is this a comment for us?

eurunuela (changes requested):
There are a couple of comments that I find impossible to understand. Could you please make them more understandable to the regular user/contributor? Thanks!

tedana/resources/decision_trees/kundu.json
tsalo: Yeah the `_comment`s are, I believe, for developers.

tedana/resources/decision_trees/kundu.json
tsalo: What about this?

```suggestion
                "log_extra_report": "Any components with (1) greater than median variance explained and (2) greater correspondence between the thresholded and binarized S0-model beta map and the thresholded and binarized F-statistic loading map than between the thresholded and binarized T2-model beta map and the thresholded and binarized F-statistic loading map (as quantified with the Dice similarity index) are rejected."
```

eurunuela:
@tsalo I've been told there's a typo in the docs:

`---manrej` should be *component indices to reject* instead of *to accept*.

Could you please fix it in this PR? Just asking because this is docs-related.

tsalo:
Done!

tedana/metrics/collect.py
```diff
@@ -61,7 +61,7 @@ def generate_metrics(
 
     if metrics is None:
         metrics = ["map weight"]
-    RepLGR.info("The following metrics were calculated: {}.".format(", ".join(metrics)))
+    RepLGR.info("\n\nThe following metrics were calculated: {}.".format(", ".join(metrics)))
```
handwerkerd, line 64: Probably not worth systematically going through all the code, but is it worth changing from `.format()` to f strings whenever we edit lines?

tsalo: I was planning to go hog-wild with style changes in #970 once the more substantive PRs were merged.

handwerkerd:
I have a general question on the goal here. We have several areas for outputs. `LGR.info` is stuff that is on the screen and output log. That log automatically generates a description of what happens in every node. My understanding is that`RepLGR` is supposed to rough paragraph of text that could potentially be included in a manuscript. Currently, that text says "This is based on the criteria of the MEICA v2.5 decision tree" and someone can look at several places to see what happens in that decision tree.  Do we want a line of text in the report explain what happens in every node?

If we're expecting people to copy the report into manuscripts (with light editing), then that's too much detail. In that case, I'd move more of your `log_extra_report` fields into `_comment`
If we're using the report as one additional location to help people understand what's happening, then I'm more ok leaving the text in `log_extra_report`.

For now, I'm going to read through and edit for accuracy and clarity, but we should decide whether this info is in the right location or should be moved to `_comment`

tedana/resources/decision_trees/kundu.json
handwerkerd: We have a lot of repeated metrics and it doesn't make sense to re-define them every time they are used. That said, I'm going through our docs and the only place we really define them is in the API ( https://tedana.readthedocs.io/en/stable/generated/tedana.metrics.dependence.html#module-tedana.metrics.dependence) and some of those definitions could use more clarity. 

I thought we had a descriptive list of all the metrics outputted in the component table, but maybe not? For now, maybe we can link to the metric descriptors in the API in a `_comment` for the whole json file (and maybe update particularly confusing descriptions int he API). That would free us to use more conceptual descriptions in the comments. For example:

```suggestion
                "log_extra_report": "For components with more than median variance, if the peak clusters for the S0-model overlap more with the peaks of the ICA component's weighting map than the peak clusters for the T2-model overlaps with the ICA weighting map, then reject.
```
(That might be better as a `_comment` that in the report)

tedana/resources/decision_trees/minimal.json
```diff
@@ -29,10 +29,12 @@
...
             "kwargs": {
-                "log_extra_report": "",
                 "clear_classification_tags": true,
-                "dont_warn_reclassify": true
-            }
+                "dont_warn_reclassify": true,
+                "log_extra_info": "",
+                "log_extra_report": "All components are initially labeled as 'unclassified'."
+            },
+            "_comment": "This step prepares the components to be classified by the decision tree."
```
handwerkerd, line 37: The info log will already output:
"Set all to unclassified"
"component classification tags are cleared"

Just to it's clear in the json file, maybe change `_comment` to "All components are set to unclassified to initialize the tree" and remove `log_extra_report`

tedana/resources/decision_trees/kundu.json
handwerkerd: I re-found the descriptive list of metrics. It's included in the outputs in the `desc-tedana.metrics.json` file. This makes a lot of sense since the specific metrics might vary depending on pipeline. That said, the descriptions could be improved. I could see explicitly referencing that file in this json (as part of this PR) and improving those descriptions either in this PR or in a separate PR.

handwerkerd:
@tsalo, I can open a PR and make multiple changes rather than having a bunch of separate requests, but I want to check if that's ok with you. My general changes would be to use the `_comments` field more and the `log_extra_report` field less. I'd link to the description of the metric names and make the comments more conceptual rather than trying to redefine each metric.

tsalo:
>Currently, that text says "This is based on the criteria of the MEICA v2.5 decision tree" and someone can look at several places to see what happens in that decision tree.

I don't think they can. Previous papers that I've read that describe the decision tree don't cover everything in sufficient detail. The main paper we cite for the v2.5 decision tree is [Kundu et al. (2013)](https://www.pnas.org/doi/abs/10.1073/pnas.1301725110), but that paper (and its supplement) seems to only describe Kappa- and Rho-based decisions. [Olafsson et al. (2015)](https://doi.org/10.1016/j.neuroimage.2015.02.052) describes each of the steps in the decision tree in some detail, but it's completely missing the additional constraint of greater-than-median variance explained in several of the nodes.

Plus, that only applies to the v2.5 decision tree. The minimal decision tree doesn't have a paper that describes each step AFAIK, nor would custom decision trees.

>Do we want a line of text in the report explain what happens in every node?

I think that is the best way to handle it.

>If we're expecting people to copy the report into manuscripts (with light editing), then that's too much detail.

I think all of this information is necessary for manuscripts, but I can also see people moving some contents into their supplementary materials. I've seen folks do this with fMRIPrep methods sections as well.

handwerkerd:
When I wrote "someone can look at several places to see what happens in that decision tree." I meant that there are several places in the existing outputs to understand what happens, but I agree these aren't detailed in any previous publication.

I'm going to disagree that all this information should be included in a manuscript. I've included the report output that would be generated by this PR for the kundu tree below. Not only is this long, but it's confusing and a poor way for anyone to understand what is happening in this process. We can edit various sentences, but I don't think anyone who reads any block of text like that would leave with a solid understanding of the process. If I reviewed a manuscript with a paragraph like that, I'd ask for revisions. [The flow chart](https://tedana.readthedocs.io/en/stable/included_decision_trees.html#kundu-decision-tree) is a much better way of sharing and understanding this process.

This is an example of the limits of manuscripts as a format and the original purpose for things like BIDS. If a manuscript includes details of every step to this level of detail, it would very long and unreadable. For example, there are many critical MRI acquisition parameters that are never reported in manuscripts, but I worked with the original BIDS group to make sure most of [this acquisition parameter list](https://www.authorea.com/doi/full/10.15200/winn.143191.17127) was in the BIDS jsons. I think we should take a similar approach here. We should have very clear output files that make it easy for people to know what happened and should expect those files to be shared with data.

Right now the relevant data is in several places. (1) [The flow chart](https://tedana.readthedocs.io/en/stable/included_decision_trees.html#kundu-decision-tree) (2) The info file `tedana_[date].tsv` as descriptors of each step (3) `desc-ICA_decision_tree.json` has additional descriptors.

Additional things we can do would be to generate a run-specific flow chart (#888) or just pull the `info` lines relating to the decision tree into a stand-alone file. I think much of the added text in this PR is useful for helping someone understand the process, but I just don't see it as useful text to include in a manuscript.

Here's what the report text currently looks like:

> TE-dependence analysis was performed on input data using the tedana workflow \citep{dupre2021te}. An initial mask was generated from the first echo using nilearn's compute_epi_mask function. An adaptive mask was then generated, in which each voxel's value reflects the number of echoes with 'good' data. A two-stage masking procedure was applied, in which a liberal mask (including voxels with good data in at least the first echo) was used for optimal combination, T2*/S0 estimation, and denoising, while a more conservative mask (restricted to voxels with good data in at least the first three echoes) was used for the component classification procedure. A monoexponential model was fit to the data at each voxel using log-linear regression in order to estimate T2* and S0 maps. For each voxel, the value from the adaptive mask was used to determine which echoes would be used to estimate T2* and S0. Multi-echo data were then optimally combined using the T2* combination method \citep{posse1999enhancement}. Global signal regression was applied to the multi-echo and optimally combined datasets.
> 
>  The following metrics were calculated: kappa, rho, countnoise, countsigFT2, countsigFS0, dice_FT2, dice_FS0, signal-noise_t, variance explained, normalized variance explained, d_table_score. Kappa (kappa) and Rho (rho) were calculated as measures of TE-dependence and TE-independence, respectively. A t-test was performed between the distributions of T2*-model F-statistics associated with clusters (i.e., signal) and non-cluster voxels (i.e., noise) to generate a t-statistic (metric signal-noise_z) and p-value (metric signal-noise_p) measuring relative association of the component to signal over noise. The number of significant voxels not from clusters was calculated for each component.
> 
>  Next, component selection was performed to identify BOLD (TE-dependent) and non-BOLD (TE-independent) components using a decision tree. This is based on the criteria of the MEICA v2.5 decision tree \citep{kundu2013integrated}. For a detailed description of the decision tree steps, with the rationale for each step, see \citet{olafsson2015enhanced}. All components are initially labeled as 'unclassified'. Any components with rho greater than kappa are rejected. Any components with more voxels that are significant based on the S0 model's F-statistics than the T2* model's are rejected, as long as there is at least one significant voxel for the T2 model. The median variance explained is calculated across all components, for later steps. Any components with higher S0 model beta map-F-statistic map Dice similarity index than T2 model beta map-F-statistic map Dice similarity index and greater than median variance explained are rejected. Any components with negative t-statistic comparing distribution of T2* model F-statistics from voxels in clusters to those of voxels not in clusters and variance explained greater than median are rejected. The kappa elbow is calculated from all components, for later steps. Unclassified components exhibiting a large step down in variance explained are classified as 'unclassified high-variance'. The 'kundu' rho elbow is calculated from all and unclassified components, for later steps. Any unclassified components with kappa greater than or equal to the kappa elbow are provisionally accepted. Any provisionally accepted components with rho greater than the rho elbow are reset to 'unclassified'. The variance explained upper threshold is calculated as the 90th percentile of variance explained from provisionally accepted components. The variance explained lower threshold is calculated as the 25th percentile of variance explained from provisionally accepted components. The 'extend factor' is determined based on the number of volumes in the run. The 'max_good_meanmetricrank' is calculated as the number of provisionally accepted components times the extend factor. The variance explained-kappa ratio is calculated from the provisionally accepted components, as the maximum kappa minus the minimum kappa, divided by the maximum variance explained minus the minimum variance explained. Any provisionally accepted, unclassified, or unclassified high-variance components with a decision table score greater than 'max_good_meanmetricrank' and variance explained greater than the variance explained upper threshold multiplied by the extend factor are rejected. Any provisionally accepted, unclassified, or unclassified high-variance components with a decision table score greater than 'max_good_meanmetricrank', variance explained less than or equal to the variance explained lower threshold, and kappa less than or equal to the kappa elbow will be accepted and labeled as 'low variance'. If no components are still labeled as unclassified or unclassified high-variance, then all remaining provisionally accepted components are accepted. If any components are still labeled as unclassified or unclassified high-variance, then a revised decision table score is calculated from the provisionally accepted, unclassified, and unclassified high-variance components. Any provisionally accepted, unclassified, or unclassified high-variance components with a revised decision tree score greater than the 'conservative_guess', variance explained-kappa ratio greater than the extend factor times two, and variance explained greater than the variance explained upper threshold times the extend factor are rejected. Any provisionally accepted, unclassified, or unclassified high-variance components with a revised decision table score greater than 'num_acc_guess' times 0.9 and variance explained greater than variance explained lower threshold times the extend factor are rejected. An updated variance explained lower threshold (25th percentile) is calculated from the 'num_acc_guess' highest variance explained components among the remaining provisionally accepted, unclassified, and unclassified high-variance components. Any provisionally accepted, unclassified, or unclassified high-variance components with a revised decision table score greater than 'num_acc_guess' and variance explained greater than the new variance explained lower threshold are accepted and labeled as 'borderline'. Any provisionally accepted, unclassified, or unclassified high-variance components with kappa less than or equal to the kappa elbow and variance explained greater than the new variance explained lower threshold are accepted and labeled as 'borderline'. All remaining unclassified, unclassified high-variance, or provisionally accepted components are accepted. Minimum image regression was then applied to the data in order to remove spatially diffuse noise \citep{kundu2013integrated}.
> 
>  This workflow used numpy \citep{van2011numpy}, scipy \citep{virtanen2020scipy}, pandas \citep{mckinney2010data,reback2020pandas}, scikit-learn \citep{pedregosa2011scikit}, nilearn, bokeh \citep{bokehmanual}, matplotlib \citep{Hunter:2007}, and nibabel \citep{brett_matthew_2019_3233118}. This workflow also used the Dice similarity index \citep{dice1945measures,sorensen1948method}.

eurunuela:
I have never used FigShare so I have no opinion on this.

handwerkerd:
@tsalo Could you try to merge Main into this PR? I tried to do it on my branch and I'm running into problems.  If you get it working, that might make it easier for me to figure out what's going wrong on my end. See: https://github.com/ME-ICA/tedana/pull/970#issuecomment-1783411010

tsalo:
@handwerkerd I updated the branch yesterday, but forgot to comment to that effect until now. Sorry about that!

handwerkerd:
Planned changes from past developers calls:
Work in progress: https://github.com/handwerkerd/tedana/tree/doc-tree-dh
- [ ] Put more text in `_comments` rather than making a hyper-detailed report
- [ ] The report log will link to a DOI where each version of the tree is detailed as a json file and a flow chart
- [ ] Use figshare for the DOI since it allows different DOIs for multiple versions of an object
- [ ] Remove log_extra_report fields from the json files but save removing the code that interacts with that field for a separate PR

handwerkerd:
Not sure if anyone by @tsalo can see this yet, but I uploaded the decision trees to figureshare in a new `tedana` project and created: 10.6084/m9.figshare.25251433 The updated data are at: https://figshare.com/account/articles/25251433

This is still private because I think there are limits to what I can change once I publish, but I can probably give individuals access to let me know if it includes what you think it should include.

tsalo:
The current approach of PRs to my branch are confusing me, so I've changed the target branch to a new one under ME-ICA (`doc-tree`). @handwerkerd you can merge into the new branch if you'd like and just modify it directly.

handwerkerd:
This is either my git skills or a permissions issue, but I'm not sure how I can push directly to https://github.com/tsalo/tedana/tree/doc-tree. My current updates are now in https://github.com/handwerkerd/tedana/tree/doc-tree
I think this is very close to ready to review/merge. I've fully removed `log_extra_report` from the code. I needed to change https://osf.io/f6g45 (data for testing reclassify) to remove that key so that tests would pass, but the change was otherwise fairly easy.

I also fixed the five-echo test failure above, but there's a new error with five-echo and I'm having trouble solving. Five-echo is the one that uses the minimal decision tree with this line for reports:
https://github.com/handwerkerd/tedana/blob/09b9e2ef7f79262fad9ab7a32207012b51304ad6/tedana/resources/decision_trees/minimal.json#L4
For some reason, it's now unhappy with the comma separate citations `\\citep{kundu2013integrated, dupre2021te}` Not quite sure why this is happening now.

tsalo:
@handwerkerd I have "allow edits by maintainers" selected, so you should be able to push to my branch if you add it as a remote, but I was originally proposing that you merge this PR into the ME-ICA/tedana@doc-tree branch (the new target branch for this PR) and make your changes there.

PR #963, opened 2023-08-04
Add prefix to all output files

### LINKED ISSUE #901: Include prefix in report, log, and figures
### Summary
I've noticed that our current output structure doesn't work well if you want to run multiple tasks/runs from the same subject/session in the same output directory (i..e, a BIDS derivatives-like structure). If we ensure that the files _without_ prefixes have them, then  it should work fine.

### LINKED ISSUE #955: need multiple reports when using Auto_tedana with multiple runs
As recommended in the [FAQ](https://tedana.readthedocs.io/en/latest/faq.html) I'm using the Auto_tedana.py script after fmriprep. I have multiple runs, but the tedana_report.html is overwritten for each run leaving me with only the last run. It would be great if the report in each run is concatenated.

tsalo:
Closes #901 and closes #955.

- Add the user-provided prefix to other outputs, including the text report, the HTML report, and figures.


eurunuela:
> It looks like the figure filenames are hardcoded into the HTML templates, so the prefixes are tough to account for.

I could have a look at that.

tsalo:
@eurunuela that would be awesome! Thanks!

eurunuela:
Let me know what you think @tsalo.

tsalo:
It looks good to me. Thanks @eurunuela!

tedana/workflows/tedana.py
```diff
@@ -445,7 +445,7 @@ def tedana_workflow(
         os.mkdir(out_dir)
 
     # boilerplate
-    basename = "report"
+    basename = f"{prefix + '_' if (prefix != '' and not prefix.endswith('_')) else prefix}report"
```
handwerkerd, line 448: Should you also include the prefix for the logfile name on line 461?
That is
461: `basename = f"{prefix + '_' if (prefix != '' and not prefix.endswith('_')) else prefix}tedana_`

handwerkerd (changes requested):
In addition to the minor suggested change, the prefix isn't added to `references.bib` or the `figures` directory name.

I'd lean towards adding it for `references.bib` since each run of the program will overwrite the file and that stylistically not great. The contents should always be identical so I'm fine if you don't want to add it.

I'd lean against adding it to `figures` since all the files should be distinct in that directory, but figured I'd still flag it.

Otherwise LGTM

tedana/workflows/ica_reclassify.py
```diff
@@ -326,10 +326,11 @@ def ica_reclassify_workflow(
         raise ValueError("The following components were both accepted and rejected: " f"{in_both}")
 
     # boilerplate
-    basename = "report"
+    prefix = prefix + "_" if (prefix != "" and not prefix.endswith("_")) else prefix
+    basename = f"{prefix}report"
```
handwerkerd, line 330:
```suggestion
    prefix_full = prefix + "_" if (prefix != "" and not prefix.endswith("_")) else prefix
    basename = f"{prefix_full}report"
```
You're overwriting `prefix` and the original version is later called by the iogenerator. That's why tests are failing.

tedana/workflows/ica_reclassify.py
```diff
@@ -326,10 +326,11 @@ def ica_reclassify_workflow(
...
     extension = "txt"
     repname = op.join(out_dir, (basename + "." + extension))
-    bibtex_file = op.join(out_dir, "references.bib")
+    bibtex_file = op.join(out_dir, f"{prefix}references.bib")
```
handwerkerd, line 333:
```suggestion
    bibtex_file = op.join(out_dir, f"{prefix_full}references.bib")
```

tedana/workflows/tedana.py
```diff
@@ -445,10 +445,11 @@ def tedana_workflow(
         os.mkdir(out_dir)
 
     # boilerplate
-    basename = f"{prefix + '_' if (prefix != '' and not prefix.endswith('_')) else prefix}report"
+    prefix = prefix + "_" if (prefix != "" and not prefix.endswith("_")) else prefix
+    basename = f"{prefix}report"
```
handwerkerd, line 449:
```suggestion
    prefix_full = prefix + "_" if (prefix != "" and not prefix.endswith("_")) else prefix
    basename = f"{prefix_full}report"
```

tedana/workflows/tedana.py
```diff
@@ -445,10 +445,11 @@ def tedana_workflow(
...
     extension = "txt"
     repname = op.join(out_dir, (basename + "." + extension))
-    bibtex_file = op.join(out_dir, "references.bib")
+    bibtex_file = op.join(out_dir, f"{prefix}references.bib")
```
handwerkerd, line 452:
```suggestion
    bibtex_file = op.join(out_dir, f"{prefix_full}references.bib")
```

handwerkerd (changes requested):

tedana/workflows/ica_reclassify.py
handwerkerd: Then this could probably be fixed by changing
https://github.com/tsalo/tedana/blob/10e294e1e6259446cb8feed952509f6ff68bbb1c/tedana/io.py#L134
to
`self.prefix = prefix + "_" if (prefix != "" and not prefix.endswith("_")) else prefix`

handwerkerd: I'm not sure why, but your new function `_infer_prefix` is outputting `None` instead of `""` for an empty prefix. That means later functions are failing to combine `None+Str` and causing a bunch of failed tests.

tsalo: Thanks for catching that @handwerkerd. The issue is that I forgot to actually return the modified prefix, so it was returning None no matter what.

handwerkerd (approved):

handwerkerd:
After approving, I noticed one minor thing that may or may not be worth changing. The four-echo integration test uses a pre-defined ICA mixing matrix:
https://github.com/tsalo/tedana/blob/1b97294eaed4e7111aba6a807c4ece38472a7270/tedana/tests/test_integration.py#L304

That file `desc-ICA_mixing_static.tsv` is copied into the output directory without any name change or prefix addition and then the same content is also appropriately named in `sub-01_desc_ICA_mixing.tsv` 

I don't see this as a major issue for this PR. The one problem I could see it happening is if someone is using another package to run ICA and sending those matricies to tedana. In that case, it would be their responsibility to appropriately name their mixing matrix files.

That said, the supplied mixing matrix is copied without any checks on whether it's overwriting anything:
https://github.com/tsalo/tedana/blob/1b97294eaed4e7111aba6a807c4ece38472a7270/tedana/workflows/tedana.py#L519-L525
This should definitely include a check for `io_generator.overwrite` That could be added here or in a stand-alone PR.

tsalo:
@handwerkerd if it's alright, I think I'd like to leave that for a separate PR. I'd like to see this one merged in and I don't have much time to work on it further at the moment.

eurunuela (dismissed):
LGTM! Thank you Taylor!

handwerkerd:
@tsalo It looks like the conflicts should be easy to resolve then merge. You want to do that?

tedana/tests/data/reclassify_debug_out.txt
```diff
@@ -1,7 +1,6 @@
 figures
-ica_reclassify_call.sh
-references.bib
-report.txt
+sub-testymctestface_references.bib
```
handwerkerd, line 2:
```suggestion
ica_reclassify_call.sh
sub-testymctestface_references.bib
```

handwerkerd (changes requested):
Noticed a few things that must have been removed with the merge that were causing tests to fail. With the changed I'm including, all the tests were passing for me.

tsalo:
@handwerkerd sorry about screwing up the merge. I thought the conflicts were spurious, since I was under the (mistaken) impression that the only PR that had been merged since I last updated was your documentation one.

handwerkerd (approved):

handwerkerd:
Now I see what happened. #747 added the new system info fields that were causing the conflict & that was merged just last week. Since that does interact with his PR, maybe it's worth waiting for @eurunuela to check nothing got messed up with the system info, but the output looks fine to me. Since he already approved, if Eneko doesn't have time for one more check during the next few days, I'm fine with you merging.

eurunuela (approved):
It looks like the conflicts were solved. Like @handwerkerd said, the outputs look fine.

PR #965, opened 2023-08-07
Remove RefLGR from code and refs field from decision trees

### LINKED ISSUE #962: Reference logger is no longer used
### Summary
The references in the decision tree JSON files are currently printed out with the references logger, but that was removed a bit before the DTM branch was merged. We can drop the `RefLGR` stuff, along with the `references` field of the decision tree JSON files, and just use `\\citep{@bibtex}` as in the reporting logger messages. The only potential issue is that the BibTeX entries would need to already be present in the boilerplate BibTeX file.

### Next Steps

1. Remove mentions/uses of `RefLGR` throughout tedana.
2. Add references to the `report` field of the decision tree JSONs as BibTeX citations.
3. Remove the `references` field.
    - What if we made the `references` field a list of BibTeX entries somehow, and then injected that into the boilerplate as we built the full references list?
4. Update the decision tree JSON documentation.

tsalo:
Closes #962.

- Remove unused `RefLGR` logger.
- Drop `refs` field from decision trees, and move the references as BibTeX citations to the `report` field.


handwerkerd:
The reason there are references in the decision trees is because specific decision trees might have their own reference list. That is we want everyone who uses this code to reference the tedana JOSS pub, but if a new decision tree combines the tedana approach with the AROMA approach then they might want people to also cite AROMA and the publication that demonstrated the combination. I didn't link the reference field to anywhere else in the code, but my ideal would be to keep the field and have any references in the decision tree combine with the static references.

tsalo:
In that case I think we need to support the references field as a BibTeX-formatted string rather than in-text citations, but I don't know if there's an established format for that in JSON files. It won't go through the RefLGR though. It'll have to get loaded and merged with the main BibTeX file as part of the report curation.

handwerkerd:
I was starting to make some suggested changes, but now I think I get this. The idea is that the report field is the text added to the report and the docs now specify that it should include references in the context of the report text. The one bit of lost (non)functionality is that the goal of the json file decision tree is that people can make their own tree without needing to open a PR to the code. If someone wants to add a new reference, they'd need to make a PR to edit https://github.com/ME-ICA/tedana/blob/main/tedana/resources/references.bib

That's not ideal, but, if someone is starting to publish with a new decision tree that uses new references, we'd probably want to have that reference included anyway.

docs/building_decision_trees.rst
```diff
@@ -202,10 +202,9 @@ that is used to check whether results are plausible & can help avoid mistakes.
     A brief description of the tree for info logging
 
 - report
-    A narrative description of the tree that could be used in report logging
-
-- refs
-    Publications that should be referenced when this tree is used
+    A narrative description of the tree that could be used in report logging.
+    This should include any citations, which must be included in the references
+    BibTeX file.
```
handwerkerd, line 207:
```suggestion
    This should include any citations, which must be included in the
     `references BibTeX file`_.
     
.. _references BibTeX file: https://github.com/ME-ICA/tedana/blob/main/tedana/resources/references.bib
```

handwerkerd: Just added the link to the file so it's more clear what we mean.

tedana/tests/test_component_selector.py
```diff
@@ -36,7 +36,6 @@ def dicts_to_test(treechoice):
         "extra_opt_param": A tree with an undefined optional parameter for a decision node function
         "missing_req_param": A missing required param in a decision node function
         "missing_function": An undefined decision node function
```
handwerkerd, line 38:
```suggestion
        "missing_function": An undefined decision node function
        "missing_key": A dict missing one of the required keys (report)
```

We want to test whether the dict is missing one of the required keys! We just need to change it to another key.

tsalo:
That's the one drawback of this PR. Would you like me to figure out a way to keep the `refs` field in the JSON as a list of BibTeX entries instead?

tedana/tests/test_component_selector.py
```diff
@@ -127,8 +125,6 @@ def dicts_to_test(treechoice):
         tree["nodes"][0]["parameters"].pop("op")
     elif treechoice == "missing_function":
         tree["nodes"][0]["functionname"] = "not_a_function"
```
handwerkerd, line 127:
```suggestion
        tree["nodes"][0]["functionname"] = "not_a_function"
    elif treechoice == "missing_key":
        tree.pop("report")
```

tedana/tests/test_component_selector.py
```diff
@@ -242,10 +238,6 @@ def test_validate_tree_fails():
     with pytest.raises(component_selector.TreeError):
         component_selector.validate_tree({})
```
handwerkerd, line 240:
```suggestion
    # A tree that is missing a required key should not be valid
    with pytest.raises(component_selector.TreeError):
        component_selector.validate_tree(dicts_to_test("missing_key"))
```

handwerkerd (changes requested):
One minor and one less minor change.
Minor is just make a bit more clear where the references BibTex file is.
Less minor is that you removed the test for if a required key is missing. I put it back in, but I'm now checking to see whether `report` is missing.

handwerkerd:
I think this is a net improvement. If someone has a use case for citing a reference not in the .bib file, it might be easier to just have a very low threshold for us to add references for people rather than designing a system to handle this situation.

If you're fine with #968, approve it so we can merge it in. Then you can use that to make this PR pass all checks & we can then approve this as well.

tedana/tests/test_component_selector.py
tsalo: The check seems to be specific to `refs`, unless I was misreading the code.

tsalo: Ah, I just saw your original comment. Never mind.

docs/building_decision_trees.rst
```diff
@@ -202,10 +202,11 @@ that is used to check whether results are plausible & can help avoid mistakes.
     A brief description of the tree for info logging
 
 - report
-    A narrative description of the tree that could be used in report logging
-
-- refs
-    Publications that should be referenced when this tree is used
+    A narrative description of the tree that could be used in report logging.
+    This should include any citations, which must be included in the
+     `references BibTeX file`_.
```
handwerkerd, line 207:
```suggestion
    `references BibTeX file`_.
```
I looked at the rendered html and we needed to remove a space to have this render correctly

handwerkerd (changes requested):
Just merge the now current main into this so that the style & RTD checks pass & this should be ready to merge.

handwerkerd (approved):
LGTM

For the merging of #968 I didn't have a strong opinion about `type() is` vs `isinstance` but if you have a preference for `isinstance` that's fine with me.

handwerkerd:
This is fairly minor. I'm fine merging with 1 review if @eurunuela or @dowdlelt don't have time to look. Maybe give until Monday for them to respond?

eurunuela (approved):
LGTM!

tsalo:
Thanks @handwerkerd and @eurunuela!

PR #969, opened 2023-08-11
Generate metrics based on decision tree

### LINKED ISSUE #921: Check metrics exist before running tree. Possibly calculate metrics from tree
### Summary

In Decision Tree Modularization (#756), the functions in [`selection_nodes.py`](https://github.com/jbteves/tedana/blob/JT_DTM/tedana/selection/selection_nodes.py) were written so that it's possible to dry-run a tree and collect all metrics that would be needed to run the tree. This check is not currently being done. Once the metrics are gathered, it would  then be possible to have tedana load a decision tree and only calculate the metrics that are requested by the tree.

### Additional Detail

Every `selection_node` function has a ` only_used_metrics` parameter. If that's true, it will output the metrics it will use (i.e. kappa, rho, etc) but not actually run anything. This dry run would be added to the initialization of the component_selector object.

### Next Steps

- Figure out where this fits in relation to other priorities.

tsalo:
Closes #921.

- This helps ensure that metrics requested in decision trees are reflected in the component table. This will allow users to develop decision trees that use metrics other than the ones currently hardcoded into the tedana workflow code, as long as the metrics are actually defined in the tedana metric collection function.
- Restructure the ComponentSelector class to take `component_table`, `cross_component_metrics`, `status_table` as parameters to `select`, rather than `__init__`. This better fits the scikit-learn organization, in which `__init__` takes hyperparameters and `transform` (which is equivalent to `select`) takes actual data.
    - Also add a trailing underscore to any attributes that are set in `select`.
- Initialize the ComponentSelector in main tedana function, then grab the necessary metrics from the object. We also have to  ensure that a few extra metrics are calculated, since they're used for the reports.
  - `ComponentSelector` initialization is done before fMRI data are loaded to save time if the user-provided tree as an error
- When defined, `n_echoes` and `n_vols` are stored in `selector.cross_component_metrics` instead of as separate parameters
- `ica_reclassify` now starts at the first node in the inputted `tree` that doesn't have an `outputs` key rather than the first node that isn't defined in an un-run tree. This potentially opens up additional functionality since `ica_reclassify` could build on multiple previous executions
- Updates to docstrings and documentation to explain the added functionality.


tsalo:
I still need to update a bunch of tests and the reclassification workflow, since many things depend on the old class.

tsalo:
One thing I'm realizing here is that the current approach of passing the selector into decision functions and modifying it within them is unfortunately fragile. Given that the decision functions primarily modify the component table, I think they could be refactored to just accept the comptable and modify that.

handwerkerd:
@tsalo I'm realizing that to specify the external metrics in the decision tree json, like you requested in https://github.com/tsalo/tedana/pull/14 it will be much easier if we merge this first. I've identified a few of the breaking test issues and I think I can get this working, but it's going to intersect with a bunch of the recent updates to main. Do you want to align with `main` or should I just do that and work from there?

tsalo:
>I've identified a few of the breaking test issues and I think I can get this working, but it's going to intersect with a bunch of the recent updates to main. Do you want to align with main or should I just do that and work from there?

I think me updating from main is the easiest way forward, so I just did that.

>Which aspect of fragility is bothering you? Having a better sense of that might help me figure out how to resolve that issue.

My concern is that the functions rely on specific attributes in the ComponentSelector. Some attributes were expected to be present after initialization, even though they depend on data that, in this PR, are provided in select.

>One other nomenclature thing. If this class is now for metrics and selection and is loaded much earlier in the workflow, is it still the Component_Selector class with the object being called the selector? My inclination is to leave that as is for now to minimize messing things up, but we might want to rename in a later PR depending how things are refactored.

I like `DecisionTree`, but agree that we can change it later.

handwerkerd:
I don't want to keep opening pull requests, but I'm working here https://github.com/handwerkerd/tedana/tree/gen-req-metrics-dh and I'll open a PR when I think this is reasonably stable.

I identified one issue that was specific to tests in that a bunch of tests need to use a component_table before it is run through the decision tree. This used to be output with initialization, but, the new version runs the selection process in the same function where the component table is added to the selector. I needed to add un-executed component table into the selector in `test_selection_utils.py/sample_selector` to fix a bunch of testing failures.

The other issue is that, in the previous version n_echoes and n_vols were both keys in `selector` and within `selector.cross_component_metrics` with the later happening only when the selection process is run. I don't think this was necessary so now they're only in `selector.cross_component_metrics`

handwerkerd (approved):

handwerkerd:
@eurunuela Do you have time to review this in the near-ish future? Both Taylor & I worked on this, so it could use one more pair of eyes.

eurunuela:
> @eurunuela Do you have time to review this in the near-ish future? Both Taylor & I worked on this, so it could use one more pair of eyes.

Absolutely! I'll have a look at it next week.

handwerkerd:
@eurunuela Flagging again in case you have a chance to look this over. The dev call is next week and I'd like to know if there's anything in this PR that needs discussion there.

eurunuela:
Yes, sorry. Will definitely review this week.

eurunuela (dismissed):
LGTM!

docs/building_decision_trees.rst
```diff
@@ -210,11 +210,9 @@ that is used to check whether results are plausible & can help avoid mistakes.
 
 - necessary_metrics
     A list of the necessary metrics in the component table that will be used
-    by the tree. If a metric doesn't exist then this will raise an error instead
-    of executing a tree. (Depending on future code development, this could
-    potentially be used to run ``tedana`` by specifying a decision tree and
-    metrics are calculated based on the contents of this field.) If a necessary
-    metric isn't used, there will be a warning.
+    by the tree. This field defines what metrics will be calculated on each ICA
+    comopnent. If a metric doesn't exist then this will raise an error instead
```
tsalo, line 214:
```suggestion
    component. If a metric doesn't exist then this will raise an error instead
```

tedana/selection/selection_utils.py
```diff
@@ -316,50 +314,37 @@ def clean_dataframe(component_table):
...
-        True if all metrics in necessary_metrics are in component_table
+        Text identifying the function name that called this function.
 
     Raises
     ------
     ValueError
-        If metrics_exist is False then raise an error and end the program
+        If ``metrics_exist`` is False then raise an error and end the program.
 
     Note
```
tsalo, line 334:
```suggestion
    Notes
```
The underline was extended, and Notes is the appropriate section title, per the numpydoc standard.

tedana/workflows/tedana.py
```diff
@@ -705,6 +706,9 @@ def tedana_workflow(
             # If we're going to restart, temporarily allow force overwrite
             if keep_restarting:
                 io_generator.overwrite = True
+                # Create a re-initialized selector object if rerunning
+                selector = ComponentSelector(tree)
```
tsalo, line 710: This shouldn't be necessary. If it is, then we should open an issue to fix it in a future PR.

tsalo (commented):
Noticed a couple of small typos and one step that seems extraneous.

tedana/selection/selection_utils.py
handwerkerd: Most of the code uses `Notes` but I/Josh used `Note` all over `selection_nodes.py` and `selection_utils.py` and once in `ica_reclassify.py`. One of us should either fix that in this PR or in a more documentation-focused PR.

tsalo: In that case, then the added `-` in the line after this one should be removed.

tedana/workflows/tedana.py
handwerkerd: This was necessary because `selector.select` looks for the last node in the decision tree that was run and then continues to run the rest of the nodes. This is how `ica_reclassify` works.

This is already happening in `main` but it is within `tedica.automatic_selection`

https://github.com/ME-ICA/tedana/blob/a21d65ec5b1682a52cb8fe0eee1689d5eb3154a3/tedana/selection/tedica.py#L69-L70

By removing the component table from the `selector` initialization and now initializing in `tedana.py` the fact that this was happening is more up-front.

If the goal is to log all failed attempts, we would probably want to save both the ica mixing matrices and the outputs. This is a bit tricky because we currently save most files at the end and not as the data in those files are generated. We could also add some way to tack on a full tree to the end of the executed tree, but that might just add confusion. Either way, this does feel like a separate PR.

tsalo: Ah, thank you for explaining. This is different from scikit-learn classes, which have `fit`, `fit_transform`, and `transform` methods that distinguish between fitting the estimator to training data and transforming new data using the parameters from the fitting data. I know we don't need to have one-to-one correspondence, but I guess I didn't intuitively grasp it.

tedana/selection/selection_utils.py
```diff
@@ -316,50 +314,37 @@ def clean_dataframe(component_table):
...
 
     Raises
     ------
     ValueError
-        If metrics_exist is False then raise an error and end the program
+        If ``metrics_exist`` is False then raise an error and end the program.
 
-    Note
+    Notes
     ----
```
tsalo, line 335:
```suggestion
    -----
```

tedana/workflows/tedana.py
handwerkerd: I'd say this is the equivalent of running `fit_transform` a second time on the same dataset using the slightly different parameters (different seed in our cause). We also have a second use case that's the equivalent of running a second transform on a dataset that's already been transformed. I'm not sure that's something that is or should be time with the methods in `scikit-learn`

handwerkerd (approved):

eurunuela (approved):

tsalo:
Can we merge this?

handwerkerd:
@tsalo You're effectively the second review since I made the last round of changes. If you're happy with this, then merge.

Issue #976, opened 2023-09-05
Fragile dependencies


effigies:
### Summary

There are currently hard pins of bokeh and jinja2, which make it very difficult to install tedana alongside other tools that may require more recent versions. For example, installing jupyterlab into an environment with tedana installed produces:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tedana 23.0.1 requires jinja2==3.0.1, but you have jinja2 3.1.2 which is incompatible.
```

This is not currently a problem for me, as I am not using these features of tedana, but using `==`, or `<` pins tends to make it difficult to resolve dependencies.

It looks like these were added in https://github.com/ME-ICA/tedana/pull/793 and https://github.com/ME-ICA/tedana/pull/870, which are 2 and 1.5 years old, respectively. It is not uncommon for tools to only support dependencies released within the last 2 years or so, so it is going to become increasingly difficult to install tedana into the same environment as other tools in the near future.

A little related to #934, but not exactly the same.

### Next Steps

IIRC, there was a weirdness with Jinja 3.0.2, so possibly that pin can just be removed at this point (or updated to `jinja2 != 3.0.2`.

I don't know anything about bokeh and how hard it will be to accommodate a more recent version.

handwerkerd:
I just updated to `jinja2`=3.1.2  and I'm getting the same error so we cannot just skip a version or two.
An issue that provides more detail is #869. The `jinja2` & `bokeh` issues are linked. `bokeh` was locked on older versions of `jinja2` & it looks like `bokeh` removed that limit about a year ago so we should able to remove our `jinja2` limit if we update `bokeh`.

The person who worked on the `bokeh` parts of the `tedana` reports moved on to other things so we'd benefit from someone with more `bokeh` experience to figure out what's still holding back a `bokeh` upgrade & fix that.

eurunuela:
I have worked a bit on the bokeh side of things in our reports in the past. I can try to have a look and see if there is anything we can do with these dependencies.

Issue #934, opened 2023-04-05
Minimal dependency installation


effigies:
### Summary

When using tedana as a library and not generating reports, it seems excessive to have bokeh and jinja2 added to the dependency graph. I wonder if there's a way to have a lightweight dependency without sacrificing the user experience you already have.

### Additional Detail

One approach could be to create a separate package `tedana-core` that has the minimum dependencies needed to actually run the numeric processing the tedana provides. Now that you're moving to `hatchling` as a backend, it's more plausible to have a monorepo containing multiple packages, which would allow something like:

**`tedana-core/pyproject.toml`**
```toml
[project]
dependencies = [
  "scipy>=1.2.0",
  "scikit-learn>=0.21",
  "nibabel>=2.5.1",
  "nilearn>=0.7",
]

[project.optional-dependencies]
pca = [
    "mapca>=0.0.3"
]
 ```

**`tedana/pyproject.toml`** 
```toml
[project]
dependencies = [
  "tedana-core[pca]",
  "matplotlib",
  ...
]
```

I'm making things up, as I don't know your what components need which dependencies, but would there be any interest in breaking things down in this way?

### Next Steps

If you do want to go down this route, I'd be interested in taking a stab at it. I haven't really done the monorepo thing. Can't really promise a timeline, though.

handwerkerd:
I'm ok with this idea in theory, but it would require a bit of discussion on what is or isn't core. I think, if someone is running tedana denoising with ICA and then component selection, the report is and should be a mandatory output. Especially when tedana is used in a larger pipeline, it is essential to save a visualization that makes it easy for a user to see if something has gone wrong. That would keep bokeh and jinja2 in the core. (Jinja2 has caused some headaches and I love a lighter-weight alternative, but that's beyond me)

The current core function that wouldn't need the reports is generating the t2star map and the optimally combined volumes. The underlying math for those steps are simple enough that, if someone really wants them to be lightweight, it would be easier to just re-code / copy code.

The real thing that could become "core" is our whole framework for taking components (ICA or any time of components), calculating metrics, and classifying components based on those metrics. That has a potential for wider uses and doesn't interantly require multi-echo data. If they were the core of tedana, others could leverage the same code, but there's a bit of a chicken & egg issue in that I'm not sure it's worth really making them a lightweight core until others express interest in using them & others probably won't have interest until we separate them.

If you are interested & decided you have time to try, I'm definitely interested in talking more about this.

PR #977, opened 2023-09-08
Split out reports dependencies

tsalo:
Related to #934 and #976. To be honest, this doesn't quite close either, but it might be a useful first step. I'm fine going the monorepo route, but since I have no experience with that approach I decided to take a first step that just splits out the reports-specific dependencies into a new dependency group.

- Move bokeh and jinja2 into a new dependency group (reports).
- Move bokeh and jinja2 imports into the relevant functions, so we don't get import errors unless we use those functions.
- Add a function that checks for the existence of these imports when no-reports isn't enabled.


eurunuela:
The PR is now up to date with `main`. We now have to decide whether we take this route or go the monorepo route.

PR #992, opened 2023-10-24
Create a GitHub Action to automatically add issues to projects

eurunuela:
Closes none.

- Creates a GitHub Action that adds issues to the documentation and datasets project boards we have, if they have the `documentation` and `dataset` tags respectively.
- I cannot add the secret key to this repository, so someone with admin permissions will have to do it.

### DIFF
--- a/.github/workflows/populate-projects.yml
+++ b/.github/workflows/populate-projects.yml
@@ -1,0 +1,26 @@
+name: Add issues to project boards
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-doc-issues-to-project:
+    name: Add documentation issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@RELEASE_VERSION
+        with:
+          project-url: https://github.com/orgs/ME-ICA/projects/7
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          labeled: documentation
+  add-data-issues-to-project:
+    name: Add dataset issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@RELEASE_VERSION
+        with:
+          project-url: https://github.com/orgs/ME-ICA/projects/5
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          labeled: datasets


eurunuela:
Would you like to try this @handwerkerd @tsalo ? Feel free to close the PR if you'd rather manually update the boards.

tsalo (approved):
I think we'll need to merge before we can test it.

BTW do you know if there's any way to automatically label incoming issues as TODO in the relevant project?

tsalo:
Ah, oh well. Setting the issues as "TODO" can be something we tackle at a later date.

eurunuela:
> Ah, oh well. Setting the issues as "TODO" can be something we tackle at a later date.

Oh, that's not what I meant. I thank if the project has a kanban board with a column called "todo", we could just use the option `column` and set it to `column: TODO`.

handwerkerd:
I'm fine trying this but I'm still thinking about how we'll use the board in practice. As an example, I guess this would automatically put all `documentation` labelled issues into one project, which wouldn't have value added beyond just searching for issues with the `documentation` label unless we use that project to organize the open issues (which would be useful to do).

From @tsalo's comment, we already have `priority` `effort` and `impact` labels` that could potentially be used to automatically sort issues on the projects into some type of TODO. For `documentation` the bigger organization that would help me would be for "tedana execution documentation" vs "tedana explanation of the method" vs "multi-echo education." We could make labels for those, but that might be a good way to manually(?) organize issues in a project.

All that said, if @tsalo is fine adding a secret key to the repo and approving this, I'm also fine giving it a try.

handwerkerd (approved):

eurunuela:
@tsalo could you please merge this when you've added the secret key? Thanks!

tsalo:
@eurunuela sorry I missed Dan's comment from earlier this week. I just bumped up your role in the org so you should be able to add the secret key now.

eurunuela:
> @eurunuela sorry I missed Dan's comment from earlier this week. I just bumped up your role in the org so you should be able to add the secret key now.

Thank you! I'll add the key and merge then.

eurunuela:
I'm closing this as it looks like I was able to automate the process through the _workflows_ menu of each project.

PR #994, opened 2023-11-01
Exclude tests folder from precommit hooks.

eurunuela:
Closes none.

- Adds an exclude clause at the top of the pre-commit config file.


handwerkerd:
I think this can also be closed since running `pip install -e .'[all]'` seemed to address the issue.

Issue #856, opened 2022-03-10
Memory issue reported with nonlinear T2* estimation in fMRIPrep


tsalo:
### Summary
At least one user has reported issues with using fMRIPrep with low-memory processing on multi-echo data, as the T2* estimation step ends up being killed.

This stems from https://github.com/nipreps/fmriprep/issues/2728.

### Additional Detail
Here's the traceback:

```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/plugins/legacymultiproc.py", line 67, in run_node
    result["result"] = node.run(updatehash=updatehash)
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
    result = self._run_interface(execute=True)
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
    return self._run_command(execute)
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 428, in run
    runtime = self._run_interface(runtime)
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 822, in _run_interface
self.raise_exception(runtime)
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 749, in raise_exception
    raise RuntimeError(
RuntimeError: Command:
[...]

PR #995, opened 2023-11-01
remove unecessary copy of large data

bpinsard:

- remove copy of the data that seems unnecessary, cuts memory usage by ~30%
- data is not modified afterward, is copied through subsetting in the following lines (362).

investigation of memory usage originates from https://github.com/nipreps/fmriprep/issues/3125

### DIFF
--- a/tedana/decay.py
+++ b/tedana/decay.py
@@ -354,7 +354,6 @@ def fit_decay(data, tes, mask, adaptive_mask, fittype, report=True):
             f"mask ({mask.shape[0]}), and adaptive_mask ({adaptive_mask.shape[0]}) do not match"
         )
 
-    data = data.copy()
     if data.ndim == 2:
         data = data[:, :, None]
 


effigies:
@tsalo It looks like you added this `.copy()` in #468, but there doesn't seem to be any discussion that indicates it's necessary. My reading is that all downstream uses should be reading, not writing this, so I think this PR is a sensible optimization, but I'm definitely not well-versed in Tedana code.

tsalo (approved):
Good catch. Thanks @bpinsard.

@effigies you're right, `data` isn't modified in place in that function, so there's no need to copy it. Not sure why I added the `copy`.

handwerkerd (approved):
Thank you for noticing and correcting this!

This feels like something where the edited `data` was being passed out of the function, but that's not happening now and going through past versions I don't see where that might have been happening.

bpinsard:
Thanks for your feedback! It was an easy catch when I quickly mem-profiled the code, the use of float64 is definitely making the workflow memory intensive. I guess there could be further optimization. 
Is float64 used only for `scipy.optimize.curve_fit`? if so maybe converting to float64 could be differed to when it's called.

handwerkerd:
Barring essential needs, I'd lead towards keeping things as float64 because the code is doing fits and divisions in multiple places and rounding the final bit of information has been shown to non-trivially reduce precision. Having float64 means are data are reliably stable to precision above float32. There is a`lowmem` option that I don't particularly like, but maybe we can change that to run with float32.

FWIW, there's another standing memory issue ( #856 ). It's possible this fix addresses that issue, but we also identified another place where we can slightly reduce memory needs (see final comment). I'm not sure how much more memory that would save, but probably still worth implementing.

Also, this PR has two approvals and can be merged. @bpinsard as the person who created the PR you are welcome to hit `squash and merge`

bpinsard:
> Barring essential needs, I'd lead towards keeping things as float64 because the code is doing fits and divisions in multiple places and rounding the final bit of information has been shown to non-trivially reduce precision. Having float64 means are data are reliably stable to precision above float32. There is alowmem option that I don't particularly like, but maybe we can change that to run with float32.

I think #856 mainly originates from a too low estimates of memory reqs passed to nipype in fmriprep. I hit such errors even with 45G of RAM. Maybe if someone familiar with the tedana internals know how many times the data get transformed/duplicated we could try to figure out a better heuristics. Though the current PR might bring the estimate to what is required.

I don't think I have permissions to merge PRs on that repo.

handwerkerd:
@bpinsard I'm working on a tedana abstract for OHBM 2024. People who contributed in the last year are welcomed as co-authors. If you want to be a co-author, let me know so that I can make sure I have appropriate info to include.

tsalo:
@all-contributors please add @bpinsard for code.

Issue #869, opened 2022-04-04
Markup not found when running pip-installed tedana


jbteves:
### Summary
On an RHEL 7 system, running
```
pip install tedana
```
and results in the following when trying to confirm the installed version number:
```
tedana --version
Module `duecredit` not successfully imported due to "No module named 'duecredit'". Package functionality unaffected.
Failed to import duecredit due to No module named 'duecredit'
Traceback (most recent call last):
  File "/data/molfesepj/conda3/bin/tedana", line 6, in <module>
    from tedana.workflows.tedana import _main
  File "/data/molfesepj/conda3/lib/python3.7/site-packages/tedana/workflows/__init__.py", line 7, in <module>
    from .tedana import tedana_workflow
  File "/data/molfesepj/conda3/lib/python3.7/site-packages/tedana/workflows/tedana.py", line 21, in <module>
    from tedana import (
  File "/data/molfesepj/conda3/lib/python3.7/site-packages/tedana/reporting/__init__.py", line 5, in <module>
    from .html_report import generate_report
  File "/data/molfesepj/conda3/lib/python3.7/site-packages/tedana/reporting/html_report.py", line 8, in <module>
    from bokeh import embed, layouts, models
  File "/data/molfesepj/conda3/lib/python3.7/site-packages/bokeh/embed/__init__.py", line 23, in <module>
    from .server import server_document, server_session
  File "/data/molfesepj/conda3/lib/python3.7/site-packages/bokeh/embed/server.py", line 25, in <module>
    from ..core.templates import AUTOLOAD_REQUEST_TAG, FILE
  File "/data/molfesepj/conda3/lib/python3.7/site-packages/bokeh/core/templates.py", line 42, in <module>
    from jinja2 import Environment, FileSystemLoader, Markup
ImportError: cannot import name 'Markup' from 'jinja2' (/data/molfesepj/conda3/lib/python3.7/site-packages/jinja2/__init__.py)
```
@pmolfese brought this to my attention.

### Additional Detail
Tried new conda env, completely fresh, issue persisted. I tried the same thing, found that Markup was not in the manifest of installed packages, however, MarkupSafe was. Turns out that what happened was that jinja2 switched to `MarkupSafe` after version 3.0.1, but the existing calls remained, and bokeh is using the old versions and didn't pin an older version, thus it is automatically pointing at the newest version. As a workaround we can pin jinja2 at 3.0.1 in our requirements.

### Next Steps
[...]

PR #998, opened 2023-11-16
MNT: Uncap dependencies

### LINKED ISSUE #976: Fragile dependencies

effigies:
I looked at #869 and #766 and I think I roughly understand the problems they were trying to fix. I think the jinja2 one was a short-lived incompatibility that just removing it should fix. The other one appeared to affect the reports. Running `make three-echo` in an environment with the latest bokeh produced one complaint that resolved to https://github.com/bokeh/bokeh/issues/12543, and otherwise just worked.

image

Closes #976.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ classifiers = [
 license = {file = "LICENSE"}
 requires-python = ">=3.8"
 dependencies = [
-    "bokeh<2.3.0",
+    "bokeh",
     "mapca>=0.0.3",
     "matplotlib",
     "nibabel>=2.5.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,6 @@ dependencies = [
     "scikit-learn>=0.21",
     "scipy>=1.2.0",
     "threadpoolctl",
-    "jinja2==3.0.1",
 ]
 dynamic = ["version"]
 
--- a/tedana/reporting/dynamic_figures.py
+++ b/tedana/reporting/dynamic_figures.py
@@ -162,8 +162,8 @@ def _create_kr_plt(comptable_cds, kappa_elbow=None, rho_elbow=None):
         ]
     )
     fig = plotting.figure(
-        plot_width=400,
-        plot_height=400,
+        width=400,
+        height=400,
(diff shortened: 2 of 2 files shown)


eurunuela (approved):
Wow thank you @effigies!

You just saved me a couple of hours of work at Brainhack next week!

I have looked at the reports (for next time, you can see them in the artifacts panel in CircleCI) and the bokeh side of things seems to be working well.

Thank you once again! I'm glad it was a simpler fix than expected. 

That said, should we have a minimum version of bokeh? I imagine that bokeh moved from `plot_width` to `width` at some point? I'm happy to approve as is, since we've had this issue open for way too long.

effigies:
I think a decent policy is to support two years back, but I reverted because the width and height options have been present the whole time I could see.

I can set a minimum, or it can be a separate issue. Your call.

eurunuela:
If the `width` and `height` options have been there the whole time, I don't know why we were using `plot_width` and `plot_height` instead 😅

Happy to leave it as is then. Thank you Chris!

tsalo (approved):
LGTM!

PR #1006, opened 2023-11-22
Remove accepted and rejected time series from the standard file outputs and change denoised time series name

### LINKED ISSUE #819: Remove accepted and rejected time series from the standard file outputs

martaarbizu:
Closes #819  .

- Remove accepted and rejected time series from the standard file outputs
- Change the name of optcomDenoised to denoised

### DIFF
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -577,16 +577,18 @@ def write_split_ts(data, mmix, mask, comptable, io_generator, echo=0):
     if len(acc) != 0:
         if echo != 0:
             fout = io_generator.save_file(hikts, "high kappa ts split img", echo=echo)
+            LGR.info(f"Writing high-Kappa time series: {fout}")
         elif io_generator.verbose:
             fout = io_generator.save_file(hikts, "high kappa ts img")
-        LGR.info(f"Writing high-Kappa time series: {fout}")
+            LGR.info(f"Writing high-Kappa time series: {fout}")
 
     if len(rej) != 0:
         if echo != 0:
             fout = io_generator.save_file(lowkts, "low kappa ts split img", echo=echo)
+            LGR.info(f"Writing low-Kappa time series: {fout}")
         elif io_generator.verbose:
             fout = io_generator.save_file(lowkts, "low kappa ts img")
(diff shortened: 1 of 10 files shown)


eurunuela (approved):
LGTM! @tsalo @handwerkerd what do you think?

handwerkerd:
Thank you for taking care of this.
I just opened a PR to @martaarbizu's branch to update the API documentation to reflect these changes.

I'd want @tsalo's option, but I'd lean against changing `optcomDenoised` to `denoised`. Other programs that use these file names, like fMRIPrep & personal scripts, would likely need to make a change to properly process files. I like `denoised` slightly more, but both names are fine and I'll lean against a change that will break other people's scripts unless we have a strong reason to do so.

That said, I am labeling this as a "breaking change" since we are altering default outputs (even if they are default outputs no one actually uses)

tsalo:
>I'd want @tsalo's option, but I'd lean against changing optcomDenoised to denoised. Other programs that use these file names, like fMRIPrep & personal scripts, would likely need to make a change to properly process files. I like denoised slightly more, but both names are fine and I'll lean against a change that will break other people's scripts unless we have a strong reason to do so.

I believe that I had this filename change in my version as well. FWIW, fMRIPrep only uses the t2smap workflow, so it doesn't produce/use the denoised file. No matter what, this will be a breaking change (there's always the possibility that some users are using the files that this PR removes), so I say we go for broke and rename everything we think would benefit from a change.

handwerkerd:
@tsalo Does BIDS care if `denoised` is or is not capitalized? If not, once my additions to the comments are added, I'm fine with this merging. When we release the next version, we should make a non-trivial note that we've changed a file name because that will affect people.

tsalo:
There's no rule about the naming, as long as we stick to alphanumeric strings. I personally prefer lower-case or lowerCamelCase, but that's not required at all.

handwerkerd:
@martaarbizu Thank you again for working on this. I made a pull request onto your PR to improve the documentation. You should be able to approve and merge https://github.com/martaarbizu/tedana/pull/1 into your PR and then this will be ready to merge into Main.

PR #1008, opened 2023-11-24
[ENH] Allow for correlations with regressors in decision tree

### LINKED ISSUE #1009: Add option to classify components based on their correlation with regressors
### Summary

### Additional Detail

eurunuela:
Closes #1009.

- Adds the `dec_correlation_higherthan_thresholds()` function to `selection_nodes.py`, which allows the user to provide a `tsv` file with regressors in its column to consider during component classification. The function also expects the number of metric labels (e.g., `["visual task", "motor task"]`) to match with the number of regressors (i.e., the number of columns in the regressors file).


handwerkerd:
Thank you for starting to work on this and sharing so that it's possible to start discussing.

I do have one concern that I wanted to bring up sooner rather than later. The way you're approaching this now, you're effectively calculating a metric per component as part of a selection process. The advantage of this is that  decision tree metrics can be tweaked using just the decision tree code, but the disadvantage is, if we start calculating metrics in the decision tree, then it's harder to understand what's happening using the decision tree outputs. That is, full columns would be added to the component table and nodes in the decision tree both make calculations and decisions. If I'm reading the draft code correctly, your not saving the fit metrics in the component table & that's important info to save.

My preference would be to add this code as a metric and then most of the decisions could be made just using `dec_left_op_right`. That would make it hard to play around with regressor fit metrics on the same components, but there are ways to address that issue. We can talk more about this later, but wanted to bring this up before you do a lot more work on this.

handwerkerd:
Also @goodalse2019 and @n-reddy both mentioned some interest in working on this, so I want to make sure they both saw you've started.

eurunuela:
That's a fair point @handwerkerd. I actually thought of using `dec_left_op_right` for this, but ended up building this other approach. It seemed to me the approach on this PR would be the easiest/quickest way to have a working version of it ready.

In any case, happy to discuss how we can adjust the code to use `dec_left_op_right`.

tsalo:
I like the idea of creating a new `metrics/external.py` module with functions to correlate the component time series with the external ones. Then the metrics could be evaluated with `dec_left_op_right` in the decision tree.

handwerkerd:
@eurunuela if you think the approach @tsalo and I are working on in #1021 is the one we'll move forward with, do you think this PR should be closed?

eurunuela:
Absolutely. Let's close this one.

PR #1014, opened 2024-01-08
Adding python library version numbers

handwerkerd:
When we closed #747 I had made a comment that we might also want to log the version numbers of key modules used by tedana. This PR does that (plus identifies & corrects an omission from that PR)

- `dataset_description.json` includes the version numbers for all libraries specified in `project.toml`
-  `utils.get_system_info` is now `utils.get_system_and_version_info`. The previous line to get the python version was moved from tedana.py and ica_reclassify.py into this function and this function now compiles the version numbers for the other packages
  - I wouldn't mind someone making sure I'm getting all the version numbers in a pythonic and not-too-inefficient way.
- In making the above changes, I noticed the `tedana` and `ica_reclassify` workflows saves the system information, but `t2smap` did not so I added that info to `t2smap`
- I also noticed that the method in the `_main` functions for getting the text of a command line input worked for an actual command line input, but not `tedana_cli` calls, which we were using in the tests. I think I fixed this is a way that won't break anything else.
  - This causes code coverage to drop since the CLI line that created garbled outputs is now skipped. Not sure how to directly call that line.
- Added library version info to `tedana_report.html` via edits to `report_info_table_template.html` and `html_report.py`

For future work (or edits to this PR). The three work flows all have a very similar block of text to specify and write `derivative_metadata` This could be it's own function. Also, `BIDSVersion` is hard-coded in all three. Given 1.5.0 is the only version we support, that's not a big deal, but, if we ever add an updated output option, we might want to make that a variable.

I was reminded to do this via https://github.com/nipreps/fmriprep/issues/3196

### DIFF
--- a/tedana/reporting/data/html/report_info_table_template.html
+++ b/tedana/reporting/data/html/report_info_table_template.html
@@ -1,7 +1,7 @@
 <div>
   <p>
     <span class="title">Tedana command used:</span>
-    <pre class="language-python">
+  <pre class="language-python">
       <code class="code_block">
         $command
       </code>
@@ -39,6 +39,10 @@
     <tr>
       <td class="title">Tedana version:</td>
       <td>$tedana</td>
+    </tr>
+    <tr>
+      <td class="title">Other library versions:</td>
+      <td>$python_libraries</td>
     </tr>
   </table>
 </div>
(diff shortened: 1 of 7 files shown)


tsalo (approved):
LGTM!

eurunuela (approved):
LGTM! Thank you Dan.

PR #1020, opened 2024-02-03
Reformat with new version of black

tsalo:
Closes none.

- Run black on package.
- Run isort on top-level files (docs).


tsalo:
@handwerkerd that's 100% right

PR #1024, opened 2024-02-12
Add Dependabot config and switch Release Drafter to Changelog config

tsalo:
Closes none.

- Add a dependabot config to automatically open PRs updating to new versions of dependencies. This will help us keep up to date _and_ flag any incompatible versions.
- Switch from release drafter to a changelog config. We can use PR labels to organize our release notes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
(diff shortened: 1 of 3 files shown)


eurunuela (approved):
This is great @tsalo, thank you!

handwerkerd (approved):
I agree with this plan so I'll approve, but I have no clue how to evaluate/test if you set this up correctly. My assumption is, if there are issues, we'll find out if we merge this first and then see what happens when the mapca version is updated.

tsalo:
Thanks @eurunuela and @handwerkerd!

@handwerkerd yeah, unfortunately it's one of those things where we'll find any problems after merging, but I did copy over these files from a repo I maintain that is working well, so I think it'll be okay.

PR #1029, opened 2024-02-15
Pin maximum versions of dependencies in pyproject.toml

### LINKED ISSUE #1028: Switch to pinning maximum versions in pyproject.toml

tsalo:
Closes #1028.

- Switch from setting minimum versions to maximum versions in the pyproject.toml. I used the versions installed from the last 3.12 CircleCI environment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,16 +22,16 @@ license = {file = "LICENSE"}
 requires-python = ">=3.8"
 dependencies = [
     "bokeh",
-    "mapca<=0.0.4",
+    "mapca>=0.0.3,<=0.0.4",
     "matplotlib",
-    "nibabel<=5.2.0",
-    "nilearn<=0.10.2",
-    "numpy<=1.26.3",
-    "pandas<=2.2.0",
+    "nibabel>=2.5.1,<=5.2.0",
+    "nilearn>=0.7,<=0.10.2",
+    "numpy>=1.16,<=1.26.3",
+    "pandas>=2.0,<=2.2.0",
     "pybtex",
     "pybtex-apa-style",
-    "scikit-learn<=1.4.0",
-    "scipy<=1.12.0",
+    "scikit-learn>=0.21, <=1.4.0",
+    "scipy>=1.2.0, <=1.12.0",
     "threadpoolctl",
 ]
 dynamic = ["version"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,12 +26,12 @@ dependencies = [
     "matplotlib",
     "nibabel<=5.2.0",
     "nilearn<=0.10.2",
-    "numpy<=1.24.4",
-    "pandas<=2.0.3",
+    "numpy<=1.26.3",
+    "pandas<=2.2.0",
     "pybtex",
     "pybtex-apa-style",
-    "scikit-learn<=1.3.2",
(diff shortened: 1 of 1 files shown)


tsalo:
The versions that work for 3.8 don't seem to be compatible with 3.12. Maybe we should drop 3.8 support?

pyproject.toml
```diff
@@ -22,16 +22,16 @@ license = {file = "LICENSE"}
 requires-python = ">=3.8"
 dependencies = [
     "bokeh",
-    "mapca>=0.0.3",
+    "mapca<=0.0.4",
```
eurunuela, line 25:
```suggestion
    "mapca<=0.0.5",
```

eurunuela: I made a new release earlier, so we could already ping to that one.

eurunuela (dismissed):

pyproject.toml
tsalo: Thanks! I actually think I should leave it as-is, since dependabot should flag it for us almost immediately after this is merged.

handwerkerd:
I'm a bit confused by this. We know that tedana won't work with all older versions of these libraries. For example, tedana is looking for outputs from `mapca` that didn't exist in older versions. For `mapca` I given the issue/bug with the newest version of `scipy` in v0.0.3 it seems to be that we should want to set it to `mapca=>0.0.5`

Similarly, by pinning maximum versions we'll have to proactively update these version maximums rather than paying attention to when new issues arise.

Like I said. I am confused. I'm I misunderstanding the reasoning for this?

tsalo:
>We know that tedana won't work with all older versions of these libraries.

We can set minimum versions if you want, but I'm more worried about breaking changes from new releases than people accidentally installing old versions.

>For mapca I given the issue/bug with the newest version of scipy in v0.0.3 it seems to be that we should want to set it to mapca=>0.0.5

I have kept it pinned to 0.0.4 to ensure that dependabot will open a PR soon.

>Similarly, by pinning maximum versions we'll have to proactively update these version maximums rather than paying attention to when new issues arise.

That's what dependabot is for. Once a new release is made, dependabot will open a pull request with the new version pinned, and if CI fails then we'll know something in the new release doesn't work with our code, rather than surprising us with a user's bug report or failing CI on an unrelated PR.

handwerkerd (changes requested):
Besides my one suggested change. I think this is fine.

Having many PRs for version updates will be annoying, but I understand what it's useful & we can deal. I'm slightly concerned what happens if/when `tedana` becomes less actively maintained, but we can deal with that in the future.

tsalo:
Settings both minimum and maximum versions seems to work well. I don't think dependabot can flag problematic minimum versions. We could probably create a minimum-version dependency set and run tests on that, like Nilearn does. That's out of scope for this PR though.

pyproject.toml
eurunuela: Good point!

handwerkerd:
Now that mapca is up to version 0.0.5 any clue how long it should take for the dependency bot to trigger?

tsalo:
I had hoped that it would happen immediately after merging this PR, but it looks like we have it set up to run weekly. So, since dependabot was first added 3 days ago, I guess it might not run until 4 days from now?

PR #1031, opened 2024-02-16
Fix error on empty input file & pre-commit updated to black version 24.2.0

### LINKED ISSUE #974: ica_reclassify Errors on Empty Input File
### Summary
The ica_reclassify workflow errors if the file provided to `--manacc` or `--manrej` is empty:
 ```
 if s[-1] == "\n":
 IndexError: string index out of range
 ```
 It runs fine if the argument is omitted, but it seems like this shouldn't be the behavior if there's an empty input file.

### Additional Detail
Tedana v23.0.1
Python v3.9.13

### Next Steps
One possible solution is to have the manually accepted and rejected files just contain every accepted or rejected component (respectively) after manual classification. Obviously, it's more space efficient to store only the modified components because a list of the original accepted and rejected components already exists. But even with hundreds of components the files would be a few KB at most, so there aren't huge space concerns. And keeping one file with the finalized list of all accepted components and one with the finalized list of all rejected components is a bit more straightforward.

handwerkerd:
Closes #974. Closes #1007.

- ica_reclassify Errors on Empty Input File fix
- This is the change @martinezeguiluz made in #1007 to fix the above issue plus merging my suggested update to that PR.
- I also merged in the last 3 months of updates to main and fixed resulting style issues.
- In trying to get this to pass the style check, I realized that pre-commit was using an older version of Black than CircleCI so I've updated the black version to use in `.pre-commit-config.yaml`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ repos:
       - id: check-yaml
       - id: check-added-large-files
   - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 23.10.0
+    rev: 24.2.0
     hooks:
       - id: black
         files: ^tedana/
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -394,7 +394,7 @@ def tedpca(
         else:
             alg_str = algorithm
         LGR.info(
-            f"Selected {comptable.shape[0]} components with {round(100*varex_norm.sum(),2)}% "
+            f"Selected {comptable.shape[0]} components with {round(100 * varex_norm.sum(), 2)}% "
             f"normalized variance explained using {alg_str} dimensionality estimate"
         )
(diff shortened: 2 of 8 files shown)


tsalo:
There's a linting issue, but other than that (maybe you need to update black?) it looks good to me.

handwerkerd:
I think I fixed the style issue. The precommit was the issue. I upgraded to black=24.2.0 and I think that's what CircleCI is using, but the precommit was using 23.10.0 and would un-edit the style change when I tried to push the commit. I needed to update the version of black in `.pre-commit-config.yaml` and now it looks like the style test is passing. I'll make a note of this in the PR title

tsalo (approved):
LGTM!

eurunuela (approved):
LGTM! Thank you Dan.

handwerkerd:
@all-contributors please add @martinezeguiluz for code

PR #1033, opened 2024-02-18
Remove accepted and rejected time series from the standard file outputs and change denoised time series name (updated)

### LINKED ISSUE #819: Remove accepted and rejected time series from the standard file outputs

handwerkerd:
Closes #819. Closes #1006.

This is an update to #1006 by @martaarbizu that includes changes to the docstrings describing the changed output file names. 

- Remove accepted and rejected time series from the standard file outputs
- Change the name of optcomDenoised to denoised

Removing files that were previously saved by default is a breaking change and renaming saved files is also a breaking change.


eurunuela (approved):
LGTM!

tedana/io.py
```diff
@@ -578,16 +589,18 @@ def write_split_ts(data, mmix, mask, comptable, io_generator, echo=0):
     if len(acc) != 0:
         if echo != 0:
             fout = io_generator.save_file(hikts, "high kappa ts split img", echo=echo)
-        else:
+            LGR.info(f"Writing high-Kappa time series: {fout}")
+        elif io_generator.verbose:
             fout = io_generator.save_file(hikts, "high kappa ts img")
-        LGR.info(f"Writing high-Kappa time series: {fout}")
+            LGR.info(f"Writing high-Kappa time series: {fout}")
```
tsalo, line 595: Unnecessary duplication here, if I'm not mistaken.

tedana/io.py
```diff
@@ -578,16 +589,18 @@ def write_split_ts(data, mmix, mask, comptable, io_generator, echo=0):
...
 
     if len(rej) != 0:
         if echo != 0:
             fout = io_generator.save_file(lowkts, "low kappa ts split img", echo=echo)
-        else:
+            LGR.info(f"Writing low-Kappa time series: {fout}")
+        elif io_generator.verbose:
             fout = io_generator.save_file(lowkts, "low kappa ts img")
-        LGR.info(f"Writing low-Kappa time series: {fout}")
+            LGR.info(f"Writing low-Kappa time series: {fout}")
```
tsalo, line 603: Unnecessary duplication.

tedana/io.py
```diff
@@ -623,19 +636,20 @@ def writeresults(ts, mask, comptable, mmix, io_generator):
...
     ---------------
 
-    =========================================    =====================================
+    =========================================    ===========================================
     Filename                                     Content
-    =========================================    =====================================
-    desc-optcomAccepted_bold.nii.gz              High-Kappa time series.
-    desc-optcomRejected_bold.nii.gz              Low-Kappa time series.
+    =========================================    ===========================================
     desc-optcomDenoised_bold.nii.gz              Denoised time series.
```
tsalo, line 642:
```suggestion
    desc-denoised_bold.nii.gz              Denoised time series.
```

tedana/io.py
handwerkerd: I think this is correct, but confusing. There's a comment in the doc string, but I'll add more comments here to clarify. `write_split_ts` is called from two places:
In `tedana.py` there's a call to `io.writeresults_echoes` that is used to write out the time series for each echo separate and that only happens when the verbose flag is used (but others could use this function to write out echo time series without the verbose flag). That triggers the `LGR.info` statement on line 592. https://github.com/handwerkerd/tedana/blob/4727198f2ef7ef8a095d1e00081d2f0231f8065e/tedana/workflows/tedana.py#L819

The second is in `writeresults` which is used to write the denoised time series and, if `verbose` then also write out the high kappa and low kappa time series. https://github.com/handwerkerd/tedana/blob/4727198f2ef7ef8a095d1e00081d2f0231f8065e/tedana/io.py#L655

As written, if it's NOT verbose, then neither the `if` nor `elif` statements will be triggered so we don't want that `LGR.info` line separately executed.

tsalo: Ah, my bad. I missed that it was changed from else to elif. Sorry for the noise!

handwerkerd: I made a fix to the docstring and added a comment here because this section of code confused me (and you) every time I've looked it it.

eurunuela (approved):

tsalo (approved):

PR #1040, opened 2024-02-20
Add T2* and S0 figures

### LINKED ISSUE #689: Add T2*/S0 maps to HTML reports
### Summary
It would be great to include the T2* and S0 maps in the HTML reports. Another useful, but not required, feature to include would be histograms of T2* and S0 values.

### Additional Detail
This stems from #450.

tsalo:
Closes #689.

- Add seaborn as a dependency.
- Add new figures to static_figures and HTML report.
    - Mosaic-format brain plots created with `nilearn.plotting.plot_stat_map`.
    - Histograms created with `seaborn.histplot`.


tsalo:
The brain images are massive in the HTML report for some reason. @eurunuela do you have any idea what might be causing that?

eurunuela:
> The brain images are massive in the HTML report for some reason. @eurunuela do you have any idea what might be causing that?

Yes. So it looks like you haven't used any CSS to limit the size of the image. It is currently taking 100% of the width of its parent div, which happens to be 100% of the page.

I can work on improving the overall layout of the report in the coming weeks.

eurunuela:
Ah, I see you made some edits to the CSS @tsalo, awesome!

Let me know if you want me to work on improving this.

tsalo:
I'm happy to just get something readable working in this PR, but I'd love to improve the report overall in a future one.

One thing that would be amazing is if we could have all of the figures be rendered with the same width. Maybe even based on the window size, if that's feasible?

eurunuela:
> One thing that would be amazing is if we could have all of the figures be rendered with the same width. Maybe even based on the window size, if that's feasible?

Yes, that's feasible and shouldn't be difficult to build.

tedana/reporting/static_figures.py
```diff
@@ -463,3 +464,69 @@ def pca_results(criteria, n_components, all_varex, io_generator):
...
+    fig.savefig(os.path.join(io_generator.out_dir, "figures", t2star_histogram))
+
+    s0_data = masking.apply_mask(s0_img, mask_img)
+    s0_histogram = f"{io_generator.prefix}s0_histogram.svg"
+
+    fig, ax = plt.subplots(figsize=(10, 6))
+    sns.histplot(data=s0_data, bins=100, kde=True, ax=ax)
+    ax.set_title("S0", fontsize=20)
+    ax.set_xlabel("Arbitrary Units")
+    fig.savefig(os.path.join(io_generator.out_dir, "figures", s0_histogram))
```
handwerkerd, line 532: For the 3 echo test data, I noticed the ranges were all flat and this is correctable. The following just saves the histograms before the maps so that I can calculate the 2nd and 98th percentile values on `t2star_data` and `s0_data` to and to use those values to scale the colormap for the brain maps. The histograms are also off and I tried to use `binrange` in `sns.histplot` but it wasn't saving correctly and I don't know why.

```suggestion
    # Plot histograms
    sns.set_style("whitegrid")

    t2star_data = masking.apply_mask(t2star_img, mask_img)
    t2star_percentiles_2_98_99 = np.percentile(t2star_data, [2, 98, 99])
    t2star_histogram = f"{io_generator.prefix}t2star_histogram.svg"

    fig, ax = plt.subplots(figsize=(10, 6))
    sns.histplot(
        data=t2star_data, bins=100, kde=True, ax=ax
    )
    ax.set_title("T2*", fontsize=20)
    ax.set_xlabel("Seconds")
    fig.savefig(os.path.join(io_generator.out_dir, "figures", t2star_histogram))

    s0_data = masking.apply_mask(s0_img, mask_img)
    s0_percentiles_2_98_99 = np.percentile(s0_data, [2, 98, 99])
    s0_histogram = f"{io_generator.prefix}s0_histogram.svg"

    fig, ax = plt.subplots(figsize=(10, 6))
    sns.histplot(data=s0_data, bins=100, kde=True, ax=ax)
    ax.set_title("S0", fontsize=20)
    ax.set_xlabel("Arbitrary Units")
    fig.savefig(os.path.join(io_generator.out_dir, "figures", s0_histogram))

    # Plot T2* and S0 maps
    t2star_plot = f"{io_generator.prefix}t2star_brain.svg"
    plotting.plot_stat_map(
        t2star_img,
        bg_img=None,
        display_mode="mosaic",
        symmetric_cbar=False,
        cmap=png_cmap,
        vmin=t2star_percentiles_2_98_99[0],
        vmax=t2star_percentiles_2_98_99[1],
        output_file=os.path.join(io_generator.out_dir, "figures", t2star_plot),
    )

    s0_plot = f"{io_generator.prefix}s0_brain.svg"
    plotting.plot_stat_map(
        s0_img,
        bg_img=None,
        display_mode="mosaic",
        symmetric_cbar=False,
        cmap=png_cmap,
        vmin=s0_percentiles_2_98_99[0],
        vmax=s0_percentiles_2_98_99[1],
        output_file=os.path.join(io_generator.out_dir, "figures", s0_plot),
    )
```

handwerkerd (changes requested):
Thank you for doing this. These are nice additions. In addition to my suggested change below, the colorbars for maps are really narrow and unreadable. I can't figure out how to make them wider.

Also, once the figures are finalized, the "ICA Components Report" section of `outputs.rst` in the docs should be updated to mention/explain the new figures.

handwerkerd:
One other Q is the addition of seaborn. It's a widely used library, but it's one more dependency. It makes the histograms slightly nicer, but I don't know if it's necessary. If you eventually plan to use seaborn to make everything look a bit nicer, that's one thing, but, for this PR, I don't think it's essential. (Not going to protest if you really want to keep the seaborn visualizations)

tsalo:
>It makes the histograms slightly nicer, but I don't know if it's necessary.

That's a good point. I've dropped seaborn.

tedana/reporting/static_figures.py
tsalo: I dropped seaborn, so there'll be a conflict with your suggestion. I can implement the general idea though.

What do you mean by the histograms being "off"?

handwerkerd: This is what the S0 histogram looks like for the three-echo tests. This is a weird truncated dataset, but it would be better to scale to the 98th or 99th percentile so that the x-axis bins are more like from 0-30000. This should be possible with the `binrange` parameter but it wasn't altering the saved plot. Maybe it will be easier to scale the range of the bins outside of seaborn.
s0_histogram

tsalo: I've added the 98th percentile as an upper limit to the figure, but TBH I think that the extreme values are informative.

handwerkerd:
This looks good. The one last thing to do before merging is to add some images and explanatory text to the "ICA Components Report" section of `outputs.rst` in the docs.

handwerkerd (commented):
I have one annoying suggestion. We have a parameter to select the colormap and we default to a red/blue map for the ICA components with positive and negative values. The T2* and S0 maps are all positive and I find are much more interpretable in grayscale. I'd suggest hard-coding them as gray scale rather than using the inputted colormap. If you disagree, I won't argue.

Also, the other text descriptions give a hint at what people should be looking for. You could add something like "The T2* map should look similar to T2 maps and be brightest in the ventricles. The S0 map should roughly follow the signal-to-noise ratio and will be brightest near the surface near receiver coils." 
The T2* histogram should have more values or a second bump to the left of the peak, but I'm not sure what to highlight to users as a warning sign of a problem.

docs/outputs.rst
```diff
@@ -527,6 +527,19 @@ should not overly focus on carpet plots and should examine these results in cont
   :height: 400px
 
 
+************************
+T2* and S0 Summary Plots
+************************
+
+Below the carpet plots are summary plots for the T2* and S0 maps.
+Each map has two figures: a spatial map of the values and a histogram of the voxelwise values.
```
handwerkerd, line 535:
```suggestion
Below the carpet plots are summary plots for the T2* and S0 maps.
Each map has two figures: a spatial map of the values and a histogram of the voxelwise values. The T2* map should look similar to T2 maps and be brightest in the ventricles and darkest in areas of largest susceptibility. The S0 map should roughly follow the signal-to-noise ratio and will be brightest near the surface near RF coils.
```
Thanks to Renzo Huber for giving me feedback on this language.

handwerkerd (changes requested):
One minor suggestion for the documentation text and my other suggestion to make the T2* and S0 maps always grayscale. I'm flexible on both, if you have other preferences.

tsalo:
@handwerkerd what do you think of the updated version?

image

If you like it, I'll update the image in the docs.

tedana/reporting/static_figures.py
```diff
@@ -463,3 +463,78 @@ def pca_results(criteria, n_components, all_varex, io_generator):
...
+
+    # Plot T2* and S0 maps
+    t2star_plot = f"{io_generator.prefix}t2star_brain.svg"
+    plotting.plot_stat_map(
+        t2star_img,
+        bg_img=None,
+        display_mode="mosaic",
+        symmetric_cbar=False,
+        cmap="binary",
```
handwerkerd, line 522:
```suggestion
        cmap="gray",
```
The binary color map is low-to-high white-to-black while the gray colormap is black-to-white, which is both how I think these images are typically viewed and how the text in the documentation describes the images.

handwerkerd (changes requested):
Sorry for prolonging this. In addition to suggested changes below, `t2s_and_s0_plots.png` should use a grayscale image. Also, I'm fine with that image just including the t2s example, but, given the file name, did you intend for it to also include s0 examples?

tedana/reporting/static_figures.py
tsalo: I'm concerned that a black-to-white colormap on a white background will result in high values disappearing, but I'm happy to try it out.

tsalo: image

I think the black-to-white colormap makes the difference between uncovered voxels and high-T2* values unclear.

handwerkerd: In the two calls to  ` plotting.plot_stat_map` you could add `black_bg=True` to make a black background and then the lowest values will fade into black.
image

handwerkerd:
Thanks for humoring my requests. I think the maps look good now so just update `t2s_and_s0_plots.png` and this should be ready.

handwerkerd (approved):
LGTM
Sorry again for creating so much back-and-forth on this one.

tsalo:
No worries. I'm glad it looks good to you.

dowdlelt (approved):
I'm happy with this - the nice visual is a useful addition and a good sanity check. Nice work ya'll - I had completely missed that this was being put together.

PR #1044, opened 2024-02-21
Output RMSE map and time series for decay model fit

### LINKED ISSUE #861: Revisiting #230 - incorporate decay curve fit information into workflow
### Summary
We're interested in how decay curve fits compare across subject age groups, from human neonates through adults. It would be great to output decay curve fit information including some or all of the measures proposed by @tsalo in the (now-closed) #230:

> - The residuals returned by [numpy.linalg.lstsq](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.lstsq.html).
> - Root mean squared error, calculated from the log data and the predicted values.
> - Something using the voxel- and volume-wise regressions. The reason I think this would be useful is that, assuming high-quality data, actual BOLD signal should increase the variability of the data within each echo, decreasing regression fit.

tsalo:
Closes #861. 

- Add a function to calculate RMSE 4D time series from monoexponential model fit.
- Add RMSE mean brain and time series plots to the HTML report.
- Combine RMSE time series and global signal time series (when calculated) into a `desc-confounds_timeseries.tsv` file. This removes the global signal time series file (`desc-globalSignal_timeseries.tsv`).


tsalo:
@madisoth gave me the thumbs up on this offline, so I'm opening it up for review.

tsalo:
Here's the mean RMSE image for the 3-echo dataset:

Screenshot 2024-04-12 at 12 09 19 PM

Here's the timeseries plot (with +/- 1 SD) for the same dataset:

Screenshot 2024-04-12 at 12 09 30 PM

handwerkerd (commented):
I've partially looked through this. One general issue I see is that you're passing the `t2s_limited` and `s0_limited` volumes. In those volumes, if there is only 1 good echo, the values are 0, but you're then fitting the data to 0's in `model_fit_decay_ts` We're outputting values for voxels with 2 good echoes, but, even then, they just won't be as accurate. If the goal is to get RMSE values then including voxels with known bad estimates will just increase error, but excluding those voxels will not represent the full volume. I don't think there is a clear right answer to this, but wanted to raise the point so we can choose an option with awareness of what we're choosing.

Also, adding a test for the new function will make it a bit easier to probe.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
     "scikit-learn>=0.21, <=1.4.2",
     "scipy>=1.2.0, <=1.13.0",
     "threadpoolctl",
+    "tqdm",
```
eurunuela, line 36: Do we really need `tqdm`? I just want to make sure we don't add a dependency that isn't necessary (not saying it isn't in this case).

handwerkerd: This is a baked-in dependency for robustica so I guess the idea is, once we start using it, we can use it. That said, the place it is used in the code is a really inefficient voxel-wise for loop. There's a way to make that for loop a lot more efficient (doing the calculation on multiple voxels for each number of echoes) so we should be able to remove the need to actually use tqdm in this PR.

tedana/workflows/tedana.py
```diff
@@ -649,6 +649,21 @@ def tedana_workflow(
             io_generator.save_file(utils.millisec2sec(t2s_limited), "limited t2star img")
             io_generator.save_file(s0_limited, "limited s0 img")
 
+    rmse_map, rmse_timeseries, rmse_sd_timeseries = decay.model_fit_decay_ts(
+        data=catd,
+        tes=tes,
+        adaptive_mask=masksum_denoise,
+        t2s=t2s_limited,
+        s0=s0_limited,
```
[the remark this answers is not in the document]
tsalo: That's a good point. I actually think it would be useful evaluate model fit when you provide your own T2* map, but that would require adding a step that estimates S0 from the data and T2*, which is just too much for this PR.

I think we'll need more advanced templating, like Jinja2, to handle missing figures. BTW I think that's already an issue with things like the T2* maps when rerunning tedana, so it's worth opening an issue about.

tedana/decay.py
```diff
@@ -460,3 +462,71 @@ def fit_decay_ts(data, tes, mask, adaptive_mask, fittype):
...
+        else:
+            s0_voxel = s0[i_voxel, :]
+            t2s_voxel = t2s[i_voxel, :]
+
+        predicted_data = monoexponential(
+            tes=tes[:n_good_echoes, None],
+            s0=s0_voxel[None, :],
+            t2star=t2s_voxel[None, :],
+        )
+        rmse[i_voxel, :] = np.sqrt(np.mean((data_voxel - predicted_data) ** 2, axis=0))
```
handwerkerd, line 527: This should be a more efficient solution that removes the need to use `tqdm`
I tried this on one dataset and `rmse` is identical for voxels with 2 or more good echoes. I decided to give no values (`nan`) for only 1 good echo since there shouldn't be any s0 and t2s estimate for those voxels and a fit to no model isn't informative. Note that this also required adding `import numpy.matlib` (Not sure why just numpy wasn't enough)

```suggestion
    rmse = np.full([n_samples, n_vols], np.nan, dtype=np.float32)
    # n_good_echoes interates from 2 through the number of echoes
    #   0 and 1 are excluded because there aren't T2* and S0 estimates
    #   for less than 2 voxels. 2 voxels will have a bad estimate so consider
    #   how/if we want to distinguish those
    for n_good_echoes in range(2, len(tes) + 1):
        # a boolean mask for voxels with a specific num of good echoes
        use_vox = adaptive_mask == n_good_echoes
        data_echo = data[use_vox, :n_good_echoes, :]
        if fitmode == "all":
            s0_echo = np.matlib.repmat(s0[use_vox].T, n_vols, 1).T
            t2s_echo = np.matlib.repmat(t2s[use_vox], n_vols, 1).T
        elif fitmode == "fit":
            s0_echo = s0[use_vox, :]
            t2s_echo = t2s[use_vox, :]
        # TODO add an else if fitmode is neither.
        # TODO Might also want to check if s0 and t2 are the right dimensions for fitmode option

        predicted_data = np.full([use_vox.sum(), n_good_echoes, n_vols], np.nan, dtype=np.float32)
        # Need to loop by echo since monoexponential can take either single vals for s0 and t2star
        #   or a single TE value. We could expand that func, but this is a functional solution
        for echo_num in range(n_good_echoes):
            predicted_data[:, echo_num, :] = monoexponential(
                tes=tes[echo_num],
                s0=s0_echo,
                t2star=t2s_echo,
            )
        rmse[use_vox, :] = np.sqrt(np.mean((data_echo - predicted_data) ** 2, axis=1))
```

tsalo:
The results are now substantially different:

image

handwerkerd:
> The results are now substantially different:
> 

I do expect the values to be non-trivially lower because you were including voxels with 1 good echo, which meant fit coefficients of 0 (i.e. the residual error would be the entire time series).

tsalo:
Ah, okay. That's a relief then.

eurunuela (approved):
LGTM!

tedana/reporting/data/html/report_body_template.html
```diff
@@ -200,6 +200,13 @@ <h2>S0</h2>
   <div class="carpet-plots-image">
     <img id="s0Histogram" src="$s0Histogram" style="height: 500px" />
   </div>
+  <h2>Model fit (RMSE)</h2>
```
handwerkerd, line 203:
```suggestion
  <h2>T2* and S0 model fit (RMSE). (Scaled between 2nd and 98th percentiles)</h2>
```

tedana/workflows/t2smap.py
```diff
@@ -7,6 +7,7 @@
 import sys
 
 import numpy as np
+import pandas as pd
```
handwerkerd, line 10:
```suggestion
```

tedana/workflows/t2smap.py
```diff
@@ -320,6 +317,22 @@ def t2smap_workflow(
...
+        t2s=t2s_limited,
+        s0=s0_limited,
+        fitmode=fitmode,
+    )
+    confounds_df = pd.DataFrame(
+        columns=["rmse", "rmse_std"],
+        data=np.column_stack((rmse_timeseries, rmse_sd_timeseries)),
+    )
+    io_generator.save_file(rmse_map, "rmse img")
+    io_generator.save_file(confounds_df, "confounds tsv")
```
handwerkerd, line 334:
```suggestion
    rmse_map, rmse_timeseries, rmse_df = decay.rmse_of_fit_decay_ts(
        data=catd,
        tes=tes,
        adaptive_mask=masksum,
        t2s=t2s_limited,
        s0=s0_limited,
        fitmode=fitmode,
    )
    io_generator.save_file(rmse_map, "rmse img")
    io_generator.save_file(rmse_df, "confounds tsv")
```

tedana/workflows/tedana.py
```diff
@@ -649,6 +649,22 @@ def tedana_workflow(
...
+            t2s=t2s_limited,
+            s0=s0_limited,
+            fitmode="all",
+        )
+        confounds_df = pd.DataFrame(
+            columns=["rmse", "rmse_std"],
+            data=np.column_stack((rmse_timeseries, rmse_sd_timeseries)),
+        )
+        io_generator.save_file(rmse_map, "rmse img")
+        io_generator.add_df_to_file(confounds_df, "confounds tsv")
```
handwerkerd, line 666:
```suggestion
        rmse_map, rmse_df = decay.rmse_of_fit_decay_ts(
            data=catd,
            tes=tes,
            adaptive_mask=masksum_denoise,
            t2s=t2s_limited,
            s0=s0_limited,
            fitmode="all",
        )
        io_generator.save_file(rmse_map, "rmse img")
        io_generator.add_df_to_file(rmse_df, "confounds tsv")
```

handwerkerd (changes requested):
I've made a bunch of suggested changes. The core of them is that I was looking at the mean & standard deviation time series in the report and didn't find the standard deviation informative, particularly since the low end was usually less than 0. I made a bunch of changes to store percentiles and plot the median and percentiles. To me showing the 2nd, 25th, median, 75th, and 98th percentiles is much more useful to understanding how RMSE varies over time & shows the extremes. For example, it make clear that the stdev was so large because the upper extrema can be very high. I also scaled the RMSE map to the 2nd-98th percentile. To make this change, it was easier to make the dataframe that is stored in the confounds file within `rmse_of_fit_decay_ts`

To highlight another comment below, the new function `add_df_to_file` had problems when used with the `--overwrite` option.

When this is all stabilized, a final example of the figures to the RTD documentation.

handwerkerd (commented):
`add_df_to_file` is not working when the model fit parameters and the global signal are both added to the confounds file.

This issue is visible in the output to the four echo integration test. The RMSE time series has an x-axis that is 2X the length of the time series. If you then look `four-echo/sub-01_desc-confounds_timeseries.tsv` it's clear what's happening. The columns are correct, but the added `global_signal` row is added as additional rows instead of as a new column in the existing rows.
```
rmse_mean	rmse_std	rmse_min	rmse_percentile02	rmse_percentile25	rmse_median	rmse_percentile75	rmse_percentile98	rmse_max	global_signal
186.89540100097656	260.73809814453125	0.6601158976554871	32.20176605224609	76.92352294921875	114.16072845458984	194.17686462402344	871.355974121094	11025.56640625	n/a
182.03968811035156	259.11785888671875	2.6228444576263428	30.978440475463863	74.24501037597656	109.76451873779295	186.53372192382812	865.50345703125	11128.96484375	n/a
183.68148803710935	258.3427734375	0.9386183619499208	31.656584701538087	75.49459838867188	111.54019165039062	189.2429656982422	868.1957470703127	11059.314453125	n/a
...
n/a	n/a	n/a	n/a	n/a	n/a	n/a	n/a	n/a	-0.5895119545327093
n/a	n/a	n/a	n/a	n/a	n/a	n/a	n/a	n/a	1.0792854501988267
n/a	n/a	n/a	n/a	n/a	n/a	n/a	n/a	n/a	-0.24254657799560972
```

tsalo:
That's fixed in my most recent commit.

tsalo:
@handwerkerd I think everything should be good now. Would you mind taking another look?

docs/outputs.rst
```diff
@@ -563,6 +565,26 @@ It is important to note that the histogram is limited from 0 to the 98th percent
...
+These plots show residual mean squared error (RMSE) values for the
+monoexponential decay model, based on the T2* and S0 maps.
+
+The first plot is the mean RMSE brain plot, which shows the mean RMSE over time for each voxel in the brain.
+
+The second plot is a time series of RMSE values across the brain, over time.
+This plot includes the median RMSE time series,
+along with an error band representing the 25th and 75th percentiles,
+and dotted lines indicating the minimum and maximum RMSE values.
+
```
handwerkerd, line 583:
```suggestion

The fit quality will vary depending on acquisition parameters and will likely be worse near signal drop-out areas. For a study with consistent acquisition parameters, relatively high RMSE values for runs or timepoints might be a marker of an underlying data quality issue.

```

docs/outputs.rst
```diff
@@ -563,6 +565,29 @@ It is important to note that the histogram is limited from 0 to the 98th percent
...
+The first plot is the mean RMSE brain plot, which shows the mean RMSE over time for each voxel in the brain.
+This plot is limited from the 2nd percentile to the 98th percentile.
+
+The second plot is a time series of RMSE values across the brain, over time.
+This plot includes the median RMSE time series,
+along with an error band representing the 25th and 75th percentiles,
+and dotted lines indicating the 2nd and 98th percentile RMSE values.
+
+The fit quality will vary depending on acquisition parameters and will likely be worse near signal drop-out areas. For a study with consistent acquisition parameters, relatively high RMSE values for runs or timepoints might be a marker of an underlying data quality issue.
```
tsalo, line 584:
```suggestion
The fit quality will vary depending on acquisition parameters and will likely be worse near signal drop-out areas.
For a study with consistent acquisition parameters,
relatively high RMSE values for runs or timepoints might be a marker of an underlying data quality issue.
```

handwerkerd (approved):
LGTM! Thanks for humoring all my suggested changes.

tsalo:
Thanks @handwerkerd!

Here's what the updated 3-echo timeseries plot looks like, for posterity.

Screenshot 2024-04-27 at 9 33 58 AM

PR #1045, opened 2024-02-21
Improve component plots

### LINKED ISSUE #1042: Check orientation of brain/acquisition in figures
check orientation of brain/acquisition when making multiplanar cuts - rotate to normal. 

Sometimes with R-L acquisitions the brain is flipped on its side, coronal isn't coronal, etc - That just adds a bit of frustration (and possible confusion) when inspecting output. May be an easy fix...

_Originally posted by @dowdlelt in https://github.com/ME-ICA/tedana/issues/383#issuecomment-535113411_

tsalo:
Closes #1042.

- Use `plot_stat_map` instead of identifying and plotting individual slices. This should address the potential orientation issues from #1042. It will also select the "best" slices for each component independently, so different components will have different slices rendered.

### DIFF
--- a/tedana/reporting/static_figures.py
+++ b/tedana/reporting/static_figures.py
@@ -268,7 +268,7 @@ def plot_component(
     ax_ts.set_yticks([])
 
     ax_freq = fig.add_subplot(gs[2])
-    ax_freq.plot(frequencies, y=power_spectrum, color=classification_color)
+    ax_freq.plot(frequencies, power_spectrum, color=classification_color)
     ax_freq.set_xlim(0, frequencies.max())
     ax_freq.set_yticks([])
 
--- a/tedana/reporting/static_figures.py
+++ b/tedana/reporting/static_figures.py
@@ -265,6 +265,7 @@ def plot_component(
     ax_ts = fig.add_subplot(gs[0])
     ax_ts.plot(component_timeseries, color=classification_color)
     ax_ts.set_xlim(0, len(component_timeseries) - 1)
+    ax_ts.set_yticks([])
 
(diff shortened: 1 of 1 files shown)


tsalo:
Argh, tight_layout doesn't help the line plots, but using `plot_stat_map` should fix any orientation issues.

tsalo:
Here's what an example component figure looks like:

image

handwerkerd (commented):
This looks much better. One thing I noticed is that the presented slices vary by component. For example, for the three-echo integration test, here are the images for components 15 & 51. The axial slice numbers are different. This is not great because we really want to compare the component maps between components, but I don't see an obvious way to fix this within `plot_stat_map`. Thoughts?
image
image

tsalo:
`plot_stat_map` will identify the "best" slices to present (probably the ones with the highest values, given a certain amount of space between slices). I think that would be better for getting a sense of the component's spatial pattern than being able to directly compare across components.

handwerkerd (approved):
LGTM

So that we remember this change, note that the presented slices may be different for each component in the opening comment for this PR.

tsalo:
Done!

dowdlelt (approved):
I'm sad to see the originals go - but happy with similarly oriented brains. 

I'm of two minds about the "best slices" approach - because it is nice to flip between components and see how the structure changes in the same area....but I'm convinced enough that it is better to have a better grasp of the individual components.

I think, based on what I've seen, that there is still enough information there to identify good/bad components, and perhaps it can do even better. Not worth holding up the overall fix, and potentially it could even improve that process.

Issue #679, opened 2021-02-11
Incorrect implementation and/or description of adaptive_mask


tsalo:
### Summary
I was reviewing [`make_adaptive_mask()`](https://tedana.readthedocs.io/en/0.0.9/generated/tedana.utils.make_adaptive_mask.html#tedana.utils.make_adaptive_mask) recently, and I noticed that the documentation and implementation don't line up. Specifically, the docstring implies that the adaptive_mask reflects the _last_ echo with good signal, while the implementation indicates that it's the _total number_ of echoes with good signal.

### Additional Detail

Here's the part that sums good echoes instead of identifying the last good one:

https://github.com/ME-ICA/tedana/blob/1a45f129c88e939eec3a5096b23b59068d382ac1/tedana/utils.py#L91-L93

When I directly checked which echoes had "bad data" in Logan's eight-echo data, according to the current implementation, the `adaptive_mask` generally had a value of 7 (indicating that the first 7 of 8 echoes are good), but the actual bad echoes were all either the first or second one!

If anyone has a good rationale for this behavior, please respond.

### Next Steps

1. Decide what we want from the adaptive mask:
    1. The first echo with bad data, minus one. I think this is the obvious interpretation of the adaptive mask, with the assumption (but not requirement) that echoes _after_ that will also be bad. However, at least in some datasets, it looks like this will eliminate a lot of voxels.
    2. The total number of echoes with good data. This is the current implementation, but IMHO doesn't make sense given how we use the mask.
    3. Which specific echoes have good data. To do this, we would need to reorganize the adaptive mask from an (S,) array of integers into an (S x E) array of booleans. Our actual masking procedure would also need to be updated. This proposal also comes up in #491. Personally, I think this is the best option.
2. Update either the documentation or implementation.

tsalo:
If we care about patterns of echoes, rather than the first X echoes, then we'll need to evaluate all subsets of echoes if we want to incorporate r-squared (i.e., #543), which is a lot to do. We'll need to figure out some shortcuts...

We'll also need to figure out the best way to incorporate our adaptive mask thresholds. For example, in a five echo dataset, if the fourth echo is the only good one, which _two_ echoes do we use for `t2smap`?

tsalo:
A summary from today's call:

[...]

PR #1057, opened 2024-03-10
Add `--masktype` option to control adaptive mask method

### LINKED ISSUE #312: Use overall signal decay in adaptive mask creation

tsalo:
Closes #312. I'd like to look into R-squared again at some point, but I think this is a solid improvement to the adaptive mask.

- Add a new CLI argument, `--masktype` to both tedana and t2smap. This parameter accepts one or more values, like `--gscontrol`. The valid values are "dropout" (the current method), "decay" (flag any echoes that do not decrease in mean value), and "none" (set adaptive mask value to all echoes, for testing and the like).
    - If you use more than one method, the function will just take the elementwise minimum value from all of the method-specific adaptive masks.
- Implement a "decay" method for the adaptive mask, which calculates an array reflecting the first echo where mean signal does not decrease from the last echo.
- Remove `getsum` parameter from `make_adaptive_mask`. We don't use `getsum=False` outside of tests, so we really don't need it.


tedana/utils.py
```diff
@@ -49,37 +49,53 @@ def reshape_niimg(data):
...
-        Boolean array of voxels that have sufficient signal in at least one
-        echo
+        Boolean array of voxels that have sufficient signal in at least ``threshold`` echos.
     masksum : (S,) :obj:`numpy.ndarray`
-        Valued array indicating the number of echos with sufficient signal in a
-        given voxel. Only returned if `getsum = True`
+        Valued array indicating the number of echos with sufficient signal in a given voxel.
+
+    Notes
+    -----
```
tsalo, line 74: Please double-check the accuracy of my description.

tedana/utils.py
```diff
@@ -108,6 +124,15 @@ def make_adaptive_mask(data, mask=None, getsum=False, threshold=1):
     # and count # of echos that pass criterion
     masksum = (np.abs(echo_means) > lthrs).sum(axis=-1)
 
+    # Determine where voxels stop decreasing in signal from echo to echo
```
tsalo, line 127: Please give feedback on the proposed method.

tsalo: Also, would it make sense to use this method _instead of_, rather than in addition to, the current method?

tedana/tests/test_utils.py
```diff
@@ -89,13 +87,11 @@ def test_make_adaptive_mask():
     # masksum has correct values
     vals, counts = np.unique(masksum, return_counts=True)
     assert np.allclose(vals, np.array([0, 1, 2, 3]))
-    assert np.allclose(counts, np.array([13564, 3977, 5060, 41749]))
+    assert np.allclose(counts, np.array([13564, 4959, 5349, 40478]))
```
handwerkerd, line 90: This test's data only has 5 volumes and the first might not be in steady state. Still, the changes in these numbers means nearly 1000 voxels dropped from 2-3 good echoes to 1 good echo in the mask. I'd expect later echoes to be higher in only 2 situations. (1) There's something wrong with the data or (2) The values are so close to the noise floor and variation across echoes is just noise.

It would be useful if several of us compared the current and new adaptive masks on real data to see if the additionally masked voxels are primarily in low signal areas.

tedana/utils.py
```diff
@@ -49,37 +49,58 @@ def reshape_niimg(data):
...
+            Basically, this identifies "exemplar" voxels reflecting the 33rd percentile.
+
+            -   The 33rd percentile is arbitrary.
+        c.  Calculate 1/3 of the mean value of the exemplar voxels for each echo.
+
+            -   This is the threshold for "good" data.
+            -   The 1/3 value is arbitrary.
+        d.  Only retain the highest value for each echo, across exemplar voxels.
+        e.  For each voxel, count the number of echoes that have a mean value greater than the
+            corresponding echo's threshold.
```
handwerkerd, line 91:
```suggestion
        a.  Calculate the 33rd percentile of values in the first echo,
            based on voxel-wise mean over time.
        b.  Identify the voxel where the first echo's mean value is equal to the 33rd percentile.
            Basically, this identifies "exemplar" voxel reflecting the 33rd percentile.

            -   The 33rd percentile is arbitrary.
            -   If more than one voxel equals the 33rd percentile, keep all of them.
        c.  Calculate 1/3 of the mean value of the exemplar voxel for each echo.

            -   This is the threshold for "good" data.
            -   The 1/3 value is arbitrary.
        d.  If there was more than one exemplar voxel, retain the the highest value for each echo.
        e.  For each voxel, count the number of echoes that have a mean value greater than the
            corresponding echo's threshold.
```

tedana/utils.py
```diff
@@ -108,6 +129,15 @@ def make_adaptive_mask(data, mask=None, getsum=False, threshold=1):
...
     masksum = (np.abs(echo_means) > lthrs).sum(axis=-1)
 
+    # Determine where voxels stop decreasing in signal from echo to echo
+    echo_diffs = np.hstack((np.full((n_samples, 1), -1), np.diff(echo_means, axis=1)))
+    diff_mask = echo_diffs >= 0  # flag where signal is not decreasing
+    last_decreasing_echo = diff_mask.argmax(axis=1)
+    last_decreasing_echo[last_decreasing_echo == 0] = n_echos  # if no increase, set to n_echos
+
+    # Retain the more conservative of the two adaptive mask estimates
+    masksum = np.minimum(masksum, last_decreasing_echo)
```
handwerkerd, line 139: Might want to make a second warning message if `last_decreasing_echo` removes nontrivially more voxels than the other threshold. I'm viewing that as a sign that something is wrong with the data.

tedana/utils.py
handwerkerd: The method looks fine. I'd keep both. Even if it's slightly arbitrary, the original method is an actual measure of dropout. The new method should only find voxels with substantial dropout and voxels where there's something wrong with the data.

handwerkerd (changes requested):
Mostly looks good. I do want several people to compare the masks on real data to make sure nothing unexpected is happening. This change really shouldn't shrink the masks much, so, if it does, that's concerning.

tedana/utils.py
tsalo: We still do need to update the original method to reflect the _last_ good echo (see https://github.com/ME-ICA/tedana/issues/679#issuecomment-1956766801).

tedana/tests/test_utils.py
tsalo: Does anyone have any multi-echo BOLD data along with high-quality quantitative T2* maps to test against? It would be awesome to see if these changes improve T2* estimation accuracy.

handwerkerd: I don't. I'm also not sure this would clearly improve accuracy. The best scenario would be voxels that were particularly inaccurate are more likely to be masked.

One thing we could test is stability. In theory, `t2smap` should give similar results for each run in a session. Does our more aggressive mask remove voxels that were less stable than others?

handwerkerd:
Thinking this over more: The code for these changes in simple, but this can non-trivially affect results in ways that are worth testing. One option to consider is that you tweak this PR to make it possible for an end-user to try the current adaptive mask, and several other options using `t2smap` (This could be hacky change to a couple of functions and the `t2smap` CLI/API, I could see either removing all of that or just the CLI/API option before merging.). It could also be tweaked to log the mask sizes & size differences with each option. Then it would be easy for several of us to run `t2smap` on a lot of data and compare the adaptive masks. 

Depending on timing we could attempt to either have results to discuss by next Thursday's dev call or prod people to try this at the dev call. Thoughts?

tedana/utils.py
tsalo: The diff here isn't clear. Could you summarize what changes you made?

Also, I'd like to expand on the logic behind this method. Does this sound right?

> This method assumes that exemplar voxels' signal at later echoes are also reasonable, and that any voxels whose values at a given echo are less than 1/3 of the exemplar voxels' values at that echo are affected by dropout.

tedana/tests/test_utils.py
tsalo: If voxels that have reached full dephasing at a given echo, then including that echo in the T2*/S0 estimation _should_ worsen model fit/accuracy, right? I think that should be our primary indicator if the adaptive masking procedure is working.

tedana/utils.py
tsalo: I should also probably make a note in the documentation that this procedure depends on the voxels included in the base mask- if your brain mask is bad, then it will impact your adaptive mask.

tedana/tests/test_utils.py
handwerkerd: I was thinking 3 echo. I agree this is true for >3 echoes. If we remove the 2nd or 3rd echo, then there is no T2* estimate to compare. To test this, we'd need to get multi-echo data with >=4 echoes and a quantitative T2* map to use as a comparison. I'm not sure if I've seen an open dataset with those properties. The closest I can think of is https://openneuro.org/datasets/ds003192/versions/1.0.1 but that has a T2 weighted FLAIR image and not a quantitative T2* map.

tedana/utils.py
handwerkerd: I mostly changed the 33rd percentiles voxel**S** to voxel and then added a bullet that, if more than one voxel is at exactly the 33rd percentile then keep all of them and pick the maximum of each echo below. Content-wise I think your original is correct, but I tried to clarify.

tsalo: Ah, thank you for the clarification. I will accept the change and add my summary of the logic.

tedana/utils.py
```diff
@@ -49,37 +49,70 @@ def reshape_niimg(data):
...
+    The adaptive mask is constructed from two methods:
+
+    1.  Count the total number of echoes in each voxel that have "good" data.
+        This method assumes that an exemplar voxel's signal at later echoes is also reasonable,
+        and that any voxels whose values at a given echo are less than 1/3 of the exemplar voxel's
+        values at that echo are affected by dropout.
+
+        This method uses distributions of values across the mask.
+        Therefore, it is sensitive to the quality of the mask;
+        a bad mask may result in a bad adaptive mask.
```
tsalo, line 84: Wait a second... it doesn't look like the adaptive mask is limited to the mask. That seems like a major issue, right?

tsalo: Most of the voxels in the data array are going to be outside the brain.

handwerkerd: um... You're right. If the function is given a mask, the percentile calculations should only be done within the mask. and the 33rd percentile would be unnecessarily low. I'd call this a bug.

This is removing voxels with a mean of 0, but, if the data are unmasked, it's not removing much. If there's a user-provided mask, it should be applied as part of this line.

https://github.com/ME-ICA/tedana/blob/a21d65ec5b1682a52cb8fe0eee1689d5eb3154a3/tedana/utils.py#L86

The rest of the function should be fine since the adaptive mask criteria can be applied to all voxels and masked later. It's only the percentile threshold that needs to be restricted by the mask.

handwerkerd: I strongly suspect this is also why we are seeing so many additional voxels being rejected by the new non-monotonic criterion vs the threshold. The percentile threshold is missing voxels outside of the brain and those are the ones flagged by the new threshold.

This means I'm less concerned about comparing old to new masks (and adding temporary options to compare different criteriea), but, given these are still arbitrary criteria, it would still be good to look at the revised adaptive masks on a bunch of data to make sure nothing undesired is happening.

tsalo: I'll open a separate issue about this then.

tsalo:
Based on our developer meeting a couple of weeks ago, I've added a `--masktype` parameter to control how the adaptive mask is generated. @handwerkerd could you take another look at some point?

tedana/utils.py
```diff
@@ -49,88 +49,144 @@ def reshape_niimg(data):
...
-    masksum : (S,) :obj:`numpy.ndarray`
-        Valued array indicating the number of echos with sufficient signal in a
-        given voxel. Only returned if `getsum = True`
+        Boolean array of voxels that have sufficient signal in at least ``threshold`` echos.
+    adaptive_mask : (S,) :obj:`numpy.ndarray`
+        Valued array indicating the number of echos with sufficient signal in a given voxel.
+
+    Notes
+    -----
+    The adaptive mask is constructed from two methods:
```
handwerkerd, line 78:
```suggestion
    The adaptive mask can remove voxels with two methods: dropout and decay.
    Both methods are applied to the mean magnitude across time for each voxel and echo.
```

tedana/utils.py
```diff
@@ -49,88 +49,144 @@ def reshape_niimg(data):
...
+
+    1.  Count the total number of echoes in each voxel that have "good" data.
+        This method assumes that an exemplar voxel's signal at later echoes is also reasonable,
+        and that any voxels whose values at a given echo are less than 1/3 of the exemplar voxel's
+        values at that echo are affected by dropout.
+
+        This method uses distributions of values across the mask.
+        Therefore, it is sensitive to the quality of the mask;
+        a bad mask may result in a bad adaptive mask.
```
handwerkerd, line 87:
```suggestion
    1.  Dropout
        Remove voxels with relatively low mean magnitudes from the mask.
        This method uses distributions of values across the mask.
        Therefore, it is sensitive to the quality of the mask.
        A bad mask may result in a bad adaptive mask.
```

tedana/utils.py
```diff
@@ -49,88 +49,144 @@ def reshape_niimg(data):
...
-    # determine samples where absolute value is greater than echo-specific thresholds
-    # and count # of echos that pass criterion
-    masksum = (np.abs(echo_means) > lthrs).sum(axis=-1)
+    assert methods, "No methods provided for adaptive mask generation."
+    assert all([method in ["decay", "dropout", "none"] for method in methods])
+
+    n_samples, n_echos, _ = data.shape
+    adaptive_masks = []
+
+    if "none" in methods:
```
handwerkerd, line 124: In playing with the `none` option, I got later steps crashing because of `NaN` values. Maybe for all options, including `none` we want to mask values that are `<=0` or `NaN`.

tedana/utils.py
```diff
@@ -49,88 +49,144 @@ def reshape_niimg(data):
...
-
-    # if multiple samples were extracted per echo, keep the one w/the highest signal
-    if lthrs.ndim > 1:
-        lthrs = lthrs[:, lthrs.sum(axis=0).argmax()]
-
-    # determine samples where absolute value is greater than echo-specific thresholds
-    # and count # of echos that pass criterion
-    masksum = (np.abs(echo_means) > lthrs).sum(axis=-1)
+    assert methods, "No methods provided for adaptive mask generation."
+    assert all([method in ["decay", "dropout", "none"] for method in methods])
```
handwerkerd, line 119:
```suggestion
    assert all([method.lower() in ["decay", "dropout", "none"] for method in methods])
```

This won't be an issue with the CLI where specific options are requires, but I'd rather the program not crash if someone capitalized one of these words

tedana/utils.py
```diff
@@ -49,88 +49,144 @@ def reshape_niimg(data):
...
+        echo_diffs = np.hstack((np.full((n_samples, 1), -1), np.diff(echo_means, axis=1)))
+        diff_mask = echo_diffs >= 0  # flag where signal is not decreasing
+        last_decreasing_echo = diff_mask.argmax(axis=1)
+        last_decreasing_echo[last_decreasing_echo == 0] = n_echos  # if no increase, set to n_echos
+        adaptive_masks.append(last_decreasing_echo)
+
+    # Retain the most conservative of the selected adaptive mask estimates
+    adaptive_mask = np.minimum.reduce(adaptive_masks)
 
     if mask is None:
```
handwerkerd, line 170: I decided to start with this masking PR, but it might be easier for me to review and have you merge #1060 first. Once #1060 is merged, this block of text would move nearer to before the different types of adaptive masks are calculated.

tedana/utils.py
```diff
@@ -49,88 +49,144 @@ def reshape_niimg(data):
...
+
+        a.  Calculate the 33rd percentile of values in the first echo,
+            based on voxel-wise mean over time.
+        b.  Identify the voxel where the first echo's mean value is equal to the 33rd percentile.
+            Basically, this identifies "exemplar" voxel reflecting the 33rd percentile.
+
+            -   The 33rd percentile is arbitrary.
+            -   If more than one voxel has a value exactly equal to the 33rd percentile,
+                keep all of them.
+        c.  Calculate 1/3 of the mean value of the exemplar voxel for each echo.
```
handwerkerd, line 99:
```suggestion
        c.  For the exemplar voxel from the first echo, calculate 1/3 of the mean value for each echo.
```

tedana/utils.py
```diff
@@ -49,88 +49,144 @@ def reshape_niimg(data):
...
+
+            -   This is the threshold for "good" data.
+            -   The 1/3 value is arbitrary.
+            -   If there was more than one exemplar voxel,
+                retain the the highest value for each echo.
+        d.  For each voxel, count the number of echoes that have a mean value greater than the
```
handwerkerd, line 105: This will change when $1061 is merged

tedana/utils.py
```diff
@@ -49,88 +49,144 @@ def reshape_niimg(data):
...
+            corresponding echo's threshold.
+    2.  Determine the echo at which the signal stops decreasing for each voxel.
+        If a voxel's signal stops decreasing as echo time increases, then we can infer that the
+        voxel has either fully dephased (i.e., "bottomed out") or been contaminated by noise.
+        This essentially identifies the last echo with "good" data.
```
handwerkerd, line 110:
```suggestion
        This essentially identifies the last echo with "good" data. For a scan that collects many echoes
        for T2* estimation or has a relatively short echo spacing, it is possible that a later echo will have
        a higher value, but the overall trend still shows a decay. This method should not be used in those
        situations.
```

handwerkerd (changes requested):
I have a bunch of comments, but, overall, this is good. The only major thing is that I don't think it will be possible to select `decay` and `dropout` through the CLI.

Also, after starting this, I realized it would be slightly nicer to merge #1060 first since you'll pretty much have to re-write #1060 to fit with these changes.

tedana/utils.py
tsalo: Great idea. I'll implement that.

tedana/utils.py
handwerkerd: See intro comment. Thinking this over again, it would be nicer to get the updated doc strings and function parameters that are in this PR changed and merged into the other masking PR.

tedana/utils.py
tsalo: True, but the docstring in #1061 reflects the changes I made there

tsalo:
I think I've addressed everything, and I updated the t2smap integration test to use both `dropout` and `decay` to ensure that the parser could handle multiple values.

tedana/utils.py
```diff
@@ -49,88 +49,159 @@ def reshape_niimg(data):
...
-    lthrs = np.squeeze(echo_means[perc_val].T) / 3
-
-    # if multiple samples were extracted per echo, keep the one w/the highest signal
-    if lthrs.ndim > 1:
-        lthrs = lthrs[:, lthrs.sum(axis=0).argmax()]
-
-    # determine samples where absolute value is greater than echo-specific thresholds
-    # and count # of echos that pass criterion
-    masksum = (np.abs(echo_means) > lthrs).sum(axis=-1)
+    assert methods, "No methods provided for adaptive mask generation."
```
handwerkerd, line 124:
```suggestion
    if (methods is None) or (len(methods) == 1 and methods[0].lower() == "none"):
        LGR.warning(
            "No methods provided for adaptive mask generation. "
            "Only removing voxels with negative or NaN values"
        )
        RepLGR.info(
            "An adaptive mask was then generated that retained echoes with negative or NaN values."
        )
    else:
        RepLGR.info(
            f"An adaptive mask was then generated using the {'+'.join(methods)} method(s), "
            "in which each voxel's value reflects the number of echoes with 'good' data."
        )
```
This wasn't letting me include the RepLGR line above, but it should be part of this suggestion as well. This is going to make sure the report text is parsed correctly if None is an input. If `methods` is None or if `none` is the only element in the `methods` list, there is one bit of output text and there's another text if `decay` and/or `dropout` are included. If someone decides to input `["decay", "dropout", "none"]` it will just ignore the `none`.

tedana/workflows/tedana.py
```diff
@@ -367,6 +378,8 @@ def tedana_workflow(
     prefix : :obj:`str` or None, optional
         Prefix for filenames generated.
         Default is ""
+    masktype : {'dropout', 'decay'} or :obj:`list`, optional
```
handwerkerd, line 381:
```suggestion
    masktype : :obj:`list` with 'dropout' and/or 'decay' or None, optional
```
I'm fairly sure the function requires a list, even for a single element so i reworded the docstring.

tedana/workflows/t2smap.py
```diff
@@ -176,6 +187,8 @@ def t2smap_workflow(
     mask : :obj:`str`, optional
         Binary mask of voxels to include in TE Dependent ANAlysis. Must be spatially
         aligned with `data`.
+    masktype : {'dropout', 'decay'} or :obj:`list`, optional
```
handwerkerd, line 190:
```suggestion
    masktype : :obj:`list` with 'dropout' and/or 'decay' or None, optional
```
I'm fairly sure the function requires a list, even for a single element so i reworded the docstring.

tedana/utils.py
handwerkerd: I think we'll want to log the voxel counts with each different method. I wrote some code for that, but it was a bit ugly and I think it might be easier to do as a separate PR once all current adaptive mask PRs are merged.

handwerkerd (changes requested):
I noticed a few more things, but this is getting close.

handwerkerd (approved):
LGTM!

PR #1060, opened 2024-03-15
Limit current adaptive mask method to brain mask

### LINKED ISSUE #1059: Adaptive mask is generated using whole acquisition matrix, instead of brain mask
### Summary

The adaptive mask identifies its exemplar voxel(s) from the 33rd percentile of the available data. Currently, the available data used include the whole acquisition matrix, but it _should_ be limited just to the brain mask.

### Additional Detail

Stems from https://github.com/ME-ICA/tedana/pull/1057#discussion_r1523611174.

### Next Steps

1. Test out current and proposed adaptive masking approaches on a range of datasets to see what effect it has.

tsalo:
Closes #1059. I built this off of #1057, so it shares many changes.

- Limit the adaptive mask to voxels in the brain mask.
- Make `mask` a required argument for `make_adaptive_mask`.
- Use `compute_epi_mask` to calculate the mask in the t2smap workflow if one isn't provided.


handwerkerd:
I haven't had a chance to do any tedana stuff today, but I'll try to get to this early next week. One quick comment. Since it's also sort-of a bug to you want to address https://github.com/ME-ICA/tedana/issues/679#issuecomment-1956766801 here and "Define good echo value in adaptive mask based on last good echo for the voxel, rather than the sum of good echoes."

tsalo:
I think I'd like to address that in a second PR so we can really pin down the effect of each change.

tedana/utils.py
```diff
@@ -49,39 +49,73 @@ def reshape_niimg(data):
...
+        "An adaptive mask was then generated, "
+        "in which each voxel's value reflects the number of echoes with 'good' data."
     )
+    mask = reshape_niimg(mask).astype(bool)
+    data = data[mask, :, :]
 
-    # take temporal mean of echos and extract non-zero values in first echo
+    # take temporal mean of echos and extract non-zero voxels in first echo
     echo_means = data.mean(axis=-1)  # temporal mean of echos
     first_echo = echo_means[echo_means[:, 0] != 0, 0]
```
tsalo, line 120: Why do we limit to non-zero voxels here?

tedana/utils.py
```diff
@@ -108,29 +142,21 @@ def make_adaptive_mask(data, mask=None, getsum=False, threshold=1):
     # and count # of echos that pass criterion
     masksum = (np.abs(echo_means) > lthrs).sum(axis=-1)
```
tsalo, line 143: I think we should raise an error if any echo means are negative, rather than taking the absolute value.

tedana/utils.py
handwerkerd: A truly zero voxel is one that was already masked out before this step so this is removing voxels with no data. I suspect this bug first arose because, at some point, the data were already masked before this step was run so that only voxels within the mask (i.e. in the brain) were used.

tedana/utils.py
handwerkerd: I don't think this should be an error, but I don't think we should take the absolute value either. At this point in the processing almost all values should be positive. It is possible that spatial interpolation will make values that are close to zero before less than zero. Those voxels should not be in the mask.

Rather than a error, I could see using `LGR.info` to store more information. At minimum, we should be logging the `lthrs` values for each echo. Maybe we also want to log the max & min values to give a sense of the range of values.

tedana/utils.py
tsalo: Thanks! I've expanded the comment a bit.

tedana/utils.py
tsalo: I started logging the lthrs, but not the max and min values.

eurunuela (dismissed):

handwerkerd:
Every time I look at the masking code, I notice something else that seems off. This time I think it's a function that is more flexible than documented (not an actual problem). I noticed that, for the masking code, [`reshape_niimg`](https://github.com/tsalo/tedana/blob/097d3a7eada0a96428effe9907c19fbe131e942e/tedana/utils.py#L29) would sometimes take as input a file name, sometimes a 3D image array, and sometimes a 1D image array. I thought this was wrong and would crash on some, but it seems to work for all those conditions.

If it's a file name, it loads the file and then reshapes and if it's already a loaded image, it just uses it. If it's already 1D, it squeezes the file and is still 1D.

Not sure whether we want to update that docstring in this PR, but I figured I'd mention.

tsalo:
@handwerkerd I've opened #1070 about the function.

eurunuela (dismissed):
LGTM! Let's remember to remove the TODO comment once the figure is added to the report.

handwerkerd:
I just checked how this affects results on one of my 3-echo datasets. I'm sharing the results here so that it's logged with this PR.

I ran 25 subjects with the current `main`, this PR with `masktype=["dropout"]` and this PR with `masktype=["dropout", "decay"]`. I used a user-provided mask for each run. This CSV file has the results [voxel_count_table.csv](https://github.com/ME-ICA/tedana/files/14962229/voxel_count_table.csv)


- The first three columns are ID, total voxels in the volume, and voxels in the user provided mask.
- The next three columns are how many voxels are lost from the adaptive mask  (i.e. no echo has good data) for `main` `PR dropout` and `PR dropout+decay`. `main` loses 0-5 voxels while this PR loses 2-147. As expected no extra voxels are lost for decay since the threshold is just the first echo.
- The next 3 columns the same except for voxels that do NOT have 2 good echoes
- The next 3 columns are the same except for voxels that do NOT have 3 good echoes
- The final 3 columns are also for 3 good echoes, but for % of voxels in the original mask. For this you can see the current `main loses less than 0.1% of the voxels (i.e. it was barely thresholding any data as bad). With this PR's fix 2-3% of voxels lost seems typical and sub-25 loses around 10% of voxels. Even for sub-25, less than 0.1% of voxels have a bad first echo so those voxels would be included in the optimally combined and denoised data, but they would not be included in the ICA or component selection steps.

Here are the maps for sub-25 (yellow is 3 good echoes, orange is 2 good echoes, and red is 1 good echo)
Main: 
sub25_main_drop
PR with dropout: 
sub25_PR1060_drop
PR with dropout+decay: 
sub25_PR1060_dd

As you can see the loses seem mostly in areas where dropout is common and ventricles.

This is a bug fix and the new results are what we've indended to do for a while, so I'm fine with these changes, but I wanted to document this and we'll want to clearly highlight this change in our next release notes.

handwerkerd:
Also, this turned out to be slightly more annoying to do than I expected because `t2smap` doesn't save the adaptive masked. I ended up pulling out the code used to create and save the adaptive mask from tedana. In case anyone else wants to play with this, here's my under-documented function:

```
import glob
import os
from nilearn.masking import compute_epi_mask
from tedana import (
    __version__,
    io,
    utils,
)

def running_adaptive_mask(data, n_echos, mask_file, masktype, out_dir, prefix):
    """ Pulling out the code from tedana.py to run and save the adaptive mask """

    catd, ref_img = io.load_data(data, n_echos=n_echos)

    io_generator = io.OutputGenerator(
        ref_img,
        out_dir=out_dir,
        prefix=prefix,
        config="auto",
        overwrite=True
    )

    first_echo_img = io.new_nii_like(io_generator.reference_img, catd[:, 0, :])
    mask = utils.reshape_niimg(mask_file)
    # Replace this line with the one above if you want to use the auto-generated mask from within tedana
    # mask = compute_epi_mask(first_echo_img) 

    mask_denoise, masksum_denoise = utils.make_adaptive_mask(
        catd,
        mask=mask,
        threshold=1,
        methods=masktype,
    )

    io_generator.save_file(masksum_denoise, "adaptive mask img")
```

handwerkerd (changes requested):
This is a TODO that suggested changing the automasking threshold from `data<0` to `data<=0`. We should either make this change here or remove the TODO.
https://github.com/tsalo/tedana/blob/3bc9ecbfca80122fe7bdf2c5bf87e3a0b8c279af/tedana/utils.py#L149-L150

Otherwise, I logged the type of differences we might expect after this PR is merged and this looks good to me.

handwerkerd (changes requested):
Looks like my really minor suggested change has altered the expected number of voxels from the adaptive mask in some of the unit tests.

handwerkerd (approved):
LGTM!

PR #1061, opened 2024-03-16
Identify the last good echo in adaptive mask instead of sum of good echoes

tsalo:
Closes https://github.com/ME-ICA/tedana/issues/679. I built this off of #1060 and #1057, so it shares many changes.

This will only ever _increase_ the adaptive mask value, as the last good echo will always be equal to, or higher than, the total number of good echoes.

Note that this does not identify the last _contiguous_ good echo (i.e., the last good echo before _any_ bad echoes). Therefore, if you have something like [good, bad, good, bad], the result with be 3, not 1.

- Identify the last good echo for the adaptive mask, instead of the total number of good echoes.


eurunuela (approved):
Changes look good to me.

tedana/tests/test_utils.py
```diff
@@ -81,72 +81,77 @@ def test_make_adaptive_mask():
...
-    assert mask.shape == masksum.shape == (64350,)
-    assert np.allclose(mask, (masksum >= 1).astype(bool))
+    assert mask.shape == adaptive_mask.shape == (64350,)
+    assert np.allclose(mask, (adaptive_mask >= 1).astype(bool))
     assert mask.sum() == 49376
-    vals, counts = np.unique(masksum, return_counts=True)
+    vals, counts = np.unique(adaptive_mask, return_counts=True)
     assert np.allclose(vals, np.array([0, 1, 2, 3]))
-    assert np.allclose(counts, np.array([14974, 3682, 5128, 40566]))
+    assert np.allclose(counts, np.array([14974, 1816, 4428, 43132]))
```
handwerkerd, line 96: I'm going to test this on more datasets tomorrow, but this is not how I'd expect the output values to change for this PR. I'd expect the value for 3 echos to remain constant since, if all three are in the mask, this change would never identify an earlier echo outside the mask. Instead, this looks like there are **more** voxels with 3 good echoes within this PR. This same direction of change happens with all the tests. Am I missing/confused about something?

tedana/utils.py
```diff
@@ -187,9 +190,17 @@ def make_adaptive_mask(data, mask, threshold=1, methods=["dropout"]):
...
+        dropout_adaptive_mask = np.zeros(n_samples, dtype=np.int16)
+        for i_voxel in range(n_samples):
+            echo_means_voxel = np.abs(echo_means[i_voxel, :])
+            echos_over_threshold = echo_means_voxel > lthrs
+            # Find the index of the last True element in a 1D boolean array
+            last_true_index = (
+                (np.where(echos_over_threshold)[0][-1] + 1) if np.any(echos_over_threshold) else 0
+            )
+            dropout_adaptive_mask[i_voxel] = last_true_index
+
```
handwerkerd, line 203:
```suggestion
        # Find the last good echo for each voxel
        dropout_adaptive_mask = np.zeros(n_samples, dtype=np.int16)
        for echo_idx in range(n_echos):
            # For voxels that were in the mask for the immediately previous echo
            # If they are still good in the current echo, increment the adaptive
            # mask value
            dropout_adaptive_mask[
                (dropout_adaptive_mask == (echo_idx)) * (echo_means[:, echo_idx] > lthrs[echo_idx])
            ] = (echo_idx + 1)
        adaptive_masks.append(dropout_adaptive_mask)

```

tsalo: I added simulated data to the adaptive mask tests in the last couple commits, and your change doesn't produce results that I'd expect.

I have one simulated voxel that is [mean of first echo, 1, mean of first echo], which _should_ give a decay-based adaptive mask value of 3 (i.e., the mean value of the first echo should definitely be good in the third echo), but your change gives a value of 1.

tedana/tests/test_utils.py
tsalo: I believe the change is from [bad, bad, good] (old value = 1, new value = 3), [bad, good, good] (old value = 2, new value = 3), and [good, bad, good] (old value = 2, new value = 3).

tedana/utils.py
handwerkerd: I think your test case is performing as I understand it should. Each echo has its own threshold. When the test is run `lthrs == [ 220.73333333, 79.73333333, 53.4  ]`
For your added test, the failing voxel is your `idx+1` where `data[idx+1,:,0] == [2442.976348096348, 1.0, 2442.976348096348]` That means the first and third echos are above threshold (`True`) but the second echo is below threshold (`False`). Since the second echo is bad, the last good echo is the first echo and the adaptive mask correctly equals 1.

tsalo: The last good echo in that case is the third echo, since 2442.976348096348 > 53.4. The descriptions from https://github.com/ME-ICA/tedana/issues/679#issuecomment-782423984 and https://github.com/ME-ICA/tedana/issues/679#issuecomment-1956766801 indicate that the last good echo should be used, rather than the last contiguous good echo.

handwerkerd: OK. Then this is a confusion on the specifications. I was thinking the "last good echo" was the one before the first bad echo while your earlier comments are saying to keep intermediate echoes if a later one is good under the assumption the threshold got something wrong in-between. That is:
```
good, good, good == 3
good, good, bad == 2
good, bad, good == (2 in current Main, 1 with my specification and 3 with your specification)
```
I think we agree that the current Main is wrong, I think either of our ideas is reasonable. I might want to dig a bit into real data to see when `good,bad,good` happens to see if the intermediate `bad` tends to be a sign that something is wrong with later echoes or if there are just cases where a middle echo is just below a threshold and we should bias towards using.

We also added my approach earlier for removing 0, negatives or Nan values. I think in that case, it makes sense to say, if we have `good, Nan, good` or `good, 0, good` the adapative mask should be 1 since the middle echo is unusable (Since we were talking past each other I wanted to make sure we both agree on that one)

https://github.com/ME-ICA/tedana/blob/682f8a61653af6a772d3f94f0b8c80f53c44bf93/tedana/utils.py#L151-L161

tsalo: > I think we agree that the current Main is wrong, I think either of our ideas is reasonable.

I agree that the sum is not the way to go, and I can see why using the last contiguous good echo makes the most sense (that was originally how I interpreted the adaptive mask method description), but I remember testing that approach and I ended up losing a lot of voxels. TBF, that might not be as much of an issue now that we have different for denoising and for ICA, but it's worth taking into account when we decide between the two proposed approaches here.

> We also added my approach earlier for removing 0, negatives or Nan values. I think in that case, it makes sense to say, if we have good, Nan, good or good, 0, good the adapative mask should be 1 since the middle echo is unusable

I also agree that having a zero or NaN in a later echo should limit the adaptive mask value. That's how it will currently work, since we take the voxel-wise minimum value from the different requested adaptive masks, and the "base" one will have a value of 1 here.

handwerkerd: I updated my above suggested code change to something that should match your specifications, but is a bit simpler to reach and avoids a for loop across voxels.

tedana/utils.py
```diff
@@ -103,9 +103,12 @@ def make_adaptive_mask(data, mask, threshold=1, methods=["dropout"]):
         -   This is the threshold for "good" data.
         -   The 1/3 value is arbitrary.
         -   If there was more than one exemplar voxel, retain the the highest value for each echo.
-    d.  For each voxel, count the number of echoes that have a mean value greater than the
+    d.  For each voxel, identify the last echo with a mean value greater than the
         corresponding echo's threshold.
 
+        -   Preceding echoes (including ones with mean values less than the threshold)
+            are considered "good" data.
```
handwerkerd, line 110:
```suggestion
        -   Preceding echoes (including ones with mean values less than the threshold)
            are considered "good" data. That means, if echoes 1-3 in a voxel are
            [good, good, bad] the adaptive mask will assign a 2, and if they are
            [good, bad, good], the adaptive mask will assign a 3.
```

tedana/tests/test_utils.py
```diff
@@ -80,73 +80,130 @@ def test_make_adaptive_mask():
...
+    assert adaptive_mask[idx + 1] == 3
+    assert adaptive_mask[idx + 2] == 1
+    assert adaptive_mask[idx + 3] == 3
+    assert adaptive_mask[idx + 4] == 0
+    assert adaptive_mask[idx + 5] == 0
+    assert mask.sum() == 49374
+    vals, counts = np.unique(adaptive_mask, return_counts=True)
     assert np.allclose(vals, np.array([0, 1, 2, 3]))
-    assert np.allclose(counts, np.array([14974, 3682, 5128, 40566]))
+    assert np.allclose(counts, np.array([14976, 1817, 4427, 43130]))
```
handwerkerd, line 124: I was trying to get a better handle with what was happening within this PR. In `make_adaptive_mask` I added `older_mask =(np.abs(echo_means) > lthrs).sum(axis=-1)` right after `dropout_adaptive_mask` was calculated and compared results in my debugger. Calls like `((older_mask==2) * (dropout_adaptive_mask==2)).sum()` show voxels that used to be 1 in the adaptive mask and are now 2.
Voxels that were 0 are still 0. (This interacts with other masking steps, but seems matched at this point in the code)
Of the 3590 voxels that were 1, 1061 are now 2, and 913 are now 3.
Of the voxels 5276 voxels that were 2, 1900 are now 3.
As expected, none of the voxels that had a higher values are now lower.

That means this change will substantively expand the number of voxels used in ICA and will balance out some of the drop caused by raising the threshold from masking. The one thing that concerns me is it turns how this test data had 913 voxels where the first and second echos were both below threshold, but the third was above threshold. That doesn't seem great, but the proposed enhancement would give this voxels a 0 in the adaptive mask, which would mean they'd be even be excluded from the optimally combined data. There's nothing to change here, but this is a discussion that will be relevant if a future enhancement is considered.

tsalo: I've opened #1083 about this. Can you follow up in that issue with your ratio idea?

Issue #704, opened 2021-03-20
Incorporate type hints when possible


tsalo:
### Summary
Should we start using type hints, when possible, in our functions?

I believe that it would reduce the amount of argument validation that we do at the beginning of each function. Plus it would make our code more transparent.

One drawback is that, in order to support type hints for numpy arrays (one of our most common argument types), we would probably need an additional dependency like [`nptyping `](https://github.com/ramonhagenaars/nptyping).

### Additional Detail
@jbteves has started using [type hints](https://docs.python.org/3/library/typing.html#module-typing) in #692, which I quite like. He also brought up style guidelines in this month's developers meeting (see #701), so I thought I'd get the ball rolling with this issue.

### Next Steps
1. Discuss
2. Possibly implement

jbteves:
Just FYI, the types aren't actually checked at runtime unless we add `mypy` as well. That's why you had to correct one of my type hints in that PR; if they were checked then I would have been getting runtime errors. We can add it if we think it's worth it, however. I included it in that PR mostly to improve documentation and also because frankly I've been using type hints in a different project and forgot we're not using them here 😆 . My view is that we should begin adding type hints in PRs as we modify functions, not do them in one big PR, and add `nptyping` as a dependency to make that possible. I need to experiment a bit with `mypy` and `npytyping` to decide if they're reliable enough to add type-checking to the whole codebase from my point of view.

Thanks for opening @tsalo !

tsalo:
> Just FYI, the types aren't actually checked at runtime unless we add `mypy` as well.

I did not know that. That's somewhat depressing and simultaneously pretty cool!

> My view is that we should begin adding type hints in PRs as we modify functions, not do them in one big PR, and add nptyping as a dependency to make that possible.

My concern about this approach (along with the same approach to reformatting with `black` or `isort`) is that there are a number of functions that are very stable, and we may not touch them (and thus get them up to snuff) for a long time. If we want to maintain a coding standard across the package, at some point we'll need to bite the bullet and apply our rules to _all_ of the code, or else we'll eventually have a bit of a Frankenstein's monster situation.

jbteves:
> That's somewhat depressing and simultaneously pretty cool!

Intriguingly that's my relationship with Python in a nutshell!

> My concern about this approach (along with the same approach to reformatting with black or isort) is that there are a number of functions that are very stable, and we may not touch them (and thus get them up to snuff) for a long time.

[...]

PR #1064, opened 2024-03-20
Generate metrics from external regressors using F stats

### LINKED ISSUE #1009: Add option to classify components based on their correlation with regressors

handwerkerd:
Closes #1009. This is an alternative approach to #1008 and #1021.

If a user provides external regressors, this will calculate a fit to those regressors to include as metrics in the component table and for use in decision trees.

- Create a new module, `metrics.external`, for calculating metrics from external regressors
  - An F statistic is used with a Legendre polynomial detrending baseline and p, F, and R2 values are saved as metrics in the component table
  - These stats are saved for the Full F model using all regressors as well as partial F models. This makes it possible to both examine when the full model fits, but to also log if it's the motion vs ROI-based vs cardiac/respiration-based regressors that fit to a specific component time series.
  - Regressor labels need to be specified, but can include regular expresses to use all regressors in a file
  - Multiple models can be included so that it is possible to create one model with nuisance regressors to reject and another with task design regressors to potentiall retain. This was a suggestion from @dowdlelt so that it would be possible to identify and conservatively retain components that fit to the overall task design.
- Add a new parameter to the tedana CLI, `--external`, to pass in a TSV with external regressors.
- Demo decision trees that incorporate external regressor correlation metrics. There are called `demo_` because they demonstrate how to use the new functionality. They might be similar to trees we'll want to recommend, but validation on actual data (and tweaking) will be easier when this is merged and more people are using it.
  - In the decision tree there is a new field `external_regressor_config` which is a list of dictionaries with the following keys:
    - `info` A description of what the config does that's saved
    - `report` A description to add to `report.txt`
    - `statistic` Currently the only option is "F" but this leaves open possibilities for additional options.
    - `detrend`: Will automatically calculate the number of polynomial detrending regressors if true, but can also be a number for the users to specify. If this is false, then will be set to 0 (mean removal) and log a warning.
    - `regressors`: A list of the regressor names to include
    - `partial_models` [optional] A list of dictionaries with titles and names for partial models, each model name is its own key and contains either a list of column labels or a regular express wildcard (i.e. `"Motion": ["^mot_.*$"]` means the Motion partial model will include any external regressor column label that begins with `mot_` and `"Motion": ["mot_x", "mot_y", "mot_z"]` specifies 3 specific column label names
  - `demo_external_regressors_motion_task_models.json` uses all the above options, uses the partial models to add a classification_tag, but not change results and retains components that correlate to the task (R2>0.5), have kappa>elbow regardless of what rho is.
  - `demo_external_regressors_single_model.json` uses the minimum number of options to run with external regressors.
- Includes tests that should cover all new code.
  - In adding testing, I started to use the `matching` parameter to confirm the correct errors are triggered and `caplog.text` to confirm expected warnings or info statements are logged.
  -   Also starting to add tests that check for specific numerical outputs.
- Documentation is updated to explain all this new functionality and outputs. This is primarily in `building_decision_trees.rst` and `included_decision_trees.rst`


handwerkerd:
I'd appreciate some reviewers for this: @tsalo @eurunuela @goodalse2019 @n-reddy @dowdlelt .
From my end, the only thing left to do is update the ReadTheDocs part of the documentation. I'm holding off on that until a couple of you agree that the overall structure makes sense.

tedana/resources/external_regressor_configs/Mot12_CSF.json
```diff
@@ -0,0 +1,27 @@
...
+        "Mot_Z",
+        "Mot_Pitch",
+        "Mot_Roll",
+        "Mot_Yaw",
+        "Mot_d1_X",
+        "Mot_d1_Y",
+        "Mot_d1_Z",
+        "Mot_d1_Pitch",
+        "Mot_d1_Roll",
+        "Mot_d1_Yaw"
```
tsalo, line 22: Are these the column names produced by AFNI?

handwerkerd: No. I think they were a tweak of the names you used in an earlier version. Fine changing, if you want to align naming with at least one program.

tedana/resources/external_regressor_configs/Mot12_CSF.json
```diff
@@ -0,0 +1,27 @@
+{
+    "regress_ID": "Motion12_CSF",
+    "info": "Contains 6 columns for motion, 6 for first derivatives of motion, and 1 for a CSF ROI time series.",
+    "detrend": true,
```
tsalo, line 4: I don't love the idea of including extra processing to external regressors within tedana. I think stuff like that should be done by the user, and possibly offloaded to `fmripost-tedana` eventually.

tedana/resources/external_regressor_configs/Mot12_CSF.json
tsalo: Awesome. I'll look up the fMRIPrep naming and propose a change soon then.

tedana/resources/decision_trees/demo_minimal_external_regressors_motion_task_models.json
```diff
@@ -0,0 +1,270 @@
...
+    "external_regressor_config": {
+        "regress_ID": "Fmodel",
+        "info": "Fits all external regressors to a single model using an F statistic",
+        "report": "Unspecified external nuissance regressors were fit to components using a linear model were rejected if they fit.",
+        "detrend": true,
+        "calc_stats": "F",
+        "f_stats_partial_models": ["Motion", "CSF"],
+        "Motion": ["^mot_.*$"],
+        "CSF": ["^csf.*$"],
+        "task_keep": ["^task.*$"]
```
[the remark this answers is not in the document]
handwerkerd: Are you referring to `"task_keep"` or `["^task.*$"]`?
I'm fine with the special key being `signal_keep` but I think `signal` is too general and confusing.
For this demo tree I'm fine if the regular expression is `["^signal.*$"]`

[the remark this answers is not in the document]
tsalo: I was referring to the regex pattern (i.e., `"^signal__.*$"`). I think `task_keep` makes sense.

tedana/resources/external_regressor_configs/Mot12_CSF.json
tsalo: Oh sorry, my bad. I misunderstood what detrend was doing.

tedana/resources/external_regressor_configs/Mot12_CSF.json
tsalo: Actually, are the configs still used in the current version of the PR?

handwerkerd: No! This info is now only used from the `decision_tree/demo...` files. That makes this an even easier fix.

tedana/resources/decision_trees/demo_minimal_external_regressors_motion_task_models.json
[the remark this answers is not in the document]
handwerkerd: Just updated this.

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
...
+            expected_regressor_names.update(tmp_partial_model_names)
+        if expected_regressor_names - regressor_names:
+            err_msg += (
+                "Inputed regressors in external_regressors do not include all expected "
+                "regressors in partial models or task\n"
+                "Expected regressors not in inputted regressors: "
+                f"{expected_regressor_names - regressor_names}\n"
+                f"Inputted Regressors: {regressor_names}"
+            )
+        if regressor_names - expected_regressor_names:
```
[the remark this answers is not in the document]
handwerkerd: That is the intended behavior. The idea is that, if you have a file with nuisance regressors and don't want to specify labels in the decision tree json, it just uses everything. It wouldn't be too hard to add another field, something like `noise_remove`, that could be used to explicitly specify which columns to use and the rest would be ignored.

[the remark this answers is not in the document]
tsalo: Thanks for clarifying. I'd rather include all regressors than add another parameter in the config file, so I'm happy with your current approach.

tedana/metrics/_utils.py
```diff
@@ -8,6 +8,29 @@
 LGR = logging.getLogger("GENERAL")
 
 
+def add_external_dependencies(dependency_config, external_regressor_config):
```
tsalo, line 11: Related to #704- can you add type annotations to all of the new or heavily modified functions?

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
+"""Metrics unrelated to TE-(in)dependence."""
```
tsalo, line 1:
```suggestion
"""Metrics based on external confounds."""
```

I can foresee adding metrics unrelated to TE dependence that don't rely on external confounds (e.g., kurtosis of the component time series, spatial distribution of the component weights).

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
...
+    Parameters
+    ----------
+    external_regressors : :obj:`pandas.DataFrame`
+        Each column is a labelled regressor and the number of rows should
+        match the number of timepoints in the fMRI time series
+    external_regressor_config : :obj:`dict`
+        Information describing the external regressors and
+        method to use for fitting and statistical tests
+    n_time : :obj:`int`
+        The number of time point in the fMRI time series
```
tsalo, line 78: We generally use `n_vols` for this in the rest of the package.

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
...
+        )
+        detrend_regressors = make_detrend_regressors(n_time, polort=0)
+
+    if external_regressor_config["calc_stats"].lower() == "f":
+        # external_regressors = pd.concat([external_regressors, detrend_regressors])
+        comptable = fit_mixing_to_regressors(
+            comptable, external_regressors, external_regressor_config, mixing, detrend_regressors
+        )
+    else:
+        # This should already be valided by this point, but keeping the catch clause here
```
tsalo, line 225:
```suggestion
        # This should already be validated by this point, but keeping the catch clause here
```

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
...
+        polort = int(1 + np.floor(n_time / 150))
+
+    # create polynomial detrending regressors -> each additive term leads
+    # to more points of transformation [curves]
+    detrend_regressors = np.zeros((n_time, polort))
+    # create polynomial detrended to the power of 0 [1's],
+    # **1 [linear trend -> f(x) = a + bx],
+    # **2 [quadratic trend -> f(x) = a + bx + cx²],
+    # **3 [cubic trend -> f(x) = f(x) = a + bx + cx² + dx³],
+    # **4 [quintic trend -> f(x) = a + bx + cx² + dx³ + ex⁴]
```
tsalo, line 266:
```suggestion
    # **4 [quartic trend -> f(x) = a + bx + cx² + dx³ + ex⁴]
```

tsalo (commented):
I went through and added comments. The method looks solid to me. I have a few thoughts on style though:

1. There's a lot of commented vestigial code.
2. The docstrings should have type annotations, per #704. I would also recommend requiring parameter named in the function calls (i.e., by putting a leading `*` before the parameters).
3. Some variable names do not match the rest of the codebase (e.g., you use `n_time` instead of `n_vols`).
4. It would be great if strings and lines in docstrings were broken on punctuation. This will result in cleaner diffs in the future.
5. We're going to need extensive tests to cover the new code.

tedana/metrics/_utils.py
```diff
@@ -8,6 +8,29 @@
...
+    """
+    Add dependency information in external regressors are inputted.
+
+    TODO Add a docstring
+    """
```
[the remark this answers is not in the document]
handwerkerd: Added in my last commit

tedana/metrics/collect.py
```diff
@@ -594,6 +635,19 @@ def get_metadata(comptable):
             },
         }
 
+    if any(col.endswith("_correlation") for col in comptable.columns):
```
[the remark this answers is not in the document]
handwerkerd: Removed in my last commit

tedana/metrics/external.py
handwerkerd: Changed to """Metrics based on fits of component time series to external time series.""" in my last commit. I didn't want to use the word "confounds" since this can also take task-locked time series.

tedana/metrics/external.py
handwerkerd: I changed `n_time` to `n_vols` everywhere in this file.

handwerkerd:
@tsalo I just added type hints to `external.py` I'm still a bit of a novice with type hints and it looks like formatting is non-trivially changing across python versions. I tried to follow guidance that's valid for python 3.8 since that's our oldest permitted version. I'll keep working in other files and on your remaining review comments, but let me know if I'm appropriately adding type hints so far. 

Update as I was writing this comment: Lots of tests just failed so the answer to that question is"no". It looks numpy v1.24.3 that's currently in my python 3.8 environment includes `numpy.typing` and not `numpy._typing` while numpy v1.26.3 in my python 3.11 environment includes `numpy._typing` and not `numpy.typing`. I suspect there's a way to set up backwards compatibility for this, but I wanted to share what I'm seeing before I switch to childcare stuff.

tedana/metrics/_utils.py
handwerkerd: All functions in _utils.py now have type hints.

handwerkerd:
@tsalo Can you explain more what you mean by "It would be great if strings and lines in docstrings were broken on punctuation. This will result in cleaner diffs in the future."

tsalo:
Absolutely. Take a paragraph in the text, like the one below:

```
This is sentence one. This is
sentence two. This is sentence
three- but it's still going.
```

Since sentences are a unit within the paragraph,  we're more likely to change them than random words.
Plus, since almost everything we write is in markdown or rst, the newlines don't impact the rendered text- just the code.

```diff
-This is sentence one. This is
+This is sentence one. This is a 
-sentence two. This is sentence
+new sentence two with other
-three- but it's still going.
+changes. This is sentence three-
+but it's still going.
```

The diff on that, where the only consideration is line length, is much more extensive (and harder to interpret for a reviewer) than the diff if we broke up the text on punctuation:

```diff
 This is sentence one.
-This is sentence two.
+This is a new sentence 
+two with other changes.
 This is sentence three-
 but it's still going.
```

tedana/selection/selection_utils.py
```diff
@@ -313,6 +317,103 @@ def clean_dataframe(component_table):
...
+    if not regex_found:
+        regex_found, out_nodes = expand_dict(node, "kwargs", metrics)
+
+    if not regex_found:
+        # Stop early and just return the node if no regular expressions were found
+        out_nodes = [copy.deepcopy(node)]
+        return out_nodes
+
+    real_out_nodes = []
+    for out_node in out_nodes:
```
handwerkerd, line 389: @tsalo I'm working on improving documentation and type hints and I'm realizing I'm confused with part of what you did here. `expand_node` passes a single `node` dictionary to `expand_node`. `expand_node` takes two fields in this dictionary (parameters and kwargs) and sends them to `expand_dict` to identify and expand any regular expressions. After this is done, `expand_node` then recurses if any regular expressions were found. I'm having trouble seeing situations where this recursion would actually happen. I vaguely remember an earlier version where recursion did the expansion, but maybe this isn't needed anymore?

There also aren't tests for these two functions so I need to better understand when recursion should happen to write a test.

tedana/selection/selection_utils.py
```diff
@@ -1,9 +1,14 @@
...
 
+import copy
 import logging
+import re
+from typing import Dict, List, Tuple, Union
 
 import numpy as np
+import pandas as pd
 
+# from tedana.selection.component_selector import ComponentSelector
```
handwerkerd, line 11: @tsalo This is a point where I'm having trouble with type hints. There are several functions that take the ComponentSelector object as a parameter or return it. I need to import ComponentSelector to have it as an option for a type hint, but I can't because ComponentSelector uses functions in this file and creates a circularity. There seem to be type hint aliases and other mechanisms in python 3.8, but I'm not sure exactly how to use them.

I've also had to import pandas in a few places solely to allow `pd.DataFrame` and `pd.DataSeries` as type hints. I'm not sure that the correct way to do this.

tedana/selection/selection_utils.py
tsalo: The idea was to use the regular expression to find a list of columns, and then to run the correlation separately for each of them, but since the list of columns is fed into a single F-test now, I don't think this step is necessary anymore.

tedana/selection/selection_utils.py
tsalo: Oof that's tough. I guess you can't have a type hint for those functions. Can you just note that in the docstrings?

>I've also had to import pandas in a few places solely to allow pd.DataFrame and pd.DataSeries as type hints. I'm not sure that the correct way to do this.

I think that's fine. Pandas is going to be imported when running tedana anyway, so I don't think it costs us anything to import it in certain files just for type hints.

tedana/metrics/external.py
tsalo: Thanks!

tedana/selection/selection_utils.py
```diff
@@ -313,19 +320,137 @@ def clean_dataframe(component_table):
...
+        List of metric names to compare regular expressions against.
+
+    Returns
+    -------
+    out_nodes or expanded_out_nodes: :obj:`list[dict]`
+        A list of dictionaries with regular expressions replaced by all
+        matching values in the 'metrics' list.
+    """
+    regex_found, out_nodes = expand_dict(node, "parameters", metrics)
+    if not regex_found:
```
handwerkerd, line 390: Should this conditional be removed? `expand_dict` is first run for "parameters". Shouldn't it also be run for "kwargs" whether or not regular expressed were expanded for "parameters"

tsalo: Maybe I'm misremembering what it does, but it seems like `expand_dict` could be removed entirely.

handwerkerd: I'm looking into this more and I don't think we're using `expand_nodes`, `expand_node`, or `expand_dict` anymore. They are designed if someone uses a regular expression as a parameter for a `selection_node` function. I think you used that in a new selection_node function, but that part of the code was removed when we removed correlations. I don't think this is possible for any of our current selection node functions since I don't think any of them can takes more than one metric in a single parameter field. I ran integration tests without these functions, and it was fine.

I'm expanding the regular expressions I am using in [`validate_extern_regress`](https://github.com/ME-ICA/tedana/blob/c26b86d97438a78f7bab7e3d99d9cfb594f8251f/tedana/metrics/external.py#L62).

tsalo: It sounds like you can completely remove those functions then. I'm totally fine with that.

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
...
+    external_regressor_config = validate_extern_regress(
+        external_regressors, external_regressor_config, n_time
+    )
+
+    return external_regressors, external_regressor_config
+
+
+def validate_extern_regress(external_regressors, external_regressor_config, n_time):
+    """
+    Confirm that provided external regressor dictionary is valid and matches data.
```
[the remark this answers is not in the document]
handwerkerd: Making this docstring more accurate rather than accepting your change.

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
...
+    # **4 [quintic trend -> f(x) = a + bx + cx² + dx³ + ex⁴]
+    for idx in range(polort):
+        # create a linear space with numbers in range [-1,1] because the mean = 0,
+        # and include the number of timepoints for each regressor
+        tmp = np.linspace(-1, 1, num=n_time) ** idx
+        if idx == 0:
+            detrend_regressors[:, idx] = tmp
+            detrend_labels = ["polort0"]
+        else:
+            # detrend the regressors by z-scoring the data (zero-mean-centered & stdev of 1)
```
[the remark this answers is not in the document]
handwerkerd: z-scoring will demean the regressors (equivalent to polort=0) and remove mean differences across echoes, but won't detrend with polort>0

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
...
+        The number of polynomial regressors to create (i.e. 3 is x^0, x^1, x^2)
+        If None, then this is set to 1+floor(n_time/150)
+
+    Returns
+    -------
+    detrend_regressors: (n_time x polort) :obj:`pandas.DataFrame`
+        Dataframe containing the detrending regressor time series
+        x^0 = 1. All other regressors are zscored so that they have
+        a mean of 0 and a stdev of 1.
+        Dataframe column labels are polort0 - polort{polort-1}
```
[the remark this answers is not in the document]
handwerkerd: This is where it happens in gscontrol_raw

https://github.com/ME-ICA/tedana/blob/325011e3c65cc6f0ea43d1c4d9b8e0367297e091/tedana/gscontrol.py#L66-L69

It's using Legendre polynomials while I'm just using a + bx + cx² + dx³ + ex⁴. Changing would involve agreeing on one or the other and moving code out of here and `gscontrol_raw` I'd prefer just keeping this in two places, but could be convinced otherwise.

tedana/metrics/external.py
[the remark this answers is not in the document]
tsalo: I think the comment needs to be updated then.

tedana/metrics/external.py
```diff
@@ -0,0 +1,650 @@
...
+            "calc_stats for external regressors in decision tree is "
+            f"{external_regressor_config['calc_stats'].lower()}, which is not valid."
+        )
+
+    return comptable
+
+
+def make_detrend_regressors(n_time, polort=None):
+    """
+    Create polynomial detrending regressors to use for removing slow drifts from data.
```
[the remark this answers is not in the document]
tsalo: I'm find keeping them separate for now, but I'd like to open an issue to push us to defend one vs. the other (or at least have a good reason to use different approaches for the two methods).

tedana/resources/decision_trees/demo_minimal_external_regressors_motion_task_models.json
```diff
@@ -0,0 +1,271 @@
...
+    "external_regressor_config": {
+        "regress_ID": "Fmodel",
+        "info": "Fits all external regressors to a single model using an F statistic",
+        "report": "Unspecified external nuisance regressors that were fit to components using a linear model were rejected if they fit.",
+        "detrend": true,
+        "calc_stats": "F",
+        "f_stats_partial_models": ["Motion", "CSF"],
+        "Motion": ["^mot_.*$"],
+        "CSF": ["^csf.*$"],
+        "task_keep": ["^signal.*$"]
```
tsalo, line 28: Our meeting today got me thinking that I'd rather see all models specifically defined, rather than using a full model and a special field (`task_keep`).

Something like this:

```suggestion
        "f_stats_partial_models": ["Full", "Motion", "CSF", "Task"],
        "Full": ["(?!signal).*$"],
        "Motion": ["^mot_.*$"],
        "CSF": ["^csf.*$"],
        "Task": ["^signal.*$"]
```

tsalo: Also, instead of having separate lookup and definitions, what about having the models in a dictionary?

```suggestion
        "models": {
            "Full": ["^(?!signal).*$"],
            "Motion": ["^mot_.*$"],
            "CSF": ["^csf.*$"],
            "Task": ["^signal.*$"]
        },
```

handwerkerd: I think we'd still need special key words because this might obscure what actually happening with the partial models.
`Full` is all nuissance regressors. and the F test for `Full` is all nuissance regressors compared to the baseline detrending regressors.
The partial F statistics are the Full model compared to the Full model **excluding** the partial regressors of interest. That is the partial F statistic for motion uses all regressors except the motion regressors and the stat is effectively testing if the full model with the motion regressors models the data significantly better than the full model excluding motion regressors.
With the ways you've set this up `Full` would still need to be a special key work to define the regressors in the full model.
Similary, `task_keep` is not a partial F stat compared to the full model. It's a completely separate F stat, so just including it in a dictionary of models would lose that distinction.
I'm fine changing to a `models` dictionary, but we'd still need to predefine `Full` and `Task` as special keywords.

We also previously discussed letting users specify which regressors are part of the full model. You said that's an unncessary feature to add. I'm happy to add if you now think it would be useful.

handwerkerd:
I think I've now addressed all feedback regarding code and I've started to add documentation to `building_decision_trees.rst` I still need to make flow charts for the new trees and add them to `included_decision_trees.rst` but otherwise this PR is (again) ready for review & for people to try it out.

tedana/resources/decision_trees/demo_minimal_external_regressors_motion_task_models.json
tsalo: >The partial F statistics are the Full model compared to the Full model excluding the partial regressors of interest. That is the partial F statistic for motion uses all regressors except the motion regressors and the stat is effectively testing if the full model with the motion regressors models the data significantly better than the full model excluding motion regressors.

I forgot that. Thanks for clarifying. Is there any particular benefit to evaluating partial F-stats instead of running full models for each of the sets of regressors?

>We also previously discussed letting users specify which regressors are part of the full model. You said that's an unncessary feature to add. I'm happy to add if you now think it would be useful.

If we're running separate full models for different sets of regressors (which is essentially what we're doing with nuisance vs. task regressors), it does seem like we should accept patterns for the different full models. I'm sorry for going back on what I originally thought on this.

I don't want to hold this PR up any further, so we can discuss this in a future PR or release, but what about something like this? 

```json
"external_regressor_config": [
    {
        "regress_ID": "F_nuisance",
        "info": "Fits all non-task regressors to a single model using an F statistic",
        "report": "Unspecified external nuisance regressors that were fit to components using a linear model were rejected if they fit.",
        "detrend": true,
        "statistics": ["F", "p", "R2"],
        "regressors": ["^(?!signal).*$"],
        "partial_models": {
            "Motion": ["^mot_.*$"],
            "CSF": ["^csf.*$"]
        }
    },
    {
        "regress_ID": "F_task",
        "info": "Fits all task regressors to a single model using an F statistic",
        "report": "External task regressors that were fit to components using a linear model were accepted if they fit.",
        "detrend": true,
        "statistics": ["F", "p", "R2"],
        "regressors": ["^signal.*$"]
    }
]
```

handwerkerd: There are several benefits of partial models. (In the docstring I'm citing https://afni.nimh.nih.gov/pub/dist/doc/manual/3dDeconvolve.pdf which has more of an explanation of page 14). For our needs, the key benefit is that we get a full fit of all nuisance regressors so that, if a combination is significantly fit with a combination of motion and respiration regressors (but not either separately) it will be identified. Also, dependence between different groups of regressors in accounted for in a single model. That is, if head motion and a CSF ROI both contain some of the same effects, the total variance explained is calculated in a single model rather than calculated twice in 2 separate models. This is an oversimplifications, but if Motion R^2=0.3 and a separate CSF R^2=0.3, that might imply these account for a combined variance of R^2=0.6 (oversimplification!), but the R^2 for a combined model would only be 0.4.

The other general benefit of partial models is that it's a single statistical test, which is mathematically more principled. That said, we're not using these for rigorous statistical thresholds, so I doubt this actually matters to us.

I do like your proposed redesign of the configuration. I think it won't be too much work, it will make the code and documentation cleaner, and will give users more flexibility in submitting external regressor files. It's worth doing. That said, to print before I travel, I need to have a finished OHBM poster by Monday, so there's no way I'm doing it before then. Even without this change, I'm doubtful we'll get this PR merged by Monday, so I'd rather get this right than set an artificial deadline that rushes review of a non-trivial code change.

For your proposed specification, the two things I'd change are:
- `"statistics"` to `"statistic"` and just include `"F"` Selecting `F` will calculate `p` and `R2` based on the F stats. Particularly since this is our only option now, adding flexibility for a range of output options just adds complexity.
- Change the `regress_ID`s to `"nuissance"` and `"task"`. That would make it slightly easier to then have the metrics outputs be `nuissance F` `nuissance R2` `nuissance p` without the repeated `F_`. at the beginning of each one.

handwerkerd: @tsalo I still need to update the RTD documentation, but this change is now fully implemented.

handwerkerd:
I think/hope this is fully done and ready for review. It's important to have people look at the code, but I'd also like to have people read the documentation (`tedana --help` https://tedana--1064.org.readthedocs.build/en/1064/building_decision_trees.html#external-regressor-configuration & https://tedana--1064.org.readthedocs.build/en/1064/included_decision_trees.html#demo-external-regressors-single-model) and try to run this to see if it both makes sense and runs as expected. (@tsalo @eurunuela @n-reddy @goodalse2019 @dowdlelt )

docs/_static/decision_tree_demo_external_regressors_motion_task_models.tex
```diff
@@ -0,0 +1,122 @@
...
+% ----- node: 8
+\node(8)[decision, below of=7,label={180:$node\ 8$}, yshift=-1.5cm]{$\kappa \geq \kappa$ elbow\\$\rho$ $<$ $\rho$ elbow};
+\node(chrej8)[changeclass, below of=8, xshift=0cm, yshift=-2cm]{Provisional reject};
+\node(chacc8)[changeclass, right of=8, xshift=3cm, yshift=0cm]{Provisional accept};
+% ----- node: 9
+\node(9)[decision, below of=chrej8,label={180:$node\ 9$},label={20: Accept even if $\rho < \rho\ elbow$},yshift=-1.5cm]{$\kappa > 2\rho$\\$\kappa \geq \kappa$ elbow};
+\node(chrej9)[changeclass, below of=9, xshift=0cm, yshift=-2cm]{Provisional reject};
+\node(chacc9)[changeclass, right of=9, xshift=3cm, yshift=0cm]{Provisional accept};
+% ----- node: 10
+\node(10)[decision, below of=chacc9,label={150:$node\ 10$},label={[align=center] 310: Reject if\\fits external\\nuissance\\regressors},yshift=-2cm]{F test for\\Nuisance Regressors\\$p_{Full} \leq 0.05$\\$R^2_{Full} \geq 0.5$};
```
tsalo, line 71:
```suggestion
\node(10)[decision, below of=chacc9,label={150:$node\ 10$},label={[align=center] 310: Reject if\\fits external\\nuisance\\regressors},yshift=-2cm]{F test for\\Nuisance Regressors\\$p_{Full} \leq 0.05$\\$R^2_{Full} \geq 0.5$};
```

docs/included_decision_trees.rst
```diff
@@ -18,7 +18,18 @@ and comprehensible, but it has not yet be extensively validated and
...
 
-Flowcharts describing the steps in both trees are below.
+With the addition of options to fit external regressors to components,
+there are two demonstration decision trees that implement this new functionality.
+While these might work well, since they have not yet been validated on data, they
+are labeled ``demo``.
+``decision_tree_demo_external_regressors_single_model``
+demonstrates fitting all nuissance regressors to a single model.
+``decision_tree_demo_external_regressors_motion_task_models``
+demonstrates fitting nuissance regressors to a model,
```
tsalo, line 28:
```suggestion
``decision_tree_demo_external_regressors_single_model``
demonstrates fitting all nuisance regressors to a single model.
``decision_tree_demo_external_regressors_motion_task_models``
demonstrates fitting nuisance regressors to a model,
```

tedana/resources/decision_trees/demo_external_regressors_motion_task_models.json
```diff
@@ -0,0 +1,293 @@
...
+        "Low variance",
+        "External regressors",
+        "Fits motion external regressors",
+        "Fits CSF external regressors",
+        "Fits task"
+    ],
+    "external_regressor_config": [
+        {
+            "regress_ID": "nuisance",
+            "info": "Fits all external nuissance regressors to a single model using an F statistic",
```
tsalo, line 34:
```suggestion
            "info": "Fits all external nuisance regressors to a single model using an F statistic",
```

tedana/resources/decision_trees/demo_external_regressors_single_model.json
```diff
@@ -0,0 +1,205 @@
...
+        "variance explained",
+        "pval nuisance model",
+        "R2stat nuisance model"
+    ],
+    "intermediate_classifications": ["provisionalaccept", "provisionalreject"],
+    "classification_tags": ["Likely BOLD", "Unlikely BOLD", "Low variance", "External regressors"],
+    "external_regressor_config": [
+        {
+            "regress_ID": "nuisance",
+            "info": "Fits all external nuissance regressors to a single model using an F statistic",
```
tsalo, line 22:
```suggestion
            "info": "Fits all external nuisance regressors to a single model using an F statistic",
```

tedana/tests/test_component_selector.py
```diff
@@ -56,6 +61,36 @@ def dicts_to_test(treechoice):
         "necessary_metrics": ["kappa", "rho"],
         "intermediate_classifications": ["random1"],
         "classification_tags": ["Random1"],
+        "external_regressor_config": [
+            {
+                "regress_ID": "nuisance",
+                "info": (
+                    "Fits all external nuissance regressors to "
```
tsalo, line 68:
```suggestion
                    "Fits all external nuisance regressors to "
```

tsalo (commented):
I just noticed that nuisance was misspelled as nuissance in a few places.

tsalo (approved):
I really don't want to hold this PR up any more. I say we merge as-is, make a release-candidate release, and deal with any bugs or typos before the next actual release.

handwerkerd:
> I really don't want to hold this PR up any more. I say we merge as-is, make a release-candidate release, and deal with any bugs or typos before the next actual release.

I'd still like to have @n-reddy and/or @goodalse2019 at least look over the documentation and run before merging. It would be good to see this understandable and executable by someone who hasn't yet look deep into the code. If they have time in the next week or so, I'm happy to wait before merging.

tsalo:
In that case, can we commit to merging on a specific date (e.g., next Friday- 08/09) if they don't provide feedback by then?

goodalse2019:
@handwerkerd and @tsalo I will have finished looked over the documentation and run by tomorrow evening. I have tested it on one subject very basic to make sure it ran and all looked like it went well! I am going to try on a couple others expanding my regressors tsv file a bit more. For testing purposes - is there a lower end of # of subjects you prefer this to be run over (e.g., at least 10?) or will just a couple do? Also - is there a good way to provide a overview of results for people to look over just to see the tests? Like uploading pdf or ppt in mattermost with some of the output figures?

handwerkerd:
I'm at a conference 8/6-9 so 8/12 would be a good day to merge if we don't get feedback requiring changes. (Earlier if we get feedback saying everything is fine.

@goodalse2019 At this point, we're not claiming the results will be better or worse, so I'd just like to see this functioning as expected on typical data. If you have more than one dataset sitting around, then trying this on 2-3 subjects with different acquisition parameters and making sure it runs as expected would be fine.
Seeing the results will be useful for the next step of trying to figure out what's actually beneficial. If it's not too large, I think you can upload a pdf as a comment on this PR, which will mean we'd be able to find it when continuing work on these methods. Thank you.

goodalse2019:
Sounds good, we have two - I can test on a couple subject from both and will upload some results asap

goodalse2019:
I've run the single model demo tree and attached are some output summaries after running on n = 2 in two me datasets:

Dataset 1A - single model 
[Dataset1_singlemodel.pdf](https://github.com/user-attachments/files/16462076/Dataset1_singlemodel.pdf)
Dataset 2 - single model
[Dataset2_singlemodel.pdf](https://github.com/user-attachments/files/16462078/Dataset2_singlemodel.pdf)

Still working on the task_motion_models demo - dataset 1 has a task and I am trying to get that to work but first few attempts were no good on my end. I can run over a few more subs in each if you think helpful and update these documents. I will be out of town in a wedding 8/2-8/4 so won't be until 8/5 that I can pick up again.

tsalo:
Thanks @goodalse2019. I think the results look reasonable, and the fact that it ran through is good news. @handwerkerd are you comfortable merging given her results?

n-reddy:
I reviewed the documentation and it makes sense to me! I just made a few small edits for clarity. Still planning to test the code if I get a chance in the next couple days.

handwerkerd:
Thank you all. I'm comfortable merging, but I'm currently traveling through Tuesday and wouldn't mind giving it one more quick check myself.  If others are comfortable merging, then give approving reviews.

tsalo (approved):

goodalse2019 (approved):
Looks good to me, single model and documentation ran successfully and was straight forward IMO.

n-reddy:
Just an update that both the single model and the task_motion_models worked well on two datasets I tested with different parameters! A small comment perhaps for future tweaks of the demo .json is that the classification tags on the task_motion_models can be a little confusing. For example, see in the pdfs linked below that the highlighted component was tagged as 'Likely BOLD, External Regressors, Unlikely BOLD' in the task_motion_model. 

Minimal model: [minimal.pdf](https://github.com/user-attachments/files/16501397/minimal.pdf)

External regressors - Motion parameters only (single model): [external_regressors_motion.pdf](https://github.com/user-attachments/files/16501398/external_regressors_motion.pdf)

External regressors - Motion parameters and task regressors: [external_regressors_motionTask.pdf](https://github.com/user-attachments/files/16501400/external_regressors_motionTask.pdf)

PR #1074, opened 2024-04-14
Update figure-generating notebook

tsalo:
Closes none. I figure that we've made a bunch of changes to the codebase that should be reflected in the documentation (e.g., the adaptive mask).

- Update the walkthrough notebook to follow the current output filename convention.
- Update the denoising approach page's figures.


docs/_static/a01_echo_timeseries.png
tsalo, line 1: I found a voxel that actually has a more interesting signal development across echoes than the old one, which was really just the checkerboard signal across all of the echoes.

docs/_static/a03a_adaptive_mask_distributions.png
tsalo, line 1: I thought this might be interesting to show the adaptive mask.

eurunuela (approved):
LGTM!

docs/approach.rst
```diff
@@ -344,57 +365,58 @@ The blue and red lines show the predicted values for the :math:`S_0` and
...
-datasets in order to remove spatially diffuse (ostensibly respiration-related)
-noise.
-Methods which have been employed in the past include global signal
-regression (GSR), minimum image regression (MIR), anatomical CompCor, Go Decomposition (GODEC), and
-robust PCA.
+:func:`tedana.gscontrol.gscontrol_raw`, :func:`tedana.gscontrol.minimum_image_regression`
+
+Due to the constraints of spatial ICA,
+TEDICA is able to identify and remove spatially localized noise components,
+but it cannot identify components that are spread out throughout the whole brain.
```
handwerkerd, line 406: I've seen this logic before, but it still makes no sense to me. ICA complete ARE spatially diffuse and ICA components can also be minimally clustered. In fact, the lack of clustering is one of our metrics to reject components. Additionally, there is little evidence besides one replicated paper that GSR or the other methods actually address this issue.
If it were up to me, I'd cut this entire section, but I'll defer if others disagree.

handwerkerd (changes requested):
This looks good overall. 

I’m having trouble putting comments directly into `plot_approach_figures.ipynb` and I can’t comment on figures so I’m writing them here:
- For loading the dataset, you have a hard coded local path. I don’t think this needs to be automatic, but it would be better to tell people where they can get the dataset and then return to a `dset_dir` variable where they can set the local location of their data. If the 8-echo dataset isn’t currently living online anywhere, we can put it here: https://osf.io/bpe8h/
- For `a03_adaptive_mask.png` either add a label for the color bar (i.e. “# of good echoes”) or add a sentence explaining the colorbar to the text in `approach.rst`
- `a13_ica_component_timeseries.png` has a slow drift component that has a relatively high kappa vs rho and is titled as “accepted”. This feels wrong. I don’t see anything obviously wrong in the notebook, but make sure I’m not missing anything and consider picking a better exemplar ICA time series. It likely is accepted because the “accepted” time series lower down on the page still contain a slow drift.

The opposite of a problem: `a05_loglinear_regression.png` shows the fit of the echoes to a line. The newer sample voxel has less variance per echo and, with 8 echoes, you can see the deviations from linearity

docs/approach.rst
```diff
@@ -181,9 +199,10 @@ This optimally combined data is written out as **desc-optcom_bold.nii.gz**
 .. note::
     An alternative method for optimal combination that
     does not use :math:`T_{2}^*` is the parallel-acquired inhomogeneity
-    desensitized (PAID) ME-fMRI combination method (`Poser et al., 2006`_).
-    This method specifically assumes that noise in the acquired echoes is "isotopic and
-    homogeneous throughout the image," meaning it should be used on smoothed data.
+    desensitized (PAID) ME-fMRI combination method :footcite:p:`poser2006bold`.
```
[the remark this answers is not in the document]
tsalo: Unfortunately it would require a new footcite formatter, which is possible if we write a new function, but would be tough.

docs/approach.rst
tsalo: It's a point that comes up in most spatial vs. temporal ICA papers I've read. I haven't directly tested it or anything, but I haven't seen anyone argue against it in a way I could cite.

tsalo:
> For loading the dataset, you have a hard coded local path. I don’t think this needs to be automatic, but it would be better to tell people where they can get the dataset and then return to a dset_dir variable where they can set the local location of their data.

I believe @dowdlelt didn't want to openly publish those data. @dowdlelt please correct me if I'm wrong. Even if we can't share the data, I'd still prefer to stick with the 8-echo data given how illustrative it is of signal decay patterns.

dowdlelt:
Yeah, I think that has to stay as it is for the moment. But perhaps I can try and obtain something new with better permissions in time. I'll add it to the list.

docs/_static/a13_ica_component_timeseries.png
tsalo, line 1: Continuing from PR comments:

>a13_ica_component_timeseries.png has a slow drift component that has a relatively high kappa vs rho and is titled as “accepted”. This feels wrong. I don’t see anything obviously wrong in the notebook, but make sure I’m not missing anything and consider picking a better exemplar ICA time series. It likely is accepted because the “accepted” time series lower down on the page still contain a slow drift.

@handwerkerd, re-running with the current version didn't change the classification. This is the component with the highest variance explained (83.78% of the variance), which I think is a good reason to select it, but I can choose the component with the second highest varex if you don't want to show this one.

I think the fact that the component is being accepted speaks to a weakness (or even built-in limitation) in tedana, which I actually would prefer to highlight. If you think it's more an edge case that should normally be rejected, what part of the decision tree would you expect to flag it? The kappa value is ~3x the rho value, so I don't think the TE-dependence models would capture it.

PR #1080, opened 2024-04-17
[DOC] desc-optcomDenoised -> desc-denoised

mvdoc:
I'm going through the docs to refresh my understanding of ME-ICA, and noticed that the output file name was not updated there.


handwerkerd (approved):
Thank you for noticing and fixing this.

PR #1086, opened 2024-04-21
Refactor `gscontrol` module

### LINKED ISSUE #1085: Minimum image regression relies on distinction between "accepted" and "ignored" components
### Summary

I'm working on refactoring the global signal regression functions in order to better understand them (particularly for #1071), and I noticed that `gscontrol.minimum_image_regression` specifically uses "accepted" components and not "ignored" components. Since we eliminated the "ignored" classification in #756, MIR likely doesn't work the way it was intended anymore.

We could probably leverage the classification tags, but those vary by decision tree and are essentially up to the user.

tsalo:
Closes #1085 (but perhaps not optimally?)

- Add type hints to `gscontrol.gscontrol_raw` and `gscontrol.minimum_image_regression`.
- Use the same variable names from metric calculation within these two functions as well. E.g., `data_cat` instead of `catd`.
- Require function calls to use keyword arguments.
- Add `verbose` parameter to `gscontrol.minimum_image_regression`. Only write out the accepted MIR outputs if `verbose=True`.
    - Note that these are only composed of non-low-variance components at the moment.
- In `minimum_image_regression`, use classification tags to split up accepted components into BOLD-like and low-variance groups, to match the old "accepted" and "ignored" classifications.


tedana/gscontrol.py
```diff
@@ -38,93 +45,106 @@ def gscontrol_raw(catd, optcom, n_echos, io_generator, dtrank=4):
...
-    dm_catd : (S x E x T) array_like
-        Input `catd` with global signal removed from time series
-    dm_optcom : (S x T) array_like
-        Input `optcom` with global signal removed from time series
+    data_cat_nogs : (S x E x T) array_like
+        Input ``data_cat`` with global signal removed from time series.
+    data_optcom_nogs : (S x T) array_like
+        Input ``data_optcom`` with global signal removed from time series.
     """
     LGR.info("Applying amplitude-based T1 equilibration correction")
```
tsalo, line 53: Is this accurate?

tedana/gscontrol.py
```diff
@@ -135,88 +155,91 @@ def minimum_image_regression(optcom_ts, mmix, mask, comptable, io_generator):
...
-    mehk_ts = np.dot(comp_pes[:, acc], mmix[:, acc].T)
+    mehk_ts = np.dot(comp_pes[:, acc], mixing[:, acc].T)
     t1_map = mehk_ts.min(axis=-1)  # map of T1-like effect
     t1_map -= t1_map.mean()
     io_generator.save_file(utils.unmask(t1_map, mask), "t1 like img")
     t1_map = t1_map[:, np.newaxis]
 
     # Find the global signal based on the T1-like effect
-    glob_sig = np.linalg.lstsq(t1_map, optcom_z, rcond=None)[0]
+    gs_ts = np.linalg.lstsq(t1_map, data_optcom_z, rcond=None)[0]
```
tsalo, line 216: I wonder, does it make sense to write out this time series to the confounds file I'm adding in #1044? @handwerkerd WDYT?

tsalo: I decided to tack it onto the confounds file.

tedana/gscontrol.py
handwerkerd: I didn't write the original function, but this looks fine to me.

handwerkerd (approved):

tedana/gscontrol.py
```diff
@@ -7,29 +7,36 @@
...
 LGR = logging.getLogger("GENERAL")
 RepLGR = logging.getLogger("REPORT")
 
 
-def gscontrol_raw(catd, optcom, n_echos, io_generator, dtrank=4):
-    """Remove global signal from individual echo ``catd`` and ``optcom`` time series.
+def gscontrol_raw(
+    *,
+    data_cat: np.ndarray,
+    data_optcom: np.ndarray,
```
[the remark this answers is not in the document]
tsalo: Going with `data_optcom` based on #1088.

PR #1089, opened 2024-04-22
Include intercept in `get_coeffs` calls

tsalo:
Closes none. I was just thinking that we should include an intercept term when calculating echo-wise betas.

-
-


tsalo:
Weirdly, including the intercept didn't change _anything_. I'm not sure why. All of the other components should have mean values of zero, right? So how could raw echo-wise time series (i.e., with different scales) have the same OLS fit with and without an intercept term?

Issue #1054, opened 2024-02-29
Detrending & tedana (or detrended variance explained)


handwerkerd:
### Summary

I was talking with @afni-rickr and he suggested detrending and possibly regressing out motion parameters before tedana. This would reduce the amount of variance tedana would need to model and potentially make it perform better.

### Additional Detail

In thinking this over, I think regressing motion would be problematic because we'd need to regress from all echoes if motion artifacts didn't all follow the regressors, it might not actually save degrees-of-freedom or improve other results. I think there's a better case for detrending. I'm not too worried about high variance trends making tedana perform worse, but they do make it harder to interpret total variance explained and accepted/rejected variance explained since the magnitude of the linear drift dominates total variance explained. This make it hard to see whether rejected components are 30% of the meaningful total variance in one population vs 50% in another.

We might want to test of tedana results are substantively different if we detrend first. If not, one way to address the above issue would be to add a new metric `detrended variance explained`. I'd need to think through the math, but we'd detrend each component's time series and scale each components "variance explained" by how much detrending reduced variance. IF we want to get extra-fancy, we could show the variance explained pie chart and component time series with or without detrending.

handwerkerd:
My top priorities are to get realignment parameters into the decision tree (#1021) and get a more stable component estimation (likely #1013). Unless someone else gets to it first, running tedana on detrended data should be fairly easy. One would need to run polynomial detrending on all echoes, but keep the mean. Without any code edits, tedana could be run with and without this detrending to see how it alters the eventual denoised time series.

afni-rickr:
Note that detrending or other regression should happen in the context of censoring, at least if time gets squeezed.  If modularity is important, censoring with spike regressors might be preferable to time squeezing.
For detrending, I would think that would still distort the relationship between the echoes.  So it would seem like detrending or other regression should happen to the OC input to ICA, rather than to the individual echoes.  Though again, I am not sure what happens after that.

CesarCaballeroGaudes:
I agree that detrending should be done to the OC input

handwerkerd:
To make the metric fits work, you'd need to detrend the separate echoes (but retain the mean). To calculate kappa & rho, the component time series are fit to each echo's data. If the OC is detrended, but the echos aren't then those calculations would definitely break down. That said, we might be losing or corrupting some echo-dependent information if we separately detrend each echo.

[...]

PR #1090, opened 2024-04-27
DNM: Detrend optimally combined data before running PCA

tsalo:
Related to #1054.

- Instead of just z-scoring the optimally combined data before running PCA, detrend it as well.

### DIFF
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -206,7 +206,7 @@ def tedpca(
     )
     data = data_oc[mask, :]
     data_z = standardize_signal(data.T, detrend=True, standardize="zscore_sample").T
-    # data_z = (data_z - data_z.mean()) / data_z.std()  # var normalize everything
+    data_z = (data_z - data_z.mean()) / data_z.std()  # var normalize everything
 
     if algorithm in ["mdl", "aic", "kic"]:
         data_img = io.new_nii_like(io_generator.reference_img, utils.unmask(data_z, mask))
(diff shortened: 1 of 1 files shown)


tedana/decomposition/pca.py
```diff
@@ -204,12 +205,11 @@ def tedpca(
         f"Computing PCA of optimally combined multi-echo data with selection criteria: {algorithm}"
     )
     data = data_oc[mask, :]
-
-    data_z = ((data.T - data.T.mean(axis=0)) / data.T.std(axis=0)).T  # var normalize ts
-    data_z = (data_z - data_z.mean()) / data_z.std()  # var normalize everything
+    data_z = standardize_signal(data.T, detrend=True, standardize="zscore_sample").T
+    # data_z = (data_z - data_z.mean()) / data_z.std()  # var normalize everything
```
tsalo, line 209: I'm just not sure what the point of z-scoring based on the overall mean and SD does.

tedana/decomposition/pca.py
```diff
@@ -204,12 +205,11 @@ def tedpca(
...
 
     if algorithm in ["mdl", "aic", "kic"]:
-        data_img = io.new_nii_like(io_generator.reference_img, utils.unmask(data, mask))
+        data_img = io.new_nii_like(io_generator.reference_img, utils.unmask(data_z, mask))
```
tsalo, line 212: Passing the detrended and z-scored data into MAPCA as well. Will this cause problems?

tsalo:
- The three-echo PCA changes from 69 to 68 components, and the ICA (which converged before on the first seed) fails to converge
- The four-echo test looks exactly the same, but that could be because it has GSR enabled.
- The five-echo PCA changes from 52 to 54 components, and the ICA (which converged before on the first seed) fails to converge

PR #1094, opened 2024-05-06
minimum nilearn to 0.10.3

handwerkerd:
@pmolfese notified me of a crash when he ran tedana with nilearn v0.10.2:

```
File ".../python3.11/site-packages/nilearn/plotting/img_plotting.py", line 77, in _get_colorbar_and_data_ranges
    raise ValueError('this function does not accept a "vmin" '
ValueError: this function does not accept a "vmin" argument, as it uses a symmetrical range defined via the vmax argument. To threshold the plotted map, use the "threshold" argument
```

In `static_figures.py` we are calling `nilearn.plotting.plot_stat_map` with the `vmin` option. Going through the nilearn releases, this was only added in v0.10.3  ( https://github.com/nilearn/nilearn/pull/3993 ). I really don't like setting our minimum nilearn version to something that was only released in January 2024. I'd welcome to code changes that address this issue and retain slightly older compatibility, but I'm not sure if there's a way to get the images we want to include in the reports using older versions.

Given this will cause crashes, it would be nice to get a solution merged before releasing our v24.0.1.

- Raises the minimum nilearn version to 0.10.3

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "mapca>=0.0.4,<=0.0.5",
     "matplotlib",
     "nibabel>=2.5.1,<=5.2.1",
-    "nilearn>=0.7,<=0.10.4",
+    "nilearn>=0.10.3,<=0.10.4",
     "numpy>=1.16,<=1.26.4",
     "pandas>=2.0,<=2.2.2",
     "pybtex",


eurunuela (approved):

tsalo (approved):
LGTM. We could probably just add an exception for the problematic version, but I'm fine with doing it this way too.

handwerkerd:
> LGTM. We could probably just add an exception for the problematic version, but I'm fine with doing it this way too.

https://github.com/nilearn/nilearn/issues/3084 Makes it sound like `plot_stat_map` didn't previously have a `vmin` option. If it did, and this is just a blip for a few versions, than I'd definitely prefer to just exclude the bad versions.

> On a related note, I think we should add a job to test using the oldest-supported versions.

Is there a way to add this into the python 3.8 unit tests? That is, have CI run once with all minimally accepted versions. FWIW I'm not sure how much this matters, but nilearn requires `numpy>=1.19.0` and I'm not sure if there are any other dependences that are also pushing up our minimal accepted versions.

handwerkerd:
I just checked and `plot_stat_maps` did not have a `vmin` option in v0.10.0 or v0.9.0 so I'm fairly sure we're using a new feature and cannot just exclude a few versions. Since it's approved, I'm going to merge this PR.

PR #1098, opened 2024-05-21
Use nearest-neighbors interpolation in `plot_component`

### LINKED ISSUE #1095: Display of IC components has a grey outline in report
### Summary
Following on from a post in neurostars. I used fmriprep,  and then tedana with my own mask. The IC components in the output figures  have a grey, brainshaped outline.  The adaptive mask plot looks fine.  It was suggested on neurostars that this may be minor bug in Nilearn’s plot_stat_map function.  The issues is not present in 23.0.2, but is present in 24.0.1.  Components are also different between versions, but that may be down to something else. 
tedana_difference

adaptive_mask

tsalo:
Closes #1095.

- Switch from the default 'continuous' resampling approach to 'nearest' in `plot_component`'s `plot_stat_map` call. This should stop non-diagonal images from being resampled to have low, non-zero values outside of the brain.

### DIFF
--- a/tedana/reporting/static_figures.py
+++ b/tedana/reporting/static_figures.py
@@ -235,6 +235,7 @@ def plot_component(
         colorbar=False,
         draw_cross=False,
         annotate=False,
+        resampling_interpolation="nearest",
     )
     display.annotate(size=30)
     example_ax = list(display.axes.values())[0]
@@ -601,6 +602,7 @@ def plot_t2star_and_s0(
         vmax=t2s_p98,
         annotate=False,
         output_file=os.path.join(io_generator.out_dir, "figures", t2star_plot),
+        resampling_interpolation="nearest",
     )
 
     s0_plot = f"{io_generator.prefix}s0_brain.svg"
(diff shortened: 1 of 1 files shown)


tsalo:
From Logan's 8-echo dataset.

Before:

image

After:

image

handwerkerd (commented):
For the three times `vmin` is set to the 2nd percentile, you would not need to set `resampling_interpolation="nearest"` and "continuous" will result in nicer looking images.
For the one situation where `vmin` isn't set, setting `vmin=np.min(stat_img.get_fdata())` might give slightly nicer visualizations, but setting interpolation to "nearest" is also fine.

tsalo:
Good point. I dropped the change for `plot_stat_map` calls outside of `plot_component`.

handwerkerd (approved):
LGTM

I also realized my suggestion for adding `vmin` to stat_map figures is a bit more complex, since there are positive & negative values. It is solvable, but this solution is fine.

tsalo:
Yeah we'd need to use `threshold` instead of `vmin`, but I played around with that and had some issues, so I think this is fine for now. If we end up overlaying our stat maps on other images (e.g., mean OC, T1w) then we might need to revisit.

tsalo:
I'd like to make a release with this change before we merging #1064. @handwerkerd any objections?

PR #1103, opened 2024-06-14
Filter out non-diagonal affine warning

### LINKED ISSUE #1092: Repeated nilearn warning message for report
### Summary

For certain anatomical images, nilearn is outputting the same warning message repeatedly (over 50 times in a run). It would be good to see if there is a way to prevent or suppress this warning.

### Additional Detail

- @mrneont  shared this with me and might be able to provide more details
- This is likely a result of #1045 which switched to using `plot_stat_map` with `nilearn` to generate brain images for the report
- The message is `/usr/local/apps/tedana/conda/envs/24.0.0/lib/python3.9/site-packages/nilearn/plotting/displays/_slicers.py:1949: UserWarning: A non-diagonal affine is found in the given image. Reordering the image to get diagonal affine for finding cuts in the slices.
  coords[direction] = find_cut_slices(`
- It looks like the warning his about how it is handling a specific situation instead of an actual problem, but I asked them to confirm the output in the report looks correct
- @mrneont thinks the issue is that nilearn is expecting data in `RAS+` orientation (right, anterior, superior), but these data were in `ASL` (anterior, superior, left) and it might not be handling the interaction between that and obliquity. If this is the case, the Q is whether we can just find a way to suppress the warning or if this is also an issue that might need to get fixed within `nilearn`

### Next Steps

- Just documenting for now.
- Hopefully someone who knows `nilearn` plotting a bit better can assess if this is easily addressable or a more complex issue

tsalo:
Closes #1092.

- Filter out the "non-diagonal affine" warning when calling `plot_stat_map`.

### DIFF
--- a/tedana/reporting/static_figures.py
+++ b/tedana/reporting/static_figures.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+import warnings
 from io import BytesIO
 
 import matplotlib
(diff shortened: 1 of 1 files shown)


tsalo:
None of the integration tests typically raise this warning, so I don't know if it's working based on that.

handwerkerd (changes requested):
I've been testing it and I'm still getting the "non-diagonal affine" warning. Looking through various help menus, it looks like `warnings.catch` then `warnings.filterwarnings` might be designed for a warning in the current code and you might need to use the `module` option to filter out warnings in a module. I tried adding `module="plotting"` to `catch_warnings` and `filterwarnings` but it's still not working (or crashing when I added it to `catch_warnings`. I might look into this more later, but I wanted to give an update.

tsalo:
@handwerkerd I think it's working now. I tested using the dataset you shared and there are now no plotting-related warnings.

handwerkerd (approved):
LGTM!

PR #1113, opened 2024-07-22
Cleaning up installation instructions

### LINKED ISSUE #1111: suggested tedana conda environmentl yml
I posted this on neurostars and someone (Steven Meisler, Penn Med postdoc) suggested I drop on tedana github as well. I did not change the distribution, so not going to do a PR, but just open as an issue. I would highly suggest this going into the readdoc.

```
name: tedana_nox
dependencies:
  - python>=3.8
  - pip
  - mdp
  - numpy
  - scikit-learn
  - scipy
  - pip:
    - mapca
    - nilearn
    - nibabel
    - tedana
```

handwerkerd:
Closes #1111

- Removed an outdated an unnecessary list of python dependencies from our installation instructions. (Unnecessary since a dependencies are automatically installed by `pip`)
- Added a link to `pyproject.toml` so that users can see the list of dependencies and we don't need to keep a parallel copy updated as part of our installation instructions
- Cleaned up the instructions to use less jargon to better explain the remaining jargon
- Added instructions for how to important workflows to run tedana within python (in addition to instructions for running from the command line)
- Added a link to our support page if a user is having problems.

@rcwelsh Does this address our concerns? Any suggestions for making this more clear for end-users who might not be strong programmers?

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -2,35 +2,47 @@
 Installation
 ############
 
-You'll need to set up a working development environment to use ``tedana``.
-To set up a local environment, you will need Python >=3.8 and the following
-packages will need to be installed:
-
-- nilearn
-- nibabel
-- numpy
-- scikit-learn
-- scipy
-- mapca
-
-After installing relevant dependencies, you can then install ``tedana`` with:
+You'll need to install ``tedana`` in your python environment.
+If you already have a python environment and you want to add tedana, type:
 
 .. code-block:: bash
 
   pip install tedana
 
-In addition to the Python package, installing ``tedana`` will add the ``tedana``
-and ``t2smap`` workflow CLIs to your path.
-You can confirm that ``tedana`` has successfully installed by launching a Python instance and running:
-
-.. code-block:: python
-
-  import tedana
-
-You can check that it is available through the command line interface (CLI) with:
+You can use a program like `conda`_ to create an environment for ``tedana`` with:
 
 .. code-block:: bash
 
-  tedana --help
+  conda create -n tedenv python=3.12 pip
+  conda activate tedenv
+  pip install tedana
+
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -2,35 +2,47 @@
...
-- numpy
-- scikit-learn
-- scipy
-- mapca
+.. code-block:: bash
+
+  pip install tedana
 
-After installing relevant dependencies, you can then install ``tedana`` with:
+You can use a program like `conda`_ to create an environment for ``tedana`` with:
```
tsalo, line 12: Maybe something:

```suggestion
As long as your python environment contains a compatible version of python, this should work.

If you want more control over your environment to ensure that every python dependency
is optimized for tedana, 
you can use a program like `conda`_ to create an environment specifically for ``tedana`` with:
```

tsalo (approved):
LGTM! Thanks for addressing this.

docs/installation.rst
```diff
@@ -2,35 +2,51 @@
...
   pip install tedana
 
-In addition to the Python package, installing ``tedana`` will add the ``tedana``
-and ``t2smap`` workflow CLIs to your path.
-You can confirm that ``tedana`` has successfully installed by launching a Python instance and running:
+The above will create a python environment for tedana called ``tedenv``,
+and you can enter that environment with ``conda activate tedenv``.
 
-.. code-block:: python
+With either of the above methods, `pip` will install python and all its dependencies.
```
eurunuela, line 27:
```suggestion
With either of the above methods, `pip` will install tedana and all its dependencies.
```

eurunuela (changes requested):
LGTM but I found a typo that would need to get fixed before merging.

eurunuela (approved):

PR #1115, opened 2024-07-29
Update list of multi-echo datasets

tsalo:
Closes none.

- Add new datasets from `open-multi-echo-data` to list of datasets in tedana's documentation.


handwerkerd (approved):
Thank you for updating all of these. It's really nice to see this section grow!

With so many, it would be nice to organize these and provide a bit more contextual info (year collected, sample size, maybe this coudl be done by justing adding a column for open data in the publications spreadsheet) but that can definitely be a separate PR.

tsalo:
For a certain level of detail, I'd rather link out to the open-multi-echo-data documentation (which is currently empty, but which should get filled in over time).

PR #1125, opened 2024-09-03
 Adding robustica option to ICA decomposition to achieve consistent results (new PR)

handwerkerd:
#1013 was mis-aligned with `main` and there were some issues with getting it aligned again, so I'm opening a new PR that has multiple updates and should be aligned with `main`

Overview from that PR: Given that the `FastICA` method depends on its seed value, in some cases the generated output can be substantially different. To overcome this issue, we added `robustica` (https://github.com/CRG-CNAG/robustica/tree/main/robustica) which is a python library based on Icasso as an option to the ICA decomposition. `robustica` clusters many iterations of `FastICA` to achieve consistent results. 

- `ica_method` is added as an option (robustica or fastica) and the user can select between the two methods. The number of runs for when robustica is used can be determined by the user through setting `n_robust_runs`. The default is set to `robustica' with 30 number of robust runs. This default value is selected based on our experience using robustica for two different datasets with over 250 subjects' data. 
- The default setting will continue to be `fastica` We should re-evaluate after more users start using `robustica` and provide feedback
- The DBSCAN clustering algorithm has been implemented in a hard-coded manner with Agglomerative clustering being used if that fails
- Explanations in the documentation are added to `faq.rst` and `approach.rst`

Note:

#1013 includes key discussions related to this PR including [an evaluation of how the initial number of components requested affects the number of stable components found](https://github.com/ME-ICA/tedana/pull/1013#issuecomment-1867853810) and an [observation of similar ICA component spatial maps with different time series](https://github.com/ME-ICA/tedana/pull/1013#issuecomment-1955777222) that shouldn't get lost with the switch to a new PR


Lestropie:
Generating a task list to make sure nothing from #1013 gets missed:

- [ ] From [this comment](https://github.com/ME-ICA/tedana/pull/1013#pullrequestreview-1791154236): Catching a failed RobustICA run is no longer a bare `except`, but it's catching all possible derivatives of `Exception`. @BahmanTahayori Do you have / can you generate a list of those exceptions raised by any of the invoked RobustICA functions, and then modify to only catch those?

- [ ] Current version of code adds "`tedana/config.py`". As per [prior comment](https://github.com/ME-ICA/tedana/pull/1013#pullrequestreview-1791228988), "`defaults.py`" might be better for some (but is less appropriate for other constants currently in that file), or these constants could just be at the head of the main entrypoint source code file.
    - [ ] There was also the question of, if such an approach for defining constants is used, whether there is code unrelated to RobustICA that should be using the same approach to prevent unintended divergences. It looks like there's been some work in this respect; don't know if someone more familiar with the code base wants to review that it would now be used everywhere it's appropriate.

- [ ] Still awaiting from [prior comment](https://github.com/ME-ICA/tedana/pull/1013#pullrequestreview-1891741672): Citation "`Anglada2022`" would be better renamed to "`anglada2022robustica`" (both removing capitalisation and adding first word of title) for consistency with the rest of the bibliography.

Looking through the code there's nothing setting off alarm bells to me, and the commit history looks appropriate.

tedana/decomposition/ica.py
```diff
@@ -37,17 +59,171 @@ def tedica(data, n_components, fixed_seed, maxit=500, maxrestart=10):
...
+        where `C` is components and `T` is the same as in `data`
+    fixed_seed : :obj:`int`
+        Random seed from final decomposition.
+    """
+    if n_robust_runs > WARN_N_ROBUST_RUNS:
+        LGR.warning(
+            "The selected n_robust_runs is a very big number! The process will take a long time!"
+        )
+
+    RepLGR.info("RobustICA package was used for ICA decomposition \\citep{Anglada2022}.")
```
Lestropie, line 123:
```suggestion
    RepLGR.info("RobustICA package was used for ICA decomposition \\citep{anglada2022robustica}.")
```

(To be applied in conjunction with a second suggestion below)

tedana/resources/references.bib
```diff
@@ -333,3 +333,14 @@ @article{tedana_decision_trees
   year    = {2024},
   doi     = {10.6084/m9.figshare.25251433.v2}
 }
+
+@Article{Anglada2022,
```
Lestropie, line 337:
```suggestion
@Article{anglada2022robustica,
```

(To be applied in conjunction with a second suggestion above)

handwerkerd:
The tests are passing locally but failing here. The one thing I see that is clearly a problem is ` ERROR: Failed building wheel for scikit-learn-extra` This is a library that is used within `robustica`. I just added it to `pyproject.toml` to see if that resolved the issue, but it did not. There are other errors, but I'm not sure if some/all are due to this problem. @tsalo, @eurunuela or others: Any ideas for how to get this built correctly on CircleCI?

pyproject.toml
```diff
@@ -30,8 +30,9 @@ dependencies = [
     "pandas>=2.0,<=2.2.2",
     "pybtex",
     "pybtex-apa-style",
-    "robustica>=0.1.3",
+    "robustica>=0.1.3, <=0.1.3",
```
tsalo, line 33: Could you pin the version?
```suggestion
    "robustica==0.1.3",
```

eurunuela:
Hey @handwerkerd, I am getting the same error on a new environment. So I don't think it's a CircleCI issue.

pyproject.toml
handwerkerd: Will this work with the version update bot?  My assumption was we needed a `<=` to give the bot something to update, but I don't really know.

tsalo: dependabot should just bump the version. So if robustica releases 0.1.4, we'd get a PR changing it to `robustica==0.1.4`.

handwerkerd: Is that what we want or would "robustica>=0.1.3, <=0.1.4", be more in line with our other modules? I'm mostly neutral on this.

eurunuela:
I hope they are still around and can fix it, but the lack of commits and PRs since October last year doesn't sound very promising.

handwerkerd:
> I hope they are still around and can fix it, but the lack of commits and PRs since October last year doesn't sound very promising.

The author responded immediately, but I was looking around and it seems like one of their dependencies, `scikit-learn-extra` has the same problem. Unfortunately it sounds like that package is no longer actively maintained and a solution might take time: https://github.com/scikit-learn-contrib/scikit-learn-extra/issues/177

BahmanTahayori:
Thanks @handwerkerd and the team. I will go through changes and will add a few other things as well. Hopefully, the issue with <scikit-learn-extra> can be resolved soon.

handwerkerd:
I've been sitting on this for a few days, but I figured I should share. robustica is a single person's project. scikit-learn-extra is an unmaintained library with a few people keeping half-an-eye on it. i want this merged, but adding these as required dependencies might end up being more work for us longer term. We should discuss here or at our Sept 19th (20th in Australia) dev call.

The options I see are:
- If/when the current scikit-learn-extra issue is resolved, we can merge and hope things are stable for a while.
- Work with robustica to remove the scikit-learn-extra dependency (and we'd also understand robustica a bit better & potentially help with future issues there)
- Add a minimum version of robustica directly into our code (with appropriate credit)
- Use another method that we either add directly to our code, like we did for mapca, or is part of a package with more active developers.

Thoughts?

BahmanTahayori:
Thanks @handwerkerd for your suggestions.  Removing the `scikit-learn-extra` dependency or making it optional seems a reasonable choice.  This library is used for two clustering methods _KMedoids_ and _CommonNNClustering_. There are 10 other stable options from `scikit-learn`. Given that `scikit-learn-extra`  is not maintained appropriately, I think we can negotiate with Miquel to see if he is happy to remove `scikit-learn-extra`.  I will liaise with him.

handwerkerd:
@BahmanTahayori fixed the original PR and has a potential solution to the scikit-learn-extra problem so I'm closing this PR and we can continue work in the original PR https://github.com/ME-ICA/tedana/pull/1013

Issue #1053, opened 2024-02-29
Censoring data within tedana


handwerkerd:
### Summary

I was just having a discussion with AFNI's @afni-rickr and @mrneont and a few others and the discussion of censoring within tedana came up.

### Additional Detail

The context of our discussion involved datasets where a non-trivial number of volumes were censored due to head motion. If tedana processes the entire time series then it will get more ICA components and unnecessarily more rejected components, which hurts statistical degrees of freedom and other stuff. 

From the practical side, it wouldn't be hard to run T2* map estimation,  ICA, and component selection on data with censored volumes. The one issue is that the denoised time series might need to exclude censored volumes so we'd end up with time series with 0 or `nan` values. AFNI could easily deal with that, but I'm not sure about other programs.

### Next Steps

- Decide if this is something work implementing

tsalo:
Is there anything in tedana that would use the full data?

My first thought is that this would be a great feature, but should go in fmripost-tedana, along with the dummy volume removal.

handwerkerd:
My assumption would be that censored volumes are calculated in AFNI or fMRIprep before tedana is run and then the list of volumes to censor would be an input to tedana.

tsalo:
If running tedana with the censored volume mask and running it directly on the censored data would be equivalent, then I think we should put that step in a postprocessing workflow (i.e., fMRIPost-tedana). If tedana needs to use both the censored and uncensored versions of the run, then I think there's a strong argument to directly incorporate it into tedana.

CesarCaballeroGaudes:
I would say that running tedana on censored data would not be equivalent to applying tedana and then censoring because censoring will block the artefacts that tedana aims to remove. Therefore, the components will be different very likely.

afni-rickr:
Hi @handwerkerd,

Actually, this is more than just censoring.  My intention (for many years now) was to clean up the data a bit before giving it to tedana, removing trends and motion parameters say, and to censor.  The hopeful benefit is that the ICA components would not be focused on the bad time points, or even things like drift or other regressors that we would plan to regress later anyway.
On my side, the intention was to not even change tedana for this, if possible.

But perhaps projecting anything out would happen differently across the echoes, and therefore probably mess up tedana.  And that might mean tedana would have to be the one to deal with such a thing.

To be clear, is the ICA step run on the OC data?

handwerkerd:
[...]

PR #1131, opened 2024-09-17
Add elbow-less minimal decision tree

### LINKED ISSUE #1129: Adding a conservative elbow-free decision tree
### Summary

This is based on an OHBM2024 discussion with @moser297 that was discussed at our [July 2024 developer call](https://docs.google.com/document/d/1o4dsvFB1r6SeqdUOWEQyIODSWKYb15-B5135KLLhGgg/edit?usp=sharing). The kappa and rho elbows do not robustly appear in all runs. In particular, an infant dataset seemed to often not generate clear elbows. This is worth investigating in general, but we could also add a conservative super-minimal decision tree that doesn't depend on any elbows or relative values that might cause issues in some datasets.

### Additional Detail

The idea would be to make a decision tree wouldn't be as aggressive at removing noise as other trees, but any user could be reasonably confident that it never removes something they care about. My initial idea would be to just include the first 6 nodes of the MEICA Decision Tree ( https://tedana.readthedocs.io/en/stable/included_decision_trees.html ) which just exclude extreme cases, like rho>kappa.

Right now, this is a placeholder issue so that we don't forget this discussion. Anyone is welcome to take this on.

### Next Steps

- Make & name a simple decision tree and test it on some data, possibly in collaboration with @moser297
- Check if the tree is too conservative (i.e. it's only removing things that would have already been removed by detrending regressors or other standard tools)
- Check if there are any criteria tweaks that could be used to make it a bit more agressive, but still reliable.

tsalo:
Closes #1129. I'll actually test it soon.

- Create a decision tree based on the minimal decision tree, with elbow nodes cut out.

### DIFF
--- a/tedana/resources/decision_trees/elbowless_minimal.json
+++ b/tedana/resources/decision_trees/elbowless_minimal.json
@@ -1,0 +1,148 @@
+{
+    "tree_id": "elbowless_minimal_decision_tree",
+    "info": "first version of minimal decision tree without elbows",
+    "report": "The elbow-less minimal decision tree \\citep{tedana_decision_trees} is a simplified version of the MEICA decision tree \\citep{kundu2013integrated,dupre2021te} without many criteria that do not rely on kappa and rho thresholds, including any criteria based on elbows calculated across components. ",
+    "necessary_metrics": [
+        "kappa",
+        "rho",
+        "countsigFS0",
+        "countsigFT2",
+        "dice_FS0",
+        "dice_FT2",
+        "signal-noise_t",
+        "variance explained"
+    ],
+    "intermediate_classifications": ["provisionalaccept", "provisionalreject"],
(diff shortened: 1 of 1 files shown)


handwerkerd:
Thank you!
This should also include a figure in `included_decision_trees.rst` (I can help make the figure) and an update to "[tedana] What are the differences between the tedana_orig, meica, and minimal decision trees?" in `faq.rst`. Hopefully we can have some people test & make sure this is useful before making those changes.

handwerkerd:
I ran tedana on some infant data where the kappa and rho elbow criteria were reported to be causing problems. In my hands, it was possible to get plausible elbow estimates. There were other issues that might be better address through things like adding censoring (#1053). It's also probably that we can improve the decision tree for an infant application, but I didn't see evidence that an elbow-less minimal tree would be the main fix for their challenges.

Unless we find a use case where this addition clearly helps, I'd suggest closing this PR and maybe the related issue for now.

Issue #919, opened 2022-12-22
harmonize terminology across codebase


handwerkerd:
### Summary

Particularly with the addition of the modularization decision tree #756, there are places where similar things are given different names in various parts of the codebase. This might create some unnecessary confusion and risks for typos

### Additional Detail

Places for harmonizing terminology includes:

  - using `component_table` rather than `comptable` everywhere
  - Decide whether to standardize on `d_table_score` vs `mean_metric_score`
  - Have the PCA component table better match the ICA component table (primarily changing from rationale to classification_tags terminology. (This is currently only an issue for the kundu selection process for PCA components.)

Places to reduce typo risks:
  - Make all classification and classification_tag labels either fully case insensitive or changed to match the capitalization of where classification and tag labels are defined.
  - Create a field in the tree json for classification_tag_explanations In it would be a dictionary for each tag in the tree and an explanation for what that tag means. Then explanations for the tags in any given tree would automatically be included in the report rather than having us need to maintain a hard coded list of explanations in the documentation.

### Next Steps

- Wait for #756 to merge
- Since this will require a lot of small changes across the code base, find a time when no one else is making dramatic code edits so that this can happen quickly.

PR #1145, opened 2024-11-05
Rename variables

tsalo:
Related to #919. Tedana development's been pretty quiet lately, so I figured it's a good time to rename the variables we've wanted to rename for a while now.

- mmix --> mixing
- mixm --> mixing_file
- data_oc --> data_optcom
- catd --> data_cat
- comptable --> component_table
- dd --> data_reduced

### DIFF
--- a/tedana/decomposition/ica.py
+++ b/tedana/decomposition/ica.py
@@ -54,7 +54,7 @@ def tedica(
 
     Returns
     -------
-    mmix : (T x C) :obj:`numpy.ndarray`
+    mixing : (T x C) :obj:`numpy.ndarray`
         Z-scored mixing matrix for converting input data to component space,
         where `C` is components and `T` is the same as in `data`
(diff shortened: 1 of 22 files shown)


tsalo:
It should only rename internal variables. I avoided changing column names or command-line parameters. There's always a chance that I missed something though...

eurunuela:
> It should only rename internal variables. I avoided changing column names or command-line parameters. There's always a chance that I missed something though...

I don't think you did. But I wanted to double-check. Thanks, Taylor!

tsalo:
I think this PR merits two reviewers.

handwerkerd (approved):
LGTM.

One very minor thing I noticed is`mmix` is always `mixing` but `ica_reclassify_workflow` has `mmix_orig` while other places use `mixing_orig`. This is really minor since this is effectively a local temp variable and I don't think this will confuse anyone.

PR #1147, opened 2024-11-16
Add dash versions of underscore params

tsalo:
Closes none. This makes the formatting of params consistent.

- Add dash versions of `--ica-method` and `--n-robust-runs`. I didn't remove the underscore versions, but we should in the next major release.


eurunuela (approved):
Makes sense. LGTM!

PR #1151, opened 2024-11-21
Add a function to check if TE values are in ms, otherwise update them

### LINKED ISSUE #1150: Add sanity check for echo times
### Summary
Sometimess folks might accidentally provide echo times as seconds instead of milliseconds, so we should check the values and raise a warning if they seem too low.

eurunuela:
Closes #1150.

- Adds a function to check if the TE values given by a user are in ms. If they are not and are below 1, it assumes they are in seconds and they get updated by multiplying by 1000. If the values are a combination of higher than one and lower than one, the function raises an error.
- Adds a unit test to ensure the functionality is correct.

### DIFF
--- a/tedana/utils.py
+++ b/tedana/utils.py
@@ -654,7 +654,7 @@ def check_te_values(te_values):
         return te_values.tolist()
     elif all((te_values > 0) & (te_values < 1)):
         # Raise a warning and convert to ms by multiplying by 1000
-        LGR.warning("TE values are not in milliseconds. Converting to ms.")
+        LGR.warning("Assuming the provided TE values are in seconds. Converting to ms.")
         return (te_values * 1000).tolist()
     else:
         raise ValueError("TE values must be positive and in milliseconds.")
--- a/tedana/utils.py
+++ b/tedana/utils.py
@@ -634,3 +634,27 @@ def get_system_version_info():
         "Python": sys.version,
         "Python_Libraries": python_libraries,
     }
+
+
+def check_te_values(te_values):
+    """Check if all TE values are in ms by checking if they are higher than 1.
+
+    Parameters
+    ----------
+    te_values : list
+        TE values to check.
+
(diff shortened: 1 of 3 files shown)


eurunuela:
I'm pretty sure this should do it but I think it would be helpful if @handwerkerd tested it on the dataset you showed earlier.

tsalo (approved):
I generally would prefer to raise a warning and continue without changing the data, but I can't imagine a scenario where a user might provide a value < 1 and actually intend milliseconds, so I'm good with your approach.

tedana/utils.py
```diff
@@ -634,3 +634,27 @@ def get_system_version_info():
...
+    -------
+    list
+        TE values in milliseconds.
+    """
+    te_values = np.array(te_values)
+    if all(te_values > 1):
+        return te_values.tolist()
+    elif all((te_values > 0) & (te_values < 1)):
+        # Raise a warning and convert to ms by multiplying by 1000
+        LGR.warning("TE values are not in milliseconds. Converting to ms.")
```
handwerkerd, line 657:
```suggestion
        LGR.warning("Assuming the provided TE values are in seconds. Converting to ms.")
```

Minor tweak to the warning text to make clear that the conversion is from seconds to milliseconds.

handwerkerd (changes requested):
Thank you for doing this! Just one minor text change.

tedana/utils.py
tsalo: Good point @handwerkerd. Now that I think about it, it should probably be made clear what the reasoning is. Something like "TE values are very low- ...".

handwerkerd: I was also considering "TE values less than 1. Assuming units are seconds. Converting to ms."

handwerkerd (approved):

tsalo (approved):

PR #1153, opened 2024-11-27
Show interactive plot in the HTML reporst for Robust ICA clusters

eurunuela:
Closes none. Stems off a discussion in #1013.

- Add a function to the dynamic reports file to generate the plot with Bokeh.
-

### DIFF
--- a/tedana/reporting/dynamic_figures.py
+++ b/tedana/reporting/dynamic_figures.py
@@ -421,3 +421,112 @@ def _link_figures(fig, comptable_ds, div_content, io_generator):
     """
     fig.js_on_event(events.Tap, _tap_callback(comptable_ds, div_content, io_generator))
     return fig
+
+
+def _create_clustering_tsne_plt(cluster_labels, similarity_t_sne, io_generator):
+    """Plot the clustering results of robustica using Bokeh.
+
+    Parameters
+    ----------
+    cluster_labels : (n_pca_components x n_robust_runs,) : numpy.ndarray
+        A one dimensional array that has the cluster label of each run.
+    similarity_t_sne : (n_pca_components x n_robust_runs,2) : numpy.ndarray
+        An array containing the coordinates of projected data.
+    io_generator : object
+        An object containing all the information needed to generate the output.
+    """
+    title = "2D projection of clustered ICA runs using TSNE"
+    marker_size = 8
+    alpha = 0.8
+    line_width = 2
+
+    # Create figure
+    p = plotting.figure(
+        title=title,
+        width=800,
+        height=600,
+        tools="pan,box_zoom,wheel_zoom,reset,save",
+    )
+    p.title.text_font_size = "16px"
+    p.xaxis.axis_label = "x1"
+    p.yaxis.axis_label = "x2"
+
+    # # Create ColumnDataSource for all points
+    # source_data = {
+    #     "x": similarity_t_sne[:, 0],
+    #     "y": similarity_t_sne[:, 1],
+    #     "cluster": cluster_labels,
+    # }
+    # source = models.ColumnDataSource(source_data)
+
+    # Plot regular clusters
+    for cluster_id in range(np.max(cluster_labels) + 1):
+        cluster_mask = cluster_labels == cluster_id
+        if not np.any(cluster_mask):
+            continue
+
+        # Get points for this cluster
+        cluster_points = similarity_t_sne[cluster_mask]
+
+        # Add scatter plot for cluster points
+        p.circle(
+            x="x",
+            y="y",
+            source=models.ColumnDataSource({"x": cluster_points[:, 0], "y": cluster_points[:, 1]}),
+            size=marker_size,
+            alpha=alpha,
+            line_color="black",
+            fill_color=None,
+            line_width=line_width,
+            legend_label="Clustered runs",
+        )
+
+        # Add hull if enough points
+        if cluster_points.shape[0] > 2:
+            from scipy.spatial import ConvexHull
+
+            hull = ConvexHull(cluster_points)
+            centroid = np.mean(cluster_points[hull.vertices], axis=0)
+            scaled_points = centroid + 1.5 * (cluster_points - centroid)
+
+            # Create hull line segments
+            xs = []
+            ys = []
+            for simplex in hull.simplices:
+                xs.extend([scaled_points[simplex[0], 0], scaled_points[simplex[1], 0], None])
+                ys.extend([scaled_points[simplex[0], 1], scaled_points[simplex[1], 1], None])
+
+            p.line(
+                x=xs,
+                y=ys,
+                line_color="blue",
+                line_dash="dashed",
+                line_width=line_width,
+                legend_label="Cluster's boundary",
+            )
+
+    # Plot noise clusters if they exist
+    if np.min(cluster_labels) == -1:
+        noise_mask = cluster_labels == -1
+        noise_points = similarity_t_sne[noise_mask]
+
+        p.x(
+            x=noise_points[:, 0],
+            y=noise_points[:, 1],
+            size=marker_size * 2,
+            alpha=0.6,
+            color="red",
+            legend_label="Unclustered runs",
+        )
+
+    # Configure legend
+    p.legend.click_policy = "hide"
+    p.legend.location = "top_right"
+
+    # Save HTML file
+    # plot_name = f"{io_generator.prefix}clustering_projection_tsne.html"
+    # plotting.save(p, os.path.join(io_generator.out_dir, "figures", plot_name))
+
+    breakpoint()
+
+    return p


eurunuela:
What do you think @handwerkerd @tsalo?

I still have to handle the case where Fast ICA is used instead. I'm open to suggestions on how to handle the variables that need to be passed down to make the Bokeh plot.

eurunuela:
You should be able to check the interactive Bokeh plot now. I will work on the edge cases that aren't covered in the other tests.

tedana/decomposition/ica.py
```diff
@@ -87,7 +88,7 @@ def tedica(
     else:
         raise ValueError("The selected ICA method is invalid!")
 
-    return mixing, fixed_seed
+    return mixing, fixed_seed, c_labels, similarity_t_sne
```
handwerkerd, line 91: I think I see why a bunch of the integration tests are failing. You added code to add the robustica QC figures and measures to the report & that code is running even when the robustica option is not selected. Here's one example where you are returning `c_labels` and `similarity_t_sne` but those won't exist if `ica_method == "fastica"`

eurunuela: Yes, I know. I was only waiting on feedback on the 5 echo before I made these changes. I thought the plot itself was the main contribution here and so I wanted to make sure folks were happy with that.

handwerkerd:
I ran this on the 5 echo test data & the figure looked a bit odd so I ran it again on some of my other data & it looked similar. It looks like all cluster boundaries connect at a single origin point and groups of x's that look like cluster aren't. Is this how the visualization is supposed to look?
image

eurunuela:
Good catch @handwerkerd! There is definitely something wrong there.

eurunuela:
This should be fixed now @handwerkerd. Clusters with only two points don't have a boundary drawn though.

handwerkerd:
I've now run it again and it looks much better, but there's some weirdness and this seems more like an issue with the algorithm that the report is highlighting. I'm testing on a 3-echo 150volume block design flashing checkerboard task. I'm running asking for 40, 50, & 60 PCA components with 100 iterations of robustICA.

Here's with 50 PCA components. The mean Index Quality is 0.943882566664395  The clusters look good, but there are clearly other clusters that are deemed "unclustered runs" my guess is that the good clusters have slightly more components in them and are slightly tighter, but not by much. That said, the overall figure is hard to interpret because of this.
image

The hover text just shows the same phrase repeatedly. Maybe, these can instead show the similarity_t_sne values (or some other metric). I also don't think those values are saved anywhere in the output.
image

I wanted to confirm the above wasn't due to some weirdness in my initialization to 50 PCA components.
Here is 60 PCA components
image

Here is 40 PCA components
image

They have the same general issue with unclustered clusters, but there are also a few stretched boundaries. I suspect these really are representing some weirdness in visualizing a higer dimensional space in 2D, but I figured I'd share.

FWIW, I'm aware I'm making a bunch of critiques after not looking at this for months. If you want me to more actively revise code, I can help.

handwerkerd:
Also, there are two bokeh warnings:
```
BokehDeprecationWarning: 'circle() method with size value' was deprecated in Bokeh 3.4.0 and will be removed, use 'scatter(size=...) instead' instead.
BokehDeprecationWarning: 'x() method' was deprecated in Bokeh 3.4.0 and will be removed, use "scatter(marker='x', ...) instead" instead.
```

We've gotten these before & it wasn't too painful to switch to `scatter`

handwerkerd:
While I'm commenting, when robustICA is run, should the figure be below the kappa & rho rank plots so it's easier to compare to the other component-related figures? It would be nice, but very **not** essential to link it so a click on a cluster highlights the components on the other figures.

eurunuela:
> Here's with 50 PCA components. The mean Index Quality is 0.943882566664395 The clusters look good, but there are clearly other clusters that are deemed "unclustered runs" my guess is that the good clusters have slightly more components in them and are slightly tighter, but not by much. That said, the overall figure is hard to interpret because of this.

I see what you mean and I thought that was a bit odd too, but was trusting Bahman's code. It may be worth displaying a couple of metrics. 

> The hover text just shows the same phrase repeatedly. Maybe, these can instead show the similarity_t_sne values (or some other metric). I also don't think those values are saved anywhere in the output.

Yes. The same phrase is shown for each dot in the plot. So, if a group of dots belong to the same cluster and you hover over them, it looks like a list of repeated items. This is in part due to the zooming levels. I can try to make it show something like `Cluster 6: <metric>` as opposed to `Cluster: Cluster 6`.

> They have the same general issue with unclustered clusters, but there are also a few stretched boundaries. I suspect these really are representing some weirdness in visualizing a higer dimensional space in 2D, but I figured I'd share.

Are those dots part of the same cluster? Or is the visualization wrong?

> FWIW, I'm aware I'm making a bunch of critiques after not looking at this for months. If you want me to more actively revise code, I can help.

To be honest, I haven't looked at the RobustICA code much myself. I have just translated the static plot that Bahman had in his PR or repository. But seeing you had many Xs very close to clusters, I think it is worth having a look. At least to extract some metrics.

eurunuela:
> Also, there are two bokeh warnings:
> 
> ```
> BokehDeprecationWarning: 'circle() method with size value' was deprecated in Bokeh 3.4.0 and will be removed, use 'scatter(size=...) instead' instead.
> BokehDeprecationWarning: 'x() method' was deprecated in Bokeh 3.4.0 and will be removed, use "scatter(marker='x', ...) instead" instead.
> ```
> 
> We've gotten these before & it wasn't too painful to switch to `scatter`

I can update the figures. It should be easy.

eurunuela:
> While I'm commenting, when robustICA is run, should the figure be below the kappa & rho rank plots so it's easier to compare to the other component-related figures? It would be nice, but very **not** essential to link it so a click on a cluster highlights the components on the other figures.

We could move it there. It would definitely make sense if we connect the figure with the others. I can try to link it with the other figures but that will be trickier.

handwerkerd:
> > Here's with 50 PCA components. The mean Index Quality is 0.943882566664395 The clusters look good, but there are clearly other clusters that are deemed "unclustered runs" my guess is that the good clusters have slightly more components in them and are slightly tighter, but not by much. That said, the overall figure is hard to interpret because of this.
> 
> I see what you mean and I thought that was a bit odd too, but was trusting Bahman's code. It may be worth displaying a couple of metrics.
> 
> > The hover text just shows the same phrase repeatedly. Maybe, these can instead show the similarity_t_sne values (or some other metric). I also don't think those values are saved anywhere in the output.
> 
> Yes. The same phrase is shown for each dot in the plot. So, if a group of dots belong to the same cluster and you hover over them, it looks like a list of repeated items. This is in part due to the zooming levels. I can try to make it show something like `Cluster 6: <metric>` as opposed to `Cluster: Cluster 6`.

Maybe something like `Cluster 6: similarity metric, # iterations in cluster` These values could also be saved in the component table since there is one value per component. I haven't played much with hover text, but I wonder if there's a way to prevent it from showing two identical lines.

I also wonder if robustICA has a threshold for making clusters and we might be able to get a metric for components that look like they are clustered, but are sub-threshold. That way we can have some way to evaluate if the missed clusters are missed for an internally consistent reason (and maybe re-evaluate thresholds). FWIW, I ran this with 100 iterations to get robust results, but something similar appeared with 50 iterations. It's possible that, with more iterations, these non-clustered groups of things that appear in only 50-70% of the iterations pop out more. Since this is getting back to robustICA performance, I'm tagging @BahmanTahayori and @Lestropie to see if they have explanations for what I'm seeing & ideas for good metrics to include. On a very positive note, including visualizations and quality checks is really useful and prompting this discussion.


> > They have the same general issue with unclustered clusters, but there are also a few stretched boundaries. I suspect these really are representing some weirdness in visualizing a higer dimensional space in 2D, but I figured I'd share.
> 
> Are those dots part of the same cluster? Or is the visualization wrong?

I think it's either a robustICA cluster creation issue or an issue with the dimensionality reduction into 2D space.
Here's a zoomed in image from the 60 PCA figure above. Three of the black circles in the overlapping boundaries are from cluster 32, which is fully shown. One black circle is from cluster 6, which is in the top right of the image.
image

I just showed this to @javiergcas and he thinks the issue might be because there is one metric used to make the clusters and another metric used to reduce dimensions. That is the clusters are created using DBSCAN and then, the silhouette values and `q` are calculated using the same distance measures.
https://github.com/eurunuela/tedana/blob/1e9977fbb41dc4faa6c73b5685c4f83fc9fcbe3d/tedana/decomposition/ica.py#L145

The neighborhood size is then recalculated using perplexity followed by tsne for dimensionality reduction.
https://github.com/eurunuela/tedana/blob/1e9977fbb41dc4faa6c73b5685c4f83fc9fcbe3d/tedana/decomposition/ica.py#L198C26-L207C6

I can dig more, but I wonder if there is information that is outputted by `evaluate_clustering` might be able to be reduced to 2D in a way that more closely matches the distance measures the clusters were originally calculated on.

Lestropie:
> So, if a group of dots belong to the same cluster and you hover over them, it looks like a list of repeated items.

Hypothetically, it could show for each item the index of the ICA run and the component index within that run, in addition to the cluster index. That would convey more about the nature of the clustering that's going on, and might help people understand what they're looking at. But might take up too much screen space?

> They have the same general issue with unclustered clusters, but there are also a few stretched boundaries.

The method for clustering components can be modified [here](https://github.com/ME-ICA/tedana/blob/24.0.2/tedana/decomposition/ica.py#L128-L161). I recall @BahmanTahayori determined that DBSCAN should be tried first followed by AgglomerativeClustering, but testing of such was limited to a smaller breadth of data than what should now be possible.

From my PhD PTSD, DBSCAN (and possibly the AgglomerativeClustering implementation here also) can fail to distinguish between samples from two adjacent clusters if there are an adequate number of samples in the space between them to "build a bridge" between the two. Maybe when this happens, it ends up with a single large cluster with too many components / that spans too much of the high-dimensional space / that includes multiple components from the same run, and the cluster is rejected, demoting all components within to "unclustered". That might explain some of the "smudge" of red X's near the centre of some of those plots (less so some of those X's that are sitting at the periphery of otherwise contiguous-looking clusters).

Looking at the [list of available algorithms in RobustICA](https://github.com/CRG-CNAG/robustica/blob/0.1.4/robustica/RobustICA.py#L16-L25), Affinity Propagation appeals to me on first principles; though from the RobustICA manuscript one would need to [check feasibility given memory and computational expense](https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-022-05043-9/figures/1) (plus the authors rated it poorly). OPTICS is purportedly a "superior" version of DBSCAN so is worth trying here.

> I also wonder if robustICA has a threshold for making clusters and we might be able to get a metric for components that look like they are clustered, but are sub-threshold.

I can't comment on whether specifically RobustICA provides anything in this regard, but from first principles at least I expect that this is going to be highly dependent on the clustering algorithm. For DBSCAN particularly, my a priori expectation is that there would be no such information: either items are aggregated together into a cluster based on the internal criteria or they are not. Any algorithm that is explicitly hierarchical is more likely to preserve information about the relationship between a component's distance from nearby clusters and the threshold used to flatten that hierarchy; but as stated, no idea whether RobustICA can forward that information.

BahmanTahayori:
Thanks @eurunuela for adding the plot to the report. Some issues with the plot is caused by going from high dimension to 2d dimension. However, I agree that some plots seem weird. Specifically the one with 60 components. I have added a few other information to the figure on my local device and when all changes are applied and tested I will push changes to github. Obviously, it doesn't mean that you need to stop merging what you already have done until I have finalised my changes.

eurunuela:
For a lack of a better metric just now, I have updated the data displayed on hover to show the similarity values.

I think I have added all suggestions made except for the connection with the other plots.

handwerkerd:
Based on the discussion in #1211, I'm fine getting this working cleanly as is, adding some documentation to the interpretation limits of this output, merging, & opening a new issue to address limits as a future PR.

I can work on the added documentation, but this will need to use the jinja2 template once we merge #1209 & there are a few things in the comments above that would need to be addressed before merging.

handwerkerd:
> I bumped into a pretty nice visualization that we could try to copy. See the MNIST embeddings plot in the link below. I think we could do something similar too, where we show the ICA component IDs when hovering over a region.
> 
> https://docs.marimo.io/#learn-more

Fun! I'd lean towards merging something similar to this PR before adding more stuff. Given past fun, I'm also a bit hesitant to add a dependency that's only 2 years old.

eurunuela:
Tests should pass now. Is there anything other than the docs that you would like to add before we merge @handwerkerd ?

handwerkerd:
I just ran the tests and then locally processed a dataset using a python 3.11 and 3.13 environment. In none of the cases is the new robustICA plot appearing in the html report.

I'm also seeing the `test_integration_five_echo` test fail because there is a different number of components. This will happen across hardware & software differences when using robustICA with only 4 iterations. #1213 will address this by no longer counting components in tests that use robustICA.

I'm still looking at this in more closely, but wanted to report the above feedback.

eurunuela:
I found out why the tSNE plot wasn't displayed. I must have missed updating the variable to use the Jinja2 format in #1209.

handwerkerd:
Besides documentation, this is ready to merge. I can draft documentation, if you prefer.

FWIW,  I was working on logging other key parameters from robustICA (mainly the number of times fastICA failed to converge) but I'm realizing it will be cleaner as a separate PR.

eurunuela:
I will leave it to you @handwerkerd then to add docs or merge and work on a different PR.

handwerkerd (approved):
LGTM!

PR #1154, opened 2024-11-27
Enforce equal figure widths in HTML report and make them adjustable by window size

### LINKED ISSUE #1050: Enforce equal figure width in HTML report
One thing that would be amazing is if we could have all of the figures be rendered with the same width. Maybe even based on the window size, if that's feasible?

_Originally posted by @tsalo in https://github.com/ME-ICA/tedana/issues/1040#issuecomment-1956722311_

eurunuela:
Closes #1050.

- Remove the fixed height from the style of each image
- Add max width to each image
- Add window width-dependent width to each image


tsalo (approved):
LGTM!

PR #1163, opened 2025-01-07
Back-fill, then forward-fill, NaNs in external regressors

### LINKED ISSUE #1152: External regressors with NaNs will crash decision tree
### Summary

I tried building a decision tree to use motion parameters from an fMRIPrep confounds file. The problem is that relative motion measures (e.g., framewise displacement) have a NaN in the first row of the confounds because they're derivatives. We could leave it up to users to deal with this, but I think an easy workaround on our end would be to backfill NaNs automatically.

### Additional Detail

We could also raise a warning if NaNs are present to ensure users know what's happening.

tsalo:
Closes #1152.

- Warn if the external regressors have NaNs.
- Back-fill (and forward-fill) cells with NaNs in the external regressors DataFrame.

### DIFF
--- a/tedana/metrics/external.py
+++ b/tedana/metrics/external.py
@@ -156,6 +156,20 @@ def validate_extern_regress(
             f"external regressor model: {sorted(extra_names)}"
         )
 
+    nan_columns = []
+    for column in external_regressor_names:
+        if external_regressors[column].isna().values.any():
+            nan_columns.append(column)
+
+    if nan_columns:
+        LGR.warning(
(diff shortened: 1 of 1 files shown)


tsalo:
Coverage didn't upload. Maybe they're disabled tokenless uploads?

eurunuela (approved):

eurunuela:
> Coverage didn't upload. Maybe they're disabled tokenless uploads?

From looking at the merges in main:

````
{"message":"Token required - not valid tokenless upload"}

400
```

PR #1166, opened 2025-01-15
Add option to overwrite file (and write out extra files with "verbose") in T2smap workflow

smoia:
Closes [I don't know if there was an issue open for this].

- T2smap workflow currently does not allow to overwrite files - but stops with runtime error if a file would be overwritten, asking to use the non-existing `--overwrite` argument. This PR solves that issue.
- I saw that in the tedana workflow parser, "overwrite" was added to the parser itself rather than to the "optional" parser group, so just quickly changed that, if desirable.

### DIFF
--- a/tedana/workflows/t2smap.py
+++ b/tedana/workflows/t2smap.py
@@ -152,6 +152,21 @@ def _get_parser():
     optional.add_argument(
         "--quiet", dest="quiet", help=argparse.SUPPRESS, action="store_true", default=False
     )
+    optional.add_argument(
+        "--verbose",
+        dest="verbose",
+        action="store_true",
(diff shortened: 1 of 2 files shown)


eurunuela (approved):
LGTM

eurunuela:
@smoia, did you end up finishing #644? Is it at a state where we could just get it up to date with `main`?

smoia:
@eurunuela TBH I totally lost track of that. I can have a look at it this weekend, would that work for you?

eurunuela:
> @eurunuela TBH I totally lost track of that. I can have a look at it this weekend, would that work for you?

That would be very much appreciated!

tsalo (approved):
LGTM too!

Issue #811, opened 2021-09-24
Calculate component-wise kappa/rho degrees of freedom based on weight maps


tsalo:
### Summary
Ever since #358, we've been calculating our voxel-wise model fit F-statistics on varying numbers of echoes. When we average across voxels, we're averaging F-statistics with different DOFs. Then, we use the full `n_echos` for calculating a single DOF for all components' averaged F-statistics. 

However, the actual degrees of freedom should vary across components. While the voxel-wise degrees of freedom will be consistent across components, the weights we use to average across voxels vary across components, which means that the ultimate average DOF will vary as well.

I very much doubt that this has a large impact on the the classification, but it seems like it would be beneficial to deal with for internal consistency.

### Additional Detail

This stems from https://github.com/ME-ICA/tedana/pull/809#discussion_r715100247.

### Next Steps

1. Add a new metric that calculates the weighted average degrees of freedom of kappa and rho values based on each component's weight map and the adaptive mask.
2. Use the new metric in `getfbounds` for . [`scipy.stats.f.ppf`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.f.html) supports float values for the `dfd` parameter.

PR #1177, opened 2025-02-25
Adding a degree of freedom (dof) flag (--n-independent-echos) to allow users to set the dof used in fstat calculations 

### LINKED ISSUE #1168: No BOLD components detected error
Hi – I am attempting to run Tedana v24.0.2 on EPTI data (https://martinos.org/~fw089/publications.html; 122 TEs, spanning 4.405 ms to 63.695 ms, 3mm isotropic voxels, TR = 1.7 seconds, MB factor = 2, ~15 minutes total). When I view the raw data, there are no obvious artifacts or indications that something went wrong with the acquisition, looks good

Hoewever, Tedana fails to classify any of the components as BOLD-like, despite there being obvious signals of interest  (see an example below; by the way, Tedana interpreted the TR in the header file as 1700 seconds, not 1700 milliseconds, which is why the FFT and time axis looks so strange, I have since changed the TR in the headers to be 1.7, which solved that issue). 

WARNING  tedana:tedana_workflow:834 No BOLD components found, but maximum number of restarts reached.
WARNING  tedana:tedana_workflow:913 No BOLD components detected! Please check data and results!

Image

One other thing to note is that in the course of my troubleshooting, I found that if I use 3 TEs only (the first, middle, and last TE), Tedana performs as expected, the signal classifications look fine, 

Does anyone have any ideas about what might be the issue here, or suggestions for things to troubleshoot when encountering this particular error? I can share the data if it would be helpful, 

Thank you!
Chuck

katielamar:
Hi everyone, 
I've updated the tedana code to address issue #1168. I've made this a draft PR since this is my first time contributing to Tedana and I'm a bit new to submitting pull requests. Let me know if you have any suggestions on how I can improve the documentation/code. 

Closes #1168.

**Summary of Updates:**

- User can now set the degree of freedom that will be used for the fstat calculation. **More info:** #1168.
- <ins>Tedana from CMD Line:</ins> Use ``` --n-independent-echos ``` to set the degrees of freedom when running  tedana workflow from the command line
- <ins>Tedana from Python:</ins> Set the variable ```echo_dof``` when calling the [```tedana_workflow```](https://github.com/ME-ICA/tedana/blob/main/tedana/workflows/tedana.py) function


tedana/metrics/dependence.py
```diff
@@ -201,7 +205,10 @@ def calculate_f_maps(
             pred_s0 = x1[:j_echo, :] * np.tile(coeffs_s0, (j_echo, 1))
             sse_s0 = (comp_betas[:j_echo] - pred_s0) ** 2
             sse_s0 = sse_s0.sum(axis=0)  # (S,) prediction error map
-            f_s0 = (alpha - sse_s0) * (j_echo - 1) / (sse_s0)
+            if echo_dof is None:
+                f_s0 = (alpha - sse_s0) * (j_echo - 1) / (sse_s0)
+            else:
+                f_s0 = (alpha - sse_s0) * (echo_dof - 1) / (sse_s0)
```
tsalo, line 211: This part is the crux of the issue, since we only use good echoes to calculate F-statistics. In the February dev call we talked about this a bit, and I think the issue with trying to infer the correct degrees of freedom for EPITI when you have a specific subset of "good" echoes is that independent DOFs contribute to multiple echoes, so there's no easy way to determine which DOFs are present in a given subset of echoes. Is that right?

handwerkerd: Since echo_dof is likely to be 3 or 4, I don't think this will be a huge issue. That said, overarching thing to address is when to mask out voxels with the adaptive mask. Right now we'll use all voxels with 3 good echoes, which means TE=1, 2, & 3ms with lots of overlapping signals EPTI. I don't think this will be a huge issue with EPTI so I don't think we need to address it, but I do have a thought for how to address it.

When echo_dof is present, instead of setting the threshold to `adaptive_mask>=3` we set it to `adaptive_mask>(2/echo_dof)`
Given EPTI data with 100 echoes, if `echo_dof==3` then the threshold is 67. If `echo_dof==4` then the threshold is 51.
To address @tsalo concern, we could also add a linear mapping so, if `echo_dof==4` then 3 is used for a value of 51-75 in the adaptive mask and 4 is used for 76-100 in the adaptive mask.

The weakness of this approach is that EPITI is solving a system of equations and info from at least three independent measurements will likely appear at a lower minimum echo. I'm not sure how much of an issue this will be. My big Q is how much dropout is observed in EPTI. For example, @katielamar  is using an EPTI sequence goes to TE>100ms. For standard multi-echo, TE>=66ms at 3T would show a lot of dropout. If EPTI is similar, my proposal would cut out a lot of voxels. If EPTI has less dropout, then this is a non-issue. If it still has non-trivial dropout at TE>=66ms then we might need to engage more with the EPTI reconstruction math to figure out a better rough estimate for when there is sufficient information from at least 3 independent data points (or add a conservative fudge factor).

katielamar: We have very minimal signal loss, so the adaptive mask consistently indicates that all of the voxels with signal have well over 100 good echoes. For EPTI, the adaptive mask isn't actually removing anything at all. So I don't think we need to worry about voxels being thrown out due to the adaptive mask.

handwerkerd: I think I mis-remembered past presentations where I thought you collected 100ms echoes (or mixing it up with something I saw Fuyixue present)

The current code is effectively ignoring the adaptive mask and any situation where there might be fewer than echo_dof independent echoes. Given your description of the lack of EPTI dropout, I'm fine with that. We might want to add an INFO or WARNING to the log to make clear that the adaptive mask isn't functioning as intended with EPTI data.

katielamar: Sounds good. I updated the documentation like we discussed. Let me know if further changes need to be made. Ill change this draft PR to a regular PR now. Also, if you let me know where a good place would be to put the INFO/WARNING I can go ahead and add that. Maybe when the tedana_workflow is first called?

handwerkerd: I'm thinking we might want this to be in `make_adaptive_mask` since that's where the issue might arise & that function is also called by the `t2smap` workflow. The annoyance is we'd need to pass `echo_dof` just to make that work. 

Here's some proposed warnings that be added after this line: https://github.com/ME-ICA/tedana/blob/894eb7719c04102c35f7e431d18b6a2a51e72b35/tedana/utils.py#L231

```
    if isinstance(echo_dof, int):
        # For EPTI sequences, the way we use adaptive mask thresholding fails
        # because sequential echoes have overlapping information.
        # Since EPTI has less dropout, it is unclear how often this will cause issues.
        # To track this, we are flagging voxels that might mark less independant signal.
        # If such voxels appear often, this would show we might need to alter how the mask is used.

        # The thresh where there might not be 3 independent sources of data within the good echoes
        # For n_echos=100 & echo_dof=3, threshold_dof=66.6.
        # For n_echos=100 & echo_dof=4, threshold_dof=50
        threshold_3dof = 2 * n_echos / echo_dof

        n_3dof_voxels = np.sum(np.logical_and(adaptive_mask < threshold_3dof, adaptive_mask >= 1))
        perc_3dof_voxels = 100 * n_3dof_voxels / np.sum(adaptive_mask >= 1)

        if perc_3dof_voxels > 0:
            LGR.warning(
                f"{n_3dof_voxels} voxels ({perc_3dof_voxels}%) have fewer than {threshold_3dof} "
                "good voxels. These voxels will be used in all analyses, "
                "but might not include 3 independant echo measurements."
            )

        # There's a separate warning about DOF if it's possible there's a DOF reduction.
        if echo_dof > 3:
            # The threshold where the loss of good echoes might affect the DOF
            # For n_echos=100 & echo_dof=4, threshold_dof=75
            threshold_dof = (echo_dof - 1) * n_echos / echo_dof
            n_dof_voxels = np.sum(
                np.logical_and(adaptive_mask < threshold_dof, adaptive_mask >= 1)
            )
            perc_dof_voxels = 100 * n_dof_voxels / np.sum(adaptive_mask >= 1)
            LGR.warning(
                f"{n_dof_voxels} voxels ({perc_dof_voxels}%) have fewer than {threshold_dof} "
                "good voxels."
                f"The degrees of freedom for fits across echoes will remain {echo_dof} even if "
                "there might be fewer independant echo measurements."
            )
```

You'd also need to pass `echo_dof` to the function (or place this somewhere else). I'll admit that I may have overengineered this, but I've also found that code for descriptive warnings is sometimes longer than the actual code changes.

I could also submit this as a PR to your PR, but figured this might be easier.

handwerkerd (changes requested):
Besides the one ongoing discussion about a warning message, the rest of this looks good.

We should definitely include some tests. I can help with those, if useful.

katielamar:
@handwerkerd  The tests/changes you made are in the latest commit. I also made a change to fix the [fstat map test](https://github.com/katielamar/tedana/blob/ab07715ec4cf455442c4e58860983a878b5c9032/tedana/tests/test_metrics.py#L232) you added for echo_dof.
```
assert np.min(f_t2_maps_orig - f_t2_maps) >= 0
assert np.min(f_s0_maps_orig - f_s0_maps) >= 0
```

One of the asserts is still failing in [fstat map test](https://github.com/katielamar/tedana/blob/ab07715ec4cf455442c4e58860983a878b5c9032/tedana/tests/test_metrics.py#L232). Its caused by the following failing. Should this test be strictly greater than 0? I think maybe the f_max may be preventing that. Can you take a look?
```
assert np.min(f_t2_maps_orig[adaptive_mask == 5] - f_t2_maps[adaptive_mask == 5]) > 0
assert np.min(f_s0_maps_orig[adaptive_mask == 5] - f_s0_maps[adaptive_mask == 5]) > 0
```

tedana/tests/test_metrics.py
```diff
@@ -207,15 +207,35 @@ def test_smoke_calculate_f_maps():
...
+        echo_dof=3,
         f_max=500,
     )
     assert f_t2_maps.shape == f_s0_maps.shape == (n_voxels, n_components)
+    # When echo_dof < the number of echoes, then f_maps_orig should have the same or larger values
+    assert np.round(np.min(f_t2_maps_orig - f_t2_maps), decimals=3) == np.round(0.0, decimals=3)
+    assert np.round(np.min(f_s0_maps_orig - f_s0_maps), decimals=3) == np.round(0.0, decimals=3)
+    # When echo_dof==3, there are 5 good echoes, and f_maps_orig>0
+    # then f_maps_orig should always be larger than f_maps with fewer DOF
+    echo5_mask = np.logical_and(np.tile(adaptive_mask == 5, (50, 1)).T, f_t2_maps_orig > 0.0)
```
handwerkerd, line 236:
```suggestion
    echo5_mask = np.logical_and(
        np.logical_and(
            np.tile(adaptive_mask == 5, (50, 1)).T,
            f_t2_maps_orig > 0.2,
        ),
        f_t2_maps_orig < 499,
    )
```

Still crashed. I figured out the issue is, if the random inputs results in an f>500, then both are set to 500, which is 0. I edited to skip this test on values greater than 499 (and also slightly increased the minimum to avoid crashes from voxels that randomly ended up with f=0.

tedana/tests/test_metrics.py
```diff
@@ -207,15 +207,41 @@ def test_smoke_calculate_f_maps():
...
+    # then f_maps_orig should always be larger than f_maps with fewer DOF
+    echo5_mask = np.logical_and(
+        np.logical_and(
+            np.tile(adaptive_mask == 5, (50, 1)).T,
+            f_t2_maps_orig > 0.2,
+        ),
+        f_t2_maps_orig < 499,
+    )
+    assert np.min(f_t2_maps_orig[echo5_mask] - f_t2_maps[echo5_mask]) > 0.0
+    assert np.min(f_s0_maps_orig[echo5_mask] - f_s0_maps[echo5_mask]) > 0.0
```
handwerkerd, line 244:
```suggestion
    t2_echo5_mask = np.logical_and(
        np.logical_and(
            np.tile(adaptive_mask == 5, (n_components, 1)).T,
            f_t2_maps_orig > 0.2,
        ),
        f_t2_maps_orig < 499,
    )
    assert np.min(f_t2_maps_orig[t2_echo5_mask] - f_t2_maps[t2_echo5_mask]) > 0.0
    s0_echo5_mask = np.logical_and(
        np.logical_and(
            np.tile(adaptive_mask == 5, (n_components, 1)).T,
            f_s0_maps_orig > 0.2,
        ),
        f_s0_maps_orig < 499,
    )
    assert np.min(f_s0_maps_orig[s0_echo5_mask] - f_s0_maps[s0_echo5_mask]) > 0.0
```
Realized that f_t2_maps was fine, but f_s0_maps was not. Frustrating that this seems to reliably fail here, but reliably passes on my computer.

handwerkerd:
Since the tests were passing on my local computer & failing here, I decided to push some commits to see if I could just solve it. I think the key problem was we have a maximum F value set to 500. That means, for a test >500, the F is a constant 500 and  the difference between DOF==3 vs 5 is 0. Since the goal of the test was to make sure the change in DOF always changed the F stats, I needed to mask out the F==500 tests for both f_t2 and f_s0.

handwerkerd (approved):

handwerkerd:
@tsalo Do you want to take another look or is this good to merge?

tedana/stats.py
```diff
@@ -11,24 +11,26 @@
...
     fmin, fmid, fmax : :obj:`float`
         F-statistic thresholds for alphas of 0.05, 0.025, and 0.01,
         respectively.
     """
-    f05 = stats.f.ppf(q=(1 - 0.05), dfn=1, dfd=(n_echos - 1))
-    f025 = stats.f.ppf(q=(1 - 0.025), dfn=1, dfd=(n_echos - 1))
-    f01 = stats.f.ppf(q=(1 - 0.01), dfn=1, dfd=(n_echos - 1))
+    f05 = stats.f.ppf(q=(1 - 0.05), dfn=1, dfd=(echo_dof - 1))
+    f025 = stats.f.ppf(q=(1 - 0.025), dfn=1, dfd=(echo_dof - 1))
+    f01 = stats.f.ppf(q=(1 - 0.01), dfn=1, dfd=(echo_dof - 1))
```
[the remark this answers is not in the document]
tsalo: Or just change the `echo_dof` variable to `n_independent_sources`/`n_independent_echos`.

tedana/utils.py
```diff
@@ -229,9 +233,41 @@ def make_adaptive_mask(data, mask, threshold=1, methods=["dropout"]):
...
+        n_3dof_voxels = np.sum(np.logical_and(adaptive_mask < threshold_3dof, adaptive_mask >= 1))
+        perc_3dof_voxels = 100 * n_3dof_voxels / np.sum(adaptive_mask >= 1)
+        if perc_3dof_voxels > 0:
+            LGR.warning(
+                f"{n_3dof_voxels} voxels ({np.round(perc_3dof_voxels, decimals=1)}%) have fewer "
+                f"than {np.round(threshold_3dof)} "
+                "good voxels. These voxels will be used in all analyses, "
+                "but might not include 3 independant echo measurements."
+            )
+        # There's a separate warning about DOF if it's possible there's a DOF reduction.
```
tsalo, line 255:
```suggestion

        # There's a separate warning about DOF if it's possible there's a DOF reduction.
```

tedana/utils.py
```diff
@@ -229,9 +233,41 @@ def make_adaptive_mask(data, mask, threshold=1, methods=["dropout"]):
             "Removing voxels from mask."
         )
         adaptive_mask[adaptive_mask < threshold] = 0
-
+    if isinstance(echo_dof, int):
+        # For EPTI sequences, the way we use adaptive mask thresholding fails
+        # because sequential echoes have overlapping information.
+        # Since EPTI has less dropout, it is unclear how often this will cause issues.
+        # To track this, we are flagging voxels that might mark less independant signal.
```
tsalo, line 240:
```suggestion
        # To track this, we are flagging voxels that might mark less independent signal.
```

tedana/utils.py
```diff
@@ -229,9 +233,41 @@ def make_adaptive_mask(data, mask, threshold=1, methods=["dropout"]):
...
+        # For n_echos=100 & echo_dof=3, threshold_dof=66.
+        # For n_echos=100 & echo_dof=4, threshold_dof=50
+        threshold_3dof = np.floor(2 * n_echos / echo_dof)
+        n_3dof_voxels = np.sum(np.logical_and(adaptive_mask < threshold_3dof, adaptive_mask >= 1))
+        perc_3dof_voxels = 100 * n_3dof_voxels / np.sum(adaptive_mask >= 1)
+        if perc_3dof_voxels > 0:
+            LGR.warning(
+                f"{n_3dof_voxels} voxels ({np.round(perc_3dof_voxels, decimals=1)}%) have fewer "
+                f"than {np.round(threshold_3dof)} "
+                "good voxels. These voxels will be used in all analyses, "
```
tsalo, line 252:
```suggestion
                f"than {np.round(threshold_3dof)} good voxels. "
                "These voxels will be used in all analyses, "
```

tedana/selection/selection_nodes.py
```diff
@@ -714,10 +714,19 @@ def calc_kappa_elbow(
...
     are called
     """
+    if (
+        "echo_dof" in selector.cross_component_metrics_.keys()
+        and selector.cross_component_metrics_["echo_dof"]
+    ):
+        echo_dof = selector.cross_component_metrics_["echo_dof"]
+    else:
+        # DOF is number of echoes if not otherwise specified
+        echo_dof = selector.cross_component_metrics_["n_echos"]
```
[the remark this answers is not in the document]
handwerkerd: ug. I'm fairly sure in all calculations that use echo_dof, it's `echo_dof - 1`. The math is correct, but the terminology is wrong. This also brings up an issue for @katielamar in that, if echo_dof is the true DOF, should we run calculations with `echo_dof` and not `echo_dof - 1`?

I think a terminology-consistent solution should be to make `echo_dof` the DOF, when we calculate this based on the number of echoes, `echo_dof = n_echoes - 1` and we'd then need to update all calculations that use these values to not subtract another `-1`.

tedana/metrics/collect.py
```diff
@@ -224,7 +230,11 @@ def generate_metrics(
 
     if "map FT2 clusterized" in required_metrics:
         LGR.info("Calculating T2* F-statistic maps")
-        f_thresh, _, _ = getfbounds(len(tes))
+        if echo_dof is None:
+            f_thresh, _, _ = getfbounds(len(tes))
+        else:
+            f_thresh, _, _ = getfbounds(echo_dof)
```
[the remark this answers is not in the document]
handwerkerd: I agree on this change.
Addressing #811 might be useful, but seems beyond the scope of this PR. If we're clarifying which inputs are supposed to be DOF, this might be easier to address.

tedana/selection/selection_nodes.py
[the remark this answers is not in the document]
handwerkerd: You comment below of changing `echo_dof` to `n_independent_sources` is another solution that would not require changing formulas.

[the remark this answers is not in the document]
katielamar: Hmm yeah I'm not sure. Im going to check with Tom on this and Ill get back to you.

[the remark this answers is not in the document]
handwerkerd: I do think @tsalo is correct here. My proposal is:
- For any stat function, like `getfbounds` the parameter should be `n_independent_sources`
- For any function that is specifically about the number of echoes (i.e. input to tedana.py) the parameter should be `n_independent_echoes` or `n_indep_echoes` or 'n_indie_echoes`

This would mean that the user facing parameter has a clearer meaning that more accurately matches what we're currently doing in the code. This might mess up existing scripts that Katie & Charles are using, but hopefully that won't be a huge problem.

@katielamar I know you're preparing your talk right now. Tell me if you want me to make this change & open another PR on your branch for you to review.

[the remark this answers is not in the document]
katielamar: Hmm the only part I'm unsure about is if we should be subtracting the 1 or if the # of components is the true dof and we should be removing the -1 part in the fstat calculation instead.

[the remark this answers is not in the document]
katielamar: Perhaps you can go ahead and make another PR with the change to `n_indep_echoes` and I can decide later on whether I should set `n_indep_echoes` to the # of bases or the # of bases + 1.

katielamar:
@handwerkerd some of the tests are failing. I merged your PR with my repo and made an update to get one of the unit tests to pass (changed np.concat to np.concatenate) and now a bunch of tests that were once passing are failing. Can you take a look? It seems to be due to an assertion you put in for the adaptive mask test.

tedana/utils.py
```diff
@@ -229,9 +233,43 @@ def make_adaptive_mask(data, mask, threshold=1, methods=["dropout"]):
...
+            # For n_echos=100 & n_independent_echos=4, threshold_dof=75
+            threshold_dof = np.floor((n_independent_echos - 1) * n_echos / n_independent_echos)
+            n_dof_voxels = np.sum(
+                np.logical_and(adaptive_mask < threshold_dof, adaptive_mask >= 1)
+            )
+            perc_dof_voxels = 100 * n_dof_voxels / np.sum(adaptive_mask >= 1)
+            LGR.warning(
+                f"{n_dof_voxels} voxels ({np.round(perc_dof_voxels, decimals=1)}%) have fewer"
+                f"than {np.round(threshold_dof)} good voxels."
+                f"The degrees of freedom for fits across echoes will remain {n_independent_echos}"
```
handwerkerd, line 269:
```suggestion
                f"{n_dof_voxels} voxels ({np.round(perc_dof_voxels, decimals=1)}%) have fewer "
                f"than {np.round(threshold_dof)} good voxels. "
                f"The degrees of freedom for fits across echoes will remain {n_independent_echos} "
```

There might be other issues, but this definitely the cause of some crashes. There are automatic style changes on commit, and I think it adjusted the lines of this message and got rid of extra spaces. The text read "fewerthan" which correctly flagged an error for the unit test.

handwerkerd:
> @handwerkerd some of the tests are failing. I merged your PR with my repo and made an update to get one of the unit tests to pass (changed np.concat to np.concatenate) and now a bunch of tests that were once passing are failing. Can you take a look? It seems to be due to an assertion you put in for the adaptive mask test.

I think found the issue. A warning message in `make_adaptive_mask` was automatically reformatted due to tedana using `pre-commit` & that removed some necessary spaces in the message. There were also some failed tests due to being unable to connect to OSF, but I think that's (again) an internet issue & not a code issue. Besides that, the code passes tests on my computer & it looks like the previously failing tests are now passing here.

@tsalo Anything else to address before merging?

tedana/utils.py
```diff
@@ -229,9 +233,43 @@ def make_adaptive_mask(data, mask, threshold=1, methods=["dropout"]):
             "Removing voxels from mask."
         )
         adaptive_mask[adaptive_mask < threshold] = 0
+    if isinstance(n_independent_echos, int):
+        # For EPTI sequences, the way we use adaptive mask thresholding fails
```
tsalo, line 237:
```suggestion

    if isinstance(n_independent_echos, int):
        # For EPTI sequences, the way we use adaptive mask thresholding fails
```

tedana/workflows/t2smap.py
```diff
@@ -132,6 +132,18 @@ def _get_parser():
         help=("Combination scheme for TEs: t2s (Posse 1999), paid (Poser)"),
         default="t2s",
     )
+    optional.add_argument(
+        "--n-independent-echos",
+        dest="echo_dof",
+        metavar="INT",
+        type=int,
+        help=(
+            "Degree of freedom to use in goodness of fit metrics (fstat)."
```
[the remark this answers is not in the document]
tsalo:
```suggestion
            "Number of independent echoes to use in goodness of fit metrics (fstat)."
```

tedana/utils.py
```diff
@@ -229,9 +233,41 @@ def make_adaptive_mask(data, mask, threshold=1, methods=["dropout"]):
             "Removing voxels from mask."
         )
         adaptive_mask[adaptive_mask < threshold] = 0
-
+    if isinstance(echo_dof, int):
+        # For EPTI sequences, the way we use adaptive mask thresholding fails
+        # because sequential echoes have overlapping information.
+        # Since EPTI has less dropout, it is unclear how often this will cause issues.
```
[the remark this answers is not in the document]
tsalo: Have you looked into this?

tedana/selection/selection_nodes.py
```diff
@@ -714,10 +714,19 @@ def calc_kappa_elbow(
...
     """
+    if (
+        "n_independent_echos" in selector.cross_component_metrics_.keys()
+        and selector.cross_component_metrics_["n_independent_echos"]
+    ):
+        n_independent_echos = selector.cross_component_metrics_["n_independent_echos"]
+    else:
+        # Number of independent echoes is the number of echoes if not otherwise specified
+        n_independent_echos = selector.cross_component_metrics_["n_echos"]
     outputs = {
```
tsalo, line 725:
```suggestion
        n_independent_echos = selector.cross_component_metrics_["n_echos"]

    outputs = {
```

tedana/utils.py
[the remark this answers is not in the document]
handwerkerd: Flagging for a response from @katielamar

tedana/metrics/collect.py
```diff
@@ -122,6 +127,10 @@ def generate_metrics(
     # Ensure that echo times are in an array, rather than a list
     tes = np.asarray(tes)
 
+    # use either the inputted number of indie echoes or the total number of echoes
+    # to calcualte the threshold for f tests
```
tsalo, line 131:
```suggestion
    # to calculate the threshold for f tests
```

tsalo (approved):
The changes look good to me. I have one small suggestion to fix a typo. If you could merge that suggestion before merging this PR, that would be great. No need to request a fresh review after that.

One code comment that I haven't gotten clarification on is the claim that "EPTI has less dropout". Is that accurate or more a factor of having more echoes at shorter echo times? For example, for an echo at 40 ms, will EPTI show less dropout than EPI?

handwerkerd:
@all-contributors please add @katielamar for code, design, ideas

PR #1179, opened 2025-03-01
Add verbose option to ica_reclassify to obtain denoised echo images

### LINKED ISSUE #1178: ica_reclassify doesn't have an option to return all denoised echoes
### Summary

Unlike the `tedana` command, which has the `--verbose` option, `ica_reclassify` doesn't provide an option to save all denoised echoes. This forces users to only use the optimally combined, which can be undesirable in some cases like hemodynamic deconvolution.

### Next Steps
 
- Replicate `tedana`'s outputting of all denoised echoes in `ica_reclassify`.

eurunuela:
Closes #1178.

- Adds verbose option to `ica_reclassify`
- Changes the way input files are registered from relative paths to absolute paths
- Adds `verbose = True` to the four echo integration test

### DIFF
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -143,9 +143,9 @@ class OutputGenerator:
             del old_registry["root"]
             for k, v in old_registry.items():
                 if isinstance(v, list):
-                    self.registry[k] = [op.join(rel_root, vv) for vv in v]
+                    self.registry[k] = [op.normpath(op.join(rel_root, vv)) for vv in v]
                 else:
-                    self.registry[k] = op.join(rel_root, v)
+                    self.registry[k] = op.normpath(op.join(rel_root, v))
 
         if not op.isdir(self.out_dir):
             LGR.info(f"Generating output directory: {self.out_dir}")
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -203,7 +203,7 @@ class OutputGenerator:
         names : list[str]
             The list of filenames being input as multi-echo volumes.
         """
-        self.registry["input img"] = [name for name in names]
+        self.registry["input img"] = [op.relpath(name, start=self.out_dir) for name in names]
 
     def get_name(self, description, **kwargs):
         """Generate a file full path to simplify file output.
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -203,7 +203,7 @@ class OutputGenerator:
         names : list[str]
             The list of filenames being input as multi-echo volumes.
         """
-        self.registry["input img"] = [op.relpath(name, start=self.out_dir) for name in names]
+        self.registry["input img"] = [name for name in names]
 
     def get_name(self, description, **kwargs):
         """Generate a file full path to simplify file output.
@@ -414,7 +414,10 @@ class InputHarvester:
             Description of the file to get the path for.
         """
         if description in self._registry.keys():
-            return op.join(self._base_dir, self._registry[description])
+            if isinstance(self._registry[description], list):
+                return [op.join(self._base_dir, f) for f in self._registry[description]]
+            else:
+                return op.join(self._base_dir, self._registry[description])
         else:
             return None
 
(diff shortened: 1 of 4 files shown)


tedana/io.py
```diff
@@ -203,7 +203,7 @@ def register_input(self, names):
         names : list[str]
             The list of filenames being input as multi-echo volumes.
         """
-        self.registry["input img"] = [op.relpath(name, start=self.out_dir) for name in names]
```
handwerkerd, line 206: My only Q is why change from a relative path to an absolute path? The key benefit of a relative path is that one can move the output directory (i.e. a BIDS derivatives directory) to somewhere else and the registry still works. Another benefit is that, for data that might be shared, an absolute path might contain identifiable info of whoever ran the script & potentially identifiable info if someone put PII higher up in a directory structure.

Any specific benefits of an absolute path?

eurunuela: The reason why I chose the absolute path is that I couldn't make it work with the relative path, as the output directory can be different for `ica_reclassify` compared to `tedana`.

handwerkerd: I'm going through the IO code and I'm not quite sure where the problem is arising. `ica_reclassify` takes both the registry from the original output directory and a separate output directly as input.

The IO object is initialized here with the full path to the registry (not the location of `out_dir` for `ica_reclassify`:
https://github.com/ME-ICA/tedana/blob/894eb7719c04102c35f7e431d18b6a2a51e72b35/tedana/workflows/ica_reclassify.py#L375

When the above is initialized, it saves `self._base_dir` which could appropriately be able to track the relative path of the input files
https://github.com/ME-ICA/tedana/blob/894eb7719c04102c35f7e431d18b6a2a51e72b35/tedana/io.py#L405

`get_file_path` does use `_base_dir` plus the relative path in the registry to locate the file.
https://github.com/ME-ICA/tedana/blob/894eb7719c04102c35f7e431d18b6a2a51e72b35/tedana/io.py#L417

This is all done without using the current directory location or `out_dir`. I haven't fully tested this myself, so maybe I'm missing something.

eurunuela: I must have missed that. I'm not very familiar with the new IO code.

I will have a look and make the changes to go back to relative paths. Thanks Dan!

handwerkerd: The IO code is very Josh, but I think I reviewed.

eurunuela: Okay. I reverted the changes to use relative paths and this time I didn't get the error I was getting before. I have no idea why relative paths didn't work in my computer last time I checked.

tsalo (approved):
LGTM!

tedana/io.py
```diff
@@ -414,7 +414,10 @@ def get_file_path(self, description):
             Description of the file to get the path for.
         """
         if description in self._registry.keys():
-            return op.join(self._base_dir, self._registry[description])
+            if isinstance(self._registry[description], list):
+                return [op.join(self._base_dir, f) for f in self._registry[description]]
+            else:
+                return op.join(self._base_dir, self._registry[description])
```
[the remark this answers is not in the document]
handwerkerd: I figured it out. Add `op.normpath` around the two `op.join` statements at this location. In `__init__`  for `OutputGenerator`

https://github.com/eurunuela/tedana/blob/0ecff202167bd772e9383d5a7f739119ec07e026/tedana/io.py#L146-L148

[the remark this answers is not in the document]
eurunuela: These relative paths have always been like this. I guess we never considered this use case of having to read all echoes back into tedana when Josh wrote the new IO.

Changes made in [3ba8c79](https://github.com/ME-ICA/tedana/pull/1179/commits/3ba8c7968885bc50f4871c3267b88a88af4954bc)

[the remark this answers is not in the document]
handwerkerd: I agree. I think this PR simply highlighted this messiness.

handwerkerd (approved):

PR #999, opened 2023-11-16
MNT: Test on 3.11 and 3.12

effigies:
Add GitHub tests for Python 3.11 and 3.12, updating pyproject.toml to reflect support.

While I'm at it, I updated the `black` target version to Python 3.8 and ran `pyupgrade --py38-plus`.

For reference, [pyupgrade](https://github.com/asottile/pyupgrade) is a small tool to migrate to new syntax or best practices as support windows drop older versions of Python.

PR #1183, opened 2025-03-07
Adding python313

handwerkerd:
Python 3.13 has been out long enough that I decided to see if tedana runs on it & it ran fine & all tests passed. 

- Adds python 3.13 to pyproject.toml as a version that can be used with tedana.
- Add python 3.13 to the circleci unit tests (if I correctly made those edits)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,6 +185,40 @@ jobs:
           paths:
             - src/coverage/.coverage.py312
 
+  unittest_313:
+    docker:
+      - image: continuumio/miniconda3
+    working_directory: /tmp/src/tedana
+    steps:
+      - checkout
+      - restore_cache:
+          key: conda-py313-v3-{{ checksum "pyproject.toml" }}
+      - run:
+          name: Generate environment
+          command: |
+            apt-get update
+            apt-get install -yqq make
+            if [ ! -d /opt/conda/envs/tedana_py313 ]; then
+              conda create -yq -n tedana_py313 python=3.13
+              source activate tedana_py313
+              pip install .[tests]
+            fi
+      - run:
+          name: Running unit tests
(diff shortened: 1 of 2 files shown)


handwerkerd:
I'm noticing this did not run unittest_313 so there's probably something else that needs to be done to get this working. I looked back to when [we added python 3.11 and 3.12](https://github.com/ME-ICA/tedana/pull/999) and I don't see anything obvious I missed, but I clearly missed something.

tsalo:
@handwerkerd you need to add the new test job to the workflow. That will get CircleCI to run it.

https://github.com/ME-ICA/tedana/blob/894eb7719c04102c35f7e431d18b6a2a51e72b35/.circleci/config.yml#L379-L389

eurunuela:
I wonder if changing our `.circleci/config.yml` to use a grid of versions would be easier to maintain going forward. What do you think?

handwerkerd:
> I wonder if changing our `.circleci/config.yml` to use a grid of versions would be easier to maintain going forward. What do you think?
> 
> Here's an example: https://github.com/eurunuela/pySPFM/blob/main/.circleci/config.yml

I'm open to making this change, but don't know enough to have a strong opinion & I'm probably not the right person to rewrite the full `config.yml`

FWIW, https://app.codecov.io/gh/ME-ICA/tedana/pull still isn't working for any of our PRs. I'm not sure if this is an issue with our PRs or how codecov is called through this config.

tsalo (approved):
LGTM. I'm happy to switch to a grid approach in a separate PR.

FWIW, we've been seeing a similar issue with CodeCov in CuBIDS and no one in Ted's lab has been able to figure it out.

eurunuela (approved):

PR #1184, opened 2025-03-09
Add `--ctab` parameter to `ica_reclassify`

### LINKED ISSUE #1182: Accept `--ctab` argument in ica_reclassify
### Summary

I am working on a workflow to combine tedana and ICA-AROMA, and there doesn't seem to be a clean way to use ica_reclassify to re-denoise the data once I've combined the classifications from tedana and ICA-AROMA.

My workflow involves running ICA-AROMA, grabbing the mixing matrix from that, and using that to run tedana. Then I take the component table from each and combine the classifications and tags. If I just rely on manacc and manrej, I lose the rationale behind each component's classification. Instead, I'd like to use a component table with the appropriate classifications and rationales (either overwritten in the original tedana directory, which is suboptimal, or as a new argument).

tsalo:
Closes #1182.

- Add a `--ctab` parameter to `ica_reclassify` that accepts a component table.
- If `--manacc`, `--manrej`, and `--ctab` are not provided, then raise an error (was being raised if `--manacc` and `--manrej` weren't provided).

### DIFF
--- a/tedana/tests/test_integration.py
+++ b/tedana/tests/test_integration.py
@@ -353,7 +353,8 @@ def test_integration_reclassify_no_manual(skip_integration):
     args = ["ica_reclassify", reclassify_raw_registry(), "--out-dir", out_dir]
 
     result = subprocess.run(args, capture_output=True)
-    assert result.returncode == 0
+    assert b"ValueError: No updated classifications provided" in result.stderr
+    assert result.returncode != 0
 
 
 def test_integration_reclassify_quiet_csv(skip_integration):
--- a/tedana/tests/test_integration.py
+++ b/tedana/tests/test_integration.py
@@ -335,13 +335,17 @@ def test_integration_three_echo_external_regressors_motion_task_models(skip_inte
     check_integration_outputs(fn, out_dir)
 
 
-def test_integration_reclassify_insufficient_args(skip_integration):
+def test_integration_reclassify_no_manual(skip_integration):
+    """Run reclassification workflow without providing any manual accept/reject arguments.
+
+    This should result in the component table being used for classification.
+    """
     if skip_integration:
         pytest.skip("Skipping reclassify insufficient args")
 
     guarantee_reclassify_data()
 
-    test_data_path, osf_id = data_for_testing_info("three-echo")
+    test_data_path, _ = data_for_testing_info("three-echo")
     out_dir = os.path.abspath(
         os.path.join(test_data_path, "../../outputs/reclassify/insufficient")
     )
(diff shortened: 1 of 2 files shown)


eurunuela (changes requested):
LGTM. I do think we should update the docs before merging this PR in.

tsalo:
Is ica_reclassify documented in detail outside of the argparse-based documentation in https://tedana.readthedocs.io/en/stable/usage.html#running-the-ica-reclassify-workflow?

tsalo:
Ironically, it looks like the [documentation on using RICA](https://tedana.readthedocs.io/en/stable/approach.html#manual-classification-with-rica) is outdated and actually matches what this PR proposes:

> Once the .tsv file containing the result of manual component classification is obtained, it is necessary to re-run the tedana workflow (see [Running the ica_reclassify workflow](https://tedana.readthedocs.io/en/stable/usage.html#running-the-ica-reclassify-workflow)) passing the manual_classification.tsv file with the --ctab option.

eurunuela:
> Is ica_reclassify documented in detail outside of the argparse-based documentation in https://tedana.readthedocs.io/en/stable/usage.html#running-the-ica-reclassify-workflow?

You may be right. I think that may be the only documentation we have. For some reason I thought I remembered a different section.

eurunuela (approved):

handwerkerd:
>  For my use case, I don't need that status table- it's still available in the original tedana directory, and I'm not planning to delete that. I just want to apply a component table + mixing matrix to my data to get an updated orthogonalized mixing matrix that I can use for later denoising. It would be nice to have some of the other bells and whistles (esp. an HTML report with component plots), but that's not a dealbreaker. It might just be easier for me to write a short script to handle this for my data.

If that's your main goal, then I'd recommend just sending the new list of components to reject to tedana and adding functionality to have an `AROMA rejected` classification tag. This is a really easy change.

You'll still have the component table you created, but it wouldn't mess with how the component table created by tedana interacts with other parts. If you want a fully combined component table, then we should do something slightly more expansive that merges the two component tables within tedana.

tsalo:
We discussed this a bit on Mattermost, so I'll try to summarize our conversation.

For most users, allowing `ica_reclassify` to apply a single tag to the rejected and/or accepted components will probably be enough. For my use case, it's just not enough information. To be honest, I find the current classification tags not particularly informative: a component being "unlikely BOLD" isn't useful, given that the component was rejected, so it has to be "unlikely BOLD". With my AROMA+tedana combination, I want to know which particular criteria resulted in each classification, so I need the AROMA rationales. Each component may be rejected for multiple reasons, and the reasons will vary across components, which would be difficult to translate to command line arguments. I recognize that most users won't want that level of information, which is part of why we moved away from the rationale codes to broader classification tags, so I'll stick with my custom code for this step.

Basically, I'll close this PR in favor of a followup PR from @handwerkerd that adds classification tag parameters, and I'll use custom code for my AROMA+tedana denoising.

PR #1190, opened 2025-03-11
Change CircleCI versioning to use version matrices

### LINKED ISSUE #1189: Switch CircleCI to matrix of versions
### Summary

Stems off #1183. I suggested changing our test setup to use a grid of versions, which should make adding new Python versions easier.

I am happy to take the lead on this.

eurunuela:
Closes #1189 

- Uses version matrices instead of duplicating code to set a new version.
- Uses single version for integration tests.
- Uses all version in matrix for unit tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,250 +6,109 @@ version: 2.1
 orbs:
   codecov: codecov/codecov@1.0.5
 jobs:
-  makeenv_38:
+  py_env:
+    parameters:
+      PYTHON:
+        type: string
+    working_directory: /tmp/src/tedana
     docker:
       - image: continuumio/miniconda3
-    working_directory: /tmp/src/tedana
     steps:
       - checkout
-      - restore_cache:
-          key: conda-py38-v3-{{ checksum "pyproject.toml" }}
-      - run:
-          name: Generate environment
-          command: |
-            if [ ! -d /opt/conda/envs/tedana_py38 ]; then
-              conda create -yq -n tedana_py38 python=3.8
-              source activate tedana_py38
-              pip install -e .[tests]
-            fi
-      - save_cache:
-          key: conda-py38-v3-{{ checksum "pyproject.toml" }}
-          paths:
-            - /opt/conda/envs/tedana_py38
-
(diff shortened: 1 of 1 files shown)


eurunuela:
I don't know how to we can get rid of these old tests that say "required".

If we want to, we can now easily run the integration tests in all these version as well.

handwerkerd:
It looks like your changes added a bunch of additional tests. I think we previously just created py_env-3.8 and now this is creating environments for all listed versions and also running some or all of the integration tests for multiple python versions. This isn't necessary a bad thing to do, but it will take a lot more time.

eurunuela:
> It looks like your changes added a bunch of additional tests. I think we previously just created py_env-3.8 and now this is creating environments for all listed versions and also running some or all of the integration tests for multiple python versions. This isn't necessary a bad thing to do, but it will take a lot more time.

It creates an environment for every version we want to test on. The unit tests are run on all those version, but I limited the integration tests to run on a single one. We could add more in the future if wanted, but thought this would take too long. Effectively, it runs the same amount of tests as before, you're just seeing the new environment creation as a separate item on the list.

eurunuela:
> I'm seeing 6 tests that have been hanging for a while that seem to be other versions of passed tests:
> 
> * `three-echo` (hanging) and `three-echo-3.9` (passed)
> * `four-echo` (hanging) and `four-echo-3.9` (passed)
> * `five-echo` (hanging) and `five-echo-3.9` (passed)
> * `unittest_38` (hanging) and `unittest-3.8` (passed)
> * `unittest_39` (hanging) and `unittest-3.9` (passed)
> * `unittest_310` (hanging) and `unittest-3.10` (passed)

The hanging tests are not in the config file. I believe CircleCI is expecting them (and says are required) because they exist in the config file in `main`.

handwerkerd:
I'm fine with this, but if we keep different names for the tests, let's merge this after #1177 just so that this doesn't mess up the tests there.

eurunuela:
@handwerkerd feel free to merge this now that #1177 got merged.

handwerkerd (approved):

PR #1199, opened 2025-03-21
Add logging for tedana newsletter subscription information

eurunuela:
Closes #1198

Changes proposed in this pull request:

- Adds a new function to `utils.py` to avoid rewriting the same logs in all 3 workflows.
- Calls the `utils.log_newsletter_info()` at the end of each workflow.

Issue #1192, opened 2025-03-18
Codecov not running through github/CircleCI


handwerkerd:
### Summary

We are supposed to automatically run codecov on PRs, but it stopped working a couple of months ago.  For most of the recent PRs ( https://app.codecov.io/gh/ME-ICA/tedana/pulls ) there's the message

```
Missing Head Commit
Unable to compare commits because the head commit was not found.
[Learn more here](https://docs.codecov.com/docs/error-reference#section-missing-head-commit)
```

### Additional Detail

I think we can still run codecov locally so it's more an issue with how it's setup to automatically run through github/CircleCI. Google is failing me for this one. I see ways to fix headless commits, but I'm just not sure what's the head for however Github interacts with CircleCI & codecov.

I've already run this by our currently more active contributors but maybe @effigies or @emdupre have suggestions.

handwerkerd:
@aptinis noticed Codecov is using a commit from May 2024 as the source/latest commit on the main branch, but doesn't see an obvious reason what it stopped updating. She also asked why we're using codecov v1.0.5 (from 2019) vs the updated v5.3.0. I remember we had a discussion a while ago about why we stuck with 1.0.5, but maybe something has changed since then & 1.0.5 is now the problem.

eurunuela:
I tried updating to the latest in one of my PRs recently and it didn't solve the issue.

handwerkerd:
I wonder if we need to update the version in `main` since that's what codecov is having trouble finding.

Also, I found our codecov token. For reference, it's https://app.codecov.io/gh/ME-ICA/tedana/config/general and should be viewable by people who can log into our codecov account. I'm not quite sure how to add this to our CircleCI secrets (or to check if it's already there).
I think this is the right guide, but I'm not quite sure how to apply it. https://circleci.com/docs/security-recommendations/

effigies:
https://app.circleci.com/settings/project/github/ME-ICA/tedana/environment-variables

handwerkerd:
Thank you @effigies. I just added the token.

We probably should also update the version. Two people said the current version is 5.3.0, but https://docs.codecov.com/changelog/changelog-update-v23xxx-and-later seems to be saying they've changed versioning to CalVer style after 5.0.1 and the current version seems to be 25.3.3. I might be overthinking here, but should we change to 5.3.0 or 25.3.3?

effigies:
[...]

PR #1200, opened 2025-03-24
Adding classification tags options to `ica_reclassify`

### LINKED ISSUE #1182: Accept `--ctab` argument in ica_reclassify

handwerkerd:
Closes #1182.
This is in response to #1182 and discussion in #1184.

- `ica_reclassify` gets two new command line options `--tagacc` and `--tagrej`. Following the confection for `--manacc`, these are `tag_accept` and `tag_reject` in the workflow
- When `ica_reclassify` is run these tags will be added to all listed components, whether or not the classification changes. That means one could add a tag like "AROMA reject" and it would be added to the existing tags whether or not a component was already classified as rejected.
  - This was always the desired functionality within the API, but I noticed & corrected an edge case where this didn't work & now it is accessible via the CLI
- The inputted tags can be either a single tag or a comma delimited string (i.e. `--tagrej "AROMA reject reason 1, AROMA reject reason 2"`). I think this makes the full functionality @tsalo wants possible, but you'd need to run & write out new outputs for every different combination of new tags. This is inefficient.
- tests added for all new functionality and some existing tests were improved to confirm details of outputs while I was editing them.


handwerkerd:
Same CircleCI testing issue (see this comment: https://github.com/ME-ICA/tedana/pull/1199#issuecomment-2749403194 )

Also, codecov has the missing head commit message again (#1192) https://app.codecov.io/gh/ME-ICA/tedana/pull/1200

handwerkerd:
@tsalo This is ready for review. I'm not sure if you're going to use this feature, but I figured you'd be a good reviewer. If you don't have time, I can swap you out for someone else.

handwerkerd:
@eurunuela after #1202, Codecov is properly integrated and I noticed a few new lines that tests missed. The small, but nice pleasure of getting 100% testing coverage. 😄

tedana/selection/selection_utils.py
```diff
@@ -253,6 +253,27 @@ def comptable_classification_changer(
...
                         )
             selector.component_table_.loc[changeidx, "classification"] = classify_if
+
+        if tag_if is not None:  # only run if a tag is provided
+            # if tag_if has commas, divide into multiple tags
+            if "," in tag_if:
+                multi_tags = tag_if.split(",")
+                tag_if = set([s.strip() for s in multi_tags])
+            else:
+                tag_if = {tag_if}
```
tsalo, line 263: Is this necessary? If the input is a comma-delimited string and the field in the component table is a comma-delimited string, do you need to modify it?

tedana/selection/selection_utils.py
```diff
@@ -253,6 +253,27 @@ def comptable_classification_changer(
...
+            for idx in changeidx:
```
tsalo, line 264:
```suggestion

            for idx in changeidx:
```

tedana/tests/test_integration.py
```diff
@@ -188,19 +188,27 @@ def test_integration_four_echo(skip_integration):
...
         out_dir=out_dir_manual,
         mir=True,
         verbose=True,
     )
 
-    # compare the generated output files
-    fn = resource_filename("tedana", "tests/data/fiu_four_echo_outputs.txt")
-
-    check_integration_outputs(fn, out_dir)
+    component_table = pd.read_csv(op.join(out_dir_manual, "desc-tedana_metrics.tsv"), sep="\t")
```
tsalo, line 206:
```suggestion
    component_table = pd.read_table(op.join(out_dir_manual, "desc-tedana_metrics.tsv"))
```

tedana/workflows/ica_reclassify.py
```diff
@@ -232,16 +258,72 @@ def _parse_manual_list(manual_list):
...
+        for x in manual_vals:
+            if not isinstance(x, str):
+                raise ValueError(
+                    "_parse_manual_list_str expected a string or a list of strings, "
+                    f"but the input is {manual_list}"
+                )
+            elif "," in x:
+                raise ValueError(
+                    "_parse_manual_list_str includes a comma in a list of multiple strings. "
+                    "Input can include a comma deliminated string, but not multiple strings. "
```
tsalo, line 306:
```suggestion
                    "Input can include a comma delimited string, but not multiple strings. "
```

tedana/workflows/ica_reclassify.py
```diff
@@ -232,16 +258,72 @@ def _parse_manual_list(manual_list):
...
         raise ValueError(
-            f"_parse_manual_list expected integers or a filename, but the input is {manual_list}"
+            "_parse_manual_list_int expected integers or a filename, "
+            f"but the input is {manual_list}"
         )
 
     return manual_nums
 
 
+def _parse_manual_list_str(manual_list):
```
tsalo, line 268: It might be good to write a unit test for this function. It would be simpler to just test this function directly than to incorporate the test into the integration tests for ica_reclassify.

tsalo (commented):
I have a couple of minor suggestions/questions, but overall this makes sense to me.

tedana/selection/selection_utils.py
handwerkerd: This makes sure the same tag isn't repeated. That is, if the existing table has `tag1,tag2` and the new tags are `tag2,tag3` the result will be `tag1,tag2,tag3` without a repetitionof `tag2`

tsalo: Ah okay, thanks. I didn't realize this happened after the old and new tags were concatenated.

handwerkerd:
I think I've addressed all of @tsalo's feedback. To simplify the tests, I moved two functions in `ica_reclassify.py` into `workflows/parser_utils.py` and then added unit tests into `test_workflows_parser_utils.py`. This also let me remove a few integration tests for `ica_reclassify` that were just being used to test failures of the above functions.

FWIW, in setting this all up, I noticed that the docstring for `parse_manual_list_int`said it would identify delimiters that were commas, spaces, or tabs, but the code only used commas for delimiters. Since this hasn't caused any problems, I just updated the docstring to only mention commas.

tsalo (approved):
LGTM. I'm happy to have this close #1182.

PR #1201, opened 2025-03-25
CircleCI testing fix

handwerkerd:
#1190 didn't quite work and now we're getting testing failures on other PR. Some tests were renamed, but we can't figure out how to get github to ignore the older test names. Everything is running fine on CircleCI

- Rename the new tests to match the names of the old tests
  - Because the old names listed python versions as 310 instead of 3.10, the consistent way to do this was to replace all lists with python versions to just be 9, 10, 11, ... and then add `3` or `3.` as fixed text.
- Also changed the codecov orb to 5.3.0 to see if that fixes the separate codecov problem (It didn't fix that issue, but keeping at 5.3.0)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@
 #
 version: 2.1
 orbs:
-  codecov: codecov/codecov@5.2.1
+  codecov: codecov/codecov@5.3.0
 jobs:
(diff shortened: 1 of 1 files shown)


eurunuela (approved):
LGTM

Issue #920, opened 2022-12-22
Refactor the tedana.py workflow to better use modularized code


handwerkerd:
### Summary

After #756 is merged, some of the newly modularized functions could be better used in the `tedana.py` workflow. It should be possible to refactor that workflow to make it easier to understand and edit

### Additional Detail

While working on modularization, it became clear that a large proportion of the tedana.py workflow as conditional statements that existed only to make it possible to manually change component classifications  (the `--manacc` option) while skipping over most of the rest of the code. Given some of the added functionality that came with modularization, keeping that all in tedana.py would have made it even messier.  We removed all those conditional statements and created a distinct workflow: [`tedana_reclassify`](https://github.com/jbteves/tedana/blob/JT_DTM/tedana/workflows/tedana_reclassify.py). Some of the ways we used modularization in tedana_reclassify.py could also be used in tedana.py.

Making these changes will also make more dynamic and informative logging possible. One key example of this is that the list of references to cite are currently hard-coded into tedana.py. Decision tree specifications also include lists of references. When every workflow is modularized, we'd be able to make reference gathering more dynamic and useful.

### Next Steps

- Decide where this fits within broader priorities

handwerkerd:
#1188 is a feature request to make it possible to rerun tedana with the same component table. This should be very possible & should be included when someone finds time to work on the modularization of `tedana.py` as mentioned in this issue.

PR #1203, opened 2025-04-02
add warning for when tedana calculates mask

### LINKED ISSUE #1176: The make_adaptive_mask method results in significant loss of voxels
I used the following command to run tedana:

`tedana -e $TE -d e1_in.nii.gz e2_in.nii.gz e3_in.nii.gz e4_in.nii.gz --out-dir="${Bids_dir}/${sub_no}/${cond}/tedana_nomask" --tedpca kundu-stabilize
`

However, I noticed that the result in desc-adaptiveGoodSignal_mask.nii.gz contains a significant loss of signal. After tracing the issue, I found that the problem lies within the compute_epi_mask method, particularly with the following code:

```
lower_cutoff = int(np.floor(lower_cutoff * len(sorted_input)))
upper_cutoff = min(int(np.floor(upper_cutoff * len(sorted_input))), len(sorted_input) - 1)
delta = (sorted_input[lower_cutoff + 1 : upper_cutoff + 1]- sorted_input[lower_cutoff:upper_cutoff])
ia = delta.argmax()
threshold = 0.5 * (sorted_input[ia + lower_cutoff] + sorted_input[ia + lower_cutoff + 1])
```

My understanding is that this code is trying to find the boundary between the background and brain tissue regions. However, in my case, the threshold value reached as high as 6000, which suggests that the ideal boundary was not found. I believe the difference between high signal areas in my images is larger than the difference between background and brain signals.

As a result, I tried lowering the upper_cutoff from 0.85 to 0.65, and the threshold dropped to 300, significantly reducing the signal loss problem.

I'm not sure if my understanding is correct. Does anyone have any suggestions on this?

Below are the results for desc-adaptiveGoodSignal_mask.nii.gz:
Left: upper_cutoff = 0.85
Right: upper_cutoff = 0.65

Image

handwerkerd:
Closes #1176.

We regularly receive user feedback that the masks calculated within tedana using `nilearns`'s `compute_epi_mask` sometimes removes too many potentially good voxels. Our documented recommendation is that users should provide their own masks.

It's useful to keep existing functionality so that tedana can run without a user-provided mask and we don't break backwards compatibility, but we should make the above message more explicit.

- Added text recommending users provide their own mask to the CLI and API descriptions for `mask`
- Changed an `LGR.info` to an `LGR.warning` with a more detailed message when an external mask is not provided
- Clarified an `LGR.info` message to state that, when a T2* map and no mask is provided, tedana assumes the T2* map is masked and uses the non-zero values of the map for the mask
- Added a test to make sure the warning appears.

FWIW, in making this PR, I noticed that our integration tests don't include providing a `mask` option besides `None` to `tedana` and don't include providing a T2* map. That's not essential to add now, & is something that should be addressed whenever we get around to doing #920

### DIFF
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -106,7 +106,7 @@ def _get_parser():
             "space as `data`. If an explicit mask is not "
             "provided, then Nilearn's compute_epi_mask "
             "function will be used to derive a mask "
-            "from the first echo's data."
+            "from the first echo's data. "
             "Providing a mask is recommended."
         ),
         default=None,
@@ -443,7 +443,7 @@ def tedana_workflow(
         then Nilearn's compute_epi_mask function will be used to derive a mask
         from the first echo's data.
         Since most pipelines use better masking tools,
-        providing a mask, rather than using compute_epi_mask is recommended.
+        providing a mask, rather than using compute_epi_mask, is recommended.
     convention : {'bids', 'orig'}, optional
         Filenaming convention. bids uses the latest BIDS derivatives version (1.5.0).
         Default is 'bids'.
--- a/tedana/workflows/tedana.py
+++ b/tedana/workflows/tedana.py
@@ -107,6 +107,7 @@ def _get_parser():
             "provided, then Nilearn's compute_epi_mask "
             "function will be used to derive a mask "
             "from the first echo's data."
+            "Providing a mask is recommended."
         ),
         default=None,
     )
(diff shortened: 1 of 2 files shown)


tedana/workflows/tedana.py
```diff
@@ -107,6 +107,7 @@ def _get_parser():
             "provided, then Nilearn's compute_epi_mask "
             "function will be used to derive a mask "
             "from the first echo's data."
```
tsalo, line 109:
```suggestion
            "from the first echo's data. "
```

tedana/workflows/tedana.py
```diff
@@ -441,6 +442,8 @@ def tedana_workflow(
         spatially aligned with `data`. If an explicit mask is not provided,
         then Nilearn's compute_epi_mask function will be used to derive a mask
         from the first echo's data.
+        Since most pipelines use better masking tools,
+        providing a mask, rather than using compute_epi_mask is recommended.
```
tsalo, line 446:
```suggestion
        providing a mask, rather than using compute_epi_mask, is recommended.
```

tsalo (approved):
LGTM!

PR #1209, opened 2025-04-17
Use Jinja2 to populate reports and deal with empty spaces when figures aren't added

### LINKED ISSUE #1084: Elegantly handle missing images in HTML report
### Summary

Some figures won't be generated, depending on the settings. For example, in the 4-echo manual selection, we won't have T2* or S0 figures:

Screenshot 2024-04-19 at 9 59 06 AM

### Next Steps

1. Replace string.Template HTML templating with Jinja2.

### LINKED ISSUE #1204: tedana workflow fails when using `--t2smap` without `--no-reports`
### Summary

When users provide a T2* map as input to tedana, no S0 map is estimated and the figure-generation step fails.

### Additional Detail

```
  File "/opt/miniconda3/envs/nipreps/bin/tedana", line 8, in <module>
    sys.exit(_main())
  File "/opt/miniconda3/envs/nipreps/lib/python3.10/site-packages/tedana/workflows/tedana.py", line 1077, in _main
    tedana_workflow(**kwargs, tedana_command=tedana_command)
  File "/opt/miniconda3/envs/nipreps/lib/python3.10/site-packages/tedana/workflows/tedana.py", line 1051, in tedana_workflow
    reporting.static_figures.plot_t2star_and_s0(io_generator=io_generator, mask=mask_denoise)
  File "/opt/miniconda3/envs/nipreps/lib/python3.10/site-packages/tedana/reporting/static_figures.py", line 571, in plot_t2star_and_s0
    assert os.path.isfile(s0_img), f"File {s0_img} does not exist"
AssertionError: File /Users/taylor/Documents/datasets/logans-dset/tedana-test2-sub-pilot_task-checkerboard/sub-pilot_task-checkerboard_S0map.nii.gz does not exist
```

eurunuela:
Closes #1204 and closes #1084 .

- Replaces assert for a warning
- T2* images are always generated
- S0 images are only generated if the S0 file exists

### DIFF
--- a/tedana/reporting/html_report.py
+++ b/tedana/reporting/html_report.py
@@ -146,28 +146,63 @@ def _update_template_bokeh(bokeh_id, info_table, about, prefix, references, boke
     # Initial carpet plot (default one)
     initial_carpet = f"./figures/{prefix}carpet_optcom.svg"
 
-    # Check if images exist in the figures directory
-    figures_dir = os.path.join(os.path.dirname(initial_carpet), "")
+    # Get the figures directory - it's in the same directory as the references file
+    figures_dir = os.path.join(os.path.dirname(references), "figures")
+
+    LGR.info(f"Figures directory: {figures_dir}")
+
+    # List all files in the figures directory
+    if os.path.exists(figures_dir):
+        files_in_figures = os.listdir(figures_dir)
+        LGR.info(f"Files in figures directory: {files_in_figures}")
+    else:
+        files_in_figures = []
+        LGR.warning(f"Figures directory not found at {figures_dir}")
 
     # Adaptive mask image
-    adaptive_mask = f"./figures/{prefix}adaptive_mask.svg"
-    adaptive_mask_exists = os.path.isfile(adaptive_mask.replace("./figures/", figures_dir))
+    adaptive_mask_filename = f"{prefix}adaptive_mask.svg"
+    adaptive_mask = os.path.join(figures_dir, adaptive_mask_filename)
+    adaptive_mask_exists = adaptive_mask_filename in files_in_figures
+    LGR.info(
+        f"Checking for adaptive mask: {adaptive_mask_filename}, exists: {adaptive_mask_exists}"
+    )
+
+    # Check if essential T2* and S0 images exist
+    t2star_brain_filename = f"{prefix}t2star_brain.svg"
+    t2star_histogram_filename = f"{prefix}t2star_histogram.svg"
+    s0_brain_filename = f"{prefix}s0_brain.svg"
+    s0_histogram_filename = f"{prefix}s0_histogram.svg"
+    rmse_brain_filename = f"{prefix}rmse_brain.svg"
+    rmse_timeseries_filename = f"{prefix}rmse_timeseries.svg"
 
     # T2* and S0 images
-    t2star_brain = f"./figures/{prefix}t2star_brain.svg"
-    t2star_histogram = f"./figures/{prefix}t2star_histogram.svg"
-    s0_brain = f"./figures/{prefix}s0_brain.svg"
-    s0_histogram = f"./figures/{prefix}s0_histogram.svg"
-    rmse_brain = f"./figures/{prefix}rmse_brain.svg"
-    rmse_timeseries = f"./figures/{prefix}rmse_timeseries.svg"
+    t2star_brain = os.path.join(figures_dir, t2star_brain_filename)
+    t2star_histogram = os.path.join(figures_dir, t2star_histogram_filename)
+    s0_brain = os.path.join(figures_dir, s0_brain_filename)
+    s0_histogram = os.path.join(figures_dir, s0_histogram_filename)
+    rmse_brain = os.path.join(figures_dir, rmse_brain_filename)
+    rmse_timeseries = os.path.join(figures_dir, rmse_timeseries_filename)
 
-    # Check if T2* and S0 images exist
     t2s0_exists = (
-        os.path.isfile(t2star_brain.replace("./figures/", figures_dir))
-        and os.path.isfile(t2star_histogram.replace("./figures/", figures_dir))
-        and os.path.isfile(s0_brain.replace("./figures/", figures_dir))
-        and os.path.isfile(s0_histogram.replace("./figures/", figures_dir))
+        t2star_brain_filename in files_in_figures
+        and t2star_histogram_filename in files_in_figures
+        and s0_brain_filename in files_in_figures
+        and s0_histogram_filename in files_in_figures
     )
+
+    LGR.info(
+        f"Checking for T2*/S0 files: t2star_brain: {t2star_brain_filename} in files: {t2star_brain_filename in files_in_figures}"
+    )
+    LGR.info(
+        f"Checking for T2*/S0 files: t2star_histogram: {t2star_histogram_filename} in files: {t2star_histogram_filename in files_in_figures}"
+    )
+    LGR.info(
+        f"Checking for T2*/S0 files: s0_brain: {s0_brain_filename} in files: {s0_brain_filename in files_in_figures}"
+    )
+    LGR.info(
+        f"Checking for T2*/S0 files: s0_histogram: {s0_histogram_filename} in files: {s0_histogram_filename in files_in_figures}"
+    )
+    LGR.info(f"T2*/S0 section will be shown: {t2s0_exists}")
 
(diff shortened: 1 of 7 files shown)


tedana/reporting/static_figures.py
```diff
@@ -574,7 +574,11 @@ def plot_t2star_and_s0(
     s0_img = io_generator.get_name("s0 img")
     mask_img = io.new_nii_like(io_generator.reference_img, mask.astype(int))
     assert os.path.isfile(t2star_img), f"File {t2star_img} does not exist"
-    assert os.path.isfile(s0_img), f"File {s0_img} does not exist"
+
+    # Check if S0 image exists, issue warning if not
+    s0_exists = os.path.isfile(s0_img)
+    if not s0_exists:
+        warnings.warn(f"File {s0_img} does not exist. S0 plots will not be generated.")
```
handwerkerd, line 581: This should be an `LGR.warning` not `warnings.warn` so that it's stored in the logger. I'd actually prefer that this is an `LGR.info` since this is just describing expected functionality and isn't actually a problem. That is "S0 maps and T2* fit metrics are not in report since a pre-existing T2* map was provided"

handwerkerd (changes requested):
There are large blank boxes for multiple missing figures in the tedana report. There are two under the `S0` header and another two under the `T2* and S0 model fit (RMSE). (Scaled between 2nd and 98th percentiles)` header. I think fixing this would requiring making https://github.com/ME-ICA/tedana/blob/main/tedana/reporting/data/html/report_body_template.html more dynamically generated or having two templates, one if a T2* is provided and another if it isn't.

Also, for re-running three echo test, I added the following lines so that this functionality is included in a test. Current code location:
https://github.com/ME-ICA/tedana/blob/4761c2acb6fbaf9e4ed5fe8aee94d042d25460db/tedana/tests/test_integration.py#L256

Lines to add below the above line:
```
        "--t2smap",
        os.path.join(out_dir, "T2starmap.nii.gz"),
```

tsalo:
>I think fixing this would requiring making https://github.com/ME-ICA/tedana/blob/main/tedana/reporting/data/html/report_body_template.html more dynamically generated or having two templates, one if a T2* is provided and another if it isn't.

I have #1084 open about this. It shouldn't be too difficult to switch from string.Template to Jinja2 at some point.

eurunuela:
> I have #1084 open about this. It shouldn't be too difficult to switch from string.Template to Jinja2 at some point.

Why don't we merge that one first and then add it to this one?

tsalo:
#1084 is just an issue, not a PR.

handwerkerd:
When you're done with active changes on this, let me know so I can review. Also make a note in the opening comment or the PR title that this also changes to Jinja2 for the report template & that it would also close #1084 . In some ways that's the more noteworthy change here than the initial goal.

eurunuela:
I think I have added all the feedback. Let me know what you think.

tedana/reporting/static_figures.py
```diff
@@ -574,7 +574,14 @@ def plot_t2star_and_s0(
     s0_img = io_generator.get_name("s0 img")
     mask_img = io.new_nii_like(io_generator.reference_img, mask.astype(int))
     assert os.path.isfile(t2star_img), f"File {t2star_img} does not exist"
-    assert os.path.isfile(s0_img), f"File {s0_img} does not exist"
+
+    # Check if S0 image exists, issue warning if not
```
handwerkerd, line 578:
```suggestion
    # Check if S0 image exists, add message to log if not
```

handwerkerd:
Not sure if you're still working on this, but I noticed the S0 and T2* plots are no longer appearing when they are supposed to appear.

eurunuela:
> Not sure if you're still working on this, but I noticed the S0 and T2* plots are no longer appearing when they are supposed to appear.

Yeah, I saw. I will try to finish tomorrow.

handwerkerd (approved):
I can't fully evaluate the changes you made for `jinja2` but everything else looks fine and the html outputs look correct on the testing data and an additional run I processed.
Your choice on if you want to get feedback from @tsalo or someone else who has a clue about jinja2 before merging.

eurunuela:
> I can't fully evaluate the changes you made for `jinja2` but everything else looks fine and the html outputs look correct on the testing data and an additional run I processed.

They don't on the CircleCI outputs though. Something is off with the paths I think.

tsalo:
I think the link is using an absolute path, which is right for the original source of the files, but wrong for the updated path for the artifacts.

tedana/reporting/html_report.py
```diff
@@ -138,44 +143,93 @@ def _update_template_bokeh(bokeh_id, info_table, about, prefix, references, boke
     -------
     HTMLReport : an instance of a populated HTML report
     """
-    resource_path = Path(__file__).resolve().parent.joinpath("data", "html")
-
     # Initial carpet plot (default one)
     initial_carpet = f"./figures/{prefix}carpet_optcom.svg"
 
+    # Get the figures directory - it's in the same directory as the references file
+    figures_dir = os.path.join(os.path.dirname(references), "figures")
```
tsalo, line 150: I think this is the culprit. You need to use a relative path, not an absolute path. This will also make the reports slightly more portable, as long as folks include the figures directory when they move the HTML file.

tsalo: That's odd. I use Jinja2 in XCP-D with relative paths without issue. Here's what I do: https://github.com/PennLINC/xcp_d/blob/2cda09ebba3187ccaa691b055cbbf579931ae83e/xcp_d/interfaces/execsummary.py#L74-L82

Maybe instead of `./figures` it should be `figures`?

tsalo:
>My first guess was that some file paths were hard-coded and CircleCI moves files. I just copied a successful execution to a new location & it still works locally, so it's not hard-coded paths.

CircleCI does move the files when it stored artifacts. Did you copy the HTML file to a new location, or did you move the whole HTML file + figures folder? In the latter scenario, relative paths will be useful. Especially if you want to share reports across machines.

tedana/reporting/html_report.py
```diff
@@ -138,44 +143,87 @@ def _update_template_bokeh(bokeh_id, info_table, about, prefix, references, boke
...
-    resource_path = Path(__file__).resolve().parent.joinpath("data", "html")
-
     # Initial carpet plot (default one)
     initial_carpet = f"./figures/{prefix}carpet_optcom.svg"
 
+    # Get the figures directory
+    figures_dir = "figures"
+
+    # List all files in the figures directory
+    files_in_figures = os.listdir(figures_dir)
```
tsalo, line 153: Any chance the cwd isn't the output directory for the report? If so, then relative paths in the HTML report would be fine, but this line will fail.

tsalo:
```suggestion
    files_in_figures = os.listdir(os.path.abspath(figures_dir))
```

tsalo: If I'm right, then this will still fail, but it'll at least be clear that the location of "figures_dir" isn't where we'd expect it because the cwd isn't the output directory.

tsalo: I was right!

> FAILED tedana/tests/test_integration.py::test_integration_four_echo - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/src/tedana/figures'

eurunuela: It should be `outputs/figures`

tedana/reporting/html_report.py
```diff
@@ -138,44 +143,69 @@ def _update_template_bokeh(bokeh_id, info_table, about, prefix, references, boke
...
+
+    # T2* and S0 images for HTML
+    t2star_brain = f"./figures/{t2star_brain_filename}"
+    t2star_histogram = f"./figures/{t2star_histogram_filename}"
+    s0_brain = f"./figures/{s0_brain_filename}"
+    s0_histogram = f"./figures/{s0_histogram_filename}"
+    rmse_brain = f"./figures/{rmse_brain_filename}"
+    rmse_timeseries = f"./figures/{rmse_timeseries_filename}"
+
+    t2s0_exists = (
```
tsalo, line 180: If you provide a T2* image to the command-line, the T2* file will exist, but the S0 image will not. Do we not generate the T2* brain and histogram images in that scenario?

eurunuela:
It seems to be working on CircleCI now. Would you like to have another look @handwerkerd? I will let you merge this if you are happy with the changes.

PR #1212, opened 2025-04-23
Properly flip ICA mixing matrix time series

### LINKED ISSUE #981: Add prefix to pre-defined ICA mixing matrix
After approving, I noticed one minor thing that may or may not be worth changing. The four-echo integration test uses a pre-defined ICA mixing matrix:
https://github.com/tsalo/tedana/blob/1b97294eaed4e7111aba6a807c4ece38472a7270/tedana/tests/test_integration.py#L304

That file `desc-ICA_mixing_static.tsv` is copied into the output directory without any name change or prefix addition and then the same content is also appropriately named in `sub-01_desc_ICA_mixing.tsv` 

I don't see this as a major issue for this PR. The one problem I could see it happening is if someone is using another package to run ICA and sending those matricies to tedana. In that case, it would be their responsibility to appropriately name their mixing matrix files.

That said, the supplied mixing matrix is copied without any checks on whether it's overwriting anything:
https://github.com/tsalo/tedana/blob/1b97294eaed4e7111aba6a807c4ece38472a7270/tedana/workflows/tedana.py#L519-L525
This should definitely include a check for `io_generator.overwrite` That could be added here or in a stand-alone PR.

_Originally posted by @handwerkerd in https://github.com/ME-ICA/tedana/issues/963#issuecomment-1711914414_

### LINKED ISSUE #1206: inconsistent flipping of ICA component signs
### Summary

@sophiatibblin, a student of @CesarCaballeroGaudes noticed something confusing. We have code to flip the sign of our ICA components so that the weight maps are more positive than negative. We also log the components where the sign is flipped, but we don't flip the time series in the ICA mixing matrix.

### Additional Detail

This is where the sign is calculated and `metric_maps["map weight"]` is flipped if the sign is negative. Later on in the same function, "map weight" is used to calculate "map Z", which is the saved component weight maps. That means the component weights are flipped before saving.
https://github.com/ME-ICA/tedana/blob/b373d0664a33e0f75ba3657a01bd5df769a1583c/tedana/metrics/collect.py#L164-L168

I don't see anywhere were the time series in the ICA mixing matrix are also flipped before saving. The time series are flipped for the static images that are generated for the report.

https://github.com/ME-ICA/tedana/blob/b373d0664a33e0f75ba3657a01bd5df769a1583c/tedana/reporting/static_figures.py#L343-L344

I'm not sure I'd call this a bug since we're internally using everything correctly, (unless I'm missing something) but this is under-documented and confusing.

### Next Steps

1. Decide if this is something that merely needs better documentation or if this is something that should be fixed in the code
2. If we decide to fix it in the code, my suggestion is to flip the time series in the mixing matrix before saving and then don't reflip in stat_figures.py

handwerkerd:
Closes #1206 and reverts the changes in #748.
Closes #981 (Update: I realized it was really easy to also address this issue within this PR)

The signs of PCA & ICA components are flipped so that there are more positive vs negative values in the component weight maps. The maps were saved with the flipped values, but the time series in the mixing matrices were not. On a second look, after opening the issue, I realized that we actually were correctly flipping the mixing matrix in `collect.py` but `generate_metrics` was not returning the updated mixing matrix to be used elsewhere & saved.

I am labelling this a `breaking change` because the time series in the mixing matrix files might change, but it should not meaningfully alter results.

- `collect.py/generate_metrics` now returns the mixing matrix with columns appropriately flipped by "Optimal Sign"
- I realized that `collect.py/generate_metrics` was returning `external_regressor_config` but it was not being altered within the function and the variable was never being updated by functions that called `generate_metrics` so I removed this returned value
- Text describing "optimal sign" in the component matrix was already correct, but I tweaked the language slightly
- Updated the test for `generate_metrics` to confirm the sign of the mixing matrix is being flipped
- Updated `static_figures.py` so that it does not reflip the sign of the mixing matrix.
- In `tedana.py`, if there was a user-provided mixing matrix, it was copied to the original and new file names and not otherwise changed. Now, just the originally named file is copied and the mixing matrix with potentially flipped values is always saved.
  - I also noticed a bug in this copying that it was supposed to compare the provided mixing matrix name to the name that would be output to tedana and only copy if they were different, but the `if` clause included the full paths so they would always be different. In practice, this worked fine because the two files were always identical, but it would no longer work when time series are flipped. I fixed this & now add "user_provided_" to the file name before copying it to the directory. This was already tested in `test_integration_three_echo`
- `test_integration_five_echo` was failing because this is now outputting 3 additional components, so I added the file names for those additional components. This is a real breaking change, but I think it's happening because this is the test using robustICA with only 4 iterations, which will be noisy. Flipping some time series for the PCA mixing matrix was enough to alter the number of stable components. robustICA with only 4 iterations is good for a test of the code, but doesn't results in robustly stable results.

### DIFF
--- a/tedana/metrics/collect.py
+++ b/tedana/metrics/collect.py
@@ -667,7 +667,9 @@ def get_metadata(component_table: pd.DataFrame) -> Dict:
             "Description": (
                 "Optimal sign determined based on skew direction of component parameter estimates "
                 "across the brain. In cases where components were left-skewed (-1), the component "
-                "time series and map weights are flipped prior to metric calculation."
+                "time series and map weights are flipped prior to metric calculation. "
+                "This sign applies to the original mixing matrix and map weights. "
+                "The outputs produced by tedana are already flipped."
             ),
             "Levels": {
                 1: "Component is not flipped prior to metric calculation.",
--- a/tedana/metrics/collect.py
+++ b/tedana/metrics/collect.py
@@ -35,7 +35,7 @@ def generate_metrics(
     external_regressors: Union[pd.DataFrame, None] = None,
     external_regressor_config: Union[List[Dict], None] = None,
     metrics: Union[List[str], None] = None,
-) -> Tuple[pd.DataFrame, Dict]:
+) -> Tuple[pd.DataFrame, npt.NDArray]:
     """Fit TE-dependence and -independence models to components.
 
     Parameters
@@ -76,9 +76,11 @@ def generate_metrics(
     component_table : (C x X) :obj:`pandas.DataFrame`
         Component metric table. One row for each component, with a column for each metric.
         The index is the component number.
-    external_regressor_config : :obj:`dict`
-        Info describing the external regressors and method used for fitting and statistical tests
-        (or None if none were inputed)
+    mixing : (T x C) array_like
+        Mixing matrix for converting input data to component space,
+        where `C` is components and `T` is the same as in `data_cat`
+        The signs of a component time series are flipped so that the components spatial map
+        has more positive voxel weights
     """
     # Load metric dependency tree from json file
     dependency_config = op.join(utils.get_resource_path(), "config", "metrics.json")
@@ -460,7 +462,7 @@ def generate_metrics(
     other_columns = [col for col in component_table.columns if col not in preferred_order]
     component_table = component_table[first_columns + other_columns]
 
-    return component_table, external_regressor_config
+    return component_table, mixing
 
 
 def get_metadata(component_table: pd.DataFrame) -> Dict:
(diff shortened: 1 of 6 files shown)


handwerkerd:
@sophiatibblin @CesarCaballeroGaudes This should address the issue you identified.

tedana/metrics/collect.py
```diff
@@ -665,7 +667,7 @@ def get_metadata(component_table: pd.DataFrame) -> Dict:
             "Description": (
                 "Optimal sign determined based on skew direction of component parameter estimates "
                 "across the brain. In cases where components were left-skewed (-1), the component "
-                "time series is flipped prior to metric calculation."
+                "time series and map weights are flipped prior to metric calculation."
```
tsalo, line 670:
```suggestion
                "time series and map weights are flipped prior to metric calculation. "
                "This sign applies to the original mixing matrix and map weights. "
                "The outputs produced by tedana are already flipped."
```

tsalo (commented):
I don't have a strong opinion on the proposed change, but the code changes look solid to me. I would suggest making it clear in the `optimal_sign` documentation that the sign has already been applied to both the mixing matrix and map weights.

sophiatibblin:
Ooh, thank you very much @handwerkerd !

tsalo (approved):
LGTM

PR #1216, opened 2025-04-24
Adding recommendations for n_robust_runs

### LINKED ISSUE #1148: Add recommendations on `--n-robust-runs` and `--maxit`
### Summary

We have some info on robustica in the FAQ, but it's not clear what folks should use for `--n-robust-runs` (and `--maxit` I suppose). My big question is whether selecting the number of robust runs is just a matter of resources and time (i.e., should you use the highest number you're willing to wait for) or if there's a sweet spot (i.e., increasing the number runs will result in fewer consistent components, which means that more is not necessarily better).

If we don't actually know the answer to this, I could try running tedana on a single run with different `--n-robust-runs` values to see how the parameter impacts the number of detected components.

handwerkerd:
Closes #1148 .

- Added recommendations for the value for `n_robust_runs to the faq

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -161,6 +161,14 @@ If users are having trouble with PCA component estimation failing on a dataset,
 we recommend using RobustICA;
 and we invite users to send us feedback on its behavior and efficacy.
 
+Also note that ``--ica-method robustica`` can be paired with ``--n_robust_runs``.
+That second option defines how many interations of ICA to run.
+The default is ``--n_robust_runs 30`` which was selected because the resulting components
+stabilized and the activation volumes plateaued during testing on a large dataset.
+If one wants to examine on other data sets, a basic guide is,
+since larger ``--n_robust_runs`` means more compute time,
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -161,6 +161,14 @@ If users are having trouble with PCA component estimation failing on a dataset,
 we recommend using RobustICA;
 and we invite users to send us feedback on its behavior and efficacy.
 
+Also note that ``--ica-method robustica`` can be paired with ``--n_robust_runs``.
```
tsalo, line 164:
```suggestion
Also note that ``--ica_method robustica`` can be paired with ``--n_robust_runs``.
```

I think we only accept underscore-delimited arguments.

tsalo: Actually, to be more accurate, it looks like some of our parameters are underscore-delimited only and some are dash-delimited only, which isn't great.

tsalo (approved):
LGTM!

PR #1217, opened 2025-04-25
Download decision trees from figshare

### LINKED ISSUE #1205: How to distribute more decision trees with tedana
### Summary

@BahmanTahayori is about to submit a manuscript for publication that uses a revision of the current decision tree. He asked me if we could add it to tedana `main` so that he can reference it in the tedana code. I think it's great to distribute more decision trees, particularly those that have been validated in manuscripts, but I want to do this in a way that won't cause future problems.

#### Concerns
1. We don't want to get in the business of judging who does or does not have a decision tree worthy of inclusion in tedana.
2. As we get more decision trees, the already cryptic names will get more cryptic. Bahman's is currently named `robust_tedana` (also the name of the overall method in the manuscript). Every new tree will be `[positive adjective]_tedana` and it will be hard to understand how to parse a long list of trees distributed with the code.

#### Proposed solutions

- For concern 1, my suggestion is to say we are willing to share any decision tree that is included in a manuscript that includes some validations against other decision trees or other methods. 
- For concern 2
  - I briefly spoke to @eurunuela and he suggested, having a place outside of the repository to share decision trees. If a decision tree isn't in the tedana repo, the code could query that repo and download the appropriate decision tree. We are already storing decision trees on figshare and have links to the json files. For example tedana_orig ( https://figshare.com/articles/code/Component_selection_decision_trees_in_tedana/25251433/2?file=45211096 ). The nice part of this approach is that figshare also have versioning so we can have multiple versions of the same trees that could be consistently rerun in the future. The minus is that this is. one more complexity to add to the code.
  - An intermediate option would be to make a subfolder of https://github.com/ME-ICA/tedana/tree/main/tedana/resources/decision_trees for user-provided trees. They'd just need distinct names and we already have a system for each one to include the reference for where it was used. They wouldn't appear in the help menu, but anyone could call them by their name. The `demo_` trees in the above folder are already like this in that we're distributing them, but they are easy to run, but aren't highlighted in the help menus.
  - For naming, if there's some division between the main ones and others, then we can be very flexible. If they are all together, we might want to lean into the original `kundu` tree and use first-author names, i.e. `tahayori` for their new tree.

### Next Steps

- Acutely @BahmanTahayori wants his json file stored somewhere consistent that he can reference in his publication he wants to submit in a couple of days. I can very quickly [ut it into a new version3 of our figshare page and then he has a clear reference. I think he'd prefer I just add it to the tedana repo. I'm fine either or both, but this is a group project and I don't want to make this decision solo. $\Large\color{red}{\textbf{I'd like others to give input on this ASAP.}}$
- Decide on a medium term approach. Proposals above are:
  - More more stuff in https://github.com/ME-ICA/tedana/tree/main/tedana/resources/decision_trees with some rule on what we'd add and some naming convention
  -  Put more trees in something like https://github.com/ME-ICA/tedana/tree/main/tedana/resources/decision_trees/published_trees with less concern for a naming convention. We could then decide to promote trees to the main directory based on usage and validated improvements over multiple publications
  - Set up a system to automatically download trees from figshare or OSF.

I think @eurunuela's suggestion to have a download system is the most elegant, but would require more time coding. I am currently leaning towards the subdirectory option with all trees also being added to our figshare site. That said, I should not make this decision solo. I'm tagging others who are or might be playing with decision trees because this might affect you & you might have opinion: @goodalse2019 @n-reddy @tsalo @CesarCaballeroGaudes @dowdlelt @smoia @mmedina002 @bf777

marlyr:
Closes #1205.

Changes proposed:

- `download_json` function that downloads decision trees from figshare to `out_dir` if not in `DEFAULT_TREES` or saved locally.
- Some relevant tests

Please note that some users on protected networks may encounter SSL issues when connecting to figshare’s API.

### DIFF
--- a/tedana/selection/component_selector.py
+++ b/tedana/selection/component_selector.py
@@ -72,15 +72,14 @@ def load_config(tree: str, out_dir: str = ".") -> Dict:
         fname = tree
     else:
         fname = download_json(tree, out_dir)
-        
+
     try:
         dectree = load_json(fname)
-    except:
+    except Exception:
         raise ValueError(
             f"Cannot find tree {tree}. Please check your path or use a "
             f"default tree ({DEFAULT_TREES}) or one from figshare."
         )
-
 
     return validate_tree(dectree)
 
--- a/tedana/selection/component_selector.py
+++ b/tedana/selection/component_selector.py
@@ -8,7 +8,7 @@ from typing import Dict, List, Union
 import pandas as pd
 from numpy import asarray
 
-from tedana.io import load_json
+from tedana.io import load_json, download_json
 from tedana.selection import selection_nodes
 from tedana.selection.selection_utils import (
     clean_dataframe,
@@ -43,19 +43,22 @@ class TreeError(Exception):
     pass
 
 
-def load_config(tree: str) -> Dict:
+def load_config(tree: str, out_dir: str = ".") -> Dict:
     """Load the json file with the decision tree and validate the fields in the decision tree.
 
     Parameters
     ----------
     tree : :obj:`str`
         The named tree or path to a JSON file that defines one
-
+    out_dir :obj:`str`, optional
+        The output directory. Default is current working directory.
     Returns
     -------
     tree : :obj:`dict`
         A validated decision tree for the component selection process.
     """
+    fname, dectree = None, None
+
     if tree in DEFAULT_TREES:
         fname = op.join(get_resource_path(), "decision_trees", tree + ".json")
     elif tree == "kundu":
@@ -65,21 +68,19 @@ def load_config(tree: str) -> Dict:
         )
         tree = "tedana_orig"
         fname = op.join(get_resource_path(), "decision_trees", tree + ".json")
-    else:
+    elif op.isfile(tree):
         fname = tree
-
+    else:
+        fname = download_json(tree, out_dir)
+        
     try:
         dectree = load_json(fname)
-    except FileNotFoundError:
+    except:
         raise ValueError(
             f"Cannot find tree {tree}. Please check your path or use a "
-            f"default tree ({DEFAULT_TREES})."
+            f"default tree ({DEFAULT_TREES}) or one from figshare."
         )
-    except IsADirectoryError:
-        raise ValueError(
-            f"Tree {tree} is a directory. Please supply a JSON file or "
-            f"default tree ({DEFAULT_TREES})."
-        )
+
 
     return validate_tree(dectree)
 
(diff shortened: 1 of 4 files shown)


eurunuela:
> Please note that some users on protected networks may encounter SSL issues when connecting to figshare’s API.

I wonder if it would be a good idea to suggest this could be the reason why people are getting errors. What do you think?

tedana/selection/component_selector.py
```diff
@@ -65,20 +73,17 @@ def load_config(tree: str) -> Dict:
         )
         tree = "tedana_orig"
         fname = op.join(get_resource_path(), "decision_trees", tree + ".json")
-    else:
+    elif op.isfile(tree):
         fname = tree
+    else:
+        fname = download_json(tree, out_dir)
 
     try:
```
handwerkerd, line 81:
```suggestion
    if fname and op.isfile(fname):
```
This can just be an if/else rather than a try/except. This means, if the file exists, but isn't a json, we'll get the error in `load_json` and if it doesn't exist we'll get the error here.

tedana/selection/component_selector.py
```diff
@@ -65,20 +73,17 @@ def load_config(tree: str) -> Dict:
...
         dectree = load_json(fname)
-    except FileNotFoundError:
+    except Exception:
```
handwerkerd, line 83:
```suggestion
    else:
```

handwerkerd (changes requested):
I'm suggesting a few changes for clarity.

In response to @eurunuela's comment, I think the message you want is here:

`        LGR.error(f"Cannot connect to figshare: {e}")`
https://github.com/marlyr/tedana/blob/cfc94b59c7f3f2ba38f36aeeb0e958077d4e7cf9/tedana/io.py#L549

I don't want to get more specific because there are many reasons someone cannot connect to a website (i.e. no internet connection) and, unless we're testing a list of them, there's no reason to guess which one it might be.

That said, the Codecov report ( https://app.codecov.io/gh/ME-ICA/tedana/pull/1217 ) says the above line and a few others aren't covered by any tests. I'm not sure if there's an easy way to add them to the tests using a mock request. I did test from behind the NIH firewall and the error does appear.

handwerkerd (approved):
LGTM!

handwerkerd:
@BahmanTahayori with this addition, people will be able to use your tree by including `--tree robust_tedana_2025` & the tree will download from figshare.

BahmanTahayori:
Thanks @handwerkerd. An excellent approach.

handwerkerd:
> Thanks @handwerkerd. An excellent approach.

Thank you to @marlyr for getting it done!

PR #1221, opened 2025-05-11
Add gscontrol figures

### LINKED ISSUE #1087: Add global signal (if `--gscontrol gsr` is used) to component plots
### Summary
I think it might be useful to include the global signal in the HTML report, and I was thinking that it could be treated as a component for the sake of the report. It would also be nice to see how much variance the global signal explains.

tsalo:
Closes #1087 (sort of).

- Plot spatial maps for GSR and MIR.
- Plot time series of global signal and MIR signal.
- Add the new plots to the HTML report.
- Print the variance explained by each step.


tedana/gscontrol.py
```diff
@@ -120,7 +120,13 @@ def gscontrol_raw(
     # Project global signal (but not Legendre bases) out of optimally combined data
     betas = np.linalg.lstsq(glbase, data_optcom_masked.T, rcond=None)[0]
     gs_fitted = np.dot(glbase[:, :1], betas[:1, :]).T
-    data_optcom_nogs = data_optcom_masked - gs_fitted + temporal_mean
+    data_optcom_nogs = data_optcom_masked - gs_fitted
+
+    # Calculate the variance explained by the global signal
+    varexpl = 100 * (1 - ((data_optcom_nogs**2).sum() / (data_optcom_masked**2).sum()))
```
tsalo, line 126: This is worth double-checking. In the 4-echo dataset this ends up being 6.68%.

tedana/gscontrol.py
```diff
@@ -257,12 +263,18 @@ def minimum_image_regression(
...
-    mehk_no_t1_gs = mehk_ts - np.dot(
-        np.linalg.lstsq(gs_ts.T, mehk_ts.T, rcond=None)[0].T,
-        gs_ts,
+    # Calculate the T1-like global signal
+    t1_gs_data = np.dot(np.linalg.lstsq(gs_ts.T, mehk_ts.T, rcond=None)[0].T, gs_ts)
+
+    # Calculate the variance explained by the T1-like global signal
+    varexpl = 100 * (1 - (((data_optcom_z - t1_gs_data) ** 2).sum() / (data_optcom_z**2).sum()))
+    LGR.info(
+        f"Variance in optimally combined data explained by T1-like global signal: {varexpl:.02f}%"
```
tsalo, line 272: This is worth double-checking. In the 4-echo dataset this ends up being 3.68%.

eurunuela (approved):
LGTM

tsalo:
I just realized I need to add information to the documentation as well. I might need to ask for another review once I add that. Sorry!

@eurunuela WDYT of calculating rho and kappa for the global signals and adding them to the donut plot like components?

eurunuela:
> @eurunuela WDYT of calculating rho and kappa for the global signals and adding them to the donut plot like components?

I fear people may get confused, but if we make it very clear it is the global signal, then I think it's useful information.

tsalo:
Maybe that part can wait until a later PR. I'll just fix up the documentation for this one.

tsalo:
Here's the documentation section: https://tedana--1221.org.readthedocs.build/en/1221/outputs.html#global-signal-removal-plots

handwerkerd:
We're starting to add general values to the cross_component_metrics file. I think it's reasonable to save the kappa and rho values for the global signal there. You could also add variance explained by the global signal. Other metrics in the component table could be adapted for other time series like global signal, but I don't know if that's worth the effort.
This is all stuff that could be added to this PR or a future PR so I don't want to hold this one up.

tsalo:
I'm happy to merge with one approval since this just adds figures.

>We're starting to add general values to the cross_component_metrics file.

I was planning to put it in a json sidecar for the confounds tsv, since that would fit well with BIDS derivatives. Something like this:

```json
{
    "global_signal": {
        "LongName": "GSR Global Signal Timeseries",
        "Description": (
            "Global signal time series, estimated using the amplitude-based T1 "
            "equilibration correction method. "
            "This involves taking the spatial minimum of the optimally combined "
            "data after detrending with a Legendre polynomial basis set of "
            "order = 0 and degree = 4. "
            "The global signal time series is then estimated by regressing the "
            "spatial global signal against the optimally combined data. "
            "The time series is z-scored prior to saving."
        ),
        "Kappa": 5,
        "Rho": 2,
        "VarianceExplained": 6.61,
    },
    "mir_global_signal": {
        "LongName": "MIR Global Signal Timeseries",
        "Description": (
            "Global signal time series, estimated using the minimum image regression method. "
            "This involves taking the voxel-wise minimum over time of the BOLD-like "
            "components after regressing out the T1-like global signal. "
            "The global signal time series is then estimated by regressing the "
            "T1-like effect against the optimally combined data. "
            "The time series is z-scored prior to saving."
        ),
        "Kappa": 5,
        "Rho": 1,
        "VarianceExplained": 3.61,
    }
}
```

PR #1226, opened 2025-05-14
Add external regressor-mixing matrix correlation heat map

### LINKED ISSUE #1223: Plot external regressor correlation heatmap
I was thinking something like the following would be nice for the report:

Image

I basically just applied hierarchical clustering to the correlations of correlations in order to sort the external regressors by similarity, then used `seaborn.heatmap` to make the plot. Should be pretty easy to generate in tedana, but we might want to split it based on the number of components or external regressors.

tsalo:
Closes #1223.

- Add seaborn as a dependency. Seaborn's only requirements are numpy and matplotlib, so I don't think this is a problem.
- Create heat map figure showing correlations between external regressors and component time series, and add it to the HTML report.

### DIFF
--- a/tedana/reporting/static_figures.py
+++ b/tedana/reporting/static_figures.py
@@ -912,7 +912,7 @@ def plot_heatmap(
     mixing: pd.DataFrame,
     external_regressors: pd.DataFrame,
     component_table: pd.DataFrame,
-    io_generator: io.OutputGenerator,
+    out_file: str,
 ):
     """Plot a heatmap of the mixing matrix and external regressors.
 
@@ -922,8 +922,10 @@ def plot_heatmap(
         Mixing matrix.
     external_regressors : (E x T) :obj:`numpy.ndarray`
         External regressors.
-    io_generator : :obj:`~tedana.io.OutputGenerator`
-        The output generator for this workflow.
+    component_table : pandas.DataFrame
+        Component table.
+    out_file : str
+        The output file name.
     """
     import re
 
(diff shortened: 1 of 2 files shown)


tsalo:
Here's what the figure looks like in the 3-echo test. Not half bad in my opinion.

image

eurunuela:
I like it!

tedana/reporting/data/html/report_body_template.html
```diff
@@ -286,6 +286,17 @@ <h3>Global Signal Timeseries</h3>
     </div>
 </div>
 {% endif %}
+{% if externalRegressorsExist %}
+<div class="carpet-plots">
```
eurunuela, line 290:
```suggestion
<div class="carpet-plots">
  <div class="carpet-plots-image">
```

tedana/reporting/data/html/report_body_template.html
```diff
@@ -286,6 +286,17 @@ <h3>Global Signal Timeseries</h3>
...
+  <h1>External Regressors</h1>
+  <img
+    id="externalRegressorCorrelations"
+    src="{{ externalRegressorCorrelations }}"
+    style="width: 95%; max-width: 1200px; height: auto"
+  />
+</div>
```
eurunuela, line 297:
```suggestion
  </div>
</div>
```

tedana/reporting/data/html/report_body_template.html
eurunuela: This should make the image as big as the carpet plots. It think it would be nice to make them the same size for consistency.

eurunuela (changes requested):
I would add an extra div to the html template to make sure the image size is consistent with the carpet plots above it.

tedana/reporting/data/html/report_body_template.html
tsalo: That's awesome, thank you!

handwerkerd:
Thoughts on adding additional rows for the summary values if external regressors are grouped? (i.e. a Motion row that shows the combination of the motion regressors)

eurunuela:
> The summary values would be F-stats or p-values, right? I think we need a separate figure for that since the scale will be misleading, but I _do_ think a figure would be great.

Do you think you could add it as a subplot below? I think it would be awesome if we could keep it all within the same figure/file.

tsalo:
I'm happy to give it a try. What makes the most sense to plot though? The F-stat, the R2 value, or -log10(p)?

tsalo:
How's this?

image

eurunuela:
Do you think it would be useful to link the plot to the others? So that when a component is selected, the others are dimmed?

Could be done in a different PR.

tsalo:
I think that would definitely be cool, but it should go in a separate PR. It would also be cool to see the external regressor time series overlaid on the component time series.

eurunuela:
Oooh I like that! Maybe we can make those two new additions in a new PR. I will open an issue.

Issue #1185, opened 2025-03-09
stringent dependencies upper pin, best practice question.


bpinsard:
### Summary

Some of the dependencies are pinned to very specific versions upper range, making it less straightforward to install alongside other packages or build minimal containers. I am not an expert in best-practice for python package dependency management, so I was wondering if there is a specific reason for that?

### Additional Detail

example:
I was building a Tedana container from the latest version using alpine, with something like 

```
FROM alpine:3.21

RUN apk add --no-cache py3-scikit-learn py3-pip
RUN pip install --break-system-packages --no-cache-dir tedana

ENTRYPOINT ["/usr/bin/tedana"]
```

Installing scikit-learn from apk makes it more efficient, avoiding build reqs, but it will install most recent numpy, scipy and pandas from alpine.
Then `pip install tedana`  will remove these to downgrade from a few patches/minors to the version pinned here:
https://github.com/ME-ICA/tedana/blob/df583477030310cc31ef6e654b99a4d0eb21af73/pyproject.toml#L23-L38
creating another large layer.

For these well established libraries, non retro-compatible changes are not expected in minor versions updates.

Is that latest-pinning strategy necessary for dependabot? As if we want to test for any increment of these libraries? In the meantime, if any other dependency pins those to lower version, the tests for dependabot dependency update PR won't necessarily test with the newest version, right?

tsalo:
Setting upper limits makes it much easier for us to determine if there's a breaking change- especially used in combination with dependabot. We do want to test each increment (at least minor releases), as we've seen bugs due to dependency changes crop up in the past, and debugging that is really difficult (e.g., you can't always tell _which_ dependency is the problem, as in #523).

We have had issues with minor releases in the past (e.g., bokeh in #1063, sklearn in #1019).

I would say that we cannot guarantee that tedana will work with package versions outside of the range tested, but there may be a more flexible way of handing the dependencies. I'm not sure what that is though. Maybe we could pin minor versions in the main dependencies and have a "strict" optional group where we pin the patch versions, but that would only work if dependabot could be configured to only bump patch versions in the strict optional group and minor versions in the main dependencies...

PR #1228, opened 2025-05-16
chore: Drop Python 3.8 support

effigies:
I'm starting to look at #1185, and I'm not particularly interested in supporting Python 3.8 while doing so.

Please close if you have a reason to continue supporting EOL Python versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,48 +6,26 @@ version: 2.1
 orbs:
   codecov: codecov/codecov@5.2.1
 jobs:
-  makeenv_38:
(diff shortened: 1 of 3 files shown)


tsalo:
I'm happy to drop 3.8 support.

The required check is set as a rule for the `main` branch, I believe. We can change it in our branch protection settings.

effigies:
My guess is that the requirements are set from the main branch, and will disappear after merge. I'm not sure. I've never set those before.

eurunuela (approved):

PR #1202, opened 2025-03-25
Returns CircleCI config to before version matrics to fix codecov

handwerkerd:
Closes #1192.

Me and @eurunuela tried to do a lot of things to get codecov working againa fter #1190 changed `.circleci/config.yml` to use version matrices. After nothing worked we decided to go back to the config from before #1190 that did work. We can then separately try to figure out how to switch back to version matrics without (again) breaking codecov interactions.

Changes proposed in this pull request:

- Replaced `.circleci/config.yml` with the version that existed before #1190

PR #1229, opened 2025-05-16
chore: Use tox-uv and uv.lock to test minimum, locked, and pre-release dependencies

effigies:
My goal is to remove dependency caps, which lead to very fragile environment resolutions. I understand that the purpose of these caps is to ensure that regressions can be localized to changes in specific dependencies, so this PR uses uv lockfiles to address that.

`uv lock [--upgrade]` inspects PyPI and finds the latest versions of tools that successfully satisfy the environment described by `pyproject.toml`, including optional dependencies. You can checkout the repository and get the same environment as anybody else with `uv sync`.

This PR also uses `tox-uv` to create environments `py<ver>-{min,latest,pre}`. `py*-min` attempts to install the minimum versions that are declared in `pyproject.toml` into an environment that created with the declared Python version. `py*-latest` will use `uv sync` to setup the environment, while `py*-pre` will look for pre-release versions of dependencies, so that you can catch breakages coming down the line.

It also sets the dependabot environment to `uv`, to auto-update your `uv.lock` file, and sets up a GitHub action to run tox, as a demonstration. If this looks like a good approach, I'll figure out how to replace your conda environments in CircleCI with `uv sync`-based ones.

Addresses #1185.


effigies:
I think you shouldn't need to change anything. Do you want to show me what you tried with the version matrices, and I can see if it'll be any easier now?

eurunuela:
You can see the changes we reverted in #1202. We thought the matrices were going to work fine but ended up having to go back to the current configuration.

effigies:
I see. Not sure I've got it in me to play with a new workflow syntax. But I'll see if I can use uv instead of conda, at least.

Test failures seem to be related to an OSF error causing 403 responses.

eurunuela:
> Test failures seem to be related to an OSF error causing 403 responses.

We've had these before. We usually just rerun the ones that failed.

effigies:
Okay, I've updated CircleCI. I think this is more-or-less usable right now, but I realize that introducing new maintenance tools is a bit like giving someone a hot potato, so I'm happy to jump on a call to talk this over with people. At the very least, hopefully this is a demo of an overall approach, and you can adapt it to suit your needs. I won't be offended if you close this PR.

eurunuela:
I think the easiest would be that you join our next devs meeting on June 19th. Unless it conflicts with OHBM.

effigies:
I should be around then. I'm not going to make it to OHBM this year.

tsalo:
Plan from today's devs call:

1. I will merge #1248 and make a new release.
2. I'll address the merge conflicts in this PR.
3. I will merge this PR and, in the near future, make an alpha release so everyone can test out this approach.

tsalo:
This is just for pre-release versions of dependencies, right? I'm fine with just testing new regular releases, like we currently have with dependabot.

Am I understanding your question correctly?

effigies:
Yes. I can just drop them.

effigies:
Yes

PR #1244, opened 2025-07-17
Fix NaNs in external regressors, take 2

bpinsard:
Turns out https://github.com/ME-ICA/tedana/pull/1163 did not fix the issue, because the modified dataframe is never returned to the workflow.

Quick fix, but adding tests would be better.

### DIFF
--- a/tedana/tests/test_external_metrics.py
+++ b/tedana/tests/test_external_metrics.py
@@ -186,7 +186,7 @@ def test_validate_extern_regress_succeeds(caplog):
 
     external_regressors, n_vols = sample_external_regressors()
     external_regressor_config = sample_external_regressor_config()
-    external_regressor_config_expanded = external.validate_extern_regress(
+    external_regressors, external_regressor_config_expanded = external.validate_extern_regress(
         external_regressors=external_regressors,
         external_regressor_config=external_regressor_config,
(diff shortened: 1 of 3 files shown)


bpinsard:
badluck: OSF is down for maintenance  https://fosstodon.org/@CenterforOpenScience

tedana/tests/test_external_metrics.py
```diff
@@ -186,14 +186,15 @@ def test_validate_extern_regress_succeeds(caplog):
...
-    external_regressor_config_expanded = external.validate_extern_regress(
+    external_regressors, external_regressor_config_expanded = external.validate_extern_regress(
         external_regressors=external_regressors,
         external_regressor_config=external_regressor_config,
         n_vols=n_vols,
         dummy_scans=0,
     )
 
     # The regex patterns should have been replaced with the full names of the regressors
+    assert external_regressors.notna().all(axis=None)
```
tsalo, line 197: I don't think there are any NaNs in this test, but if you replace the dt columns' first rows (currently zeroes) with NaNs in the test TSV file, that should work.

bpinsard:
Seems like CircleCI hits rate limit with OSF.

bpinsard:
all green, should be ready now.

PR #1248, opened 2025-08-21
Add 'exclude' option to t2smap workflow

### LINKED ISSUE #1247: Add `t2smap` option to provide T2* map or select volumes for T2* estimation
### Summary

I'm thinking of the potential effect of non-steady-state volumes (a.k.a. dummy volumes) on T2* and S0. Theoretically, if one wanted to estimate T2*/S0 from steady-state volumes, but wanted to combine all volumes, it would be hard to actually do that without the ability to either (1) run t2smap once with the `--dummy-scans` flag to get T2*/S0 and then again, passing in the T2* map and _no_ `--dummy-scans` flag to get the optimally combined data across the full run or (2) run t2smap with a flag that only uses some volumes for T2*/S0 estimation, but still uses all of them for optimal combination (e.g., `--ignore`).

tsalo:
Closes #1247. I'm happy to work on a more advanced method (e.g., something that accepts an array of values) if people would like that.

- Add `--exclude` option to `t2smap`. It accepts a comma-separated list of values (e.g., `1,3,5`), a start-inclusive and end-inclusive range (e.g., `0:5`), or a mix of the two (e.g., `0,2,4,6:10`).

### DIFF
--- a/tedana/tests/utils.py
+++ b/tedana/tests/utils.py
@@ -57,7 +57,7 @@ def data_for_testing_info(test_dataset=str):
     osf_id : str
        The ID for the OSF file.
        Data download link would be https://osf.io/osf_id/download
-       Metadata download link would be https://osf.io/osf_id/metadata/?format=datacite-json
+       Metadata download link would be https://osf.io/metadata/osf_id?format=datacite-json
        If "path" is specified, this value is not returned.
     """
 
@@ -107,7 +107,7 @@ def download_test_data(osf_id, test_data_path):
     """
 
     try:
-        datainfo = requests.get(f"https://osf.io/{osf_id}/metadata/?format=datacite-json")
+        datainfo = requests.get(f"https://osf.io/metadata/{osf_id}/?format=datacite-json")
     except Exception:
         if len(listdir(test_data_path)) == 0:
             raise ConnectionError(


tsalo:
Thank you for sharing that @handwerkerd! I didn't realize how different AFNI's syntax was from Python's. I decided to go with a Python list-like approach for now.

My current draft supports the following:
- 0-based indexing (i.e., 0 is the first volume)
- Ranges that are start-inclusive and end-exclusive (e.g., `0:5` = `0,1,2,3,4`)
- Comma-separated lists (e.g., `0,1,2,3`)
- Combinations of the two (e.g., `0,1,2,5:10`)

It doesn't currently support:
- Open-ended ranges (e.g., `:5`, `100:`)
- Explicit step sizes (e.g., `0:10:2` to get `0, 2, 4, 6, 8`)
- Negative indices (e.g., `-1, -2` to get the last two volumes in the run)

tedana/workflows/t2smap.py
```diff
@@ -232,6 +250,13 @@ def t2smap_workflow(
         aligned with `data`.
     dummy_scans : :obj:`int`, optional
         Number of dummy scans to remove from the beginning of the data. Default is 0.
+    exclude : :obj:`str`, optional
+        Volume indices to exclude from adaptive mask generation and T2* and S0 estimation,
+        but which will be retained in the optimally combined data.
+        Can be individual indices (e.g., '0,5,10'), ranges (e.g., '5-10'),
+        or a mix of the two (e.g., '0,5-10,15').
```
tsalo: Need to update this.

tedana/workflows/t2smap.py
```diff
@@ -315,6 +344,20 @@ def t2smap_workflow(
...
 
+    if fitmode == "ts" and n_exclude > 0:
+        raise ValueError(
+            "Excluding volumes is not supported for fitmode='ts'. "
+            "Please set fitmode='all' or remove the exclude argument."
+        )
+
+    if n_exclude > 0:
+        LGR.info(f"Excluding volumes: {exclude_idx}")
+        # Adjust exclude indices for dummy scans that are already removed
```
tsalo, line 358: data_cat has dummy_scans dropped already so need to adjust based on that as well.

tsalo:
Thanks!

tedana/workflows/t2smap.py
```diff
@@ -232,6 +250,13 @@ def t2smap_workflow(
         aligned with `data`.
     dummy_scans : :obj:`int`, optional
         Number of dummy scans to remove from the beginning of the data. Default is 0.
+    exclude : :obj:`str`, optional
```
handwerkerd, line 254:
```suggestion
        dummy_scans are excluded from the optimally combined data.
    exclude : :obj:`str`, optional
```

Reading the API & help text, this was confusing. Both here and for argparse, add something to say that dummy scans are fully removed to distinguish from `exclude` which affects the T2* estimates, but doesn't remove volumes from the time series.

tedana/workflows/t2smap.py
```diff
@@ -315,6 +344,22 @@ def t2smap_workflow(
     info_dict = utils.get_system_version_info()
     info_dict["Command"] = t2smap_command
 
+    if fitmode == "ts" and n_exclude > 0:
+        raise ValueError(
+            "Excluding volumes is not supported for fitmode='ts'. "
+            "Please set fitmode='all' or remove the exclude argument."
+        )
+
+    if n_exclude > 0:
```
handwerkerd, line 356:
```suggestion
    if dummy_scans > 0:
        LGR.warning(f"Removing the first {dummy_scans} volumes as dummy scans.")
    if n_exclude > 0:
```
Removal of dummy scans wasn't logged anywhere and this makes the text for dummy scans overlapping with excluded volumes more clear.

tedana/workflows/t2smap.py
```diff
@@ -346,8 +391,23 @@ def t2smap_workflow(
...
+    if n_exclude > 0 and np.max(exclude_idx) > n_vols:
+        raise ValueError(
+            f"The maximum exclude index ({np.max(exclude_idx)}) is greater than the number of "
+            f"timepoints in the data ({n_vols})."
+        )
+    elif n_exclude > 0:
+        LGR.info(f"Excluding {n_exclude} volumes from adaptive mask and T2*/S0 estimation")
+        use_volumes = np.ones(n_vols, dtype=bool)
+        use_volumes[exclude_idx] = False
+        data_for_mask = data_cat[:, :, use_volumes]
```
handwerkerd: Minor suggestion to take or ignore. `data_for_mask` is also data for `decay.fit_decay_ts` so it's a confusing name. Maybe change to something like `data_without_excluded_vols`

tsalo: Sounds good! I just pushed that change.

handwerkerd (approved):

tsalo:
@handwerkerd do you mind if I merge and then make a new release for fMRIPrep to use?

tsalo:
It looks like OSF is acting up again.

PR #1250, opened 2025-08-27
Report sort external regressors rows alphabetically in correlation matrix

bpinsard:
Currently, the correlation matrix orders the regressors in an arbitrary order, which seems to lose the order of the file it was loaded from, here is an example:
image

This is a simple fix that simply sorts the index, but maybe fixing the correlation function to keep the loaded TSV order would make more sense.


tsalo:
I'm pretty sure we run hierarchical clustering so that similar regressors are presented together.

PR #1262, opened 2025-10-17
Remove maximum version restriction for nilearn

tsalo:
Closes none. To make the initial change, I did the following:

1. Remove the maximum version for nilearn from `pyproject.toml`.
2. Run `uv lock --upgrade` to update the lock file. Fortunately, it only updated nilearn.

@effigies does that process make sense?

- Remove upper limit on nilearn version from pyproject.toml.
- Update tested nilearn version from 0.11.1 to 0.12.1 and mapca from 0.0.5 to 0.0.6.
- Address resulting bugs.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1450,7 +1450,7 @@ wheels = [
 
 [[package]]
 name = "mapca"
-version = "0.0.5"
+version = "0.0.6"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "nibabel" },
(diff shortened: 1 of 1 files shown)


effigies:
I'd just note for the sake of the changelog that you haven't updated your nilearn requirement to 0.12.1 but just uncapped it. The `uv.lock` does not impact downstream packages at all, it just means your workflows won't be broken by upgrading a dependency unless you actively upgrade them (or allow dependabot to do so).

PR #1266, opened 2025-11-20
Parallelize nonlinear T2*/S0 estimation and RobustICA

### LINKED ISSUE #1091: Nonlinear Curve fitting is slow, and other anatomical T2star fitting issues
### Summary
t2smap is a nice tool for estimating T2* parameters, however, for anatomical data it falls short in several ways. Curve fit is slow, you have to split up the (usually) single 4D data file just to pass the data in and there is no feedback on how long it is expected to take. 

### Additional Detail
Its a minor use case, but its mine. For my data, it would take ~60 hours or more to do the curve fitting and I find that offensive. And I think it may be worth speeding up the curve fit in general. I had put in the progress bar on a branch a while back and never opened a pull request. Want to fix that. 

### Next Steps
I've made modifications to t2smap that:
use joblib Parallel and delayed to run curve fitting - progress bars cleanly show how long it will take, and testing confirms this is an embarrassingly parallel problem - 24 cores provides a ~20x speed up. 
tqdm over the voxel loop makes it easy to see progress. Likely we should print what each internal loop is doing (looping over number of good echoes). For anatomical data with many echoes, you get a lot of progress bars back to back and that can be confusing
image
Probably want to silence covariance warnings ahead of time?
Modified the input parser such that, if a dataset is a 4D file, with timeseries length = n_echoes, then treat that like it was multiple seperate files on the command line. 

So I plan to open a pull request for this - but I wanted to go ahead and spit this out there while I finished up editing the doc strings.

tsalo:
Closes #1091.

- Use joblib to parallelize the curve_fit function calls.
- Use tqdm to make pretty progress bars.
- Pass the `n_threads` parameter to RobustICA as well.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -40,7 +40,7 @@ jobs:
       DEPENDS: ${{ matrix.dependencies }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           submodules: recursive
           fetch-depth: 0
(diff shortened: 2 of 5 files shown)


eurunuela (approved):
LGTM!

PR #1271, opened 2025-12-04
Only mean centering PCA components

handwerkerd:
In playing around with ICA convergence failures @marms7 and I realized that not variance scaling the data pre & post PCA seems to remove almost all convergence failures and (n=1) might even give better final ICA components for denoising. This is a work in progress and we're still testing, but I wanted to share here in case others have thoughts.

- Added a `--mean_center_only` option that removes variance scaling and zscoring of data
  - I don't love this because, if we decide this will eventually be the default, we'll need to change the input option, but it's enough for me (and others to start testing now)
- Switched RobustICA to using `unit-variance` for whitening to match the recommended default fastICA setting and to now match what we're doing in our single call to fastICA (@BahmanTahayori & @Lestropie any reason this might be an issue with robustICA?)

### DIFF
--- a/tedana/decomposition/ica.py
+++ b/tedana/decomposition/ica.py
@@ -161,7 +161,7 @@ def r_ica(data, n_components, fixed_seed, n_robust_runs, max_it):
             robust_ica = RobustICA(
                 n_components=n_components,
                 robust_runs=n_robust_runs,
-                whiten="arbitrary-variance",
+                whiten="unit-variance",
                 max_iter=max_it,
                 random_state=fixed_seed,
                 robust_dimreduce=False,
@@ -320,6 +320,7 @@ def f_ica(data, n_components, fixed_seed, maxit, maxrestart):
             n_components=n_components,
             algorithm="parallel",
             fun="logcosh",
+            whiten="unit-variance",
             max_iter=maxit,
             random_state=fixed_seed,
         )
--- a/tedana/decomposition/pca.py
+++ b/tedana/decomposition/pca.py
@@ -60,6 +60,7 @@ def tedpca(
     kdaw=10.0,
     rdaw=1.0,
     low_mem=False,
+    mean_center_only=False,
 ):
     r"""Use principal components analysis (PCA) to identify and remove thermal noise from data.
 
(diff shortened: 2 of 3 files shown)


Lestropie:
More familiar with how PCA is used in the upstream sliding-window thermal denoising than how it's used in tedana, but I'll play out this train of thought and see if it's applicable.

Some will be familiar with the noise map estimate that comes out of the sliding window noise level estimation. This can be strongly influenced by scanner B1- bias field correction, which makes images homogeneous in brightness in the brain but heteroscedastic. Each of those individual noise level estimates is however derived assuming that the noise level within the finite sliding window patch is constant, which for strong bias fields / large patches may be violated. This can be addressed through use of a variance-stabilising transform, which rescales (and potentially non-linearly transforms) the data based on some prior estimate of the noise level map prior to PCA. My own denoising tool does this in an iterative fashion, using the noise estimate from the previous iteration to standardise variance in the current iteration. From a cursory inspection, I think you're trying to achieve a similar thing here, only that you're using the variance of the empirical data, which is likely not as suitable a reference magnitude for such a transformation as is a principled noise level estimate. So permitting the propagation of a noise level estimate from a prior pipeline step through to inform this step may be an overall more robust approach. Where this is not implemented or unavailable, I don't have any fundamental objections to `unit-variance`, but am not well placed to discover any fault with such.

One thing to watch out for is image voxels that are consistently zero-filled. I've recently seen data from multiple vendors where voxels outside the brain are filled with zeroes across all volumes, and this breaks PCA. Just be sure it's not the Z-transform dividing by zero and introducing NaNs before you even attempt PCA.
https://github.com/Lestropie/dwidenoise2/issues/1

I also recall when investigating demeaning & whitening for fixing MP-PCA issues that it was somewhere advocated to do an iterative double-centering approach of demeaning across both axes multiple times, as otherwise there could be order-dependency issues. 
https://github.com/MRtrix3/mrtrix3/pull/2363
Not sure if that's what this line is doing internally?
`kept_data = stats.zscore(kept_data, axis=None) # variance normalize everything`
But if so, it would seem to make the prior line:
`kept_data = stats.zscore(kept_data, axis=1) # variance normalize time series`
redundant?

PR #1282, opened 2025-12-18
Track decay model fit failures

### LINKED ISSUE #1267: Track and write out spatial map of model fit failures
### Summary

Currently we just let users know how many model fit failures there are in the nonlinear T2*/S0 estimation step, and patch in the log-linear estimates into failing voxels, but I think it would be useful to write out a boolean map indicating if the model was successful or not.

tsalo:
Closes #1267.

- Write out map of model fit failures when using `curvefit`.


eurunuela (approved):
LGTM

PR #1284, opened 2025-12-18
Add Kappa-Rho difference metric

### LINKED ISSUE #1279: Calculate KappaRhoDifference metric
> One metric we tested was a kappa rho difference score. For every component, we calculated a kappa & rho value and the score was `abs(kappa-rho)/(kappa+rho)`. That is, a component that is more purely kappa OR rho has a higher score, which would mean that there is less "mixing" of kappa and rho signal sources within each component. 

_Originally posted by @handwerkerd in https://github.com/ME-ICA/tedana/issues/1271#issuecomment-3666607246_

tsalo:
Closes #1279.

- Add option for a new metric, `kappa_rho_difference`.

### DIFF
--- a/tedana/resources/config/metrics.json
+++ b/tedana/resources/config/metrics.json
@@ -46,6 +46,10 @@
         "normalized variance explained": [
             "map weight"
         ],
+        "kappa_rho_difference": [
+            "kappa",
+            "rho"
+        ],
         "d_table_score": [
             "kappa",
             "dice_FT2",


tsalo:
Adding a new metric makes me think that we should have documentation on the different metrics, with examples and use-cases.

eurunuela:
> Adding a new metric makes me think that we should have documentation on the different metrics, with examples and use-cases.

I think that would be useful. I think a good number of users don't know what kappa and rho actually represent 😅

handwerkerd:
Overall a nice addition. One Q is if we want to add this to the included decision trees or just leave it as a semi-hidden option for now.

PR #1287, opened 2025-12-19
Add auto-detection of T2* map and echo time units

### LINKED ISSUE #1286: Supplying pre-calculated T2* map
Tedana expects -e in ms, but the precalculated T2* appears to be interpreted in seconds. This causes the OCME weighting to be extremely biased, OCME time-series is dominated by last echo. To avoid this issue, I would suggest clarifying in -h / documentation the expected units (sorry if its already there, but I dont see it at first glance) or standardize to have units match (either both ms, or both seconds), or add some kind of unit check that automatically detects this and converts to the correct units on the fly

eurunuela:
## Summary

- Adds `check_t2s_values()` function to auto-detect if precalculated T2* maps are in seconds (per BIDS convention) or milliseconds
- Updates `check_te_values()` to prefer seconds per BIDS convention with deprecation warning for millisecond inputs
- Updates CLI help text for `-e` (echo times) and `--t2smap` to clarify expected units
- Applies unit validation when loading precalculated T2* maps in tedana workflow
- Adds `check_te_values()` to t2smap workflow for consistency

### New function: `check_t2s_values()`
- Auto-detects T2* map units based on median non-zero values
- T2* in seconds (0.01-0.1s): Converts to ms with info message
- T2* in ms (10-100ms): Uses as-is with warning about BIDS convention
- Invalid values (≥1000): Raises error

### Updated: `check_te_values()`
- Echo times in seconds (<1): Converts to ms with info message  
- Echo times in ms (>1): Uses as-is with deprecation warning
- Updated docstring to reflect BIDS preference

### CLI updates
- `-e` help text now shows seconds as preferred with examples
- `--t2smap` help text specifies seconds with auto-detection note

## Test plan

- [x] Added `test_check_t2s_values()` with comprehensive test cases
- [x] Updated `test_check_te_values()` tests
- [x] All existing tests pass (131 tests)
- [x] Flake8 and black checks pass

Fixes #1286

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tedana/tests/test_utils.py
+++ b/tedana/tests/test_utils.py
@@ -485,7 +485,7 @@ def test_check_te_values(caplog):
 
     # Check that the error is raised when TE values are in mixed units
     with pytest.raises(ValueError):
-        utils.check_te_values([0.5, 1, 2.5])
+        utils.check_te_values([0.5, 2, 3])
 
 
 def test_check_t2s_values(caplog):
--- a/tedana/utils.py
+++ b/tedana/utils.py
@@ -789,7 +789,7 @@ def check_te_values(te_values):
     The heuristic used is:
     - If all TE values are between 0 and 1: values are assumed to be in seconds
       (correct per BIDS), converted to milliseconds and returned
-    - If all TE values are > 1: values are assumed to be in milliseconds, a
+    - If all TE values are >= 1: values are assumed to be in milliseconds, a
       deprecation warning is logged, and values are returned as-is
     - Mixed values or negative values raise an error
     """
@@ -798,7 +798,7 @@ def check_te_values(te_values):
         # Values appear to be in seconds (expected per BIDS)
         LGR.info("TE values appear to be in seconds. Converting to milliseconds for internal use.")
         return (te_values * 1000).tolist()
-    elif all(te_values > 1):
+    elif all(te_values >= 1):
         # Values appear to be in milliseconds (deprecated)
         LGR.warning(
             "TE values appear to be in milliseconds. Per BIDS convention, echo times should "
@@ -809,7 +809,7 @@ def check_te_values(te_values):
     else:
         raise ValueError(
             "TE values must be positive and either all in seconds (values < 1, preferred per "
-            "BIDS convention) or all in milliseconds (values > 1, deprecated)."
+            "BIDS convention) or all in milliseconds (values >= 1, deprecated)."
         )
 
 


tedana/tests/test_utils.py
```diff
@@ -413,13 +413,47 @@ def test_millisec2sec():
...
     """Ensure that check_te_values returns the correct values."""
-    assert utils.check_te_values([2, 3, 4]) == [2, 3, 4]
+    # Values in seconds (preferred per BIDS) - should be converted to milliseconds
+    assert utils.check_te_values([0.015, 0.039, 0.063]) == [15, 39, 63]
     assert utils.check_te_values([0.15, 0.35, 0.55]) == [150, 350, 550]
-    # Check that the error is raised when TE values are in different units
+
+    # Values in milliseconds (deprecated) - should be returned as-is with warning
+    assert utils.check_te_values([15, 39, 63]) == [15, 39, 63]
+    assert utils.check_te_values([2, 3, 4]) == [2, 3, 4]
```
tsalo, line 481: Can you use `caplog` to check for the actual warning text?

tsalo: Also it could be good to include an EPTI echo time list here.

tedana/utils.py
```diff
@@ -569,6 +569,93 @@ def millisec2sec(arr):
     return arr / 1000.0
 
 
+def check_t2s_values(t2s_map):
```
tsalo, line 572: We should check the units in the header first. If those aren't available or are unclear, then we should check the values in the array.

tedana/tests/test_utils.py
eurunuela: Done in https://github.com/ME-ICA/tedana/pull/1287/commits/6fc126f3a0c43e90de92b5984d9f4aeaf49f61f5

tedana/utils.py
eurunuela: Done in https://github.com/ME-ICA/tedana/pull/1287/commits/6fc126f3a0c43e90de92b5984d9f4aeaf49f61f5

tedana/utils.py
```diff
@@ -569,6 +569,125 @@ def millisec2sec(arr):
...
+    biased optimal combination weighting.
+    """
+    # Check NIfTI header for time units if image is provided
+    header_units = None
+    if hasattr(t2s_map, "header"):
+        try:
+            _, time_units = t2s_map.header.get_xyzt_units()
+            if time_units in ("sec", "msec", "usec"):
+                header_units = time_units
+                LGR.info(f"T2* map header specifies time units as '{time_units}'.")
```
tsalo: Oh sorry, this is my bad. I forgot that the xyzt units indicate the spatial and time units, not the units of the actual data. Sorry about that!

eurunuela: Would you rather we drop this bit?

tsalo: Yeah, unfortunately the logic doesn't work since the header units here don't indicate anything about the units of the data. As an example, a 4D T2* NIfTI can have time-steps (TRs) in milliseconds but the T2* values might be in seconds. Also, for most T2* maps, they're going to be 3D so I don't think they'll even have a time-step in the NIfTI header.

PR #1298, opened 2026-01-08
Consolidate gscontrol plots

### LINKED ISSUE #1295: Reformat gscontrol plots into component-like plots
### Summary

I just think it would be nice to have the gscontrol (gsr and mir) plots (currently a weight map and a separate time series plot that has both gsr and mir) be more like the component plots.

### Additional Detail

Something like these:

Image

We could also add the gscontrol plots to the donut.

tsalo:
Closes #1295.

- Make tag check in `minimum_image_regression` robust to letter case.
- Use the component plot structure for GSR and MIR plots.

### DIFF
--- a/tedana/gscontrol.py
+++ b/tedana/gscontrol.py
@@ -232,7 +232,9 @@ def minimum_image_regression(
 
         # Select rows where the 'classification_tags' column contains any of the ignore tags
         ign = component_table[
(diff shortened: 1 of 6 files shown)


tsalo:
Here are examples of the new plots

Screenshot 2026-01-08 at 3 25 44 PM
Screenshot 2026-01-08 at 3 25 42 PM

I think they look pretty good!

handwerkerd (approved):
LGTM!

PR #1301, opened 2026-01-09
Add other time series to component plots

### LINKED ISSUE #1296: Add MIR-modified and orthogonalized versions to component plots
### Summary
I think we need to show the MIR and orthogonalized components in some way. At one point we accidentally had those as the actual component plots (#243), but decided not to have it that way.

We could just have the new time series included in the top portion, and maybe the new FFTs in the bottom, but I think the component weight maps would change too and I don't think there's a good way to add them.

tsalo:
Closes #1296.

- Modify the component plots to include time series and FFTs for processed (post-orthogonalization if tedort is enabled, post-MIR if that is enabled) versions of the mixing matrix.

### DIFF
--- a/tedana/reporting/static_figures.py
+++ b/tedana/reporting/static_figures.py
@@ -191,6 +191,7 @@ def plot_component(
     stat_img,
     component_timeseries,
     power_spectrum,
+    labels,
     frequencies,
     tr,
     classification_color,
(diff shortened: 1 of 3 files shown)


tsalo:
Here are a couple of examples. The time series are so similar I'm not sure how useful they are...

Screenshot 2026-01-09 at 11 59 22 AM

Screenshot 2026-01-09 at 11 59 34 AM

tsalo:
Closing. We can figure out a better way to visualize these components.

PR #1303, opened 2026-01-09
Document metrics in more detail

### LINKED ISSUE #1285: Add detailed explanations and examples for metrics
We have some information on Kappa and Rho, but something like `countnoise` could do with an explanation and example images.

Stems from #1284.

tsalo:
Closes #1285.

See https://tedana--1303.org.readthedocs.build/en/1303/dependence_metrics.html#other-available-metrics.

- Add section on other metrics to dependence_metrics page.


eurunuela (commented):
This is great, thanks! My only question would be: do we want to have equations for each metric?

eurunuela (approved):
Happy with the changes. OK to merge once you update the branch with main.

PR #1305, opened 2026-01-12
Operate entirely on masked data

### LINKED ISSUE #425: Only operate on masked data in array format and keep masks as images

### LINKED ISSUE #1070: `reshape_niimg()` may be too flexible
### Summary

`reshape_niimg` is more flexible than documented. `reshape_niimg` will sometimes take as input a file name, sometimes a 3D image array, and sometimes a 1D image array. It seems to work for all those conditions.

If it's a file name, it loads the file and then reshapes and if it's already a loaded image, it just uses it. If it's already 1D, it squeezes the file and is still 1D.

### Additional Detail

Stems from https://github.com/ME-ICA/tedana/pull/1060#issuecomment-2043671230.

tsalo:
Closes #425.

- Add `mask` attribute for OutputGenerator with base mask image.
- Mask the input data, keeping everything in `M x E x T` instead of `S x E x T`.
- Remove `reshape_niimg` (closes #1070).


tedana/decomposition/pca.py
```diff
@@ -214,10 +218,10 @@ def tedpca(
     data_z = (data_z - data_z.mean()) / data_z.std()  # var normalize everything
 
     if algorithm in ["mdl", "aic", "kic"]:
-        data_img = io.new_nii_like(io_generator.reference_img, utils.unmask(data, mask))
-        mask_img = io.new_nii_like(io_generator.reference_img, mask.astype(int))
+        # Use un-normalized data for maPCA to match GIFT results
```
tsalo, line 222: Comment stemming from cdac098b37f9c5154c85ce40d46e6379669effdb.

tsalo:
Okay, tests are passing, coverage is up, the direct comparison with main is successful, and I ran a head-to-head profile comparison with main that showed about 45% reduction in memory with minimal increase in duration! This is officially ready for review.

eurunuela (approved):
This is awesome. Thank you @tsalo !

tedana/io.py
```diff
@@ -318,7 +325,12 @@ def save_img(self, data, name):
             data = np.float32(data)
 
         # Make new img and save
-        img = new_nii_like(self.reference_img, data)
+        mask = mask or self.mask
+        img = masking.unmask(data.T, mask)
+        if img.ndim == 4:
+            # Only set the TR for 4D images.
+            img.header.set_zooms(self.reference_img.header.get_zooms())
```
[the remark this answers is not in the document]
tsalo: reference_img is created from the input fMRI files, so they shouldn't ever be 3D.

tedana/io.py
```diff
@@ -1124,3 +1030,151 @@ def _infer_prefix(prefix):
...
+    if len(data) == 1:
+        # z-cat data
+        data_img = nb.load(data[0])
+        n_z = data_img.shape[2] // n_echos
+        arr = data_img.slicer[:, :, :n_z, :].get_fdata()
+        # Using slicer to create the image messes up the affine, so we need to create the
+        # image manually. Use a header copy with dimensions updated to match the ref array,
+        # since the original header describes the full z-concatenated volume.
+        ref_img = nb.Nifti1Image(arr, data_img.affine)
+
```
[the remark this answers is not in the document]
tsalo: I don't think this fix is necessary.

tedana/workflows/tedana.py
```diff
@@ -639,8 +639,41 @@ def tedana_workflow(
     LGR.info("Initializing and validating component selection tree")
     selector = ComponentSelector(tree, out_dir)
 
+    # Initialize OutputGenerator with reference image
+    # XXX: This doesn't support AFNI data yet.
```
handwerkerd: I just ran this with AFNI's HEAD files and it seemed to work fine. Any reason this might introduce a problem with AFNI or anything specific I should check?

tsalo: Oh, I think I fixed it and forgot to remove that comment

tsalo: I just removed the comment.

tedana/io.py
handwerkerd, line 312:
```suggestion

        mask : nibabel.Nifti1Image
            A 3D NIfTI1 format image with the same number of True values as values in the only or second dimension of data.
            Used to expand data into a 3D volume for saving.
```

handwerkerd:
```suggestion
        data : nibabel.Nifti1Image or (Mb,) array_like or (Mb x T) array_like
            Data to save to a file.
            If this is img_like then it is treated as unmasked data.
            If this is array_like, it will be expanded to a 3D volume using the mask in the io_generator or the provided mask.
```

I realized this docstring wasn't updated and I was having trouble figuring out how this function was used across the code. Confirm these updates are accurate and linted corrected.

More narratively, I see you're using `save_img` in 3 ways:
1. If data is already a 3D image, then no unmasking
2. If data is a 1D or 2D vector and `mask` is not provided, then expand the data using `self.mask` which is the intiial user-provided mask (or the optimally combined mask? need to clarify)
3. When you're using data for denoising, you've shrunk data even more so you separately pass a volume with the appropriate mask.

handwerkerd (changes requested):
I'm still going through the code and it seems to be working fine, but I'm seeing something that may be unnecessarily confusing. The images are saved with `save_img` in `io.py` You added nice functionally for that function to use masks, but you use it inconsistently. Some of this variation is that you kept the initial functionally so you didn't have to change all calls to `save_file` I'm not sure if that's worth changing.

One suggestion is, everytime you use the option `mask` option for `save_img` you end up taking the mask in `io_generator.mask` and use it to expand a more restrictive mask as part of the function call. Then, in `save_img` you use either `io_generator.mask` or the supplied mask.  
For example the following expands the mask before sending it to another function.

https://github.com/tsalo/tedana/blob/219e5c7ca790b14ffc3e5a7fb7434acae4e7a6a5/tedana/workflows/tedana.py#L911

Why not just give the extra-restrictive mask (`mask_clf` in the above example) to `img_save` and then `unmask(mask_clf, io_generator.mask)` is run within `img_save`. My sense is that would allow for cleaner code in a bunch of places. This is only a suggestion.

tedana/io.py
```diff
@@ -318,7 +325,12 @@ def save_img(self, data, name):
             data = np.float32(data)
 
         # Make new img and save
-        img = new_nii_like(self.reference_img, data)
+        mask = mask or self.mask
```
handwerkerd:
```suggestion
        if mask:
            mask = unmask(mask, io_generator.mask)
```
I haven't tested this, but this is my rough suggestion from my iniital review comment. The idea being the supplied mask is expanded here instead of as part of the input.

handwerkerd (approved):
LGTM.

FWIW, I ran a 5 echo dataset on my Mac with `main` and this PR. The run time was nearly identical, but the max memory usage was ~2GB vs ~1GB.

PR #1310, opened 2026-01-13
Rescale normalized variance explained to 0 - 100

### LINKED ISSUE #1307: Set normalized variance explained to same range as unnormalized variance explained
### Summary

Currently, normalized variance explained ([`calculate_varex_norm`](https://tedana.readthedocs.io/en/stable/generated/tedana.metrics.dependence.calculate_varex_norm.html#tedana.metrics.dependence.calculate_varex_norm)) ranges from 0 to 1, with the values summing to 1, while unnormalized variance explained ([calculate_varex](https://tedana.readthedocs.io/en/stable/generated/tedana.metrics.dependence.calculate_varex.html#tedana.metrics.dependence.calculate_varex)) ranges from 0 to 100, with the values summing to 100. I propose that we scale normalized variance explained to 0 - 100.

### Next Steps

1. In `calculate_varex_norm`, multiply the resulting varex values by 100.
2. Update any decision trees that use normalized variance explained to work with the new scale.

tsalo:
Closes #1307.

- Remove `calculate_varex_norm` and just use `calculate_varex` for both normalized and unnormalized variance explained.
- Multiply PCA-estimated variance explained by 100.


tedana/metrics/collect.py
```diff
@@ -525,8 +525,8 @@ def get_metadata(component_table: pd.DataFrame) -> Dict:
             "LongName": "Normalized variance explained",
             "Description": (
                 "Normalized variance explained in the optimally combined "
-                "data of each component."
-                "On a scale from 0 to 1."
+                "data of each component. "
```
handwerkerd: It's bothered me for a while that the difference between variance and normalized variance isn't obvious from our descriptors. I just dug into the code (again). It looks like non-trivially different code is used for each, but they both end up running `getcoeff` for a fit between the data and the mixing matrix. The difference is that normalized zscores the mixing matrix along one dimemsion (though I think it was already scaled) and then zscores the data across both dimensions.

While this is being updated, maybe update this description to something like: `Variance explained for data that has been normalized (zscaled) across time and voxels.`

handwerkerd (approved):
I have one minor suggested change, but this is otherwise fine.
I did add a "breaking change" tag and we should make sure to highlight this break in the next release notes.

tedana/metrics/collect.py
tsalo: Well the normalized one _does_ use computefeats2 instead of just getcoeff, and that includes extra steps beyond z-scoring the data. It also crops the parameter estimates to between -0.999 and 0.999 and applies a Fisher's r-to-z transform.  From what I can recall, this is supposed to mimic z test statistics, even though it's not actually valid, which comes back to #178.

tsalo: Oh, also the mixing matrix _is_ z-scored over time before either version of varex is calculated, and I don't think the data are z-scored across both directions for the normalized varex (just over time).

eurunuela (approved):
LGTM, thanks @tsalo !

tedana/metrics/collect.py
handwerkerd: I'm fairly sure the mixing matrix should already be zscored so the extra zscoring might not do anything. Overall, this is very confusing. Ideally we should clean up that part of the code, but secondarily, we should probably just document the differences between the two somewhere obvious (especially if we also add in #1311 which would increase confusion).

tsalo: Agreed. I just don't know how best to describe the normalized variance explained because it isn't what one would think it would be (varex from standardized parameter estimates) and what it _actually_ is isn't well-defined. I can fill in the docstring so at least what's happening is described accurately though.

tsalo: I've updated the metadata description. Do you want to re-review or should I merge?

handwerkerd: You can merge.

PR #1311, opened 2026-01-13
Add metrics for common measures of explained variance

### LINKED ISSUE #1308: Calculate raw variance explained
### Summary
I think it would be more useful to show the variance explained by a component in the optimally combined data, rather than just in the reconstructed ICA data (i.e., the ICA betas), in the component plots. I'd also like to see the variance explained in the absence of other components, rather than the variance explained when accounting for variance explained by other components. This means the values would sum to >100, but I think it's more informative when looking at components in isolation, as in the component plots.

### Next Steps

Here's a draft of a function to calculate this:

```python
def calculate_varex_raw(
    *,
    data_optcom: np.ndarray,
    mixing: np.ndarray,
) -> np.ndarray:
    """Calculate variance explained by the mixing matrix from the optimally combined data.

    Parameters
    ----------
    data_optcom : (S x T) array_like
        Optimally combined data.
    mixing : (T x C) array_like
        Mixing matrix.

    Returns
    -------
    varex : (C) array_like
        Variance explained for each component.
    """
    assert data_optcom.shape[1] == mixing.shape[0]
    data_optcom_z = stats.zscore(data_optcom, axis=1)
    total_var = (data_optcom_z ** 2).sum()

    varex = np.zeros(mixing.shape[1])
    for i_comp in range(mixing.shape[1]):
        ts = mixing[:, i_comp][:, None]
        beta = get_coeffs(data_optcom_z, ts)
        varex[i_comp] = (1 - ((data_optcom_z - beta.dot(ts.T)) ** 2.0).sum() / total_var) * 100
    return varex
```

tsalo:
Closes #1308.

For example, if we have two components that each explain 20% of the optimally combined data's variance, but also share 50% of their variance (and thus the total variance explained by the two would be 30%), this varex value will be 20% for each of those, while "semi-partial R-squared" would be 10% for each, and the "normalized coefficient energy" (what we call "normalized variance explained") would be ~50% (dividing the shared variance equally).

- Add the following new metrics:
    - "marginal R-squared": The "variance explained" for each component without controlling for other components.
    - "partial R-squared": The variance explained by each component after regressing the other components out of the data *and* the component itself. It is a conditional effect size.
    - "semi-partial R-squared": The variance explained by each component after regressing out the other components from the target component. It indicates the incremental increase in R-squared when adding the target component to the model.


tsalo:
I asked ChatGPT about the different varex measures we have and what they should actually be called:
- "variance explained" would be better named "relative standardized coefficient energy (RCE)" or "relative standardized beta energy"
- "normalized variance explained would be "relative Fisher-z–transformed coefficient energy" or "relative Fisher-z beta energy"
- my new "raw variance explained" should be "regressor-wise marginal R²" or "univariate explained variance"
- If we calculated variance explained like I do with "raw variance explained" (i.e., against optcom), but with a single model fit so that shared variance is split up evenly across components, then it would be called "partial R²"
- What I was planning to call "unique variance explained" should be called "semi-partial R²"
- Also there's "partial R²", which I think is what we should be using instead of "normalized variance explained" for statistical inference, but I'm not sure.

tsalo:
After reading through the ChatGPT summary I'd rather change our naming convention than try to fit this new measure into it. What we're calling variance explained just isn't variance explained. If we don't feel a need for any actual measures of variance explained I think that's actually fine, as long as we name what we _do_ use correctly. I can address that in a separate PR though- I'll just rename the new measure in this PR to be correct.

tedana/metrics/collect.py
```diff
@@ -534,6 +566,60 @@ def get_metadata(component_table: pd.DataFrame) -> Dict:
             ),
             "Units": "percent",
         }
+    if "relative variance explained" in component_table:
+        metric_metadata["relative variance explained"] = {
+            "LongName": "Relative variance explained",
+            "Description": (
+                "Relative variance explained by each component in the optimally combined data. "
```
eurunuela:
```suggestion
                "Relative variance explained by each component relative to the optimally combined data. "
```

eurunuela: Maybe this makes it clearer?

eurunuela (changes requested):
This all looks good. I added a comment for one of the metric descriptions hoping it will make it easier to understand.

Other than that, I think it would be good to have these in the docs as well. I don't know if having their own page within the docs would be the way to go. Any ideas?

tsalo:
It might be best to merge #1303 first, since it adds a section that describes metrics. Then I could add descriptions of these metrics to that.

tedana/metrics/collect.py
```diff
@@ -534,6 +566,60 @@ def get_metadata(component_table: pd.DataFrame) -> Dict:
             ),
             "Units": "percent",
         }
+    if "relative variance explained" in component_table:
+        metric_metadata["relative variance explained"] = {
+            "LongName": "Relative variance explained",
+            "Description": (
+                "Relative variance explained by each component relative to the optimally combined data. "
```
tsalo:
```suggestion
                "Relative variance explained by each component relative to the optimally "
                "combined data. "
```

tedana/metrics/dependence.py
```diff
@@ -412,13 +414,168 @@ def calculate_varex(
...
+
+    n_vols = mixing.shape[0]
+
+    # Z-score over time
+    mixing = stats.zscore(mixing, axis=0)
+    data_optcom = stats.zscore(data_optcom, axis=1)
+
+    # S x C correlation matrix of each component with the data
+    # correlation = (1/(T-1)) * dat_z @ mix_z
+    # NOTE: We use population scaling here to match the behavior of np.corrcoef.
```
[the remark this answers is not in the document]
tsalo: This is incorrect. The results are the same as corrcoef. I don't know if that's a bug in corrcoef or if you're just wrong, but the results match.

eurunuela (approved):
I really like the new explanations of what exactly is each of the metrics. Thank you for looking into this @tsalo!

Issue #1265, opened 2025-11-20
Dependabot is not working with uv


tsalo:
### Summary

I haven't seen any Python package-related dependabot PRs since we merged #1229, so I'm a little worried that it's not working. We're still seeing action dependency-related dependabot PRs, so dependabot is still working generally. It might just not be targeting `uv` as expected.

PR #1316, opened 2026-01-15
chore: uv lock --upgrade

effigies:
Closes none.

Given that Dependabot is failing to run (#1265, https://github.com/dependabot/dependabot-core/issues/12042), this is just a bump to see if the latest version of everything is still passing.


tsalo:
That would be great, thank you!

tsalo (approved):
LGTM!

Issue #1312, opened 2026-01-14
January 2026 Developers' call


handwerkerd:
Our next developers call is scheduled for: Thursday, January 15th at 7:00 AM PST / 8:00AM MST / 9:00AM CST / 10:00 AM EST / 3:00PM UTC / 4:00 PM CET / 2:00AM (Friday) AEDT. [You time here](https://www.timeanddate.com/worldclock/fixedtime.html?msg=tedana+dev+call+January+2026&iso=20260115T10&p1=263&ah=1).

[Teams meeting link](https://teams.microsoft.com/l/meetup-join/19%3ameeting_ZDRhOGJjZjctNzcyZi00NDNiLTk4MDQtZjE4NGRmN2EyODQ3%40thread.v2/0?context=%7b%22Tid%22%3a%2214b77578-9773-42d5-8507-251ca2dc2b06%22%2c%22Oid%22%3a%22d0454ceb-fd20-4027-9d7b-1c7806332d29%22%7d)
Meeting ID: 248 529 883 113 09
Passcode: 9LS6Ge6N

[Agenda](https://docs.google.com/document/d/1cAmPHXO827mLMhgA_UZANEMdTs4hmWNN8vNxPfW0ZGA/edit?usp=sharing)

PR #1319, opened 2026-01-15
Deprecate support for z-concatenated inputs

handwerkerd:
As discussed at in our dev call #1312, we do not think anyone is using the zcat option for input anymore. Since creates a non-trivial amount of special cases and may cause problems with some future changes, we plan to deprecate this option.

- If a user provides zcat data as input, it puts up a warning that this option will be deprecated at the end of 2026 and asks the user to contact the tedana developers (so that we know if anyone is still using this option)
- Removes the descriptive text for the zcat option from the command line help (though it will still be functional). It is still mentioned in function docstrings.
- The command line doc strings implied that the echoes and data files could be in any order, but the function docstrings said they need to be in ascending order. I'm not sure if any order actually works in all our code, so I tweaked the command line docstring to also specify ascending order.

### DIFF
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -870,6 +870,11 @@ def load_data(data, n_echos=None, dummy_scans=0):
                 raise TypeError(f"Unsupported type: {type(item)}")
 
         if len(data) == 1:  # a z-concatenated file was provided
+            LGR.warning(
+                "DEPRECATION WARNING: We are planning to deprecate supplying a single "
+                "z-concatenated data file at the end of 2026. "
+                "If you are using this option, contact the tedana developers ASAP."
+            )
             data = data[0]
         elif len(data) == 2:  # inviable -- need more than 2 echos
             raise ValueError(f"Cannot run `tedana` with only two echos: {data}")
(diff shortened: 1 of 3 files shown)


tedana/io.py
```diff
@@ -870,6 +870,11 @@ def load_data(data, n_echos=None, dummy_scans=0):
                 raise TypeError(f"Unsupported type: {type(item)}")
 
         if len(data) == 1:  # a z-concatenated file was provided
+            LGR.warning(
+                "DEPRECATION WARNING: We are planning to deprecate supplying a single "
+                "z-concatenated data file at the end of 2026. "
+                "If you are using this option, contact the tedana developers ASAP."
```
tsalo: It would be great to have the URL to a specific issue here.

```suggestion
                "If you are using this option, please comment on https://github.com/ME-ICA/tedana/issues/1313."
```

tsalo (approved):
LGTM!

tsalo (changes requested):
Oh whoops sorry I just noticed the linter error. Can you fix that before merging? It's probably my fault (line too long).

tsalo (approved):
LGTM!

PR #1327, opened 2026-01-21
fix: Add cache-control for data files and improve Rica modal accessibility

### LINKED ISSUE #1323: Rica not clearing cache from previously loaded dataset
### Summary

I just tried to use #1288 after merging. I tried to open up a new dataset. RICA was showing the name of the new dataset, but the ICA visualizations of the previous dataset I had tested it on.

### Additional Detail

I opened http://localhost:8000/rica/index.html in a different browser (original browser was Firefox & second one was Chrome) and the ICA visualizations rendered correctly. I think what's happening is that it's not just the RICA app that's saved in the cache, but also the data from when RICA was put into the cache. @javiergcas May be having a similar issue on his data.

### Next Steps

- Others should test this to confirm this is a general & reproducible problem. Run tedana on one dataset and then run `python open_rica_report.py`. Close that interactive report and end + the server. Run tedana on another dataset and then run `python open_rica_report.py` Are you seeing the ICA components from the new dataset or the previous one?

eurunuela:
## Summary
This PR fixes the browser caching issue reported in #1323 and addresses PR review comments for accessibility improvements.

## Cache Control Fixes (#1323)
When users run tedana on different datasets and open Rica reports consecutively, the browser was caching data files (TSV, PNG, SVG, NIfTI) from the previous dataset because they have identical filenames.

**Changes:**
- Add `DATA_EXTENSIONS` tuple to `RicaHandler` to identify data files that should not be cached
- Extend cache-control headers to all data file types (`.tsv`, `.png`, `.svg`, `.json`, `.txt`, `.nii`, `.gz`)
- Restrict CORS to localhost origins only for security
- Add depth limit (3 levels) to file search to avoid performance issues on large directories
- Use atomic temp file downloads to avoid inconsistent cache state if download fails

## Accessibility Improvements
**HTML changes:**
- Add `role="dialog"`, `aria-modal="true"`, `aria-labelledby` to Rica modal
- Add `aria-label="Close modal"` to close button
- Add `aria-haspopup="dialog"` to trigger button

**JavaScript changes:**
- Move focus to close button when modal opens
- Return focus to trigger element when modal closes
- Trap Tab key focus within modal while open

## Defense in Depth
A complementary fix has been submitted to Rica (ME-ICA/rica#99) that adds client-side cache-busting.

## Test plan
- [ ] Run tedana on dataset A, open Rica via `python open_rica_report.py`
- [ ] Close Rica and stop the server
- [ ] Run tedana on dataset B, open Rica via `python open_rica_report.py`
- [ ] Verify dataset B's data is displayed (not dataset A's cached data)
- [ ] Test modal accessibility with keyboard navigation

Fixes #1323

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/tedana/reporting/data/html/report_body_template.html
+++ b/tedana/reporting/data/html/report_body_template.html
@@ -307,16 +307,16 @@
 
 <!-- Rica Button -->
 <div class="rica-button-container">
-  <button type="button" class="rica-button" onclick="openRicaModal()">
+  <button type="button" class="rica-button" onclick="openRicaModal()" aria-haspopup="dialog">
     🔍 Open in Rica
   </button>
 </div>
 
 <!-- Rica Modal -->
-<div class="rica-modal-overlay" id="ricaModal">
+<div class="rica-modal-overlay" id="ricaModal" role="dialog" aria-modal="true" aria-labelledby="ricaModalTitle">
   <div class="rica-modal">
-    <button type="button" class="rica-modal-close" onclick="closeRicaModal()">×</button>
-    <h2>Open in Rica</h2>
+    <button type="button" class="rica-modal-close" onclick="closeRicaModal()" aria-label="Close modal">×</button>
+    <h2 id="ricaModalTitle">Open in Rica</h2>
     <p>
       <strong>Rica</strong> is an interactive web-based tool for exploring ICA components
       with 3D brain visualization and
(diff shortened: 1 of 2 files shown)


handwerkerd (approved):
This worked for me. The computing cluster @javiergcas had the problem on is down through Monday (scheduled maintenance) so he can't test until then. I'm fine merging this and, if there are continued problems, we'll let you know.

PR #1328, opened 2026-01-22
Add option to calculate valid univariate z statistics in decision tree

tsalo:
Related to #178. I can do some comparisons between the old and new approaches if anyone wants that.

- Add function `voxelwise_univariate_zstats`, which calculates valid univariate z-statistic maps for regressors by separately regressing the data on each regressor's time series, rather than running a single regression with  all of the regressors.
    - Based on my understanding, this is not useful for ICA, where the component time series are only valuable in the context of the full mixing matrix, but it could be useful for (1) external regressors or (2) evaluating correlations with removed components after denoising.
- Add metric "map univariate Z statistics" for the valid, univariate z-statistic maps. This is distinct from "map Z" and "map weight", which should be removed and/or renamed (see #1332).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,6 +3,10 @@
 # Please see the documentation for all configuration options:
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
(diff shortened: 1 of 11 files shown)


eurunuela (approved):
LGTM, thanks!

PR #1332, opened 2026-01-23
Replace pseudo-z-statistics with standardized parameter estimates

### LINKED ISSUE #178: Calculating coefficients in computefeats2

### LINKED ISSUE #1331: Replace pseudo-z-statistic maps with standardized parameter estimate maps
### Summary
As I've been working on #1328, I've been comparing the values between the proposed approach (valid z-statistics) and existing approaches (what we call "betas", "weights" and "z-statistics"). One thing I've realized is that valid z-statistics will decrease as the amount of variance explained by the component decreases. This makes perfect sense, of course, since a component that doesn't explain much of the variance won't have many significant voxels. However, it means that we wouldn't get a meaningful measure of "speckliness" for components that don't explain much variance, since they won't have many, if any, significant voxels to group into clusters vs. noise.

I still think it's probably useful to have the ability to calculate valid z-statistics, but we _shouldn't_ be using them for the metrics that we are using them for.

In a test dataset, it looks like the proportion of significant voxels in each pseudo-z map is between 4 and 6%. I'm thinking we should use standardized parameter estimate maps and select the top 5% of voxels as our "thresholded" map.

tsalo:
Closes #1331. Closes #178.

The old approach did the following to calculate the "weight maps":

1. Calculate standardized parameter estimate maps.
2. Crop the values to -0.999 - 0.999.
3. Fisher's z-transform the values.

The old approach then calculated the "Z maps" by doing the following to the weight maps:

1. Z-score the Fisher's z-transformed values.
2. Crop the values to -8 to 8.

These "Z maps" were treated as valid z-statistic maps, even though they weren't.

Now, we just calculate and use standardized parameter estimate maps. We use a proportional threshold of the top 5% (two-sided) of voxels for cluster-extent thresholding, rather than a hardcoded value threshold of 1.96 (p < 0.05, two sided).

- Remove `computefeats2`.
- Calculate standardized parameter estimates by z-scoring the mixing matrix and the data and then using `get_coeffs`.
- Use a proportional threshold for standardized parameter estimate maps (95th percentile, using the absolute values of the maps) instead of a pseudo-z-statistic threshold (1.95) that preserved roughly the same proportion of voxels.
- Modify `tedana.metrics.dependence.threshold_map`, `tedana.metrics.dependence.compute_signal_minus_noise_z`, `tedana.metrics.dependence.compute_signal_minus_noise_t`, and `tedana.metrics.dependence.compute_countnoise` to accept either a value threshold or a proportion threshold from which to calculate a value threshold.


tedana/io.py
```diff
@@ -780,7 +781,10 @@ def writeresults(ts, mask, component_table, mixing, io_generator):
         LGR.info(f"Writing denoised ICA coefficient feature set: {fout}")
 
         # write feature versions of components
-        feats = computefeats2(split_ts(ts, mixing, mask, component_table)[0], mixing[:, acc], mask)
+        accepted_ts = stats.zscore(ts, axis=-1)
+        mixing_z = stats.zscore(mixing, axis=0)
+        feats = get_coeffs(accepted_ts[mask, :], mixing_z)
+        feats = feats[:, acc]
```
tsalo: I'm not sure this is actually valid. Would it be easier to just load the z-scored ICA components img and select the accepted components from it?

tedana/metrics/collect.py
```diff
@@ -191,7 +191,7 @@ def generate_metrics(
 
     if "map Z" in required_metrics:
         LGR.info("Calculating z-statistic maps")
-        metric_maps["map Z"] = dependence.calculate_z_maps(weights=metric_maps["map weight"])
+        metric_maps["map Z"] = metric_maps["map weight"].copy()
```
tsalo: I think we can deprecate "map Z" and just use the standardized parameter estimates.

tedana/tests/test_metrics.py
```diff
@@ -319,7 +311,7 @@ def test_smoke_compute_signal_minus_noise_z():
         z_maps=z_maps,
         z_clmaps=z_clmaps,
         f_t2_maps=f_t2_maps,
-        z_thresh=1.95,
+        value_threshold=1.95,
```
eurunuela: Did you not mention changing this to 1.96?

tedana/io.py
eurunuela: I think so. I would avoid recomputing.

eurunuela (changes requested):
Just a couple of questions that need clarification but otherwise looks good to me.

tedana/tests/test_metrics.py
tsalo: I did, although that was when I thought we'd get z test statistics. Happy to change it in the tests though.

tedana/io.py
tsalo: I just moved the step where we calculate the full set of standardized PE maps from the tedana workflow into writeresults, which makes it easier to select the accepted components and skip any recomputing.

tedana/tests/test_metrics.py
eurunuela: Ah okay! In that case I'm happy with the changes.

eurunuela (approved):
LGTM, thanks!

PR #1337, opened 2026-01-29
Use standardized parameter estimates for metrics instead of unstandardized ones

### LINKED ISSUE #1336: Remove/replace metrics calculated from unstandardized parameter estimates
### Summary

Unstandardized parameter estimates (what we call "map optcom betas" in [`generate_metrics()`](https://github.com/ME-ICA/tedana/blob/75a0d48518cb1c0507be8db5384fa36cd4cda667/tedana/metrics/collect.py)) are scaled by the variance of the data, which means that the parameter estimate maps will be confounded by the scale of the data. I don't think it makes sense to calculate dependence metrics, such as "variance explained" (more appropriately called coefficient energy), "dice_FT2", and "dice_FS0", with maps that have that confound in them. Instead, I propose that we use the standardized parameter estimate maps added in #1332 for these measures.

I do think we probably want to use the unstandardized maps for figures, but I could be wrong about that. (EDIT: I tried showing the standardized maps in the figures and they looked terrible, so I'm happy with using the unstandardized ones)

### Additional Detail

Related to #1309, #1315, and #1332.

tsalo:
Closes #1336.

- Add new metrics:
    - map SPE T2 clusterized: Use the standardized parameter estimate maps instead of the unstandardized ones.
    - map SPE S0 clusterized: Use the standardized parameter estimate maps instead of the unstandardized ones.
    - dice_FT2_SPE: Use the standardized parameter estimate maps instead of the unstandardized ones.
    - dice_FS0_SPE: Use the standardized parameter estimate maps instead of the unstandardized ones.
- Deprecate the following metrics with warnings.
    - map beta T2 clusterized
    - map beta S0 clusterized
    - dice_FT2
    - dice_FS0
    - varex kappa ratio
    - d_table_score

I might need to figure out alternatives to `varex kappa ratio` (which is calculated within the decision trees) and `d_table_score` (both calculated outside and within decision trees).


eurunuela (approved):
These look good to me, thanks. That said, I have not tried running it on data. I wonder how much it changes the classification, and therefore the denoising. Have you seen big changes? If so, it might be worth posting some comparison figures here.

tsalo:
>I'm a bit hesitant to change a key parameter without giving users a way to compare. I'm fine making this a breaking change and making this change the default, but maybe retain a way to calculate and save the older metrics.

I agree. I don't love having a metric that doesn't make sense, but given how much using a more appropriate metric changes things, it seems like we still need to support the old one. I can have the original ones called `<measure>_pe` and use them in `tedana_orig`.

tsalo:
I've reduced the changes to (1) add warnings when using the deprecated metrics and (2) add new metrics that use the standardized parameter estimates. The bad metrics are still incorrectly named (i.e., using "betas", which implies standardized parameter estimates instead), but I'm not sure there's a good way around that while still keeping things working as they did in the past.

PR #1338, opened 2026-01-29
Refer to unstandardized PEs as PEs and standardized PEs as betas

### LINKED ISSUE #1309: Rename betas and weights
### Summary

We currently refer to parameter estimates calculated from the z-scored mixing matrix and the un-normalized optimally combined data as "betas" and parameter estimates calculated from the z-scored mixing matrix and the z-scored optimally combined data (followed by `numpy.arctanh` and variance rescaling) as "weights". In general, I believe most people use the word "beta" to refer to standardized parameter estimates instead of unstandardized ones (originally brought up in #448).

To be honest, I think we should also just simplify `computefeats2` to just z-score the data and mixing matrix, without the weird extra steps (cropping values >0.999 or <-0.999, then `numpy.arctanh`, then variance rescaling). It's not clear to me if we want standardized parameter estimates (betas) or z-statistics from a significance test. If we want the latter, I think we can simplify the code.

Related to #178.

tsalo:
Closes #1309.

- Consistently refer to unstandardized parameter estimates (which we'd been calling "betas") as "PEs", "pes", "parameter estimates", or "unstandardized parameter estimates".
- Consistently refer to standardized parameter estimates (with which we'd recently replaced our Fisher's Z-transformed standardized parameter estimates [see #1332] and had been calling "weights" or "Z-statistic maps") as "betas" or "standardized parameter estimates".
- Remove the now-unused (as of #1332) `calculate_z_maps` function.

As a result, a number of metrics have been renamed:
- "map Z": "map optcom standardized parameter estimates"
- "map Z clusterized": "map optcom standardized parameter estimates clusterized"
- "map weight": "map optcom standardized parameter estimates"
- "map weight clusterized": "map optcom standardized parameter estimates clusterized"
- "map optcom betas": "map optcom parameter estimates"
- "map echo betas": "map echo parameter estimates"
- "map beta T2 clusterized": "map PE T2 clusterized"
- "map beta S0 clusterized": "map PE S0 clusterized"


tsalo:
Some things will only make sense once #1337 is merged, so I'll hold off on requesting reviews until that's done.

PR #1340, opened 2026-01-30
fixing a pandas string to bool issue

handwerkerd:
@marlyr was running tedana with the newest version of pandas and tedana crashed. We think the issue arises in https://pandas.pydata.org/docs/whatsnew/v3.0.0.html due to `The str dtype can only hold strings (or missing values), in contrast to object dtype. (setitem with non string fails)`

- We identified one place in the code where we copied a pandas series will `str` values and then overwrote the values as `bool`. There was no reason to do this. This PR creates a numpy bool array and then assigns new values there.

### DIFF
--- a/tedana/selection/component_selector.py
+++ b/tedana/selection/component_selector.py
@@ -653,12 +653,12 @@ class ComponentSelector:
     @property
     def likely_bold_comps_(self) -> int:
         """A boolean :obj:`pandas.Series` of components that are tagged "Likely BOLD"."""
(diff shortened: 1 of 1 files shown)


tsalo:
I think you can accomplish what you want with the following:

```python
def likely_bold_comps_(self) -> pd.Series:
    return self.component_table_["classification_tags"].str.contains("Likely BOLD")
```

tsalo (approved):
LGTM!

PR #1348, opened 2026-02-04
Fix handling of op3 in dec_left_op_right

tsalo:
Closes none, but addresses a bug I identified while working on a decision tree. In nodes with three conditionals, the operator from the second conditional (op2) was being used in place of the third conditional's operator (op3).

Fortunately, the only decision tree impacted by this bug (i.e., the only one where op2 and op3 were different) is `demo_external_regressors_motion_task_models`, for the "Fits motion external regressors" (< vs >), "Fits CSF external regressors" (< vs >), and "Fits task" (just > vs >=) nodes.

- Use `op3` where appropriate in `dec_left_op_right`.
- Compare left3, right3, and op3 against None instead of implicit boolean, which would be a problem for thresholds of 0, to match handling of first and second conditionals.

### DIFF
--- a/tedana/selection/selection_nodes.py
+++ b/tedana/selection/selection_nodes.py
@@ -342,8 +342,8 @@ def dec_left_op_right(
                 f"left2={left2}, right2={right2}, op2={op2}"
             )
 
-    # If any of the values for the second boolean statement are set
-    if left3 or right3 or op3:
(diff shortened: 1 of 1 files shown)


eurunuela (approved):
LGTM. Thanks for catching this!

PR #1350, opened 2026-02-06
Calculate dependence metrics for modified mixing matrices and confounds

### LINKED ISSUE #1300: Calculate dependence metrics for external regressors and gscontrol time series
### Summary

I think it would be useful to have dependence metrics for external regressors. We would need a new output though. It would be like the metrics TSV file, but regressor by metric instead of component by metric.

### LINKED ISSUE #1341: Recalculate dependence metrics for orthogonalized and post-MIR mixing matrices
### Summary
It just occurred to me that we might want to do this, even if only to validate that these steps are improving things. For example, I would _hope_ that orthogonalizing rejected components with respect to accepted components would make the rejected components more TE-independent (i.e., higher rho and lower kappa).

tsalo:
Closes #1300 and closes #1341.

- Run `generate_metrics` on the following:
    - The orthogonalized mixing matrix if `--tedort` is enabled.
    - The post-MIR mixing matrix if `--gscontrol mir` is enabled.
    - Any external regressors provided by the user.
    - The `gsr` and `mir` time series if `--gscontrol gsr` and/or `mir` are enabled.
- Add `use_multivariate` argument to `generate_metrics`. It should be true for mixing matrices and false for external regressors and confounds. It shouldn't matter for PCA, since orthogonal components should produce the same component weights whether or not a multivariate model is used.


tedana/workflows/tedana.py
```diff
@@ -1076,6 +1076,28 @@ def tedana_workflow(
             "series."
         )
 
+        # Generate a new component table from the orthogonalized mixing matrix
+        LGR.info("Generating metrics for TEDORT-orthogonalized mixing matrix")
+        orig_verbose = io_generator.verbose
+        io_generator.verbose = False
+        orth_component_table, _ = metrics.collect.generate_metrics(
```
tsalo, line 1086: One thing I learned from trying out this branch on real data is that the dependence metrics will change only for accepted components when using the orthogonalized mixing matrix. Given how orthogonalization is meant to be used (i.e., you would do denoising using only the orthogonalized rejected components), it might make more sense to calculate the dependence metrics on a mixing matrix that only contains the orthogonalized rejected components.

tsalo: I tried generating the metrics using just the orthogonalized rejected components, and the dependence metrics stayed the same as the original values for those components! I'm a little surprised.

PR #1351, opened 2026-02-06
Generalize the generate_decision_table_score function

tsalo:
Closes none, but should make calculating variants of the `d_table_score` metric in #1337 easier..

- Modify the `generate_decision_table_score ` function to accept two lists of 1D arrays (ascending and descending) instead of specific metrics.


eurunuela (approved):
LGTM

PR #1367, opened 2026-02-19
Possible fix for CircleCI testing file download problem

### LINKED ISSUE #1330: Reduce number of data downloads during tests
### Summary
Our tests are regularly failing because OSF is rejecting download attempts. We should reduce the number of data downloads.

1. Unit tests are run twice (once on GH with Tox and once on CircleCI).
2. I think the test data download code checks that the files haven't been updated, but I think the check isn't working as intended and the data are downloaded every time they're used (i.e., multiple times in a suite of tests).
3. We could save the data as a cache on CircleCI so it just grabs those files instead of downloading them from OSF. The cache can have a version number that we change if and when we update the test data.

handwerkerd:
Closes #1330

CircleCI seems to periodically fail when trying to download data from osf. I suspect putting a call to `raise_for_status` into a `try` `except` clause might resolve this. It works on my local system so I figured I'd open up a PR to see if it functions well in interaction with CircleCI.

- A `try` `except` clause is used `download_test_data` to see if `requests.get` failed, which was the main issue behind local filewalls. On the assumption that this might not immediately crash on CircleCI, but the subsequence `raise_for_status` call fails, that second statement is moved into the `try` clause.

### DIFF
--- a/tedana/tests/utils.py
+++ b/tedana/tests/utils.py
@@ -108,6 +108,7 @@ def download_test_data(osf_id, test_data_path):
 
     try:
         datainfo = requests.get(f"https://osf.io/metadata/{osf_id}/?format=datacite-json")
+        datainfo.raise_for_status()
     except Exception:
         if len(listdir(test_data_path)) == 0:
             raise ConnectionError(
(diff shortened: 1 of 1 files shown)


tsalo (approved):
LGTM

PR #1370, opened 2026-02-22
Check that classification tags are used in decision trees and that each decision tree includes the Likely BOLD tag

### LINKED ISSUE #1346: Use classification, not tags, to determine if BOLD components found
### Summary

I am currently playing with a decision tree that gives detailed classification tags (much like the old "rationales") instead of "Likely BOLD" and "Unlikely BOLD", and I noticed that tedana was trying to rerun the ICA step because no BOLD components were found, even though many components were accepted. I think we should check against the classification (accepted vs. rejected) instead of the classification tags, which are not required to include "Likely BOLD", when determining if BOLD components have been found.

tsalo:
Closes #1346. This should make it easier to use tedana with decision trees that don't use the Likely BOLD tag.

- Use accepted classification to trigger ICA restart in tedana workflow, instead of the Likely BOLD tag. This was already the case in the ica_reclassify workflow, so I just needed to change the main tedana workflow.

### DIFF
--- a/tedana/selection/component_selector.py
+++ b/tedana/selection/component_selector.py
@@ -156,6 +156,7 @@ def validate_tree(tree: Dict) -> Dict:
         default_classifications
     )
     all_decide_comps = set(tree.get("intermediate_classifications")) | set(default_decide_comps)
+    all_tagset = set()
     for i, node in enumerate(tree["nodes"]):
         # Make sure each function defined in a node exists
         try:
@@ -240,12 +241,21 @@ def validate_tree(tree: Dict) -> Dict:
                 tagset.update({node["kwargs"]["tag_if_false"]})
             if "tag" in node.get("kwargs").keys():
                 tagset.update({node["kwargs"]["tag"]})
+
+            all_tagset.update(tagset)
             undefined_classification_tags = tagset.difference(set(tree.get("classification_tags")))
             if undefined_classification_tags:
                 LGR.warning(
                     f"{sorted(tagset)} in node {i} of the decision tree includes a classification "
                     "tag that was not predefined"
                 )
+
+    unused_tags = set(tree.get("classification_tags")) - all_tagset
+    if unused_tags:
+        LGR.warning(
+            "The following classification tags are defined for the decision tree, "
+            f"but are not used in any node: {', '.join(sorted(unused_tags))}"
+        )
 
     # If there is an external_regressor_config field, validate it
     if tree["external_regressor_config"] is not None:
--- a/tedana/selection/component_selector.py
+++ b/tedana/selection/component_selector.py
@@ -147,6 +147,10 @@ def validate_tree(tree: Dict) -> Dict:
             f"Decision tree includes fields that are not used or logged {sorted(unused_keys)}"
         )
 
+    # Check for required classification tags
+    if "Likely BOLD" not in set(tree.get("classification_tags")):
+        raise TreeError("The classification tag 'Likely BOLD' is required for all decision trees.")
+
     # Combine the default classifications with the user inputted classifications
     all_classifications = set(tree.get("intermediate_classifications")) | set(
         default_classifications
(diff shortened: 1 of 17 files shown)


handwerkerd:
I just remembered / realized why this is a (solvable) problem.

Node 11 in the meica decision tree ( https://tedana.readthedocs.io/en/stable/included_decision_trees.html#meica-decision-tree ) is replicated in the variants of that tree (I'm fairly sure not the variations of miminal). The flow-chart is accurate, but it hides the mechanics of what is happening. For example here:

https://github.com/ME-ICA/tedana/blob/main/tedana/resources/decision_trees/meica.json#L155-L167

If there are less than 2 `provisionalaccept` components, this node classifies everything as `accepted`. Then, at the point if the code where you are making edits here, it looks for likely BOLD components and restarts if it doesn't find them. It cannot look for no `accepted` components because they will be accepted. This was done this way because it means, when the 10th iteration of ICA is reached, it won't rerun ICA and it won't need to make any additional alterations to the decision tree.

The proposed change here will break the existing functionality.

Proposed solution: That decision tree node tags the components as `No provisional accept`. Instead of looking for no likely BOLD components, look for accepted components that are **not** tagged `No provisional accept`. That would mean it would rerun if there are no accepted components and if all accepted components also have the `No provisional accept` tag. I think that would give you the flexibility to not require any specific tags while still maintaining this quirky functionality in the meica tree variants.

tedana/selection/component_selector.py
```diff
@@ -238,13 +241,22 @@ def validate_tree(tree: Dict) -> Dict:
...
                     f"{sorted(tagset)} in node {i} of the decision tree includes a classification "
                     "tag that was not predefined"
                 )
 
+    unused_tags = set(tree.get("classification_tags")) - all_tagset
+    if unused_tags:
+        LGR.warning(
+            "The following classification tags are defined for the decision tree, "
+            f"but are not used in any node: {', '.join(sorted(unused_tags))}"
+        )
```
tsalo, line 258: To be honest, I feel like this should be an error. @handwerkerd WDYT?

tedana/selection/component_selector.py
```diff
@@ -238,13 +241,22 @@ def validate_tree(tree: Dict) -> Dict:
...
             if "tag" in node.get("kwargs").keys():
                 tagset.update({node["kwargs"]["tag"]})
+
+            all_tagset.update(tagset)
             undefined_classification_tags = tagset.difference(set(tree.get("classification_tags")))
             if undefined_classification_tags:
                 LGR.warning(
                     f"{sorted(tagset)} in node {i} of the decision tree includes a classification "
                     "tag that was not predefined"
                 )
```
tsalo, line 251: It also seems like this should be an error instead of a warning.

tedana/selection/component_selector.py
handwerkerd: I'm very weakly inclined to keep this one a warning because the full decision tree should run as designed. If someone is actually trying to design a decision tree, I'd lean towards letting things run vs ending with an error. That said, if "Likely BOLD" is listed but never use it, it will trigger the exact issue this PR is trying to resolve. I'll defer to your preference.

tedana/selection/component_selector.py
handwerkerd: Same as above, I'm weakly inclined to letting something run that is capable of running as intended, but also fine if you want to more aggressively enforce the specification requirements with an error.

tsalo: I've kept it as a warning, but I'd like to have some way to treat it as an error if we end up creating a CLI for validation (#1378).

tedana/selection/component_selector.py
tsalo: I decided to go with an error for this one, but we can revisit before the next release if you end up preferring a warning.

handwerkerd (approved):
The changes look good and this runs as expected.

One thing I'll highlight because it might interact with your plans for #1360 is that `ica_reclassify` currently works by adding new nodes to the previously executed tree and then running the new full tree. The validator only runs on the intially loaded tree and, as long as that tree includes `Likely BOLD` there no reason for noded added through `ica_reclassify` to include `Likely BOLD`. Right now `ica_reclassify` is only adding nodes via `add_manual()` but, if you want to add something more complex, you might want to run the validator & that wouldn't necessarily need the `Likely BOLD` tag.

tsalo:
Thanks @handwerkerd! I wasn't really thinking of `ica_reclassify`, except when someone wants to run it with a new tree. I don't think we'll need to validate the expanded trees as long as the base trees are valid, but I might be misunderstanding.

PR #1373, opened 2026-02-24
Set zooms in reference image

### LINKED ISSUE #1372: TR information lost when saving files
### Summary

tedana is dropping the TR information (and possibly other header information?) when saving files.

### Additional Detail

I suspect this is due to the file IO changes made in #1305. This is viewable in the testing data. Using AFNI commands:
```
$ 3dinfo -TR three-echo/TED.three-echo/three_echo_Cornell_zcat.nii.gz 
2.000000
$ 3dinfo -TR outputs/three-echo/desc-denoised_bold.nii.gz 
1.000000
```

### Next Steps

- Figure out where the header info me being dropped and make sure to save it.
- Fix this bug before the next version release

tsalo:
Closes #1372.

- Set zooms in reference image for BRIK/HEAD inputs.
- Set zooms in reference image for z-concatenated inputs.

### DIFF
--- a/tedana/io.py
+++ b/tedana/io.py
@@ -1162,9 +1162,12 @@ def _convert_to_nifti1(img, dtype=None):
     else:
         data = np.asarray(img.dataobj, dtype=dtype)
     affine = img.affine
+    zooms = img.header.get_zooms()
 
     # Try to preserve header information where possible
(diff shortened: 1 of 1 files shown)


handwerkerd (approved):
I checked on the test data and my AFNI format data and this resolves the bug.

PR #1381, opened 2026-03-12
Convert 3D AFNI volumes to actual 3D

### LINKED ISSUE #1380: adverse interaction with nibabel & AFNI volumes
### Summary

tedana v26.0.0 was crashing for @mauspad. I isolated it to nibabel loading the inputted mask with different dimensions if it's a nii file vs an AFNI HEAD/BRIK file.

### Additional Detail

The crashes output:
```
"tedana/reporting/static_figures.py", line 804, in plot_adaptive_mask
                      ob.add_contours(
  File "/nilearn/plotting/displays/_slicers.py", line 412, in add_contours
    self._map_show(img, type="contour", threshold=threshold, **kwargs)
  File "nilearn/plotting/displays/_slicers.py", line 475, in _map_show
                      data_bounds = get_bounds(data.shape, affine)
                    File "nilearn/image/resampling.py", line 179, in get_bounds
                      adim, bdim, cdim = shape
ValueError: too many values to unpack (expected 3)
```

I realized that the mask in `io_generator.mask` was 4D, with a 4th dimension of length 1, while the other masks used to generate contours were 3D. This caused the crash.

This goes back to when the mask is initially loaded:

https://github.com/ME-ICA/tedana/blob/586542b37420db065636361abf3e5c26414b54e0/tedana/io.py#L226

The original mask was in AFNI HEAD/BRIK format. I converted it to nii and ran the following

```python
>>> import nibabel as nb
>>> mask_nii = nb.load("mask_epi_anat.sub-11423.nii.gz")
>>> mask_afni = nb.load("mask_epi_anat.sub-11423+tlrc.HEAD")
>>> print(f"{mask_nii.get_fdata().shape=}\n{mask_afni.get_fdata().shape=}")
mask_nii.get_fdata().shape=(86, 102, 86)
mask_afni.get_fdata().shape=(86, 102, 86, 1)
```

Just changing from AFNI to NII removed that additional dimension and tedana no longer crashed.

### Next Steps

Figure out if:
1. This is a bug in nibabel that we should report to them and code a work-around in tedana to fix this issue until the `nibable` bug is fixed
2. We're mis-using `nb.load` and there's a built-in way for us to handle AFNI images correctly.

handwerkerd:
Closes #1380.

nibabel reads in 3D AFNI volumes with a 4th dimension of length 1. This causes problems with some nilearn commands.

- Added `max_dims` option to `io._convert_to_nifti1` If a volume has more than the desired dimensions, it will remove trailing dimensions of length 1 or raise an error
- When data is loaded for a mask, set `max_dims` to 3.
- `io._convert_to_nifti1` was inconsistently used with `nb.load` For any call of `nb.load` that isn't inherently for a nifti file previously outputted by tedana, wrap it with `io._convert_to_nifti1`


mrneont:
+1 for reproducibility!

handwerkerd:
Instead of adding AFNI-writing functionality to nibabel, I decided to just add a small AFNI-formatted dataset to `tedana/tests/data` and write a test. This is now ready for review.

tedana/io.py
```diff
@@ -1173,6 +1186,18 @@ def _convert_to_nifti1(img, dtype=None):
...
+    if max_dim:
+        img_shape = new_img.header.get_data_shape()
+        if len(img_shape) > max_dim:
+            if np.sum(img_shape[max_dim:]) == len(img_shape) - max_dim:
+                # if all the dimensions beyond the first 3 are of length 1, convert to 3D
+                img_data = np.squeeze(new_img.get_fdata(), axis=tuple(range(3, len(img_shape))))
+                tmp_img = new_img_like(new_img, img_data)
+                new_img = tmp_img
+            else:
+                raise ValueError(f"{img.get_filename()} has more than {max_dim} dimensions")
```
tsalo:
```suggestion
    if max_dim and (new_img.ndim > max_dim):
        new_img = nb.funcs.squeeze_image(new_img)
        if max_dim and (new_img.ndim > max_dim):
            raise ValueError(f"{img.get_filename()} has more than {max_dim} dimensions")
```

tedana/io.py
```diff
@@ -19,6 +19,7 @@
 import requests
 from nilearn import __version__ as nilearn_version
 from nilearn import masking
+from nilearn.image.image import new_img_like
 from packaging.version import Version
```
tsalo:
```suggestion
from packaging.version import Version
```

If you like the suggestion below.

tedana/tests/test_io.py
```diff
@@ -319,6 +319,18 @@ def _flaky_load(path):
     assert np.allclose(out, expected)
 
 
+def test_convert_to_nifti1():
+    """Test whether the dimensionality issues with AFNI values are properly handled."""
+
```
tsalo:
```suggestion
    """Test whether the dimensionality issues with AFNI values are properly handled."""
```

tsalo (commented):
Overall it looks like a solid solution to this issue. I have a minor suggestion to streamline the image squeezing code.

tedana/io.py
handwerkerd: Using `ndim` is definitely cleaner so I'll adapt that change. I was cautious with squeeze because, if someone submitted a single-slice image (64x64x1x1), I didn't want to reduce to 2D. It's unlikely anyone would do this, but I also didn't want to reduce (64x1x20x1) to (64x20)

tsalo: `squeeze_image` will only reduce to 3D. In example 2 in [the docs](https://nipy.org/nibabel/reference/nibabel.funcs.html#nibabel.funcs.squeeze_image) they go from 10x1x1 to 10x1x1 (i.e., it doesn't drop the singleton second or third dimensions).

handwerkerd: Fully accepted this suggestion except I removed  the first `max_dim` from `if max_dim and (new_img.ndim > max_dim)` since it's nested in another clause that required `max_dim`

tedana/tests/test_io.py
handwerkerd: No change here. Was this suggestion an accident or did you mis-type?

tsalo: I was trying to remove the space between the docstring and the first line of code, since PEP8 doesn't recommend having one for functions (but does recommend having one for class methods).

tedana/io.py
tsalo: That's a great point

tedana/tests/test_io.py
handwerkerd: fixed

tsalo (approved):
LGTM!

PR #1391, opened 2026-03-30
Block Nilearn 0.13.0/0.13.1 and uv lock

### LINKED ISSUE #1389: Dealing with image header info, nibabel, & nilearn
### Summary

Some of our recent changes (or changes to niilearn) seem to be garbling header info. I'm not yet sure what is or isn't wrong, but this is causing some problems.

### Additional Detail

This issue is most directly prompted by https://neurostars.org/t/tedana-26-0-1-t2-maps-with-extreme-outliers-compared-to-version-24-0-1/35893

Between tedana v24.0.1 and v26.0.1 The T2* map in a user's data seems to be saved as ints with a scaling factor rather than floats and that's causing other issues.

I tried to replicate his issue on my data and also noticed that an AFNI header field `TAXIS_NUMS` is retained as the number of volumes, even though there is only volume. I'm also seeing a warning with AFNI's `3dinfo`
`** WARNING: NIfTI file [path]/tedana_out/v26.0.1_curvefit/T2starmap.nii.gz dimensions altered since AFNI extension was added`

The image is rendering correctly, which is why this wasn't caught earlier, but clearly something is off. FWIW, I'm giving tedana .nii.gz files that were previously AFNI BRIKs so some of our code to trigger the convert to NIFTI that also copied AFNI headers may not have been run.

### Next Steps

- Look at our load/copy/save code to see places where problems might happen
- Look at outputs to see what header fields are being lost or garbled
- @mrneont is trying to make a better mapping for header information. He might be able to help identify what fields we might need to copy/edit that are not being dealt with in our current code.

tsalo:
Closes #1389. This addresses the issue raised in https://neurostars.org/t/tedana-26-0-1-t2-maps-with-extreme-outliers-compared-to-version-24-0-1/35893. My fix for the bug (https://github.com/nilearn/nilearn/pull/6157) has been merged into Nilearn, so when they release 0.13.2/0.14.0 it should work for tedana.

- Specify nilearn shouldn't be 0.13.0 or 0.13.1.
- Run `uv lock`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     "mapca>=0.0.7",
     "matplotlib>=3.7",
     "nibabel>=2.5.1",
-    "nilearn>=0.11.0",
+    "nilearn>=0.11.0,!=0.13.0,!=0.13.1",
     "numpy>=1.21",
     "pandas>=2.0",
     "pybtex>=0.25",
--- a/uv.lock
+++ b/uv.lock
@@ -1700,8 +1700,7 @@ version = "0.0.7"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "nibabel" },
-    { name = "nilearn", version = "0.12.1", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.10'" },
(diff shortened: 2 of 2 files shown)


handwerkerd (approved):
I think this resolves the specific issues I brought up in #1389 so you can close that with this PR. I think a general check of headers is useful, but I'll open a separate issue/PR on that once I have a more specific plan/idea.

PR #1394, opened 2026-03-30
Drop support for Python 3.9

### LINKED ISSUE #1388: Drop support for Python 3.9
### Summary

Python 3.9 is no longer supported (security support ended in October 2025; see https://endoflife.date/python) and it is preventing some dependencies from being updated, such as flake8-black. I propose that we drop support for it.

### Additional Detail

To be honest, this is something we should have planned out before releasing 26.0.0, but it never even occurred to me.

### Next Steps

1. Remove 3.9 from pyproject.toml, tox.ini, and tox.yml.
2. Change 3.9 to 3.10 in circleci config.yml.

tsalo:
Closes #1388.

- Drop support for Python 3.9 and run `uv lock`.


tox.ini
handwerkerd:
```suggestion
py3{10,11,12,13,14}-latest
py310-min
```

handwerkerd (changes requested):
I found one reference to python 3.9 that you missed in tox.ini. Otherwise this looks good.

tsalo:
Thanks for catching that @handwerkerd!

PR #1395, opened 2026-03-30
Add information about new variance outputs

### LINKED ISSUE #1390: Outputs not documented
### Summary
When running tedana with the option --fittype curvefit, tedana generates additional outputs such as t2s_var.nii.gz and t2s_s0_covar.nii.gz. I cannot find any information in the tedana docs regarding what these outputs are.

### Additional Detail
This might be true for other outputs.

tsalo:
Closes #1390.


eurunuela (approved):
LGTM, thanks!

PR #1396, opened 2026-03-30
Remove CircleCI unit test steps

### LINKED ISSUE #1377: Remove duplicate CircleCI unit tests
### Summary

We have Tox up and running, and it runs more quickly than using CircleCI, so I think we should drop any duplicate tests and just use GitHub Actions for unit tests and style checking, and CircleCI for integration tests.

tsalo:
Closes #1377.


handwerkerd:
Reading the changes, how does this affect code coverage? It looks like lines of code that are only run within unit tests might not enter the codecov evaluation. I don't completely understand how tox +GitHub actions are working so maybe it's separately added to codecov via `tox.yml`?

tsalo:
Tox runs the unit tests across a range of Python versions and pushes the coverage reports to CodeCov. CodeCov will merge those coverage reports with the integration test coverage reports coming from CircleCI. CircleCI was previously also running the unit tests and pushing their coverage reports to CodeCov, which was redundant with Tox. The coverage shouldn't change at all, so I think the CodeCov change in this PR is spurious.

handwerkerd (approved):
Assuming tests finish passing here, LGTM. The tests passed locally for me. I suggest merging this, and then merging into #1394. We can then confirm this didn't mess up any tests or coverage reports for that other PR & them merge that one too.

PR #1402, opened 2026-04-03
Account for dummy scans in workflows

### LINKED ISSUE #1401: Dummy scans not removed from data in tedana workflow
### Summary

I think this is a bug that I introduced in #1305. Dummy scans appear to be account for in external regressors, but are not removed from the actual data. I will open a PR to fix the bug ASAP.

tsalo:
Closes #1401. One thing that I have been failing to do is write and push tests that fail before implementing a bug fix. I am trying to adopt that approach more consistently, starting in this PR. Note that the first commit fails due to #1401 because I added an explicit check, and then the next commit fixes it.

- Update tests to fail when dummy scans are not accounted for correctly (i.e., the optcom file should be `dummy_scans` shorter than the input data).
- Remove `dummy_scans` from `data_cat` in t2smap and tedana workflows. It looks like ica_reclassify is already good.

### DIFF
--- a/tedana/tests/test_integration.py
+++ b/tedana/tests/test_integration.py
@@ -190,6 +190,7 @@ def test_integration_five_echo(skip_integration):
     suffix = ".sm.nii.gz"
     datalist = [prepend + str(i + 1) + suffix for i in range(5)]
     echo_times = [15.4, 29.7, 44.0, 58.3, 72.6]
+    dummy_scans = 2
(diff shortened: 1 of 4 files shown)


eurunuela:
Happy to give this a quick review once you have fixed the tests.

tsalo:
@eurunuela that would be great, thank you!

eurunuela (approved):
LGTM. I like the idea of making tests fail for bug fix PRs.

PR #1425, opened 2026-05-15
Use seconds for internal calculations

### LINKED ISSUE #1422: Consistently use TEs in seconds internally
### Summary

This shouldn't change the workflow behavior or outputs, but I think it would simplify the code, as we'd only do a single unit check at the beginning and then consistently use seconds throughout the code.

tsalo:
Closes #1422.

- Change check functions to output seconds instead of milliseconds.
- Remove sec2millisec and millisec2sec functions.
- Update docstrings to specify seconds instead of milliseconds.
- Update tests to use seconds instead of milliseconds.


eurunuela (approved):
LGTM

tsalo:
Thanks!

Issue #217, opened 2019-02-12
Optionally supplement component selection with AROMA


tsalo:
### Summary
In [their 2017 paper](https://journals.plos.org/plosone/article/file?id=10.1371/journal.pone.0173289&type=printable), Dipasquale et al. compared AROMA run on optimally combined data to ME-ICA. While they ultimately determined that ME-ICA was the best choice, given that it did a better job than AROMA at reducing functional connectivity-motion associations. However, AROMA did quite well overall and definitely improved TSNR _far_ more than ME-ICA.

### Additional Detail
[AROMA](https://github.com/maartenmennes/ICA-AROMA) is implemented in pure Python and only uses four metrics for component classification, so it should be fairly straightforward to incorporate into `tedana`. One benefit to incorporating AROMA into `tedana` instead of allowing users to simply run it separately (e.g., at part of fMRIPrep) is that we can perform a single ICA decomposition and classify components according to both sets of criteria at once, rather than running an ICA for `tedana` and then another one for AROMA. A drawback would be that we will need more inputs in order to support AROMA, including motion parameters and transforms to standard space. Luckily, I think that these inputs are things we probably also want for comprehensive reports, so we may want to support them as options anyway.

### Next Steps
1. Determine how best to abstract AROMA methods away from FSL. This could involve contributing to AROMA directly or incorporating its code in `tedana` (AROMA has an Apache license).
     - We need to compute metrics and call the component selection without calling MELODIC, without using FSL's directory structure/filenaming conventions, and without writing out files, if possible (e.g., by returning DataFrames instead of writing out text files).
2. Add `--tedica` argument to the `tedana` workflow. We already plan to ultimately support both v2.5 and v3.2 of Kundu's component selection approach, so this will need to be added at some point anyway.
    - The options could be `kundu2` (default) and `kundu2+aroma`. Or something.
3. Add options for motion parameters and transforms.
    - We at least want motion parameters to improve our visual reports. Transforms will be a bit more problematic, although we could probably use tissue probability maps or masks instead.
4. Add function in `selection` for running AROMA on our ICA results and outputting a DataFrame that we can merge with the `tedana` one we build already.
5. Validate approach.

handwerkerd:
[...]

Issue #1378, opened 2026-03-04
Add CLI to validate decision trees


tsalo:
### Summary
We could add a command-line interface just for validating decision trees. If possible, it would be good to upgrade any warnings to errors in the validation CLI.

This relates to #1370.

PR #1428, opened 2026-05-15
Add high-frequency content and motion parameter correlation metrics from AROMA

tsalo:
Closes none, but relates to #217. I copied the AROMA metrics over from fMRIPost-AROMA. The other two metrics, CSF and edge fraction, would require external masks and are highly dependent on the space and resolution (MNI152NLin6Asym 2 mm isotropic).

- Add new "HFC" (high-frequency content) metric from ICA-AROMA. In ICA-AROMA, components are rejected using a hardcoded HFC threshold, so this is the one metric we are actually porting over.
- Add new "max_RP_corr" (maximum motion-parameter correlation) metric from ICA-AROMA. In ICA-AROMA, this is used in combination with edge fraction, so we can't actually use it the same way ICA-AROMA does.


tedana/resources/config/metrics.json
```diff
@@ -75,6 +77,14 @@
             "countnoise",
             "countsigFT2"
         ],
+        "HFC": [
+            "mixing",
+            "tr"
+        ],
+        "max_RP_corr": [
```
tsalo: I think we can generalize this to accept any external metric regressor-set ("regress_ID") instead of having a dedicated motpars argument.

It might require some changes to how we handle external regressors though. I don't think we have any way to pass them to specific metrics at the moment.

tsalo:
I tested the AROMA decision tree locally and it worked, though the motion metric threshold is arbitrary. In ICA-AROMA, the motion metric (max_RP_corr) is used in conjunction with the edge fraction metric, which we can't reproduce within tedana, so there's no way to implement that classification node properly. But at least we can add the metric itself!

eurunuela (changes requested):
Overall this is a solid PR — the pluggable handler architecture is a nice design and HFC is a well-motivated addition. I have three blockers and a few suggestions below.

**Blockers** (must fix before merge):
1. `calculate_hfc` uses magnitude spectrum but the AROMA-derived threshold assumes power spectrum
2. `add_external_dependencies` hardcodes statistic names, defeating the pluggable handler abstraction
3. `calculate_max_rp_corr` uses an unseeded RNG, making results non-reproducible

**Suggestions** (not blocking, but worth addressing):
- `docs/building_decision_trees.rst` line 311 still says the only valid `statistic` is `"F"` — `max_rp_corr` needs to be added there and the config schema documented
- `external_aroma.json` isn't referenced in `docs/included_decision_trees.rst` or `docs/building_decision_trees.rst` — users won't know it exists
- HFC isn't documented in `docs/dependence_metrics.rst` — would be good to describe what it measures, its range, and note that the 0.35 threshold comes from single-echo AROMA and hasn't been validated on ME data
- The `max_rp_corr` threshold (0.5) and the 6-regressor assumption (the expanded model has 6×N columns) should be documented together — if someone passes N≠6 regressors, the threshold may not be calibrated correctly for their model size
- Is there a tracking issue for re-deriving the HFC and max_RP_corr thresholds on ME-fMRI data before `external_aroma.json` graduates from demo status?

tedana/metrics/frequency.py
```diff
@@ -0,0 +1,58 @@
...
+    -------
+    hfc : (C,) numpy.ndarray
+        High-frequency content score for each component, on a scale from 0
+        (all power at ``f_hp``) to 1 (all power at Nyquist).
+    """
+    mixing = np.asarray(mixing, dtype=float)
+
+    # One-sided magnitude spectrum; skip the DC bin (index 0)
+    fft_vals = np.fft.rfft(mixing, axis=0)
+    mixing_fft = np.abs(fft_vals[1:, :])  # (F, C)
```
eurunuela: The docstring says "power spectrum" but `np.abs()` gives the amplitude (magnitude) spectrum, not power. The original AROMA implementation (Pruim et al. 2015) uses the power spectrum (`np.abs()**2`), and the 0.35 threshold in the decision tree was calibrated against power. Magnitude and power produce different score distributions, so the threshold won't transfer cleanly between them. This should either be changed to `np.abs(fft_vals[1:, :])**2` to match AROMA, or the deviation needs to be justified and the threshold re-derived empirically using this magnitude-based definition.

tedana/metrics/external.py
```diff
@@ -666,3 +666,187 @@ def compute_external_regressor_correlations(
...
+    dependency_config : :obj:`dict`
+        Updated dictionary with external-regressor metric dependencies added.
+    """
+    dependency_config["inputs"].append("external regressors")
+
+    for config_idx in range(len(external_regressor_config)):
+        regress_id = external_regressor_config[config_idx]["regress_ID"]
+        statistic = external_regressor_config[config_idx]["statistic"].lower()
+
+        if statistic == "f":
```
eurunuela: This function hardcodes `if statistic == 'f' / elif statistic == 'max_rp_corr'` in parallel with the `STATISTIC_HANDLERS` dict defined just above it. The whole point of that dict is that adding a new statistic only requires one change, but `add_external_dependencies` breaks that — a third statistic would need edits in both places with no static check to enforce consistency. Could the dependency-registration logic be moved into each handler (or into a dict of dependency builders keyed by statistic name), so this function can just iterate over `STATISTIC_HANDLERS` without needing to know what each statistic produces?

tedana/metrics/external.py
```diff
@@ -666,3 +666,187 @@ def compute_external_regressor_correlations(
...
+    rp2_1bw = np.vstack((rp2[1:], np.zeros(2 * n_params)))
+    rp_model = np.hstack((rp2, rp2_1fw, rp2_1bw))  # (T, 6 * n_params)
+
+    n_volumes, n_components = mixing.shape
+    n_rows_to_choose = int(round(0.9 * n_volumes))
+    n_splits = 1000
+
+    max_correlations = np.empty((n_splits, n_components))
+    for i_split in range(n_splits):
+        chosen_rows = np.random.choice(n_volumes, size=n_rows_to_choose, replace=False)
```
eurunuela: `np.random.choice` here uses the global legacy RNG with no seed, so `max_RP_corr` scores will vary slightly between runs on identical data. For a component classification tool this is a reproducibility problem — the same dataset can yield different component decisions depending on random state. The tests work around it with `np.random.seed(0)` before each call, which shows the issue is known but hasn't been fixed in the API. The function should accept an optional `seed` parameter (e.g. `rng = np.random.default_rng(seed)`) and use that for all sampling, so callers can get reproducible results without reaching into global state.

tedana/metrics/external.py
```diff
@@ -666,3 +666,187 @@ def compute_external_regressor_correlations(
     )
 
     return correlation_df
+
+
+def calculate_max_rp_corr(*, mixing: np.ndarray, regressors: np.ndarray) -> np.ndarray:
```
eurunuela: Suggestion: the expanded model has 6×N columns (raw N + derivatives N + ±1 TR shifts = 6N), and the 0.5 rejection threshold in `external_aroma.json` was calibrated for N=6 motion parameters from the original AROMA paper. The expected maximum correlation under the null grows with the number of regressors, so that threshold will be too lenient when N≠6. Worth adding a note to the docstring flagging this assumption — e.g. that the demo tree's threshold assumes exactly 6 input regressors — so users who supply a different number know they may need to recalibrate.

tedana/resources/decision_trees/external_aroma.json
```diff
@@ -0,0 +1,217 @@
+{
+    "tree_id": "demo_external_regressors_single_model",
```
eurunuela: Two related suggestions:

1. This tree isn't referenced in `docs/included_decision_trees.rst` or `docs/building_decision_trees.rst` — users won't discover it, and the latter still says the only valid `statistic` is `"F"` (line 311), which is now wrong since `max_rp_corr` is supported.

2. Since the HFC (0.35) and max_RP_corr (0.5) thresholds were taken directly from the single-echo AROMA paper and haven't been validated on ME-fMRI data, it would be worth adding a `"notes"` or `"description"` field here explicitly flagging that — and ideally a tracking issue for re-deriving the thresholds on ME data before this tree graduates from `demo_` status.

tedana/metrics/frequency.py
```diff
@@ -0,0 +1,58 @@
+"""Frequency-related metrics."""
+
+import numpy as np
+
+
+def calculate_hfc(*, mixing: np.ndarray, tr: float, f_hp: float = 0.01) -> np.ndarray:
```
eurunuela, line 6: Suggestion: HFC doesn't appear in `docs/dependence_metrics.rst`, so users have no documentation on what the metric measures, what its [0, 1] range means in practice, or why 0.35 was chosen as a threshold. Would be good to add a section there alongside kappa/rho, and note that the threshold comes from single-echo AROMA and hasn't been validated on ME data yet.

tedana/metrics/frequency.py
tsalo: Thanks! Fixed!

tedana/metrics/external.py
tsalo: Fix now.

tedana/metrics/external.py
tsalo: I passed the seed from the CLI into this step so now it's reproducible.

tedana/metrics/external.py
tsalo: The 0.5 rejection threshold is entirely arbitrary. The ICA-AROMA classifier uses a hyperplane with an intercept and slopes for max_RP_corr and edge_fract, and we can't reproduce edge_fract within tedana without accepting an edge mask.

tedana/resources/decision_trees/external_aroma.json
tsalo: I don't plan to merge this decision tree into `main`. I will remove it once everyone has had a chance to review, and can incorporate either of the metrics into an existing decision tree if anyone has a good spot for them.

tedana/metrics/frequency.py
tsalo: I added a section on the metric to the documentation, though I don't think we need to be concerned about validating the frequency content of ICA components in multi-echo data.

docs/building_decision_trees.rst
```diff
@@ -331,7 +337,22 @@ and each specified partial model. This can be used to potentially reject compone
...
+
+When ``statistic`` is ``"max_rp_corr"``, tedana builds an expanded regressor model
+from the selected ``regressors`` and adds a ``max_rp_corr {regress_ID} model`` column
+to ``selector.component_table_``.
+The expanded model contains the original N regressors, their derivatives, and both sets
+shifted forward and backward by one TR, for 6*N model columns.
+The metric also compares squared component and regressor time series, for 12*N total
+correlation comparisons per random split.
+The random subsampling uses the seed provided to the tedana workflow.
+Thresholds taken from ICA-AROMA, such as ``max_rp_corr > 0.5``, assume the original six
```
handwerkerd: Really minor but we should probably add the AROMA reference to `resources/references.bib` and cite it here.

docs/dependence_metrics.rst
```diff
@@ -160,6 +160,47 @@ Other Available Metrics
...
+of the maximum absolute Pearson correlation between the component time series and
+an expanded regressor model.
+The expanded model contains the original N regressors, their derivatives, and both
+sets shifted forward and backward by one TR, for 6*N model columns.
+Correlations are calculated for the raw time series and for element-wise squared
+time series, giving 12*N total comparisons per split.
+
+Values are bounded between 0 and 1, where higher values indicate stronger
+association between the component and at least one expanded regressor.
+Thresholds inherited from ICA-AROMA, such as ``max_rp_corr > 0.5``, assume the
```
handwerkerd: cite AROMA here.

docs/dependence_metrics.rst
```diff
@@ -160,6 +160,47 @@ Other Available Metrics
...
+The component's one-sided power spectrum is calculated, frequencies at or below
+the high-pass cutoff are removed, and HFC is the normalized frequency where the
+cumulative retained power is closest to 50%.
+Values are bounded between 0 and 1.
+Values near 0 indicate that the retained power is concentrated near the high-pass
+cutoff, while values near 1 indicate that the retained power is concentrated closer
+to the Nyquist frequency.
+
+The commonly used threshold of ``HFC > 0.35`` is inherited from single-echo
+ICA-AROMA. It has not yet been validated as a tedana ME-fMRI component
```
handwerkerd: cite AROMA here

tedana/resources/decision_trees/external_aroma.json
handwerkerd:
```suggestion
    "tree_id": "demo_external_aroma",
```

You'd also want to change this file name to `demo_external_aroma.json`

I'm fine including this demo tree along with the other demo trees to give people a reference for how to use these additional metrics.

tedana/resources/decision_trees/external_aroma.json
```diff
@@ -0,0 +1,217 @@
...
+    "info": "Demonstration based on the minimal decision tree that uses an F stat on a model with multiple external regressors.",
```
handwerkerd:
```suggestion
    "info": "Demonstration of a tree that merges some multi-echo decision criteria with head motion and temporal frequency criteria from ICA-AROMA.",
```

tedana/resources/decision_trees/external_aroma.json
```diff
@@ -0,0 +1,217 @@
...
+    "report": "This is based on the minimal criteria of the original MEICA decision tree \\citep{kundu2013integrated} without the more aggressive noise removal steps \\citep{dupre2021te}.",
```
handwerkerd:
```suggestion
    "report": "This is based on the minimal criteria of the original MEICA decision tree \\citep{kundu2013integrated} with some of the criteriea from ICA-AROMA REFERENCE.",
```

tedana/selection/component_selector.py
handwerkerd, line 37:
```suggestion
    "demo_external_regressors_motion_task_models",
    "demo_external_aroma",
```

tedana/resources/decision_trees/external_aroma.json
handwerkerd: I also tried to add an integration tested based on `test_integration_three_echo_external_regressors_single_model` but it's failing. Another integration test that uses a predefined mixing matrix should be relatively quick to run and some of the changes here (static figures) would only appear in an integration test so I think it would be a good addition.

handwerkerd (changes requested):
Looks good. If you're adding all of this, I'd suggest including a `demo_external_aroma.json` that actually runs & that people can use for a reference and also an integration test to make sure the static figures are appropriately generated.

docs/building_decision_trees.rst
tsalo: Will add, though I also need to fix this description. `max_rp_corr > 0.5` does not come from ICA-AROMA. It's an arbitrary threshold.

tsalo: I fixed the descriptions of the threshold throughout and added the ICA-AROMA reference as well.

tsalo:
I have run the updated decision tree locally and it runs fine. I don't think we need to test each decision tree in an integration test though. The integration tests already take a long time, and I think it would be better to implement #1378 at some point instead.

tedana/resources/decision_trees/external_aroma.json
tsalo: Since both you and Dan requested it, I've added a section on the tree, though I haven't created a workflow figure.

tedana/metrics/external.py
tsalo: I made it clearer in the text that 0.5 is arbitrary.

eurunuela (commented):
Ran the demo_external_aroma tree end-to-end locally to check the static-figure path, and I think I found why the integration test was failing for @handwerkerd.

The tree's external_regressor_config requests regressors: ["trans_x", "trans_y", "trans_z", "rot_x", "rot_y", "rot_z"] (fMRIPrep naming), but our only external-regressor test file, tedana/tests/data/external_regress_Ftest_3echo.tsv, has columns Mot_X…Mot_Yaw. So any integration test pointed at that TSV can't resolve the regressors — fit_max_rp_corr_to_regressors does external_regressors[config["regressors"]] and KeyErrors before it computes anything.

Once I renamed those six columns to match, the workflow runs clean (exit 0): HFC and max_rp_corr motion model both populate, 10 accepted / 33 rejected, and confound_correlations.svg renders correctly via the new n_models == 0 branch in plot_heatmap. So the tree logic and the figure path are fine.

We may want to build a wrapper around it or force users to follow the fMRIprep naming convention.

tsalo:
I'm fine changing the demo tree to work with our test data. Since it's not a real decision tree, I expect people to tinker with it before applying it to their data, which might be from fMRIPrep or afni_proc.py (or something else, I suppose).

handwerkerd (commented):
Overall this is good. One thing in your replies is that this isn't the full AROMA and we're not even using the same thresholds as AROMA. Do you think including `aroma` in the name of the decision tree `json` will cause unnecessary confusion? One reason this is on my mind is I added an integration smoke test and the highest kappa component was rejected due to the new motion criterion. As a `demo` tree we're releasing this without claims it is useful, but maybe calling it `aroma` might cause unnecessary headaches.

Image

Here's the integration test I added. It was fairly quick & could be added to the PR.

```
def test_integration_three_echo_aroma_external(skip_integration):
    """Integration test of tedana workflow with aroma extern regress."""

    if skip_integration:
        pytest.skip("Skipping three-echo with aroma external regressors integration test")

    test_data_path, osf_id = data_for_testing_info("three-echo")
    out_dir = os.path.abspath(
        os.path.join(test_data_path, "../../outputs/three-echo-aroma-externalreg")
    )

    if os.path.exists(out_dir):
        shutil.rmtree(out_dir)

    # download data and run the test
    # external_regress_Ftest_3echo.tsv has 12 columns for motion, 1 for CSF, and 1 for task signal
    # The regressor values and expected fits with the data are detailed in:
    # tests.test_external_metrics.sample_external_regressors
    download_test_data(osf_id, test_data_path)
    tree_name = "resources/decision_trees/demo_external_aroma.json"
    tedana_cli.tedana_workflow(
        data=f"{test_data_path}/three_echo_Cornell_zcat.nii.gz",
        tes=[14.5, 38.5, 62.5],
        out_dir=out_dir,
        tree=files("tedana") / tree_name,
        external_regressors=files("tedana") / "tests/data/external_regress_Ftest_3echo.tsv",
        mixing_file=f"{test_data_path}/desc_ICA_mixing_static.tsv",
        low_mem=True,
    )
```

Note, to run with the motion regressors (and their labels) we're already including in tests, in `demo_external_aroma.json` the `external_regressor_config`->`regressors` field should be `"regressors": ["Mot_X", "Mot_Y", "Mot_Z", "Mot_Pitch", "Mot_Roll", "Mot_Yaw"]`

tsalo:
I've updated the decision tree and added your integration test. I also expanded the description of the demo tree to make it clear that it only includes a partial port of ICA-AROMA, the thresholds aren't validated, and no one should use it as-is. I think the comment is clear enough that no one will use it as a production-ready decision tree.

eurunuela (approved):

handwerkerd (approved):
LGTM. Thank you again for your patience with all my suggestions.

PR #1429, opened 2026-05-16
Deprecate T2* input and generate R2* outputs

### LINKED ISSUE #1424: Accept and prioritize R2* over T2*
### Summary

If nothing else, R2* is easier to average or analyze across subjects, because it's less susceptible to outliers than T2* (i.e., low R2* can become very high T2*).

### Next Steps

1. Add `--r2smap` parameter to CLIs and make it mutually exclusive with `--t2smap`.
2. Internally convert loaded T2* map (when provided) to R2* map.
3. Modify internal logic to use R2* instead of T2*.
4. Write out R2* map in s^-1 instead of T2* map. This would be the only breaking change, but I don't think it's a big deal.

tsalo:
Closes #1424. I still need to add T2* outputs back into the workflows and change the figures back to T2*.

- Deprecate t2smap input to CLIs.
- Add r2smap input to CLIs. It is mutually exclusive with t2smap input.
- Write out R2* maps in addition to T2* maps. The R2* maps are in s^-1 units.


tedana/decay.py
```diff
@@ -17,73 +17,70 @@
...
-    temp_arr[:, nonzerovox] = np.exp(-echo_times / t2s[nonzerovox])  # (E x V) array
-    bad_voxel_idx = np.any(temp_arr == 0, axis=0) & (t2s != 0)
+    eps = np.finfo(dtype=r2s.dtype).eps
+    nonzerovox = r2s != 0
+    temp_arr = np.zeros((len(echo_times), len(r2s)))
+    temp_arr[:, nonzerovox] = np.exp(-echo_times * r2s[nonzerovox])  # (E x V) array
+    bad_voxel_idx = np.any(temp_arr == 0, axis=0) & (r2s != 0)
     n_bad_voxels = np.sum(bad_voxel_idx)
     if n_bad_voxels > 0:
         n_voxels = temp_arr.size
```
eurunuela, line 53: `temp_arr.size` returns `E * V` (echoes × voxels), so `n_voxels` here is not the voxel count — it's the total number of array elements. The log message on line 56 then reports `n_bad_voxels / (E * V)` as the percentage, which is off by a factor of `E` (number of echoes).

Suggested fix:
```python
n_voxels = temp_arr.shape[1]  # number of voxels
```

tedana/utils.py
```diff
@@ -611,6 +611,57 @@ def check_t2s_values(t2s_map):
...
+            "R2* maps should be in s⁻¹ (typical values: 10-100 s⁻¹). "
+            "If you have a T2* map in seconds, use --t2smap instead."
+        )
+    elif median_r2s > 1000:
+        raise ValueError(
+            f"R2* map median value is {median_r2s:.2f} s⁻¹, which is too large. "
+            "R2* maps should be in s⁻¹ (typical values: 10-100 s⁻¹). "
+            "Please check your R2* map units."
+        )
+    elif not (2 <= median_r2s <= 200):
```
eurunuela, line 654: The `[2, 200] s⁻¹` warning band works well for 3T brain data, but at other field strengths it can produce spurious warnings for perfectly valid data:
- 1.5T: typical R2* ≈ 7–50 s⁻¹ — falls below the lower bound
- 7T: R2* can reach 150–180 s⁻¹ in gray matter — approaches the upper bound

Please add a note to the docstring documenting this assumption, e.g.:
```
Notes
-----
The typical range of [2, 200] s⁻¹ is calibrated for 3T brain data.
Users at 1.5T or ≥7T may see this warning even with correctly-scaled maps.
```
No code change is strictly needed, but without this note the warning will confuse non-3T users.

eurunuela (changes requested):
Thanks for this thorough refactor — the core of the change is correct and the architecture is solid. I'm requesting changes on two specific issues before merging.

**Mathematical correctness ✓**
The transformation `S(TE) = S0 * exp(-TE * R2*)` is exactly equivalent to the old `exp(-TE / T2*)` when `R2* = 1/T2*`. The optimal combination weights (`tes * exp(-tes * fr2s)`) are also correct per Posse et al. (1999).

**Deprecation path ✓**
`--t2smap` is properly marked as deprecated in the help text, mutually exclusive with `--r2smap`, and `check_t2s_values()` auto-converts ms→s with a clear warning. `load_mask()` correctly handles all flag combinations and converts to R2* before returning.

**Documentation ✓**
Formulas and units in `approach.rst` and `outputs.rst` are accurate. The BIDS alignment is appreciated.

**Issues that need to be addressed (see inline comments):**
1. **Bug in `_apply_r2s_ceiling()` log message** — `n_voxels` is computed as `temp_arr.size` (echoes × voxels) instead of the voxel count alone, so the reported percentage is off by a factor of E (number of echoes).
2. **`check_r2s_values()` warning threshold is undocumented** — the `[2, 200] s⁻¹` warning band is calibrated for 3T. Users at 1.5T (typical R2* ≈ 7–50 s⁻¹) will hit the lower warning bound; at 7T (R2* up to ~180 s⁻¹) they approach the upper bound. The threshold itself is defensible, but the field-strength assumption should be documented in the docstring so users know why they might see the warning.

**Non-blocking suggestions:**
- `test_decay.py` has no tests for realistic high R2* values (100–300 s⁻¹) or very low values (< 1 s⁻¹). Worth a follow-up issue.
- `test_utils.py` lacks an isolated unit test for the ms T2* → seconds conversion path in `check_t2s_values()`.

handwerkerd:
As I mentioned at our dev call, my sense is that most people in the field are used to looking at T2* contrasts and might have more trouble visually interpreting the R2* contrast. I'm asking some power users near me if they have opinions on this topic to check if my assumption is or is not accurate.

handwerkerd:
I'm still waiting to hear back from a few others, but I've now heard from 3 people (+ me) who are against this change. 

Because:
1. This is a breaking chance that will mess up interactions with existing data interfaces (i.e. everyone using their own T2* map as input

2. T2 maps really are a more standard and interpretable contrast.
For our five_echo integration tests, here's with T2* and then R2*. The big thing that jumps out better in R2* is the bright area around the dropout-region, but this is also visible in our adaptive mask and the T2*/R2* estimate variance image. This brightness hides some of the grey/white contrast that is visible in the T2*. Being able to see if we get good gray/white/structural contrast is the value of this QC image. The five-echo data has some, but not a lot of contrast, and that's valuable info.

image

image

3. I saw a modest clarity benefit of consistently using R2* under-the-hood, but I don't see a clear benefit for this externally facing change.

Since the file won't be large, I'm find saving an R2* map in addition to the T2* map and it wouldn't be hard to add an R2* input option in addition to the T2* input option rather than doing a replacement. Then it would be slightly easier to compare and see if there might be a benefit that I'm currently not seeing.

tsalo:
@handwerkerd maybe I should close this then. Instead, I could open a documentation-related PR to add a note somewhere that, if your goal is to average or statistically analyze T2* maps, you should calculate R2* first.

handwerkerd:
Sorry for the work you put into this and my initial delay responding.
Since it is a relatively small file, I could see just saving the R2* either by default or with the verbose flag.

PR #1434, opened 2026-05-20
Change T2* ceiling to seconds

tsalo:
Closes none, but patches something I missed in #1425.

- Replace infinite T2* values with 0.5 (500 ms) instead of 500 (artifact from when we had ms-based T2*).


eurunuela (approved):
LGTM. Good catch!

Issue #531, opened 2020-02-18
Add method for T2* estimation/correction from complex data


tsalo:
### Summary
Per the resources that @agahkarakuzu kindly shared with us in https://github.com/ME-ICA/tedana/issues/464#issuecomment-583893763, T2* estimates should be corrected for background field gradients using the phase data.

### Additional Detail
There is a MATLAB implementation [here](https://github.com/neuropoly/t2star-mapping), as well as in [qMRLab](https://github.com/qMRLab/qMRLab), that we could translate to Python for tedana.

### Next Steps
1. Translate T2* correction method to Python.
2. Test against original MATLAB version with ME-GRE data, given that high-resolution GRE data with many echoes is what the method was developed on.
3. Apply to ME-EPI data with much fewer echoes (e.g., 4) to determine if the method works well with ME-fMRI.
4. Support in t2smap and tedana workflows.
5. Add documentation.
6. Add tests.

tsalo:
@dowdlelt I only have access to complex, four-echo data. Is there any chance that you reconstructed phase data for any of the eight-echo datasets you've run and that you could share those data for testing this feature?

Alternatively, if anyone has any ME-GRE data we could use, that would also work.

dowdlelt:
I did not, unfortunately. Just had magnitude. I'll keep this in mind if I get in the scanner to do some testing, but its not super likely in the short term.

agahkarakuzu:
@tsalo I’ll get an example dataset soon, I think I can share with you. As a side note, processing time is going to be quite long. BTW is it not possible to obtain phase images from the complex data you have in hand, using e.g. ‘fslcomplex’ module?

tsalo:
@agahkarakuzu That would be amazing!

I do have phase data, but only four echoes. In order to test out the method as I manage to translate it to Python, I'd prefer to use some data with a relatively large number of echoes.

agahkarakuzu:
@tsalo 4 echoes are not that bad btw, the MEGRE data I’ll have is also 4 echoes. Have you visually inspected the decay curve of your magnitude images across 4 echoes?

tsalo:
Ah, but my data are ME-EPI, not GRE, so it's relatively low-resolution. Four-echo MEGRE data would definitely be better for testing the code.

I haven't looked at the signal decay for my data yet, but I recall our multi-echo fMRI denoising did a pretty good job on it.

agahkarakuzu:
[...]

PR #1451, opened 2026-06-25
Add option to estimate S0/T2* from complex-valued data

tsalo:
Relates to #531. This adds the ability to include phase data in the T2*/S0 estimation step in `t2smap` (not `tedana`). It does not include the magnitude correction requested in #531.

I just let Opus implement this based on a mishmash of my own edits and other Opus edits on a separate clone of tedana, so I'm keeping it as a draft until I have the chance to go through it line by line.

- Add `--phase` argument to `t2smap`. This accepts phase data to match the magnitude data from `-d`. Because we typically apply `t2smap` to motion-corrected, and even distortion-corrected, magnitude data, we'll need to think about the best way to handle phase data. Right now, I think the best workflow is:
    1. Rescale phase to -pi - pi.
    2. Convert magnitude and phase to real and imaginary.
    3. Apply transforms to real and imaginary files (probably not including slice timing correction).
    4. Convert BOLD reference-space real and imaginary files back to magnitude and phase.
    5. Run `t2smap`.
- Add `--fittype nlls` option to `t2smap`. This will use the phase data in the T2*/S0 estimation step.
- Add `--fitmode varys0` option to `t2smap`. This triggers a fitting approach where S0 is allowed to vary across volumes, but T2* is kept static. It should be faster than `--fitmode all`.

### DIFF
--- a/docs/superpowers/plans/2026-06-23-t2smap-complex-nlls.md
+++ b/docs/superpowers/plans/2026-06-23-t2smap-complex-nlls.md
@@ -1,0 +1,1195 @@
+# t2smap Complex NLLS Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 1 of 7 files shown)


tsalo:
Test failures are due to a timeout. I can just reduce the data used for the nlls tests (maybe 5x5x5x15 array?).

tsalo:
Per yesterday's dev call, this is worth getting in. I'll deal with the merge conflicts and get the tests passing.

PR #1457, opened 2026-06-29
Fix T2* and S0 variance assignment and empty T2* variance plot

### LINKED ISSUE #1440: S0 variance map all black in report
### Summary

For the five echo integration test, I noticed that the `S0 Estimate Variance` map is all black. The other figures look fine. This is a figure that should appear with `-- fittype curvefit`

### Additional Detail

This was an addition in #1290 
I looked at `stat-variance_desc-s0_statmap.nii.gz` and there is spatial structure, but the values range from 10^-7 to 10^-10. I assume, because the values are so close to 0, `plotting.plot_stat_map` is messing up the range in some internal way.

T2* Estimate Variance:

Image

S0 Estimate Variance

Image

### Next Steps

Figure out if these values are correct or if there's a scaling issue. For a non-linear fit of 5 values, it is plausible that the intercept is fairly stable. If this is some quirk of the fitting method and these values will be very close to 0, it's unclear if this is a useful metric. If there is a scaling issue, figure out what it is!

tsalo:
Closes #1440.

- Correctly assign T2* and S0 variance to the right files. They were previously swapped.
- Set threshold to 0 for variance plots, which should fix the empty T2* variance figure (#1440).

### DIFF
--- a/tedana/decay.py
+++ b/tedana/decay.py
@@ -242,8 +242,8 @@ def fit_monoexponential(data_cat, echo_times, adaptive_mask, report=True, n_thre
             s0_voxel,
             t2s_voxel,
             failure,
-            t2s_var_voxel,
             s0_var_voxel,
+            t2s_var_voxel,
             t2s_s0_covar_voxel,
         ) in results:
             if failure:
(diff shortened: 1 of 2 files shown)


tedana/decay.py
```diff
@@ -678,6 +682,11 @@ def rmse_of_fit_decay_ts(
     for n_good_echoes in range(2, len(tes) + 1):
         # a boolean mask for voxels with a specific num of good echoes
         use_vox = adaptive_mask == n_good_echoes
+        if n_good_echoes == 2:
+            # Voxels with a single good echo are fit using the first two echoes
+            # (mirroring fit_monoexponential), so include them here rather than
+            # leaving their RMSE as NaN (which renders as empty voxels).
+            use_vox = use_vox | (adaptive_mask == 1)
```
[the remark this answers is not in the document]
tsalo: Since the "full" T2*/S0 maps use the estimates from the first two echoes for voxels with an adaptive mask value of 1, it makes sense to show that in the RMSE map. Rejecting this suggestion.

tedana/decay.py
handwerkerd: Everything looks good. One really minor suggestion. The return docstring for the offending function does not mention it's also returning covariance values. To hopefully decrease future confusion, add them.

tsalo: Good catch! Should be fixed now.

handwerkerd (approved):
Thank you for fixing this.

PR #1458, opened 2026-07-02
Remove unnecessary reclassify download

handwerkerd:
In dealing with firewall issues I realized that `test_integration_reclassify_quiet_csv` adds two files to `.testing_data_cache/reclassify`. When `guarantee_reclassify_data` checks if the downloaded data are changed, it finds a change, deletes the data, and then re-downloads those data. For general use, this re-download is simply a waste. For someone offline or behind an agressive firewall, this deletes the data and then it cannot be downloaded to run the tests.

- `test_integration_reclassify_quiet_csv` creates `accept.csv` and `reject.csv` in the test's output directory instead of the downloaded data directory. In this function `check_integration_outputs` is given the names of these added files so that the file registry check passes.


tsalo (approved):
LGTM!

PR #1459, opened 2026-07-02
Expand interpolation option to perform two-stage interpolation+fitting

tsalo:
Closes none. Stems from https://github.com/ME-ICA/tedana/pull/1427#issuecomment-4835499372.

- Instead of just interpolating over failing voxels, the new version (1) interpolates over failing voxels, (2) re-runs the curvefit function on those failing voxels with the interpolated values as new initial values, and (3) interpolates over any voxels that fail the second curvefit step.
- I also moved the T2* estimation steps into a new function since so much is shared between t2smap and tedana.


eurunuela (approved):
LGTM and it is a good start towards better modularising the main workflow.

tsalo:
Thank you for reviewing @eurunuela! I think this PR requires two reviews. @handwerkerd would you mind reviewing at some point as well?

handwerkerd (approved):
LGTM!

As an aside I noticed that `t2smap_workflow` isn't outputting a log file. I think all the `LGR` calls are in it, but the logger was never initialized. If I'm not just missing something, we should fix that in a separate PR.

tsalo:
Good catch!

PR #1460, opened 2026-07-02
Modify maxrestart to reflect number of restarts, not number of ICA attempts

### LINKED ISSUE #1347: Description and name of `maxrestart` conflict
### Summary
The name `maxrestart` implies that it determines the number of additional attempts at ICA that can be performed in the case of nonconvergence or no BOLD components being found; however, the description and behavior both say that it is the number of attempts total.

As a user, I would expect setting `maxrestart 0` to mean that only one ICA attempt, with no restarts, will be performed. In fact, that will raise an error (and not an informative one, which is its own issue), since _no_ ICA attempt will be performed.

We based `maxrestart` on MELODIC's `maxrestart`, and MELODIC's help text says "--maxrestart	maximum number of restarts", so I'm incline to change the behavior so that `--maxrestart 0` means run ICA once, and `--maxrestart 1` means try _one_ restart. This would be a breaking change, but only a minor one. I don't think many users actually play with this parameter, and even if they do it would only mean that their hardcoded maxrestart values will result in one additional restart if necessary.

Related to #1346.

tsalo:
Closes #1347.

- This changes the behavior of `--maxrestart` by running ICA a maximum of 11 times by default (`--maxrestart 10`) and 1 time if set to 0.


eurunuela (approved):
LGTM

PR #1461, opened 2026-07-06
Draft slice banding metrics

### LINKED ISSUE #1415: Slice artifact metric
### Summary

A fairly obvious spatial pattern for components that should be rejected is a multiband artifact that follows slices.

Image

### Additional Detail

I don't know of a good way to quantify the artifact, but I think we should try to figure something out.

tsalo:
Closes #1415. This drafts a couple of metrics that might be useful for flagging multiband slice leakage artifacts.

- Add `spike` metric, which looks at kurtosis in a time series to flag components dominated by single large signal spikes.
- Add `slice_banding` and `slice_leakage` metrics, which might help with multiband acceleration artifacts.


tsalo:
I ran tedana on the first 20 runs of ds006185 (PAFIN fMRIPrep derivatives), then found the run with the component with the highest value for each component and the component with the lowest value in that same run (so we're comparing apples to apples), then plotted the relevant data.

Hopefully this helps give an idea of what these metrics are capturing.

### Spike: the component time series

spike

### Slice banding: the squared beta map (cropped to 2nd - 98th percentile)

slice_banding

### Slice leakage: the squared beta map (cropped to 2nd - 98th percentile)

slice_leakage

### Slice leakage, aliasing component: the squared beta map (cropped to 2nd - 98th percentile)

slice_leakage_aliasing_z

### Slice leakage, periodicity component: the squared beta map (cropped to 2nd - 98th percentile)

slice_leakage_periodicity_z

handwerkerd:
Here's more general feedback vs a full review of a draft.
* Since "slice banding" looks at every other slice without using the MB factor, it might actually be showing head motion from an interleaved sequence. It's still an artifact, but it might not be a MB artifact.
* I think "Slice leakage" is looking at the overall magnitude but, for ICA spatial maps, the magnitudes might vary more in the peaks than on average. I could see weighting voxels based on higher mean magnitudes. One rough idea for this is to spatially smooth each component and take the difference between the smoothed and unsmoothed data. If the magnitude of the difference varies by MC slice, that might be a more precise measure.
* "aliasing_z" might not be accurate because all MB sequences add a phase shift to improve reconstruction (see fig 4 of https://pmc.ncbi.nlm.nih.gov/articles/PMC3323676/ ) That means any spatial ghosting across slices should never be in the exact same location within each slice.
* For spiking, AFNI's 3dDespike looks for large differences between the L1 fit & the data. The main benefit of this approach is that it not only says if a time series is spikey, but it gives value for the number of spikes.
* I thought there was a leaking metric in Olafson's https://www.sciencedirect.com/science/article/pii/S1053811915001536 but the part of the paper might just be qualitative. Should read closer

More brainstorming that might be beyond the scope of this PR:
* I think there are already metrics for slice leakage that use the full voxels x time data. Maybe we can use those metrics to get a spatial map of where leakage is happening & then identify components with similar spatial maps
* The huge benefit we have here is multiple echoes. For the spiking artifact, we can look if a spike is in one or all echoes (fit component map to get a mean time series for each echo) It might then be possible to use Cesar's MEPFM method to identify spikes that cannot be well-modeled by relationships across echoes ( https://afni.nimh.nih.gov/pub/dist/doc/program_help/3dMEPFM.html & https://ieeexplore.ieee.org/document/8363649)
* Similar for the leakage measures, I think a leakage artifact should be similar across echoes. I'd need to think more, but there might be a way to use the multi-echo info to get a cleaner metric... maybe the fit of the time series to each echo & then look for magnitude diffs?

PR #1467, opened 2026-07-14
add logging to t2smap 

### LINKED ISSUE #1466: t2smap does not generate a log file
As an aside I noticed that `t2smap_workflow` isn't outputting a log file. I think all the `LGR` calls are in it, but the logger was never initialized. If I'm not just missing something, we should fix that in a separate PR.

_Originally posted by @handwerkerd in https://github.com/ME-ICA/tedana/pull/1459#pullrequestreview-4688878128_

handwerkerd:
Closes #1466 .

- The LGR logger in `t2smap_workflow` is now initialized with the same style of log file name as `tedana_workflow`
- The integration test for `t2smap` was not using `check_integration_outputs` and replicated a bunch of the code there. Added a param to that function to check for a log function that begins with `t2smap_` instead of `tedana_` and edited `test_integration_t2smap` to now use that to check the file registry.


tsalo (approved):
LGTM!

PR #1475, opened 2026-07-19
Use colorblind encoding in HTML report

tsalo:
Closes none.

- Switch red/green colors for colorblind-accessible colors. Also change the markers for rejected components to squares.
    - Maybe I should turn the accepted components to diamonds, so they're still distinct from the rejected components, but are more comparable to the squares.


tsalo:
### Old version

image

### New version

image

handwerkerd:
I like the approach and agreed about changing accepted to diamonds. FWIW I think the adaptive mask already uses colorblind friendly colors. 

I'm currently in transit but will try to find time to review/test code soonish.

eurunuela (approved):
LGTM! I will try to make the changes to Rica as well.

tedana/reporting/_palette.py
```diff
@@ -0,0 +1,47 @@
...
+
+    color: str
+    bokeh_marker: str
+    mpl_linestyle: str
+
+
+CLASSIFICATION_STYLES = {
+    "accepted": ClassStyle("#009E73", "circle", "solid"),
+    "rejected": ClassStyle("#D55E00", "square", "dashed"),
+    "ignored": ClassStyle("#0072B2", "triangle", "dotted"),
```
handwerkerd:
```suggestion
```
We don't use `ignored` anywhere so there's no reason to define a color. I also don't see where the `other` condition will come up, but it's benign and probably fine to include to avoid a crash.

tedana/tests/test_reporting_palette.py
```diff
@@ -0,0 +1,68 @@
+"""Tests for tedana.reporting._palette."""
+
+from tedana.reporting import _palette
+
+
+def test_known_classes_have_distinct_color_and_marker():
+    classes = ["accepted", "rejected", "ignored"]
```
handwerkerd:
```suggestion
    classes = ["accepted", "rejected"]
```

tedana/tests/test_reporting_palette.py
```diff
@@ -0,0 +1,68 @@
...
+
+def test_unknown_class_falls_back_to_other():
+    style = _palette.get_style("brand-new-label")
+    assert style is _palette.CLASSIFICATION_STYLES["other"]
+
+
+def test_helpers_match_get_style():
+    assert _palette.color_for("accepted") == _palette.get_style("accepted").color
+    assert _palette.marker_for("rejected") == _palette.get_style("rejected").bokeh_marker
+    assert _palette.linestyle_for("ignored") == _palette.get_style("ignored").mpl_linestyle
```
handwerkerd:
```suggestion
```

tedana/tests/test_reporting_palette.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def test_okabe_ito_values():
+    assert _palette.color_for("accepted") == "#009E73"
+    assert _palette.color_for("rejected") == "#D55E00"
+    assert _palette.color_for("ignored") == "#0072B2"
```
handwerkerd:
```suggestion
```

handwerkerd (changes requested):
Looks good. The only critique is that we haven't used "ignored" since #756. I opened #1503 to note that we should remove vestigial references to it, but no need to propagate "ignored" in new PRs.

PR #1479, opened 2026-07-20
Add PCA tab to HTML report

### LINKED ISSUE #1471: Add PCA tab to HTML report
### Summary

If nothing else, this can include the PCA scree plots, but we could also consider adding an option to generate PCA component selection-related figures.

tsalo:
Closes #1471. We could expand this in the future to include component maps (e.g., if verbose is True) and other PCA-related figures.

- Create new PCA tab in HTML report for the already-generated PCA-related figures.


tsalo:
Here's what the new tab looks like

Screenshot 2026-07-20 113621

eurunuela (approved):
LGTM. Will have to fix merge conflicts.

handwerkerd (changes requested):
This is a great addition. My one request is to add to outputs.rst examples have good and bad outputs. This is an unfortunately common problem and it would be useful to have one place to point people. I have a dataset with some useful exemplars so I'll try to pull a few out ASAP & will write some explanatory text.

tsalo:
I don't typically refer to the PCA figure so if you're comfortable writing the summary and creating a good vs. bad figure that would be great. Do you want to open a PR to my branch so it all ends up bundled in this PR, or would you rather we merge this and you make a followup documentation PR before the next release?

handwerkerd (approved):
I'll approve & can add to `outputs.rst` as part of a separate PR.

PR #1480, opened 2026-07-20
Reorder layers in adaptive mask plot

tsalo:
Closes none, but spins off a small bugfix from #1478.

- Reverse order of mask lines in `plot_adaptive_mask` so that initial mask is on the bottom and the classification mask is on the top.

### DIFF
--- a/tedana/reporting/static_figures.py
+++ b/tedana/reporting/static_figures.py
@@ -804,10 +804,10 @@ def plot_adaptive_mask(
             cut_coords=5,
         )
         ob.add_contours(
-            mask_clf,
+            io_generator.mask,
             threshold=0.2,
             levels=[0.5],
-            colors=[color_dict["Classification, OC & Initial"]],
+            colors=[color_dict["Initial mask only"]],
             linewidths=1.5,
         )
         ob.add_contours(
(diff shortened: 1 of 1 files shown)


tsalo:
Here are the figures from the five-echo test. You'll be able to see that the majority of the initial mask is retained in the final mask, and the new plot correctly shows a blue line (Classification, OC & Initial) for that mask instead of a red line (Initial mask only).

#### Before

adaptive_mask_before

#### After

adaptive_mask_after

eurunuela (approved):
LGTM! Thanks!

handwerkerd:
@tsalo I was traveling, but I wanted to mention we did originally compare the two color contours (see : https://github.com/ME-ICA/tedana/pull/1333#issuecomment-3797358300 )

The reason I prefer the initial mask as the top layer is because that makes the user-provided mask very clear and highlights any deviations from the initial mask.

That is, if you look at the following, it's clear that a non-trivial number of voxels in the user-provided mask were not included in the ICA mask.
image

For the flipped version, there's a little dot of magenta but it seems to be a bit less clear how much is lost.
image

If others really like the now-current flipped ordering, I won't push back more than this.

tsalo:
I can revert the change, but we should fix the labels. The "initial mask only" label is accurate with the new order because any time you see magenta it means that that bit corresponds to voxels that were only present in the initial mask. In the old version that label wasn't accurate, since it reflected the initial mask, but not _only_ the initial mask.

